Issue #20, opened 2016-10-07
Incorrect density channel indexes for CVs at branching points [4]


bcumming:
The FVM formulation does not correctly map mechanism types to control volumes at branching points (including the soma).

At such CVs, there should be contributions from the density channels that are defined on each branch, however we currently only consider contributions from density channels defined on the "root" segment.

The solution will require that each density channel instance has an aditional state variable corresponding to the surface area of the CV that the channel covers.

We should hold off implementing this until @halfflat has finished merging his validation branch. We have this ticket, and the `fvm_multi.mechanism_indexes` unit test that fails.

Note that the solution still converges to the correct solution as we refine the mesh, however for course discretisations the errors are large.

bcumming:
This needs to be resolved as part of the GPU work, because it would be best to fix before moving stimulii into the backend.

bcumming:
I managed to get 1 day of work in on this ticket before the end of the sprint

https://github.com/eth-cscs/nestmc-proto/milestone/2

So I will move the remaining [2] story points over to the next sprint

https://github.com/eth-cscs/nestmc-proto/milestone/3

PR #22, opened 2016-10-07
add unit test that shows bug in fvm

bcumming:
* add unit test fvm_multi.mechanism_indexes that fails to show
  bug in ticket #20
* add interface to fvm_multicell for getting list of stimuli
* add unit test that checks stimuli are correctly added to
  lowered cell

### DIFF
--- a/tests/unit/test_fvm_multi.cpp
+++ b/tests/unit/test_fvm_multi.cpp
@@ -239,20 +239,13 @@ TEST(fvm_multi, mechanism_indexes)
     // iterate over mechanisms and test whether they were assigned to the correct CVs
     // TODO : this fails because we do not handle CVs at branching points (including soma) correctly
     for(auto& mech : fvcell.mechanisms()) {
-        auto const& ni = mech->node_index();
+        auto const& n = mech->node_index();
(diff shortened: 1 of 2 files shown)


tests/unit/test_fvm_multi.cpp
```diff
@@ -140,3 +140,119 @@ TEST(fvm_multi, multi_init)
...
+
+    // make vectors with the expected CV indexes for each mechanism
+    std::vector<unsigned> hh_index  = {0u, 4u, 5u, 6u, 7u, 8u};
+    std::vector<unsigned> pas_index = {0u, 1u, 2u, 3u, 4u, 9u, 10u, 11u, 12u};
+    // iterate over mechanisms and test whether they were assigned to the correct CVs
+    // TODO : this fails because we do not handle CVs at branching points (including soma) correctly
+    for(auto& mech : fvcell.mechanisms()) {
+        auto const& ni = mech->node_index();
+        if(mech->name()=="hh") {
+            EXPECT_EQ(ni.size(), hh_index.size());
```
halfflat: How about copy indices locally to a vector and then using gtest's vector comparison?

PR #23, opened 2016-10-11
Enhanced validation tests.

halfflat:
Validation of simulation results by comparing L-infinity distance and maximum peak displacement in voltage traces taken at mid-points and end-points of sections.

Note that these tests will broadly _fail_ at this point.

Reference data generation and visualization:
* Use voltage traces as validation data from NEURON; refactor generation scripts.
* Generate validation data as a dependency of `validation.exe`.
* Store validation data in a `tsplot`-compatible format.
* Extend `tsplot` to handle multiple time series per file, a non-time 'time' axis (see `-A` option), filtering of time series by meta data queries (see `-s` option).
* Remove pre-generated spike-time validation data files.

Core extensions:
* Add a single-cell `singleton_recipe` recipe wrapper.
* Allow cell descriptions to be cloned (explicitly) for use with this wrapper.
* Add `simple_sampler` which maintains a trace of scalar observables as a vector of (time, value) pairs.
* Allow cell groups to be exposed from `model` (used for post-synaptic event injection in synapse validation test.)

Utility extensions:
* Add `math::signum` function.
* Fix `value_type` determination in `util::transform_iterator` where the underlying iterator dereferences to a reference type.
* Add `util::max_element_by` and `util::max_value` sequence utilities to `rangeutil.hpp`.
* Small bug fixes for `util::optional` (one const correctness case, and proper behaviour for implicit conversion of `nothing_t` to `optional<void>`.)
* Generic `util::first` and `util::second` accessor functional objects for reference access to `std::pair` fields in the absence of generic lambda functions.
* Add `util::path` for path manipulation, corresponding to a subset of C++17 `std::filesystem::path` functionality. Only POSIX path semantics supported at this point.

Validation tests:
* Rework all validation tests as comparisons of voltage trace data against a (presumably converged) reference trace generated by NEURON.
* Add options to `validation.exe` driver for verbose output of test results (`-v`), limiting the compartment number in convergence tests (`-m`) and saving the generated traces to a `tsplot`-compatilble JSONfile (`-o`).
* Gather trace comparison metrics and analysis into `trace_analysis.hpp` and `trace_analysis.cpp`.
* Add super simple command line argument parsing utility `tinyopt.hpp`.
* Make a consistent naming of `ball_and_3stick` model across sources.
* New `ball_and_taper` model.


src/util/rangeutil.hpp
```diff
@@ -117,6 +118,77 @@ sort_by(const Seq& seq, const Proj& proj) {
...
+    return std::accumulate(std::begin(canon), std::end(canon), base);
+}
+
+// Maximum element by projection `proj`
+// - returns an iterator `i` into supplied sequence which has the maximum
+//   value of `proj(*i)`.
+
+template <typename Seq, typename Proj>
+typename sequence_traits<Seq>::iterator
+max_element_by(Seq& seq, const Proj& proj) {
```
bcumming: do we need the const and non-const versions of this? the const version ought to be enough.

src/util/sentinel.hpp
```diff
@@ -163,6 +161,16 @@ class sentinel_iterator {
     bool operator>(const sentinel_iterator& x) const {
         return !(x<=*this);
     }
+
+    // access to underlying iterator/sentinel
+
+    bool is_sentinel() const { return e_.index()!=0; }
+    bool is_iterator() const { return e_.index()==0; }
```
bcumming: would it be better to define one of the above, and write the other in terms of it? Ignore this if you can justify your actions!

tests/validation/validation_data.hpp
```diff
@@ -1,28 +1,83 @@
...
+    ~trace_io() {
+        if (f_) {
+            write_traces();
+        }
+    }
 
 private:
-    std::string path_=DATADIR "/validation";
+    util::path datadir_ = DATADIR "/validation";
+    std::ofstream f_;
```
bcumming: maybe a slightly longer name... I use `fid` everywhere.

nrn/nrn_validation.py
```diff
@@ -0,0 +1,251 @@
...
+import numpy as np
+import neuron
+from neuron import h
+
+# This is super annoying: without neuron.gui, need
+# to explicit load 'standard' hoc routines like 'run',
+# but this is chatty on stdout, which means we get
+# junk in our data if capturing output.
+
+def hoc_setup():
```
bcumming: see Aaron's hint about using a temporary variable to `/dev/null` stdio
```
temp = h.load_file('stdrun.hoc')
```

halfflat: I just tried this, but unfortunately still get "1" emitted to stdout.

src/util/sentinel.hpp
halfflat: This form reflects the state of the underlying variant type, but I'm not strongly committed to this form instead of what you suggest.

src/util/rangeutil.hpp
halfflat: Update includes test that demonstrates need for non const version (in short: need mutable iterator returned from call with non-const container).

nrn/nrn_validation.py
bcumming: So, we go with the original for now?

halfflat: Yep. Can't really see a better alternative.

PR #29, opened 2016-10-17
Fixes Clang 3.7 failure to compile.

vkarak:
* Container assignment requires that the elements of the target container are copy constructible, whereas `unique_ptr`s in Clang 7.0 are only move constructible, which I believe it's the correct behaviour according to the standard, preventing `rangeutil.hpp` from compiling.
* Includes correct header (`<numeric>`) that provides `std::accumulate()`.

### DIFF
--- a/src/cell.hpp
+++ b/src/cell.hpp
@@ -90,8 +90,11 @@ public:
         spike_detectors_(other.spike_detectors_),
         probes_(other.probes_)
      {
-        util::assign_by(segments_, other.segments_,
-            [](const segment_ptr& p) { return p->clone(); });
+         // unique_ptr's cannot be copy constructed, do a manual assignment
+         auto siter = segments_.begin();
+         for (const auto& s : other.segments_) {
+             *siter = std::move(s->clone());
+         }
      }
 
     /// add a soma to the cell
--- a/src/cell.hpp
+++ b/src/cell.hpp
@@ -242,4 +242,3 @@ cable_segment* cell::add_cable(cell::index_type parent, Args ...args)
 
 } // namespace mc
 } // namespace nest
-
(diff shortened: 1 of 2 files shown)


halfflat:
I'm a bit worried about the semantics: being able to move assign from a range means that the range must have ownership semantics, while we're assuming that they are not.

Can the new `assign` overload be over `Seq&&` instead, with a restriction that `Seq` be a container (that is, a sequence with ownership)? While we have an `is_sequence` type trait, we don't have a test for containership yet. It might be as straightforward as insisting that all non-owning sequences such as ranges derive from a tag class (e.g. `view_tag` or `non_owning_sequence_tag` …) and testing against that.

vkarak:
It seems then that there is a more fundamental design issue then. STL `assign` requires that the each element of the target container is constructible from a reference to the corresponding element of the source constructor (at least is this what the Clang implementation checks for). The problem of the ownership you are mentioning was solved in the original design using the `transform_view` to return a copy of each `segment_ptr`, wasn't it? But this still doesn't meet the requirement of the STL assign. What if the lambda returning a clone of `segment_ptr` created a copy locally and then moved that? Haven't tried it and may not work..

bcumming:
Also, the latest version of Clang is 3.9. You should be careful to differentiate Apple Clang 7.0 from mainline Clang.

vkarak:
Updated title accordingly. I haven't found a straightforward way to check the mapping between the two.

vkarak:
I reverted the semantics of `assign_by` back to their original. More from my last commit message:

The problem described in previous commit was caused only inside the cell's copy constructor, where an assignemt for `std::unique_ptr` was attempted. Since `std::unique_ptr`'s cannot be copy-assigned, I have made that assignment manually moving the just cloned other's segment to the cell being constructed. The original `assign_by` method, relying on STL's `assign()` cannot accommodate this without changing its semantics (as pointed out by @halfflat), which would then also conflict with STL's semantics.

halfflat:
Looks good to me! We can consider move semantics vs transform in more depth anon.

Issue #12, opened 2016-10-06
Update read.me to reflect the new installation procedure


w-klijn:
- github init for externals
- modparser installation

w-klijn:
Fixed in #31

PR #31, opened 2016-10-19
Update README.md with current installation procedure.

w-klijn:
Fixes issue #12 

Update README.md with current installation procedure.
  - Add Julia installation steps and example run
  - New Modcc build instructions
  - small typos

Also included in this PR:
  - Silence variable not used warning in validate.cpp
  - Add missing include to rangeutil.hpp

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,15 +3,24 @@
 This is the repository for the NestMC prototype code. Unfortunately we do not have thorough documentation of how-to guides.
 Below are some guides for how to build the project and run the miniapp.
 Contact us or submit a ticket if you have any questions or want help.
+https://github.com/eth-cscs/nestmc-proto
+
+1. Basic installation
+2. MPI
+3. TBB
+4. TBB on Cray systems
+5. Targeting KNL
+6. Building and running on HBP pcp machines
+ a. Julia
+
+
+
+
 
 ```bash
-# clone repo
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -59,7 +68,7 @@ cmake <path to CMakeLists.txt> -DWITH_TBB=ON
 ### TBB on Cray systems
 
 To compile with TBB on Cray systems, load the intel module, which will automatically configure the environment.
-The guide below shows how to use the version of TBB that is installed as part of the Intel compiler toolchain.
+The guide below shows how to use the version of TBB that is installed as part of the Intel compiler tool-chain.
```
bcumming: it should be toolchain

tests/validation/validate.cpp
```diff
@@ -54,8 +53,9 @@ int main(int argc, char **argv) {
             }
         }
 
-        rv = RUN_ALL_TESTS();
+        if (RUN_ALL_TESTS());   // Wrap with empty if to silence output not used warning
```
bcumming: if you just want to ignore the return value of `RUN_ALL_TESTS` just run it inline as

```
RUN_ALL_TESTS();
```

But in this case we want to return this value to the shell, i.e.

```
return RUN_ALL_TESTS();
```

PR #32, opened 2016-10-19
add reference solution for hh soma test

bcumming:
* julia script that uses SUNDIALS to solve the single compartment
  model with HH channel. The model has 4 variables (voltage and
  3 gating variables), which are coupled in the solver (i.e. not
  split like in Neuron / nest-mc.
* SUNDIALS uses variable order stepping and variable size time steps
  with error detection to accurately integrate in time


halfflat:
Awesome!

PR #33, opened 2016-10-19
Add analytic simple cable solver.

halfflat:
* New file (Julia module): PassiveCable.jl
* Update script documentation for profstats
* Add (too brief) documentation for PassiveCable.jl
* Fix `env`, `python` invocation on python scripts.

### DIFF
--- a/scripts/PassiveCable.jl
+++ b/scripts/PassiveCable.jl
@@ -86,7 +86,7 @@ end
 #     computed potential at (x,t) on cable.
 
 function cable(x, t, L, lambda, tau, r, V, I; tol=1e-8)
-    scale = -I*r*lambda;
+    scale = I*r*lambda;
     if scale == 0
         return V
     else
@@ -135,7 +135,7 @@ function rallpack1(x, t; tol=1e-8)
     lambda = sqrt(d/4 * RM/RA)
     tau = CM*RM
 
-    return cable(x, t, L, lambda, tau, r, EM, I, tol=tol)
+    return cable(x, t, L, lambda, tau, r, EM, -I, tol=tol)
 end
 
 end # module PassiveCable
--- a/scripts/PassiveCable.jl
+++ b/scripts/PassiveCable.jl
@@ -62,7 +62,7 @@ end
 #     I:  injected current on the left end (x = 0) of the cable.
 
 function cable(x, t, L, lambda, tau, r, V, I; tol=1e-8)
-    scale = I*r*lambda;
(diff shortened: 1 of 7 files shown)


scripts/PassiveCable.jl
```diff
@@ -0,0 +1,104 @@
...
+# In the physical model, the parameters correspond to the following:
+#
+#     L:  length of cable
+#     r:  linear axial resistivity
+#     g:  linear membrane conductance
+#     c:  linear membrane capacitance.
+#     V:  membrane reversal potential
+#     I:  injected current on the left end (x = 0) of the cable.
+
+function cable(x, t, L, lambda, tau, r, V, I; tol=1e-8)
```
bcumming: some of the indentation is a bit iffy in this routine and elsewhere. Is it your intention to left justify return statements?

scripts/README.md
```diff
@@ -1,5 +1,5 @@
 tsplot
```
bcumming: because this is markdown, how about making these headers, i.e. 
```
tsplots
=====
```
can be replaced with
```
#tsplots
```

scripts/PassiveCable.jl
```diff
@@ -0,0 +1,104 @@
...
+end
+
+
+# Rallpack 1 test
+#
+# One sided cable model with the following parameters:
+#
+#     RA = 1 Ω·m    bulk axial resistivity
+#     RM = 4 Ω·m²   areal membrane resistivity
+#     CM = 0.01 F/m²  areal membrane capacitance
```
bcumming: The nomenclature used in most documents I have seen is for upper case `C_m` being the membrane capacitance (i.e. units `F`), and lower case `c_m` being the specific membrane capacitance, i.e. units `F/area`. This is what we use elsewhere in our code, though it might not be what the rallpack documentation uses.

halfflat: The use here is to align with the naming used in the Rallpacks paper. Elsewhere in the script, the lower case, unsubscripted variables are used to denote linear rather than specific resistivity, capacitance etc., and so use a different name.

I'll update the comments in the script to reflect this.

scripts/PassiveCable.jl
halfflat: Mea culpa.

scripts/README.md
halfflat: Both are valid markdown, but I'll change to '#' for consistency with the other documents.

PR #34, opened 2016-10-20
Fix markdown of README.md

w-klijn:
- Fix broken markdown layout
- merge Julia installation proc with the KNL section
- Changed building of on HBP systems to example environment load


bcumming:
Thanks for that. Let's update these with examples of how to run with MPI and TBB on KNL when we have gotten to the bottom of the odd performance on Julia.

Issue #16, opened 2016-10-06
modcc: Declaring a state variable with (nA) unit leads to error


w-klijn:
When attempting to declare a state variable with (nA) as the unit an error occurs:

```
 ./modcc -t cpu -o file.txt generic_neuron_iaf_cell.txt
error: generic_neuron_iaf_cell.txt:(line 64,col 8)
  '(' is not a valid name for a state variable

63 STATE {
 64   vI (nA)   :
 65  }
```
Problem also occurs with (mA).
Removing the unit results 'fixes' the issue

[generic_neuron_iaf_cell.txt](https://github.com/eth-cscs/nestmc-proto/files/514079/generic_neuron_iaf_cell.txt)

bcumming:
@vkarak, this should be cut and paste from other variable declaration blocks where units are assigned to variables.

PR #35, opened 2016-10-22
modcc: Support for units syntax within a state block.

vkarak:
Fixes issue #16.

### DIFF
--- a/data/test.mod
+++ b/data/test.mod
@@ -10,8 +10,8 @@ NEURON  {
 }
 
 STATE {
-    h (mA)
-    m (mV) r (S)
+    h (nA)
+    m r
 }
 
 UNITS   {
--- a/data/test.mod
+++ b/data/test.mod
@@ -10,8 +10,8 @@ NEURON  {
 }
 
 STATE {
-    h
-    m r
+    h (mA)
+    m (mV) r (S)
 }
 
 UNITS   {
--- a/modcc/module.cpp
+++ b/modcc/module.cpp
@@ -31,21 +31,18 @@ Module::Module(std::string const& fname)
     buffer_[size] = 0; // append \0 to terminate string
 }
 
-Module::Module(std::vector<char> const& buffer)
-{
+Module::Module(std::vector<char> const& buffer) {
     buffer_ = buffer;
 
     // add \0 to end of buffer if not already present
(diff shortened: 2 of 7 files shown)


data/test.mod
```diff
@@ -10,8 +10,8 @@ NEURON  {
 }
 
 STATE {
```
bcumming: The test now checks only the case where units are provided.

A more low level unit test where we test correct parsing of state variable definitions with and without units is a good idea.

data/test.mod
```diff
@@ -10,7 +10,7 @@ NEURON  {
...
-    h
+    h (nA)
```
vkarak: I could probably completely revert this file to its original status.

modcc/module.cpp
```diff
@@ -40,6 +40,15 @@ Module::Module(std::vector<char> const& buffer)
         buffer_.push_back(0);
 }
 
+Module::Module(const char* buffer, size_t count)
+{
```
bcumming:
```C++
auto size = std::distance(buffer, std::find(buffer, buffer+count, '\0'));
buffer_.insert(buffer_.end(), buffer, buffer+size);
buffer_.push_back(0);
```

vkarak: I can see the C++ salt here, but we are scanning twice the `buffer`, first time in `std::distance` and second time in `insert()`. Of course, that won't be a performance issue for the buffer sizes we really get.

Hmm, I could use though @halfflat's sentinel iterators.

bcumming: If you want to rangify it, go for it! We are all in the same source tree now.

On the performance side, calculating the size of the buffer to be inserted before inserting means that we have fewer `malloc` than with `push_back` one by one.

vkarak: > On the performance side, calculating the size of the buffer to be inserted before inserting means that we have fewer malloc than with push_back one by one.

👍  That's a valid argument. I would go for it then + the reserve before pushing back. Rangifying it would be fun, but I will need to rewrite [this](https://github.com/eth-cscs/nestmc-proto/blob/master/tests/unit/test_range.cpp#L146) unless we move it to util. But again, that won't fix the possible reallocation you just mentioned. So I'll go with your original suggestion + the `reserve()`.

PR #36, opened 2016-10-25
Address bug in fix for cell clone ctor.

halfflat:
* Add test for cell clone semantics.
* Fix `cell::cell(clone_cell_t, const cell&)` for correct copy semantics.


vkarak:
Looks fine to me.

PR #37, opened 2016-10-26
Feature/cable validation

halfflat:
Refactor trace convergence validation tests, adding Rallpack1 test.

* New test cell: `make_cell_simple_cable()`. Emulates Rallpack1 model electrical behaviour, while working around current inability to set membrane conductance.
* Re-organize validation data generation: moves validation reference models to `validation/ref`, and adds appropriate amounts of `CMake` cruft to generate the data as required.
* Tidy up `BUILD_VALIDATION_DATA` and `VALIDATION_DATA_DIR` testing and setting in CMake. (These are now cache variables.)
* Bug fixes in numeric models: voltage scaling in `hh_soma.jl` (since rolled into a Julia module `HHChannels.jl`; improper scaling by length in `PassiveCable.jl` (not convinced it is correct now, but certainly it is less wrong.)
* Use numeric solution for reference data for `validate_soma.cpp`
* Bug fix in validation trace comparison, when neither trace has any local maxima.
* Consolidate common trace convergence testing code in `validate_ball_and_stick.cpp` (this file should probably be renamed to something better representing its contents).
* Allow multiple `-s` selection options for `tsplot`, acting disjunctively.


tests/test_common_cells.hpp
```diff
@@ -54,10 +61,15 @@ inline cell make_cell_ball_and_stick(bool with_stim = true) {
     auto soma = c.add_soma(12.6157/2.0);
     soma->add_mechanism(hh_parameters());
 
-    auto dendrite = c.add_cable(0, segmentKind::dendrite, 1.0/2, 1.0/2, 200.0);
-    dendrite->add_mechanism(pas_parameters());
-    dendrite->mechanism("membrane").set("r_L", 100);
-    dendrite->set_compartments(4);
+    c.add_cable(0, segmentKind::dendrite, 1.0/2, 1.0/2, 200.0);
+
+    for (auto& seg: c.segments()) {
```
bcumming: Is the loop necessary, when we have just one dendrite? I would prefer
```C++
soma->mechanism("membrane").set("r_L", 100);

auto dendrite = c.add_cable(0, segmentKind::dendrite, 1.0/2, 1.0/2, 200.0);
dendrite->mechanism("membrane").set("r_L", 100);
dendrite->add_mechanism(pas_parameters());
dendrite->set_compartments(4);

File: tests/test_common_cells.hpp [hunk 2]
```diff
@@ -90,10 +106,15 @@ inline cell make_cell_ball_and_taper(bool with_stim = true) {
     auto soma = c.add_soma(12.6157/2.0);
     soma->add_mechanism(hh_parameters());
 
-    auto dendrite = c.add_cable(0, segmentKind::dendrite, 1.0/2, 0.2/2, 200.0);
-    dendrite->add_mechanism(pas_parameters());
-    dendrite->mechanism("membrane").set("r_L", 100);
-    dendrite->set_compartments(4);
+    c.add_cable(0, segmentKind::dendrite, 1.0/2, 0.2/2, 200.0);
+
```
bcumming: see above comment.

tests/test_common_cells.hpp
```diff
@@ -149,6 +172,76 @@ inline cell make_cell_ball_and_3stick(bool with_stim = true) {
...
+ 
+        if (seg->is_dendrite()) {
+            seg->set_compartments(4);
+        }
+    }
+
+    if (with_stim) {
+        // stimulus in the middle of our zero-volume 'soma'
+        // corresponds to proximal end of cable.
+        c.add_stimulus({0,0.5}, {0.,  HUGE_VAL, I});
```
bcumming: `stimulus::value_type` is double now, but if that was changed to `float`, `HUGE_VAL` could be a problem.
How about `std::numeric_limits<nest::mc::stimulus::value_type>::max()`?

tests/validation/validate_ball_and_stick.cpp
```diff
@@ -21,66 +21,75 @@
 
 using namespace nest::mc;
 
-// TODO: further consolidate common code
+struct sampler_info {
+    const char* label;
```
bcumming: You use `const char*` instead of `std::string` in a lot of places here. What is the reason for that?

tests/validation/validate_ball_and_stick.cpp
```diff
@@ -21,66 +21,75 @@
...
 
-    float sample_dt = .025;
-    std::pair<const char *, simple_sampler> samplers[] = {
-        {"soma.mid", simple_sampler(sample_dt)},
-        {"dend.mid", simple_sampler(sample_dt)},
-        {"dend.end", simple_sampler(sample_dt)}
-    };
+        EXPECT_TRUE(run_validation);
+    }
+    catch (std::runtime_error&) {
```
bcumming: After gtest has been informed of the failure, the simulation is still run (i.e. we don't just return). Is the idea that we still want to collect the results from NestMC?

validation/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+# Helper function to add a data generation script that writes to standard output.
+# e.g.:
+#     add_validation_data(OUTPUT foo_model.json DEPENDS foo_model.py common.py COMMAND python foo_model.py)
+
+include(CMakeParseArguments)
+function(add_validation_data)
+    cmake_parse_arguments(ADD_VALIDATION_DATA "" "OUTPUT" "DEPENDS;COMMAND" ${ARGN})
+    set(out "${VALIDATION_DATA_DIR}/${ADD_VALIDATION_DATA_OUTPUT}")
+    string(REGEX REPLACE "([^;]+)" "${CMAKE_CURRENT_SOURCE_DIR}/\\1" deps "${ADD_VALIDATION_DATA_DEPENDS}")
+    add_custom_command(
```
bcumming: The indentation is a bit off.

validation/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+
+add_custom_target(validation_data)
+
+# Helper function because ffs CMake.
+
+function(make_unique_target_name name path)
+    # Try and make a broadly human readable target name if possible.
+    string(REGEX REPLACE ".*/" "" leaf "${path}")
+    string(REGEX REPLACE "[^a-zA-Z0-9_.+-]" "_" canon "${leaf}")
+
```
bcumming: This is better documentation that the official CMake docs.

validation/ref/neuron/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+     ball_and_stick
+     ball_and_3stick
+     ball_and_taper
+     simple_exp_synapse
+     simple_exp2_synapse
+     soma)
+
+foreach(model ${models})
+    set(script "${model}.py")
+    add_validation_data(
```
bcumming: gargh indentation! CMake also confuses text editors.

validation/ref/numeric/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# note: function add_validation_data defined in validation/CMakeLists.txt
+
+add_validation_data(
+    OUTPUT numeric_soma.json
+    DEPENDS numeric_soma.jl HHChannels.jl
+    COMMAND julia numeric_soma.jl)
```
bcumming: do we check for julia, in case it has not been installed on the target system?

validation/ref/numeric/HHChannels.jl
```diff
@@ -0,0 +1,143 @@
...
+        el     = -54.3mV,
+        q10    = 1
+    ) = new(c_m, gnabar, gkbar, gl, ena, ek, el, q10)
+
+end
+
+immutable Stim
+    t0        # start time of stimulus
+    t1        # stop time of stimulus
+    j         # stimulus current density
```
bcumming: `j` is not very descriptive. How about `current`? You could also use `i_e`, which is used widely in the literature, but I haven't seen `j` before.

validation/ref/numeric/CMakeLists.txt
halfflat: Ah, we (er, I) don't, no. Good point. We also depend on some Julia packages — I'll see how hard it would be to have CMake complain usefully if they are not installed.

validation/ref/numeric/HHChannels.jl
halfflat: Used `j` to emphasise that it is a current density rather than a current. Lower case `i` for current density is a safe convention? Happy to change.

tests/validation/validate_ball_and_stick.cpp
halfflat: Only for simplicity: lots of `const char*` literals, and the struct is only used within this one source file, so it didn't need to be more generic.

validation/ref/numeric/numeric_rallpack1.jl
```diff
@@ -0,0 +1,60 @@
...
+ts = collect(0s: 0.025ms: 250ms)
+trace = Dict(
+    :name => "membrane voltage",
+    :sim => "numeric",
+    :model => "rallpack1",
+    :units => "mV",
+    :data => Dict(
+        :time => map(t->t/ms, ts),
+        symbol("cable.x0.0") => map(v->v/mV, run_cable(0, ts)),
+        symbol("cable.x0.3") => map(v->v/mV, run_rallpack1(0.3, ts)),
```
halfflat: This should be `run_cable`, not `run_rallpack1`, at least until we can ensure nestmc validation test and reference model are the same.

tests/test_common_cells.hpp
halfflat: I earlier missed the setting of `r_L` for the soma segment, and thought this made it clearer that all (admittedly two) segments should have the same `r_L` parameter set.

halfflat:
This PR doesn't have the partial frustrum compartment modelling, so `ball_and_taper` results still poor. Hope that merge of complex compartment code will demonstrate a closer match with NEURON. With some more maths work (for the future!) could derive taper version of analytic cable model.

validation/ref/numeric/HHChannels.jl
bcumming: lower case `i` is used for current density, upper case `I` is current. I would change to `i_e`, which is current density of electrode current.

tests/test_common_cells.hpp
bcumming: make a `math::max` helper `constexpr` function. There is no guarentee that  `HUGE_VAL` can be meaningfully cast to `float`.

halfflat: I think actually the float cast is safe by the standard, but will make a `math` helper. `max` has some unfortunate overloading, so I'll stick with `infinity` if that's cool.

PR #38, opened 2016-10-26
Omp

Ivanmartinezperez:
1-OpenMP back end working without tasks (-DWITH_OMP).
2-Extrae+Paraver support added, now just working on Marenostrum due to the path specified(-DWITH_EXTRAE).


run-once.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
bcumming: The `run-once.sh` and `run_instrumented.sh` scripts
* they don't belong in the root path.
* have system specific information (i.e. Mare Nostrum).
* the naming, with underscore and hyphen, is inconsistent.

I would make a new path `scripts/system-marenostrum` and put them in their (along with `extrae.xml`)

cmake/CompilerOptions.cmake
```diff
@@ -4,6 +4,8 @@ set(CXXOPT_DEBUG "-g")
 set(CXXOPT_PTHREAD "-pthread")
 set(CXXOPT_CXX11 "-std=c++11")
 set(CXXOPT_WALL "-Wall")
+set(BSCOPT_EXTRAE "-O -g -I $(EXTRAE_HOME)/include")
```
bcumming: These options have three bits of information
    1. debug flags `-g`
    2. optimization flags `-O`
    3. an include path

When `WITH_EXTRAE` is set in the main CMakeLists, you should check if `-g` has been added to `CMAKE_CXX_FLAGS`, and add it if needed.

You should not specify the optimization flag, which might conflict with the default set by the system.

Set the include path using the CMake ` include_directories(${EXTRAE_INCLUDE})` in the main CMakeLists.txt. Make the user provide `EXTRAE_INCLUDE` (maybe an `EXTRAE_ROOT` like for TBB)

cmake/CompilerOptions.cmake
```diff
@@ -4,6 +4,8 @@ set(CXXOPT_DEBUG "-g")
...
+set(BSCLINK_EXTRAE "-Wl,-rpath -Wl,/apps/CEPBATOOLS/extrae/latest/impi5_mt+libgomp4.9/64/lib -L/apps/CEPBATOOLS/extrae/latest/impi5_mt+libgomp4.9/64/lib")
```
bcumming: System specific information like this is very brittle. I would make the user provide an `EXTRAE_LINK` with this information. You could include a READMD.md in the `scripts/system-marenostrum` path (see my comment below), that explains how to build.

CMakeLists.txt
```diff
@@ -39,6 +39,16 @@ if(WITH_TBB)
     add_definitions(-DWITH_TBB)
 endif()
 
+# OpenMP support
+set(WITH_OMP OFF CACHE BOOL "use OpenMP for on-node threading" )
```
Ivanmartinezperez: If the user specifies something different (i.e. THREADING_MODEL=foo) We load serial or just throw and error?

PD: If THREADING_MODEL is not specified, serial is the default model

bcumming: I think that it is reasonable to default to `serial`, and then throw an error if the user provides an unsupported/unknown back end. Silently just using serial if an invalid back end is requested will cause confusion.

cmake/CompilerOptions.cmake
bcumming: The way to do this properly would be to write a CMake module, something like `extrae.find`, to automate this process.

Ivanmartinezperez:
1- We need to modify the [README.md](https://github.com/eth-cscs/nestmc-proto/blob/master/README.md), right now we use `-DTHREADING_MODEL=omp` or `-DTHREADING_MODEL=tbb` to select the threading model.

2-Extrae linked through an script using `LD_PRELOAD` environment variable, if you want to extract a trace, just execute the scripts inside Extrae folder.

3-I need to write a readme inside the Extrae folder!

CMakeLists.txt
```diff
@@ -31,24 +31,28 @@ if(WITH_TRACE)
     add_definitions("-DWITH_TRACE")
 endif()
 
-# TBB support
-set(WITH_TBB OFF CACHE BOOL "use TBB for on-node threading" )
-if(WITH_TBB)
+#threading model selection
```
bcumming:
```
 set(THREADING_MODEL "serial" CACHE STRING "set the threading model, one of serial/tbb/omp")
```

CMakeLists.txt
```diff
@@ -31,12 +31,26 @@ if(WITH_TRACE)
...
+if(THREADING_MODEL MATCHES "tbb")
+    # TBB support
+    set(WITH_TBB OFF CACHE BOOL "use TBB for on-node threading" )
```
bcumming: the `WITH_TBB` and `WITH_OMP` variables below are no longer need if we use `THREADING_MODEL` to describe the threading backend.

PR #45, opened 2016-10-27
Part 1 of the big GPU merge

bcumming:
This PR is part of the gpu feature merge. The GPU implementation is not implemented here. Instead, we focus on refactoring of the original "multicore" back end so that it is ready for adding the GPU back end.

This is a big and messy change, for which I am sorry.
## build System
- A `WITH_CUDA` option has been added to the main CMakeLists. This finds the CUDA toolkit, and sets CUDA compiler flags, and will build unit tests for the gpu back end.
- The CMakeLists that generates mechanisms with modcc has been updated to generate CUDA mechanisms.
- the library is now named `libnestmc` instead of `libcellalgo`
- merge the external libraries that are optionally linked againts (tbb, libunwind, etc) into a single
  `EXTERNAL_LIBRARIES` list for ease of linking
## modcc
- the cprinter and cudaprinter have had small changes to generate mechanism files that are compatible with the refactored library.
## algorithms
- the indexes into algorithm was "rangified". An algorithm `index_into_iterator` takes two ranges as inputs to make a range that lazily generates the index of sub into super set.
## backends
- made a new path `src/backends/` for backend specific type and implementation code.
- currently:
  - complete support for the `multicore` and `gpu` backends
  - `gpu` back end is not optimized or validated
- the back end implementations are in `src/backends`
- a single `backend` class, `nest::mc::{multicore,gpu}::backend`, is provides all backend specific type and implementation details fro each backend
  - storage containers
  - Hines matrix assembly for FVM method
  - Hines matrix solver
  - mechanism "factory"
## lowered fvm cells
- removed `fvm_cell` because this can be modelled with an `fvm_multicell` with one cell.
- refactored to use backend type and implementation from `fvm_policy`
- use `std::vector` instead of containers in `nest::mc::memory::` where possible when building cells.
## memory library
Refactor the "memory" library, making it much simpler and better integrated into the rest of the application. However, it is still far from perfect. The `Coordinator` approach needs to be improved, most likely by putting target-specific wisdom into pointers (which could obviate the need for a `const_array_view` type.
1. renaming and moving
   - move from `vector/` to `src/memory`
   - move into the `nest::mc` namespace, i.e. all types and functions are now in `nest::mc::memory`
   - change from camel case nameing scheme to NestMC style naming.
2. simplification
   - remove the CRTP cruft that was used to make operator overloading work for operations like copying from one range into another, and filling a range with a constant value. These have been replaced with `memory::fill()` and `memory::copy()` helper functions. This simplified the code _a lot_, and makes code clearer in user land.
   
   ```
   // before
   vec(0, 5) = other;
   // now
   memory::copy(other, vec(0, 5));
   ```
   - add some wrappers in `src/memory/wrappers.hpp` that help with making views. These are particularly useful for passing `std::vector` through interfaces that expect a view.
## debug backtraces
Added stack traces for debugging.
- support for OSX and Linux via libunwind
- backtraces can be generated manually `nest::mc::util::backtrace().print()`
  - creates a new file and dumps trace into file
  - prints message to `stderr` with file name and instructions on how to analyse
- backtraces are also automatically generated when an assertion `EXPECTS` statement fails
- a python script in `scripts/print_backtrace` pretty prints the output with file name, line number and demangled symbols
## util simplification and consolidation
The `src/util.hpp` file was removed
- much of its contents were dead code and just removed
- useful components like `pprintf` and `make_unique` were moved into the `src/utils` path in standalone files
There was a lot of overlap between functionality provided in `src/memory/util.hpp` and existing functions/types in the `nest::mc::util` namespace. The `memory` implementations were removed, and their `nest::mc::util` counterparts used. There is still some work remaining, namely moving the rest of the `src/memory/util.hpp` into `src/util/...`


src/memory/gpu.hpp
```diff
@@ -0,0 +1,71 @@
...
+template <typename T>
+void memcpy_d2h(const T* from, T* to, std::size_t size) {
+    auto bytes = sizeof(T)*size;
+    if (size==0) return;
+    auto status = cudaMemcpy(
+        reinterpret_cast<void*>(to), reinterpret_cast<const void*>(from),
+        bytes, cudaMemcpyDeviceToHost
+    );
+    if(status != cudaSuccess) {
+        LOG_ERROR("cudaMemcpy(d2h, " + std::to_string(bytes) + ") " + cudaGetErrorString(status));
```
halfflat: We've now got multiple error, verbosity macros in use under `src/`; these should be consolidated in one place with consistent usage across the code.

src/spike_source.hpp
```diff
@@ -12,7 +12,8 @@ class spike_detector {
 public:
     using cell_type = Cell;
 
-    spike_detector(const cell_type& cell, typename Cell::detector_handle h, double thresh, float t_init) :
+    spike_detector(
+        const cell_type& cell, typename Cell::detector_handle h, double thresh, float t_init) :
```
halfflat: Parameters line up with initializers, without any intervening punctuation. I think it's a confusing compromise. We can nut this out on our coding guidelines bike shed channel.

src/util/rangeutil.hpp
```diff
@@ -194,3 +195,15 @@ Value max_value(const Seq& seq, Compare cmp = Compare{}) {
...
+// convenience helper for printing a range
+
+template <typename T>
+std::ostream& operator<<(std::ostream& o, const nest::mc::util::range<T>& rng) {
+    o << "{";
+    for (auto const& value: rng) {
+        o << value << " ";
+    }
+    return o << "}";
+}
```
halfflat: I'd prefer to have a pretty printer wrapper for this sort of thing, as the one formatter for all ranges is a bit of a Procrustean bed. Use would be nearly as simple though, e.g.
```
    std::cout << range_printer(some_range) << ...
```

tests/unit/test_fvm_multi.cpp
```diff
@@ -75,7 +75,30 @@ TEST(fvm_multi, init)
...
+    // test matrix diagonals for sign
+    auto is_pos = [](decltype(J.l()) v) {
+        for(auto val : v) if(val<=0.) return false;
+        return true;
+    };
+    auto is_neg = [](decltype(J.l()) v) {
+        for(auto val : v) if(val>=0.) return false;
+        return true;
+    };
+    EXPECT_TRUE(is_neg(J.l()(1, J.size())));
```
halfflat: Formatting of test lambdas. (Consider using `std::any_of`/`std::all_of`?)

src/util/rangeutil.hpp
bcumming: I see your point. That Procrustes sounds like quite an unsavoury character, so I am sorry if my one-size-fits-all helper function caused harm to any travellers crossing our code base.

modcc/cprinter.cpp
```diff
@@ -102,8 +102,8 @@ CPrinter::CPrinter(Module &m, bool o)
 
     text_.add_line();
     text_.add_line("// allocate memory");
-    text_.add_line("data_ = vector_type(field_size * num_fields);");
-    text_.add_line("data_(memory::all) = std::numeric_limits<value_type>::quiet_NaN();");
+    text_.add_line("data_ = array(field_size * num_fields);");
+    text_.add_line("memory::fill(data_, std::numeric_limits<value_type>::quiet_NaN());");
```
bcumming: Yep, it was sloppiness. I have fixed it by using just one line:
```
text_.add_line("data_ = array(field_size*num_fields, std::numeric_limits<value_type>::quiet_NaN());");
```

src/algorithms.hpp
```diff
@@ -283,43 +285,101 @@ bool is_unique(const C& c)
     return std::adjacent_find(c.begin(), c.end()) == c.end();
 }
 
+template <typename R1, typename R2>
+class index_into_iterator:
+    public std::iterator<std::forward_iterator_tag, typename R1::value_type>
+{
+public:
+    using value_type = typename R1::value_type;
```
halfflat: If the class is ultimately going to be defined over iterator types rather than range types, then `iterator_traits<super_iterator>::difference_type` will be the way to go. Analogous to indices for e.g. `std::string`, an unsigned type might be preferable.

CMakeLists.txt
```diff
@@ -39,6 +39,18 @@ if(WITH_TBB)
     add_definitions(-DWITH_TBB)
 endif()
 
+# CUDA support
+set(WITH_CUDA OFF CACHE BOOL "use CUDA for GPU offload" )
+if(WITH_CUDA)
+    find_package(CUDA REQUIRED)
+    # the vector library has a compiled component when using the CUDA backend
+    include(ExternalProject)
```
bcumming: removed

src/algorithms.hpp
```diff
@@ -283,43 +285,101 @@ bool is_unique(const C& c)
...
+    mutable value_type super_idx_;
+
+public:
+    index_into_iterator(super_iterator sup, super_iterator sup_end, sub_iterator sub) :
+        super_it_(sup), super_end_(sup_end),
+        sub_it_(sub), super_idx_(0)
+    {}
+
+    index_into_iterator(const R1& super, const R2& sub):
+        super_it_  (std::begin(super)),
```
bcumming: removed.

src/algorithms.hpp
```diff
@@ -283,43 +285,101 @@ bool is_unique(const C& c)
...
+{
+public:
+    using value_type = typename R1::value_type;
+
+private:
+    using super_iterator = typename R1::const_iterator;
+    using sub_iterator = typename R2::const_iterator;
+
+    mutable super_iterator super_it_;
+    const super_iterator super_end_;
```
bcumming: I agree.
* `index_into_iterator` is now templated on the 3 iterator types
* updated unit test to check that arrays work specifically

src/algorithms.hpp
```diff
@@ -283,43 +285,101 @@ bool is_unique(const C& c)
...
+    {}
+
+    index_into_iterator(const R1& super, const R2& sub):
+        super_it_  (std::begin(super)),
+        super_end_ (std::end(super)),
+        sub_it_    (std::begin(sub)),
+        super_idx_ (0)
+    {}
+
+    value_type operator* () {
```
bcumming: I agree.

src/algorithms.hpp
```diff
@@ -283,43 +285,101 @@ bool is_unique(const C& c)
...
+
+    void advance_super() {
+        while(super_it_!=super_end_ && !is_aligned()) {
+            ++super_it_;
+            ++super_idx_;
+        }
+
+        // this indicates that no match was found in super for a value
+        // in sub, which violates the precondition that sub is a subset of super
+        EXPECTS(!(super_it_==super_end_));
```
bcumming: I have added an `index_into_iterator<>::npos` type as you described it.

src/algorithms.hpp
```diff
@@ -283,43 +285,101 @@ bool is_unique(const C& c)
...
 ///
 /// Both sets are sorted and have unique entries.
 /// Complexity is O(n), where n is size of super
-template<typename C>
-// C::iterator models forward_iterator
-// C::value_type is_integral
-C index_into(const C& super, const C& sub)
+template<typename R1, typename R2>
+util::range<index_into_iterator<R1, R2>>
+index_into(const R1& super, const R2& sub)
```
bcumming: I made the change.

src/algorithms.hpp
```diff
@@ -283,43 +285,101 @@ bool is_unique(const C& c)
...
 {
-    //EXPECTS {s \in super : \forall s \in sub};
     EXPECTS(is_unique(super) && is_unique(sub));
     EXPECTS(is_sorted(super) && is_sorted(sub));
     EXPECTS(sub.size() <= super.size());
```
bcumming: They fit the preconditions stated above. I agree that it could be used for something more general, but we have only got tests that check the "ordered subset of another ordered subset" use case. I would propose keeping it like this, and extending it in the future if we find a use case for a more generic form.

src/cell.hpp
```diff
@@ -58,7 +58,7 @@ constexpr clone_cell_t clone_cell{};
 /// high-level abstract representation of a cell and its segments
 class cell {
 public:
-    using index_type = cell_lid_type;
+    using iarray = cell_lid_type;
```
bcumming: fixed.

src/cell_group.hpp
```diff
@@ -23,7 +23,7 @@ namespace mc {
 template <typename LoweredCell>
 class cell_group {
 public:
-    using index_type = cell_gid_type;
+    using iarray = cell_gid_type;
```
bcumming: fixed. simple text substitution error.

src/fvm_multicell.hpp
```diff
@@ -628,9 +638,17 @@ void fvm_multicell<T, I>::advance(T dt) {
...
+        // is a kernel actually needed?
+        // for now I only make the update if the injected current in nonzero to
+        // avoid a redundant host->device copy on the gpu
+        //
         // note: current_ in [mA/cm^2], ie in [nA], cv_areas_ in [µm^2].
         // unit scale factor: [nA/µm^2]/[mA/cm^2] = 100
-        current_[loc] -= 100*ie/cv_areas_[loc];
+        //current_[loc] -= 100*ie/cv_areas_[loc];
+        if (ie!=0.) {
+            current_[loc] = current_[loc] - 100*ie/cv_areas_[loc];
```
bcumming: Yes: the types returned by `operator[]` for `DeviceVector` are proxies, which are converted to `value_type` on asignment. Operators like `-=` have not been defined for these proxies.

This isn't a problem any more, because the gpu and cpu code for this loop has now been refactored into target specific loops/kernels.

src/gpu/util.hpp
```diff
@@ -0,0 +1,47 @@
...
+}
+*/
+
+namespace nest {
+namespace mc {
+namespace util {
+namespace gpu {
+
+
+
```
bcumming: This is a mess! Before Pascal GPUs, there was no `atomic_add` for `double`, so we had to roll our own with `CAS`. Pascal now natively supports atomic operations for `double`, so these are defined by the compiler. Unfortunately they are defined as "magic functions" inside the cuda 8 compiler, and are present even when compiling for non-Pascal architecture... mess ensues.

I have a hacked solution where `atomic_add` was defined inside the namespace of each mechanism, and forgot to remove this file.

I have removed it now.

src/memory/array_view.hpp
```diff
@@ -0,0 +1,540 @@
...
+    size_type        size_;
+};
+
+// export is_array_view helper
+using impl::is_array_view;
+
+} // namespace memory
+} // namespace mc
+} // namespace nest
+
```
bcumming: yes, they are.

src/memory/gpu.hpp
halfflat: Commenting just to test slack trigger.

src/backends/catalogue_multicore.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include "memory_multicore.hpp"
+
+namespace nest {
+namespace mc {
+namespace multicore {
+
+class catalogue : public memory_traits {
+public:
+    using base = memory_traits;
```
bcumming: This pattern has been removed. Everything is done via template parameters now.

src/backends/catalogue_multicore.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <ion.hpp>
+#include <mechanism.hpp>
+
+#include "memory_multicore.hpp"
+
+namespace nest {
+namespace mc {
+namespace multicore {
+
+class catalogue : public memory_traits {
```
bcumming: I have moved the catalogue code directly into the `backend` classes. They are so small that it really is a zero-sum game making them generic, and I think it would actually make the code harder to follow (which is important, because every time I refactor, this part of the code breaks). Given that, I prefer to minimise abstraction.

src/backends/fvm_multicore.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include "catalogue_multicore.hpp"
+#include "matrix_multicore.hpp"
+
+namespace nest {
+namespace mc {
+namespace multicore {
+
+struct fvm_policy : public memory_traits {
```
bcumming: I have moved everything into one monolithic backend class for each backend. They aren't actually so big, with all the GPU-specific backend implementation taking less than 250 LOC.

The correct generic abstraction is difficult to know before we try to implement another solver. Something like FD would look very similar, and I prefer to wait until such time as we try to implement a new solver before worrying too much about making it generic.

src/fvm_multicell.hpp
```diff
@@ -287,45 +286,46 @@ void fvm_multicell<T, I>::compute_cv_area_unnormalized_capacitance(
...
-    size_type probes_count = 0u;
-    size_type detectors_size = size(detector_handles);
-    size_type targets_size = size(target_handles);
-    size_type probes_size = size(probe_handles);
+    auto detectors_count = 0u;
+    auto targets_count = 0u;
+    auto probes_count = 0u;
+    auto detectors_size = size(detector_handles);
+    auto targets_size = size(target_handles);
+    auto probes_size = size(probe_handles);
```
bcumming: fixed

src/indexed_view.hpp
```diff
@@ -1,27 +1,26 @@
 #pragma once
 
-#include <vector/include/Vector.hpp>
+#include <memory/memory.hpp>
```
bcumming: It is now templated on `backend`. Because it is only used in mechanism implementations with the `memory::` library, I don't feel the need to make it more generic.

So I move it into the memory library.

src/util/unwind.hpp
```diff
@@ -14,28 +9,31 @@ namespace mc {
 namespace util {
 
 #ifdef WITH_UNWIND
```
halfflat: Given that the only reason `libunwind.h` needs to be included here is to get the definition of `unw_word_t`, which itself is an integral type, could we keep the #ifdefs and such inside `unwind.cpp`, hidden from the rest of the code, and use e.g.
```
struct source_location {
    std::string name;
    std::uintptr_t address;
};
```
which does not depend upon the architecture-specific int size used for `unw_word_t`?

This has the nice side effect of keeping the global namespace free of the libunwind-specific enums and typedefs.

src/util/unwind.cpp
```diff
@@ -51,8 +53,8 @@ std::string demangle(std::string s) {
     int status;
     char* demangled = abi::__cxa_demangle(s.c_str(), nullptr, nullptr, &status);
 
-    // __cxa_demangle only returns a valid string if the identifier in s was
-    // a mangled c++ symbol (i.e. returns an empty string for normal c symbols)
+    // the string returned by __cxa_demangle is on valid if it was passed a valid
```
bcumming: I don't like your tone.

src/util/unwind.hpp
bcumming: done. I was thinking about this as I was implementing this. I have assumed that it is a `std::uint64_t`, which is how it is defined in libunwind, with a `static_assert` in `unwind.cpp`, that will break the build if that assumption is every broken.

halfflat: Any reason not to use `uintptr_t`?

PR #47, opened 2016-10-27
Feature/new test model

halfflat:
Add `ball_and_squiggle` model; fix `ball_and_taper`.

* Make `test_common_cells.hpp` and `ball_and_taper.py` agree.
* Add `ball_and_squiggle` model that has a tapering undulating profile.


bcumming:
Looks good to me.

PR #46, opened 2016-10-27
More range functionality, unit tests.

halfflat:
Part of complex compartments feature update.

* Tests for `math::pi`, `math::lerp`, `math::area_frustrum` and `math::volume_frustrum`
* Fix `math:pi<long double>()`.
* New `filter` view: lazily selects based on predicate.
* Generic `front` and `back` for sequences.
* New range STL wrappers `stable_sort_by`, `all_of`, `any_of`.
* Consolidate common utility unit testing structures into `tests/unit/common.hpp`

PR #49, opened 2016-10-27
Feature/divided compartments

halfflat:
**Note: Includes PR #46 — resolve that PR first**

New compartment info structure for FVM.

* Make `algorithm::sum`, `algorithm::mean` more generic, allowing use with array types.
* Add `div_compartment` compartment representation, that holds geometric information for each half of a compartment that will then be used in calculating control volumes.
* Add three compartmentalisation schemes/policies that discretize a segment into `div_compartment` objects:
    * `div_compartment_by_ends` divides based only on the segment end points and radii.
    * `div_compartment_sampler` forms frusta by sampling the segment radius at each compartment boundary
    * `div_compartment_integrator` computes the compartment areas and volumes exactly by summing all frustra in the intersection of the segment and the compartment span.

### DIFF
--- a/src/algorithms.hpp
+++ b/src/algorithms.hpp
@@ -6,8 +6,9 @@
 #include <type_traits>
 #include <vector>
 
-#include "util.hpp"
-#include "util/debug.hpp"
+#include <util.hpp>
+#include <util/debug.hpp>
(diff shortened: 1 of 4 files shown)


halfflat:
Rebased on master.

PR #54, opened 2016-10-30
Complex compartments

halfflat:
- Use divided compartments to determine FVM coefficients.
- Pick correct control volume in FVM from sgement position (avoids off-by-half error.)
- Add colour override functionality to tsplot: `--colour` option.
- Add const accessor for cell soma.
- Source formatting, comments in `math.hpp`
- Fix `range_view`: was using incorrectly named type trait.
- Add unit test for `range_view`.
- Allow points of discontinuity to be omitted from L-infinity norm calculations.
- Add `-d, --min-dt` option to `validate.exe` to control time step in validation convergence tests.
- Add validation test: confirm divided compartment policy does not effect results on simple frustrum dendrites.
- Change default max compartments on validation tests to 100 (ad hoc observed convergence limit at dt circa 0.001 ms; finer spatial division would required much finer dt.)
- Make NEURON validation data generation scripts use CVODE by default, and with `secondorder=2` when non-zero `dt` is given.

### DIFF
--- a/src/fvm_multicell.hpp
+++ b/src/fvm_multicell.hpp
@@ -45,7 +45,7 @@ inline int find_cv_index(const segment_location& loc, const compartment_model& g
     return index;
 };
 
-template <typename Value, typename Index, typename DivPolicy=div_compartment_integrator>
+template <typename Value, typename Index>
 class fvm_multicell {
 public:
     fvm_multicell() = default;
@@ -247,16 +247,12 @@ private:
 //////////////////////////////// Implementation ////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////
 
-template <typename T, typename I, typename DivPolicy>
-void fvm_multicell<T, I, DivPolicy>::compute_cv_area_unnormalized_capacitance(
+template <typename T, typename I>
+void fvm_multicell<T, I>::compute_cv_area_unnormalized_capacitance(
     std::pair<size_type, size_type> comp_ival,
     const segment* seg,
     index_type &parent)
 {
-    using util::left;
-    using util::right;
-    using util::size;
-
     // precondition: group_parent_index[j] holds the correct value for
     // j in [base_comp, base_comp+segment.num_compartments()].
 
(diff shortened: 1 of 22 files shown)


src/fvm_multicell.hpp
```diff
@@ -233,8 +247,8 @@ class fvm_multicell {
 //////////////////////////////// Implementation ////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////
```
bcumming: The `DivPolicy` is used in one location. I wonder if it needs to be its own template parameter? Maybe when we construct a cell we can pass the policy as a parameter to the constructor?

src/fvm_multicell.hpp
```diff
@@ -233,8 +247,8 @@ class fvm_multicell {
...
 
-template <typename T, typename I>
-void fvm_multicell<T, I>::compute_cv_area_unnormalized_capacitance(
+template <typename T, typename I, typename DivPolicy>
+void fvm_multicell<T, I, DivPolicy>::compute_cv_area_unnormalized_capacitance(
     std::pair<size_type, size_type> comp_ival,
     const segment* seg,
     index_type &parent)
```
bcumming: I don't think you need the `using util::left/right` statements below (I can't seem to leave a comment on the lines themselves)

tests/validation/trace_analysis.hpp
```diff
@@ -21,6 +21,11 @@ namespace mc {
 
 double linf_distance(const trace_data& u, const trace_data& ref);
 
+// Compute linf distance as above, excluding samples near
```
bcumming: The comments above function definitions is inconsistent. in `trace_analysis.cpp`
```
/*
 * Description of function
 */
```
Then here we are using single `//` comments.
Elsewhere we have used `///` to indicate a descriptive comment above functions (and their prototypes).
I certainly prefer that to the 'noisy' style in `trace_analysis.cpp`.

src/fvm_multicell.hpp
halfflat: Yup, let's bake in the default for now and add it back as part of a backend policy class.

tests/validation/trace_analysis.hpp
halfflat: Sure. I don't have a strong preference one way or the other.

PR #59, opened 2016-11-01
Mechanisms unit tests

vkarak:
* Adds a unit test templated over the mechanism to be tested.
* A "prototype" mechanism is also need so that we can verify the results of the mechanism generated by modcc. As prototype mechanisms I used the current versions of the generated mechanisms.
* Aliasing of indexes can also be introduced, which is in fact the case of the expsyn and exp2syn mechanisms unit tests.


tests/unit/test_mechanisms.cpp
```diff
@@ -34,3 +39,97 @@ TEST(mechanisms, helpers) {
...
+    EXPECT_TRUE((std::is_same<typename proto_mechanism_type::index_type,
+                              typename mechanism_type::index_type>::value));
+    EXPECT_TRUE((std::is_same<typename proto_mechanism_type::value_type,
+                              typename mechanism_type::value_type>::value));
+    EXPECT_TRUE((std::is_same<typename proto_mechanism_type::vector_type,
+                              typename mechanism_type::vector_type>::value));
+
+    auto num_syn = 8;
+
+    // Indexes are aliased
```
vkarak: This comment is stale.

tests/unit/test_synapses.cpp
```diff
@@ -93,6 +99,7 @@ TEST(synapses, expsyn_basic_state)
     ptr->net_receive(3, 1.04);
     EXPECT_EQ(ptr->g[1], 3.14);
     EXPECT_EQ(ptr->g[3], 1.04);
+
```
vkarak: I should cleanup a bit this file and revert it to its original version, since now mechanisms tests are moved to `test_mechanisms.cpp`.

bcumming:
There have been quite a few changes to the generated mechanisms in the gpu branch. The front and middle of modcc are unchanged, however the cprinter and cudaprinter generate different code to accommodate the refactoring elsewhere in NestMC.

I think that this is a very, very good idea to test that mechanisms generated by modcc are correct. But I wonder if it would be better to introduce it after the gpu branch is merged?

vkarak:
Sure, this PR can wait until the big GPU merge is done.

Issue #62, opened 2016-11-04
Add KINETIC block parsing support to `modcc`


halfflat:
Point estimate: 6
Feature: #60 

* Extend lexer to support KINETIC-specific syntax: `KINETIC` keyword; `CONSERVE` keyword; `~` reaction introduction symbol; `<->` reaction arrows symbol; distinguish real and integral numbers.
* Extend parser to treat integral number expressions appropriately.
* Extend parser with an `Expression` sub-type for reactions and add any further sub-types for sub-components of a reaction.
* Add unit tests covering above.

Example kinetic block for validation:
```
PROCEDURE rates {
    : compute f1(v), r1(v) etc.
    ...
}

KINETIC kin {
    rates(v)
    ~ s1 <-> s2 (f1, r1)
    ~ s2 <-> s3 (f2, r2)
    ~ s2 <-> s4 (f3, r3)
    CONSERVE s1 + s3 + s4 - s2 = 1
}
```

bcumming:
I think it is missing
- a brief description of what is being solved.
- breakdown of time to be spent on each of the steps.

bcumming:
looks good.

halfflat:
Will also support at the parsing phase multimolecular reaction forms, e.g.
```
~ 2c1 + c2 <-> c3
```

halfflat:
PR #71 merged!

PR #71, opened 2016-11-07
Add simple KINETIC block parsing

halfflat:
Targets issue #62.

New tokens added:
* keywords `tok::kinetic`, `tok::conserve`;
* symbols `tok::tilde`, `tok::arrow`;
* split `tok::number` into `tok::real` and `tok::integer` (stoichiometric expressions may have only integer coefficients).

New expression sub-types added, together with parsing:
* `IntegerExpression` (derives from `NumberExpression`);
* `ReactionExpression` — represents a line of the form "~ a + b <-> c (f, r)";
* `StoichTermExpression` — represents a single term in the lhs or rhs of a reaction as an integral coefficient and identifier, for example "3c" or "-2 sp2";
* `StoichExpression` — represents a possibly empty series of `StoichTermExpression`s, corresponding to the lhs or rhs of a reaction, or the lhs of a `CONSERVE` declaration;
* `ConserveExpression` (derives from `BinaryExpression`) — represents a line of the form "CONSERVE a + 2b - c = expr".

Unit tests added for parsing and lexing of the above, with some clean up of the testing code:
* verbose testing selectable at run time with `-v`, instead of being a compile time decision;
* code duplication reduced in `test_parser.cpp` through the use of helper functions `check_parse` and `check_parse_fail`;
* handle verbose output from lexer tests by subclassing the lexer; additionally, emit tokens from `Lexer::parse()` in verbose output.


halfflat:
Commenting just to test slack trigger (on master).

tests/modcc/test_lexer.cpp
```diff
@@ -1,20 +1,45 @@
...
+    std::cout << "________________\n" << string << "\n________________\n";
+}
+
+void verbose_print(const Token& token) {
+    if (!g_verbose_flag) return;
+    std::cout << "tok: " << token << "\n";
+}
+
+class VerboseLexer: public Lexer {
+public:
```
bcumming: I like the verbose lexer... very useful for debugging.

tests/modcc/test_parser.cpp
```diff
@@ -15,481 +79,413 @@ TEST(Parser, full_file) {
...
     }
+}
 
-    {
-        Parser p("LOCAL x, ");
-        auto e = p.parse_local();
+// test parsing of line expressions
+TEST(Parser, parse_line_expression) {
+    const char* good_expr[] = {
+        "qt=q10^((celsius-22)/10)"
```
bcumming: Is this line missing a comma at the end?

tests/modcc/test_parser.cpp
```diff
@@ -15,481 +79,413 @@ TEST(Parser, full_file) {
...
-    {
-"+x             ",
-"-x             ",
-"(x + -y)       ",
-"-(x - + -y)    ",
-"exp(x + y)     ",
-"-exp(x + -y)   ",
+TEST(Parser, parse_stoich_term) {
+    const char* good_pos_expr[] = {
+        "B", "B3", "3B3", "0A", "12A"
```
bcumming: what about `2e4` or `1E3` for example? These are ambiguous, if we have a variable named `e4` or `E3` respectively. From looking at the lexer and parser code, I think these might be a problem.

tests/modcc/test_parser.cpp
```diff
@@ -15,481 +79,413 @@ TEST(Parser, full_file) {
...
-"foo(x+3, y, bar(21.4))",
-"y=exp(x+3) + log(exp(x/y))",
-"a=x^y^z",
-"a=x/y/z"
+// test parsing of stoich and reaction expressions
+TEST(Parser, parse_reaction_expression) {
+    const char* good_expr[] = {
+        "~ A + B <-> C + D (k1, k2)",
+        "~ 2B <-> C + D + E (k1(3,v), k2)",
+        "~ <-> C + D + 7 E (k1, f(a,b)-2)",
```
bcumming: Here you have a space between the number and the `E`. We discussed this didn't we, and decided that the compiler error message about an unexpected real number would be enough, didn't we?

tests/modcc/test_parser.cpp
```diff
@@ -15,481 +79,413 @@ TEST(Parser, full_file) {
...
-#ifdef VERBOSE_TEST
-        if(e) std::cout << e->to_string() << std::endl;
-#endif
-        EXPECT_NE(e, nullptr);
-        EXPECT_EQ(p.status(), lexerStatus::happy);
+    const char* bad_expr[] = {
+        "~ A + B <-> C + D (k1, k2, k3)",
+        "~ A + B <-> C + (k1, k2)",
+        "~ 2.3B <-> C + D + E (k1(3,v), k2)",
+        "~ <-> C + D + 7E",
```
bcumming: Aha, we did decide that after all...

tests/modcc/test_parser.cpp
halfflat: Indeed, yes!

PR #78, opened 2016-11-15
bugfix for missing mechanism paths

bcumming:
The mechansims/gpu and mechanisms/multicore paths were not being generated
by CMake during configuration, and modcc was silently ignoring that it was
not able to write to the nonexistant target paths.
I have updated CMake to generate the missing gpu and multicore target paths.


vkarak:
Looks fine to me.

Issue #67, opened 2016-11-04
Inital GPU support


halfflat:
Goal is to support having all the state evolution numerical work able to be performed on the GPU for the models currently implemented in the miniapp.

Specifically:
* CUDA implementations and glue for all matrix and mechanism computations.
* The miniapp able to use these GPU implementations for all of its supported models. (Compile-time only or support run-time selection?)
* GPU results validated against CPU results.

Tasks
#61 - clean up GPU branch [10]

PR #84, opened 2016-11-16
Feature/gpu validation issue #68

### LINKED ISSUE #68: GPU Validation [10]
point estimate: 10 points
feature: #67

The GPU branch is compiling and generating numerically stable results, however the miniapp is not generating spikes. This task will aim to reproduce correct results on the GPU.

The following steps are proposed:

1. reproduce the hh-soma validation test on GPU [4]
  * this is the simplest spot to start because it does not require events and doesn't have spatial terms
  * ensure that GPU mechanisms are correct
  * ensure that voltage and current updates are correctly implemented
2. reproduce the ball and stick model on GPU [2]
  * introduces spatial terms
  * ensure that the fvm model generation is correctly performed on GPU
3. reproduce miniapp spike chains [4]
  * ensure that the spike detection and deliver works on GPU

This will require running and debugging tests and miniapp, creating and updating existing validation and unit tests for the GPU along the way.

bcumming:
feature: #67

The GPU branch is compiling and generating numerically stable results, however the miniapp is not generating spikes. This task will aim to reproduce correct results on the GPU.

The following steps were performed:
1. reproduce the hh-soma validation test on GPU
  * this is the simplest spot to start because it does not require events and doesn't have spatial terms
  * ensure that voltage and current updates are correctly implemented
2. reproduce the ball and stick model on GPU
3. reproduce miniapp spike chains
  * ensure that the spike detection and deliver works on GPU

Add cell_group unit test to the cuda unit tests
* builds simple ball and stick model and integrates
  for 50ms and records how many spikes occur
* is a simple early warning that something is broken
  * no substitute for the validation tests

Update the validate_soma, validate_ball_and_stick and
validate_synapses validation tests for gpu
* refactor indivudual tests into test runner functions
  that are templated on lowered cell type
* for each of the original validation tests we now have
  a cuda (.cu) and c++ (.cpp) implementation
* add additional "backend" meta data to the validation
  output (either "gpu" or "cpu")
* the soma and ball_and_stick validation passes
* the synapses validation fails due to lack of net_receive
  on gpu mechanism implementations

The CUDA back end was using the default net_receive implementation,
which was a "nop". This is why event based simulations were not
validating on the GPU.
* use a CPrinter to generate the same net_receive block that is
  used for the multicore backend. This is not efficient, because
  each read/write requires a cudamemcpy between host and device
  memory, however it allows us to pass all unit and validation tests.
* A more efficient GPU-specific implementation is left for later
  optimization work.

fixes #68

### DIFF
--- a/modcc/cudaprinter.cpp
+++ b/modcc/cudaprinter.cpp
@@ -129,12 +129,10 @@ CUDAPrinter::CUDAPrinter(Module &m, bool o)
 
         // print stubs that call API method kernels that are defined in the
         // kernels::name namespace
-        auto proctest = [] (procedureKind k) {
-            return is_in(k, {procedureKind::normal, procedureKind::api});
-        };
         for(auto const &var : m.symbols()) {
             if (var.second->kind()==symbolKind::procedure &&
-                proctest(var.second->is_procedure()->kind()))
+                is_in(var.second->is_procedure()->kind(),
+                      {procedureKind::normal, procedureKind::api}))
             {
                 var.second->accept(this);
             }
--- a/modcc/cudaprinter.cpp
+++ b/modcc/cudaprinter.cpp
@@ -412,11 +412,9 @@ CUDAPrinter::CUDAPrinter(Module &m, bool o)
 
     //////////////////////////////////////////////
     //////////////////////////////////////////////
-
-    auto proctest = [] (procedureKind k) {return k == procedureKind::api;};
     for(auto const &var : m.symbols()) {
         if( var.second->kind()==symbolKind::procedure && 
-            proctest(var.second->is_procedure()->kind()))
+            var.second->is_procedure()->kind()==procedureKind::api)
         {
             auto proc = var.second->is_api_method();
             auto name = proc->name();
--- a/modcc/cudaprinter.cpp
+++ b/modcc/cudaprinter.cpp
@@ -1,5 +1,6 @@
 #include <algorithm>
 
+#include "cprinter.hpp" // needed for printing net_receive method
 #include "cudaprinter.hpp"
 #include "lexer.hpp"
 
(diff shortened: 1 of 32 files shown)


modcc/cudaprinter.cpp
```diff
@@ -431,8 +417,8 @@ CUDAPrinter::CUDAPrinter(Module &m, bool o)
 
     auto proctest = [] (procedureKind k) {return k == procedureKind::api;};
```
halfflat: The name `proctest` is now rather confusing: e.g. line 420, the test reads as if we're checking if it's a procedure, then testing again if it is a procedure. Rename?

tests/unit/test_cell_group.cu
```diff
@@ -0,0 +1,36 @@
...
+#include <fvm_multicell.hpp>
+#include <util/rangeutil.hpp>
+
+#include "../test_common_cells.hpp"
+
+using fvm_cell =
+    nest::mc::fvm::fvm_multicell<nest::mc::gpu::backend>;
+
+nest::mc::cell make_cell() {
+    using namespace nest::mc;
```
halfflat: `using namespace` here, but `nest::mc::cell` below, and the whole file is not itself an include file...
Can we just have a `using namespace nest::mc;` after the '#include' directives and use a different variable name than `cell` inside `make_cell()`? There is an overabundance of colons.

tests/validation/convergence_test.hpp
```diff
@@ -2,10 +2,11 @@
 
 #include <util/filter.hpp>
 #include <util/rangeutil.hpp>
+#include <cell.hpp>
 
 #include <json/json.hpp>
 
-#include "gtest.h"
+#include "../gtest.h"
```
halfflat: The `tests/CMakeLists.txt` file adds the directory with `gtest.h` (and the `test_common_cells.hpp` etc.) to the include path. Use of implicit include path vs. relative-to-source path in the test sources is inconsistent, but there shouldn't be any need to change code to use the relative path if it's already correctly finding `gtest.h` inter alia.

tests/validation/validate_soma.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <recipe.hpp>
+#include <simple_sampler.hpp>
+#include <util/rangeutil.hpp>
+
+#include "../test_common_cells.hpp"
+#include "convergence_test.hpp"
+#include "trace_analysis.hpp"
+#include "validation_data.hpp"
+
+using namespace nest::mc;
```
halfflat: `using` should be moved within the `validate_soma()` method.

tests/validation/validate_ball_and_stick.hpp
```diff
@@ -0,0 +1,157 @@
...
+#include <util/path.hpp>
+
+#include "../gtest.h"
+
+#include "../test_common_cells.hpp"
+#include "convergence_test.hpp"
+#include "trace_analysis.hpp"
+#include "validation_data.hpp"
+
+using namespace nest::mc;
```
halfflat: `using` should be moved within the methods below.

tests/validation/convergence_test.hpp
bcumming: It should be either `<gtest.h>` or `"../gtest.h" in my opinion. Isn't using `"..."` for paths relative to this one, and `<...>` for paths set with `-I -isystem` flags consistent?

The pattern "../gtest.h" in all of the unit tests.

modcc/cudaprinter.cpp
bcumming: I just removed the lamda, and used it directly, because it is only used once.

tests/validation/convergence_test.hpp
bcumming: I have fixed the `CMakeLists.txt` file to not include the base path, and adjusted all files in the tests path to use `../xyz.hpp`.

tests/unit/test_cell_group.cu
bcumming: done.

tests/validation/validate_soma.hpp
bcumming: agreed, and done!

tests/validation/validate_ball_and_stick.hpp
bcumming: agreed, and done!

Issue #94, opened 2016-11-21
Bug: modcc: incorrect parsing of negative terms in arithmetic expression


halfflat:
The expression "2 * 7 - 3 * 11 + 4 * 13" incorrectly parses as:
```
((2 * 7) - ((3 * 11) + (4 * 13)))
```
The correct parse should be:
```
(((2 * 7) - (3 * 11)) + (4 * 13))
```

Issue #63, opened 2016-11-04
modcc: transform parsed KINETIC representation to DERIVATIVE representation


halfflat:
Point estimate: 6
Feature: #60 

* Implement an AST transformation for `modcc` that rewrites a parsed kinetic block as a derivative block.
* Insert transformation into modcc processing.
* Validate with visitor; incorporate as unit test.
* Incorporation of `CONSERVE` semantics is *not* in scope for this task.

Expect a system such as
```
~ a + b <-> c (x, y)
~ 2c <-> a  (u, v)
``` 
to be transformed to a representation equivalent to that produced from the derivative expressions
```
a' = -a*b*x + c*y + c*c*u - a*v
b' = -a*b*x + c*y
c' = a*b*x - c*y - 2*c*c*u + 2*a*v
```

PR #95, opened 2016-11-21
Add KINETIC block rewriter (issue #63)

halfflat:
Adds a new `KineticRewriter` visitor that transforms (after semantic analysis) a parsed `KINETIC` procedure into an equivalent `DERIVATIVE` procedure. The visitor takes a `ProcedureExpression` and composes the equivalent procedure, available via the `as_procedure()` method on the visitor object.

* Move common functinality for 'local' variable insertion during transformation phase to new files `astmanip.?pp`.
* Add `Expression` method for directly setting scope.
* Use `scope_ptr` type alias widely.
* Implement correct `clone()` behaviour for `DerivativeExpression`
* Implement `KineticRewriter` transforming visitor class.
* Add equivalence test for `KineticRewriter`: the test incorporates a simple ad-hoc algebraic expression simplifier.
* Add unit test to `Parser.parse_binop` to exercise bug #94


modcc/kinrewriter.hpp
```diff
@@ -0,0 +1,196 @@
...
+    // Contribution to final ODE for each species is given by
+    // net_rate * net_delta.
+
+    for (auto& p: net_delta) {
+        if (p.second==0) continue;
+
+        auto term = make_expression<MulBinaryExpression>(
+            loc,
+            make_expression<IntegerExpression>(loc, p.second),
+            net_rate->clone());
```
halfflat: I think possibly should be `net_rate_sym` instead of `net_rate`, to avoid excess evaluations. Will confirm.

modcc/parser.cpp
```diff
@@ -753,9 +753,9 @@ symbol_ptr Parser::parse_procedure() {
             break;
         default:
             // it is a compiler error if trying to parse_procedure() without
-            // having DERIVATIVE, PROCEDURE, INITIAL or BREAKPOINT keyword
+            // having DERIVATIVE, KINETIC, PROCEDURE, INITIAL or BREAKPOINT keyword
             throw compiler_exception(
-                "attempt to parser_procedure() without {DERIVATIVE,PROCEDURE,INITIAL,BREAKPOINT}",
+                "attempt to parser_procedure() without {DERIVATIVE,KINETIC,PROCEDURE,INITIAL,BREAKPOINT}",
```
halfflat: speling errro.

tests/modcc/test_kinetic_rewriter.cpp
```diff
@@ -0,0 +1,507 @@
...
+        case tok::exp:
+            return prodterm("exp("+to_string(inner)+")");
+        case tok::log:
+            return prodterm("log("+to_string(inner)+")");
+        case tok::sin:
+            return prodterm("sin("+to_string(inner)+")");
+        case tok::cos:
+            return prodterm("cos("+to_string(inner)+")");
+        default:
+            throw std::runtime_error("unrecognized binop");
```
halfflat: 'binop' should be 'unaryop' in error text.

vkarak:
Trying to compile the branch, `modcc` segfaults when compiling `hh.mod`:
```
lldb) target create "./modcc/modcc"
Current executable set to './modcc/modcc' (x86_64).
(lldb) settings set -- target.run-args  "-t" "cpu" "mechanisms/mod/hh.mod" "-o" "hh.hpp"
(lldb) r
Process 90332 launched: './modcc/modcc' (x86_64)
Process 90332 stopped
* thread #1: tid = 0x1342d9f, 0x000000010004003c modcc`Symbol::kind(this=0x0000000000000000) const + 12 at expression.hpp:216, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x70)
    frame #0: 0x000000010004003c modcc`Symbol::kind(this=0x0000000000000000) const + 12 at expression.hpp:216
   213 	    }
   214 	
   215 	    symbolKind kind() const {
-> 216 	        return kind_;
   217 	    }
   218 	
   219 	    Symbol* is_symbol() override {
```

halfflat:
Yep, that would be a bug. Thanks for finding it! I admit some embarrassment.

tests/modcc/test_kinetic_rewriter.cpp
```diff
@@ -0,0 +1,507 @@
...
+#include <stdexcept>
+#include <string>
+
+#include "expression.hpp"
+#include "kinrewriter.hpp"
+#include "parser.hpp"
+
+// Simple algebraic term expansion/collection routines.
+
+namespace alg {
```
vkarak: We don't indent namespaces, do we?

vkarak: These algorithms look quite useful. What do you think of moving them to a separate header file inside test?

halfflat: I've historically indented 'internal' namespaces, but wouldn't indent the namespace that everything is in. This is a totally arbitrary distinction, and I'm not going to defend it.

Putting it into its own header is a good idea. I will move it out.

tests/modcc/test_kinetic_rewriter.cpp
```diff
@@ -0,0 +1,507 @@
...
+    std::map<std::string, prodsum> exmap;
+
+    for (const auto& stmt: stmts) {
+        if (auto assign = stmt->is_assignment()) {
+            auto lhs = assign->lhs();
+            std::string key;
+            if (auto deriv = lhs->is_derivative()) {
+                key = deriv->spelling()+"'";
+            }
+            else if (auto id = lhs->is_identifier()) {
```
halfflat: It should be taken in the e.g. the analysis of the example `DERIVATIVE` block, where in one instance we have an assignment to `a'` (the first path) and two lines below, an assignment to `rev2` (the second path).

vkarak:
@bcumming Looks fine for me. If you agree we can merge it.

PR #96, opened 2016-11-21
Feature/mechanisms unit tests

vkarak:
Resubmitting the mechanisms' unit test PR adapted to the post GPU-merge status.

* Prototype mechanisms are stored under `test/mechanisms`
* Only multicore mechanisms are considered in this PR
* A unit test templated over the mechanism type is provided for testing any given mechanism
  * Currently tested: expsyn, exp2syn, pas and hh
* The tests have no physical meaning. They just check numerical consistency.


bcumming:
@halfflat, can you please have a look at this?

halfflat:
Broadly looks good! A couple of things below:

* Owing to the potential of vectorization bugs of various sorts, I think it would be valuable to perturb the initial state across instances of the same mechanism, e.g. by having variation across the initial voltage vector.
* Similarly, having an irregular grouping of point mechanisms per compartment could help catch problems with gather/scatter optimizations.
* Having explicit prototype mechanisms checked into the source tree feels a bit brittle. What do you think of the following?
   * We modify `modcc` so that we can override the mechanism name and class name at mechanism creation time.
   * We get CMake to build regular and optimized versions of the test mechanisms using this option so that we can incorporate them into the same test executable.
   * The validation test is then performed between the regular and optimized mechanisms.

  Correct mathematical behaviour of various mechanisms could then be left to our validation testing…

vkarak:
> Owing to the potential of vectorization bugs of various sorts, I think it would be valuable to perturb the initial state across instances of the same mechanism, e.g. by having variation across the initial voltage vector.

Good point. I would not vote though for continuous math functions (e.g., sin/cos), but I would rather use a step-wise one with breakpoints at odd numbers. I think it would be easier to debug than having full blown real numbers as initial values.

> Having explicit prototype mechanisms checked into the source tree feels a bit brittle. What do you think of the following?

I totally agree. I knew my solution was a bit ugly, but I needed some feedback. I'll probably come back to you for the CMake magic.

vkarak:
@halfflat Addressed PR comments.

halfflat:
Looks good! But I was getting concerned by the proliferation of copy-paste CMake code and was thinking also that having another in-source build for mechanisms was not ideal. 

I'll submit a PR to your branch with a proposed alternative to the CMake set up, which also simplifies (I think) the existing mechanism building process.

vkarak:
I've merged your PR.

PR #114, opened 2016-11-28
Fixes icc compilation of uninitialized

vkarak:
It was introduced with PR #101.


halfflat:
Yup, totally mea culpa!

Issue #113, opened 2016-11-28
Bug: isolate and work around xlC++ compilation-related issues. [4]


halfflat:
Point estimate: 4
Feature: #65 

Some unit tests are failing when compiled with xlC++ 13 on POWER8, and there appear to be other compiler errors exercised by the code:
* `{}` as default argument to a `std::vector` parameter was inducing an ICE.
* The `test_uninitalized` unit test exhibits a failure where an `apply` method invocation is being ordered before the preceding copy construction is completed.

Tasks:
* Isolate failing unit tests, confirm if possible that these are compiler issues rather code issues.
* Prepare defect report to forward through to IBM if we can make an isolated demo.
* Implement work-arounds as required.

The issue title says 'bug', but it's not clear if it's _our_ bug or _their_ bug.

halfflat:
Isolated three issues as follows:
1. std::end() returns incorrect pointer when applied to an array struct member with -O2:
   ```
   #include <cassert>
   #include <iterator>

   struct X {
       float x[1];
   };

   int main() {
       X a;
       assert(std::end(a.x) == &a.x[0]+1);
   }
   ```
    Workaround: write our own `std::end` equivalent and use it consistently throughout the code. This is clearly not very attractive.
2. ICE when compiling code which instantiates a templated class with a member that a const `vector` reference argument with default initializer `{}`.
   ```
   #include <vector>

   template <typename P>
   struct X {
       void g(const std::vector<float>& ={}) {}
   };

   void test() {
       X<double> u;
       u.g();
   }
   ```
    Workaround: don't use the default initializer for the method that triggers this ICE in our code; already in.
3. Incorrect optimization can reorder initialization and access of object constructed in `std::aligned_storage` (or, in fact, a `char` array with correct alignment).
   ```
   #include <cassert>
   #include <new>
   #include <type_traits>

   struct uninitialized_int {
       typename std::aligned_storage<sizeof(int), alignof(int)>::type data;

       int& ref() { return *static_cast<int*>(static_cast<void*>(&data)); }
       void construct(int x) { new(&data) int(x); }
   };

   struct increment {
       int operator()(int& a) const { return ++a; }
   };

   int main() {
       uninitialized_int ua;
       ua.construct(11);

       increment A;
       A(ua.ref());
       assert(12==ua.ref());
   }
   ```
[...]

PR #115, opened 2016-11-29
Work around xlC 13.1.4 compiler bugs.

halfflat:
Addresses in part issue #113.

* Make compatibility wrappers/functions that dance around xlC bugs. Wrappers are provied in `util/compat.hpp` and live in the `compat` namespace.
   * `compat::end` reimplements `std::end` but in a way that apparently does not trigger the xlC bug.
   * `compat::compiler_barrier_if_xlc_leq()` inserts a compiler reordering barrier if the compiler is xlC and the version less than or equal to that specified. Name is deliberately verbose.
   * `compat::isinf()` is an inline wrapper around `std::isinf()`, which apparently is sufficient to defuse an evaluation order bug with `std::isinf()` in switch statements.
* Use `compat::compiler_barrier_if_xlc_leq()` in `util::unitialized` reference access methods to avoid improper reordering with -O2.
* Use `compat::isinf()` in `test_math.cpp` to defuse improper reordering within `EXPECT_EQ` gtest macro of `std::isinf()`.
* Use `compat::end()` in `util::back()` and `util::cend()` to avoid incorrect `std::end()` behaviour with -O2.
* Use `util::cend()` in `algorithms::sum()`, again to avoid incorrect `std::end()` behaviour with -O2.

### DIFF
--- a/src/algorithms.hpp
+++ b/src/algorithms.hpp
@@ -29,7 +29,7 @@ typename util::sequence_traits<C>::value_type
 sum(C const& c)
 {
     using value_type = typename util::sequence_traits<C>::value_type;
-    return std::accumulate(std::begin(c), std::end(c), value_type{0});
+    return std::accumulate(util::cbegin(c), util::cend(c), value_type{0});
 }
 
 template <typename C>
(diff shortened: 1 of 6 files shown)


bcumming:
Looks good to me. Thanks for doing this Sam!

Issue #99, opened 2016-11-25
Prototype Gaussian elimination schema from linear ODE sparsity pattern. [8]


halfflat:
Point estimate: 8
Feature: #60

## Background

Initial kinetic scheme support requires the solution of linear, non-diagonal ODEs. and an implicit Euler scheme will be the first implementation.

The sparsity of the coefficient matrix is known at mechanism compile time, and so in principle the sequence of operations required to solve for one time step can be determined in advance, avoiding any explicit looping.

## Task

Given a sparsity pattern (ultimately determined by the AST produced by `modcc`), prototype a transformation that produces the correct sequence of arithmetic operations that will solve one step of implicit Euler integration for the corresponding system.

## Example

The kinetic scheme
```
~ a <-> b (r,s)
~ b <-> c (u,v)
```
will lead to the solution of a linear system of the form
```
[ b11 b12 0   ] [ x1 ] = [ r1 ]
[ b12 b22 b23 ] [ x2 ] = [ r2 ]
[ 0   b32 b33 ] [ x3 ] = [ r3 ]
```
and ultimately to a sequence of operations as follows (in single assignment form, following a simple left-to-right elimination scheme)
```
c22 = b11*b22-b21*b12
c23 = b11*b23
s2  = b11*r2 - b21*r1
c33 = c22*b33 - b32*c23
s3  = c22*r3 - b32*s2
x3  = s3 / c33
x2  = (s2 - c23*x3) / c22
x1  = (r1 - b12*x2) / b11
```

## Acceptance

The completed task will comprise a stand-alone piece of code that demonstrates the production of a sequence of operations corresponding to the Gaussian elimination process generically across a set of sample sparsity patterns and matrices of size 2x2 through 10x10. The matrices will always have non-zero diagonal, but need not be symmetric.

The correctness of the sequences can be validated with numerical checks with the inclusion of a simple expression evaluator that can operate over the produced operation sequences. For validation purposes, the matrix can be taken as being strictly diagonally dominant.

PR #118, opened 2016-12-04
Prototype synthetic Gaussian elimination (issue #99)

halfflat:
Adds a stand-alone prototype application `symge-demo` with source under `proto/symge`.
Adresses issue #99.

`symge-demo` depends upon `util/optional.hpp` but is otherwise stand-alone.

Demo application:
* Validates symbolic GE across a range of strictly positive definite matrices of size up to 10x10.
* Reports (with `-v`) number of required extra single-assignment symbols and numeric operations.
* With `-d`, print detailed information on test matrices and symbolic computations to standard error.

Note that the pivot selection can be simplified under the assumption that diagonal pivots should always be chosen in preference.


proto/symge/symbolic.hpp
```diff
@@ -0,0 +1,212 @@
...
+
+struct symbol_error: public std::runtime_error {
+    symbol_error(const std::string& what): std::runtime_error(what) {}
+};
+
+// Symbols are either primitive or represent the difference
+// between two product terms.
+
+template <typename S>
+struct symbol_term_ {
```
bcumming: impl namespace

proto/symge/symge-demo.cpp
```diff
@@ -0,0 +1,387 @@
...
+        join_impl(ss << std::forward<Head>(head) << sep, sep, std::forward<Args>(tail)...);
+}
+
+template <typename... Args>
+std::string join(const std::string& sep, Args&&... items) {
+    std::stringstream ss;
+    join_impl(ss, sep, std::forward<Args>(items)...);
+    return ss.str();
+}
+
```
bcumming: comments please. Just explain what it does. This might be useful for naming temporary variables in modcc?

proto/symge/symge-demo.cpp
```diff
@@ -0,0 +1,387 @@
...
+        if (l==0) return "a";
+
+        unsigned i = l-1;
+
+        char l0='a', l1='z';
+        char u0='A', u1='Z';
+        char d0='0', d1='9';
+        for (;;) {
+            char& c = s[i];
+            if ((c>=l0 && c<l1) || (c>=u0 && c<u1) || (c>=d0 && c<d1)) {
```
bcumming: range switch might work.

proto/symge/symge-demo.cpp
```diff
@@ -0,0 +1,387 @@
...
+            if (c==m.augcol()) o << " | ";
+            o << std::setw(12) << m[r][c];
+        }
+        o << " |\n";
+    }
+    return o;
+}
+
+template <typename Sep, typename V>
+struct sepval_t {
```
bcumming: I think we have a convention of using `_type`.
This should go in the coding standards, not that anybody reads them.

proto/symge/symge-demo.cpp
```diff
@@ -0,0 +1,387 @@
...
+        symbol s = syms[i];
+        if (s.def()) o << s << ": " << s.def() << "\n";
+    }
+    return o;
+}
+
+// Symbolic GE
+
+using symmrow = msparse::mrow<symbol>;
+using symmatrix = msparse::matrix<symbol>;
```
bcumming: underscore, e.g. `sym_mrow` etc.

proto/symge/symge-demo.cpp
```diff
@@ -0,0 +1,387 @@
...
+    struct pq_order_t {
+        bool operator()(const pq_entry& a, const pq_entry& b) const {
+            // The last condition preferences pivots on diagonal elements.
+            return a.key>b.key ||
+                  (a.key==b.key && a.mincol<b.mincol) ||
+                  (a.key==b.key && a.mincol==b.mincol && a.row!=a.key);
+        }
+    };
+
+    std::priority_queue<pq_entry, std::vector<pq_entry>, pq_order_t> pq;
```
bcumming: I like the generality, however we are working with SPD matrices, which don't require pivoting.
I would keep it simple and naive for now (which makes things easier for me to understand ;) ).

proto/symge/symge-demo.cpp
```diff
@@ -0,0 +1,387 @@
...
+    unsigned n;
+    unsigned nnz;
+    unsigned nmul;
+    unsigned nsub;
+    unsigned nsym;
+    double relerr;
+};
+
+template <typename Rng>
+symge_stats run_symge_validation(Rng& R, unsigned n, bool debug = false) {
```
bcumming: This would correspond to the most detailed level of unit testing?
With some smaller tests for individual components, it should be good to go!

proto/symge/symge-demo.cpp
halfflat: Turns out it's a non-standard extension; I'll stick with the `if` as the gain in clarity from a range switch is relatively small.

proto/symge/symge-demo.cpp
halfflat: For the demo, I don't think we need the unit tests, but certainly would for anything going into `modcc`.

proto/symge/symge-demo.cpp
halfflat: I think this is the first instance in the code of a top-level helper class like this; I'll change it to `sepval_proxy`.

halfflat:
Resubmitting PR to different branch.

PR #119, opened 2016-12-04
Cyclic iterators and cyclic range views

vkarak:
Support of cyclic iterators and cyclic views that can endlessly iterate
over a sequence, wrapping around at the end (or at the beginning in case
of reverse iteration).

### DIFF
--- a/tests/unit/test_cycle.cpp
+++ b/tests/unit/test_cycle.cpp
@@ -1,0 +1,179 @@
+#include "../gtest.h"
+
+#include <string>
+
+#include "common.hpp"
+#include <util/cycle.hpp>
+#include <util/meta.hpp>
+
+using namespace nest::mc;
+
+TEST(cycle_iterator, construct) {
+    std::vector<int> values = { 4, 2, 3 };
+    auto cycle_iter = util::make_cyclic_iterator(values.cbegin(), values.cend());
+
+    {
+        // copy constructor
+        auto cycle_iter_copy(cycle_iter);
+        EXPECT_EQ(cycle_iter, cycle_iter_copy);
+    }
+
+    {
+        // copy assignment
+        auto cycle_iter_copy = cycle_iter;
+        EXPECT_EQ(cycle_iter, cycle_iter_copy);
+    }
+
+    {
+        // move constructor
+        auto cycle_iter_copy(
+            util::make_cyclic_iterator(values.cbegin(), values.cend())
+        );
+        EXPECT_EQ(cycle_iter, cycle_iter_copy);
+    }
+}
+
+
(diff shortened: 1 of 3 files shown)


tests/unit/test_cycle.cpp
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+
+    {
+        std::vector<int> values_new(10, -1);
+
+        auto cycle_iter = util::cyclic_view(values).cbegin();
+        for (auto &v : values_new) {
+            v = *cycle_iter++;
+        }
```
vkarak: I was trying now to rewrite this in a more C++ fashion as follows:
```c++
        std::vector<int> values_new;
        std::copy_n(util::cyclic_view(values).cbegin(), 10,
                    std::back_inserter(values_new));

        EXPECT_EQ(10, values_new.size());
```

Interestingly, the above expectation fails. More interestingly, I realised that Clang 3.9, which I'm using, implements `copy_n()` by first getting an "end" iterator as `std::next(begin, 10)` and then trying to copy elements. However, moving the cycle iterator 10 times, will bring you just to the second element, that's why you end up with only one element in the new vector. The proposed implementation [here](http://en.cppreference.com/w/cpp/algorithm/copy_n) works fine with the above code and meets in any case the expected criterion of `copy_n()` of "Copies exactly count values from the range beginning at first to the range beginning at result". I can understand Clang wouldn't expect a "cyclic" range to be passed through `copy_n()` but I don't believe my view is at fault here. Opinions? 

I will try it also with another compiler. Alternatively, if we want to `copy_n()` portably (?), I should augment the internal state of the `cyclic_iterator` by a `num_wrap_arounds` to keep track of the "wrap-arounds". Gets ugly?

vkarak: Or I could probably make the equality check for the `cycle_iterator` stricter.

vkarak: I've just tested icc and gcc: all of them treat the `copy_n()` the same way. I could then add `num_wraps_` member variable and also check against it inside the `==` operator. I will need to handle properly the overflows though, since a cycle iterator by definition cycles forever...

halfflat: Yup, the current implementation (I'm still looking at it, sorry!) breaks a possibly expected invariant, that `distance(i, i+n) == n`. Rather than counting wrap-arounds, how about keeping track of the raw offset?

I'll have more comments soon…

halfflat:
I made a PR on your branch that addresses the problem with `copy_n` (plus removes a signed/unsigned warning.) Otherwise I think it is good to go!

vkarak:
Done.

PR #122, opened 2016-12-09
Add initializer_list version for is_in in modccutil.hpp

apeyser:
is_in was templated on (T thing, const T (&list)[N])
which fails to match an initializer list on clang-3.7.1
as in is_in(a, {procedureKind::...)

templating on (T thing, std::initializer_list<T> list)
works on clang, and continues to work on gcc.

### DIFF
--- a/modcc/modccutil.hpp
+++ b/modcc/modccutil.hpp
@@ -4,10 +4,21 @@
 #include <memory>
 #include <sstream>
 #include <vector>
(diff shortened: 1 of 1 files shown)


halfflat:
Looks good to me! We can use a more generic approach (e.g. wrapping `std::find` with a sequence-friendly interface) when we split out the range utilities and allow `modcc` to depend upon it.

PR #123, opened 2016-12-12
Bug/issue#20

### LINKED ISSUE #20: Incorrect density channel indexes for CVs at branching points [4]

### LINKED ISSUE #120: Update LaTeX docs for FVM scheme
The calculation of conductances was updated recently (weighted harmonic means etc.), such that the implementation no longer looks much like the formulation in `/docs/model`.

Specifically, the symbols used (e.g. `c` for conductance, `h` for distance between nodes, `R` for bulk resistivity etc) should match the documentation.

bcumming:
This PR addresses two issues that were closely related
* correctly accounting for the current contribution of density mechanisms to CVs at branch points, where the density mechnanism is not present on all branches. This was discussed in issue #20.
* adding support for weighting of current densities calculated from density mechanism. This is required to weight the current contribution to CVs in issue #20.

## small updates

* update CMake rules for finding libunwind, because it broke for some reason.
* add `binary_find` and unit tests to the algorithms library
    * returns an iterator, as opposed to `std::binary_search`, which returns a boolean.
    * works with ranges.
* added `subrange_view` specialization that takes a subrange specified by a pair of indexes
* added `assign_from` to range utils
    * a helper function that returns a proxy type that can be copied into a container
    * evaluate a range and store contents in a container, with minimal verbosity in user code
    * simple syntax for initializnig a container where it is declared, e.g. `std::vector<int> vec = util::assign_from(...);`
* update the LaTeX documentation for the FVM scheme
* fix bug when an event has to be delivered exactly 0 ms in the future in `cell_group`
    * avoid divide by zero on the diagonal of the linear system in the new formulation.

## updated FVM formulation

Most of the changes were in the `fvm_multicell` type.
* The FVM formulation was changed slightly, moving parameters (e.g. dividing both sides of equation by dt)
    * to ensure symmetric positive definate matrix property with new partial weights
    * to give the terms in the linear syste, i.e. the matrix, solution and rhs vectors more natural units    * the system is no `G*v = i`, where `G` is conductance matrix with units [uS], `v` is voltage [mV] and `i` is current [nA].
    * change names of fields from non-descriptive things like `face_alpha` to `face_conductance`
    * add more comments that explicitly give the expected units of fields inside the back end (to help future generationas trying to understand the code... and to help me understand it three weeks from now) 
* keep additional information about the surface area of sub-control-volumes at the start and end of each segment when calcuating CV areas, capacitances and face conductances
* use this information to ensure that current contributions from density channels on branching points are properly accounted for by weighting
* remove weighting from point process currents, because they are calculated with the correct units of nA
* plumbing work to add support for user-supplied weights
    * the backend code for multicore and gpu now supports weights for mechanism generation
    * update `cprinter` and `cudaprinter` to generate kernels that use user-supplied weights for density channels
* backend modifications to generate matrices and RHS vectors according to the new formulation

fixes #20
fixes #120


src/fvm_multicell.hpp
```diff
@@ -223,12 +223,30 @@ class fvm_multicell {
 
     std::vector<std::pair<const array fvm_multicell::*, size_type>> probes_;
 
+    struct segment_cv_range {
```
bcumming: add comments that explain why and how each bit of information herein is needed/used.

src/fvm_multicell.hpp
```diff
@@ -198,16 +198,16 @@ class fvm_multicell {
...
 
-    /// cv_capacitance_[i] is the capacitance of CV i per unit area (i.e. c_m) [F/m^2]
-    array cv_capacitance_;
+    /// cv_capacitance_[i] is the capacitance of CV membrane [pF]
+    ///     C_m = area*c_m
+    array cv_capacitance_; // units [µm^2*F*m^-2 = pF]
 
-    /// the average current density over the surface of each CV [mA/cm^2]
-    /// current_ = i_m - i_e
+    /// the average current density over the surface of each CV [nA]
```
bcumming: this is no longer the average.

src/fvm_multicell.hpp
```diff
@@ -238,11 +256,12 @@ class fvm_multicell {
 //////////////////////////////// Implementation ////////////////////////////////
 ////////////////////////////////////////////////////////////////////////////////
 template <typename Backend>
-void fvm_multicell<Backend>::compute_cv_area_unnormalized_capacitance(
+typename fvm_multicell<Backend>::segment_cv_range
+fvm_multicell<Backend>::compute_cv_area_unnormalized_capacitance(
```
bcumming: does this need `unnormalized` in the title (hint, it doesn't)?

src/fvm_multicell.hpp
```diff
@@ -501,62 +545,106 @@ void fvm_multicell<Backend>::initialize(
...
         const auto& mech_name = syni.first;
         size_type mech_index = mechanisms_.size();
 
-        auto comp_indices = syn_mech_map[syni.second];
-        size_type n_indices = size(comp_indices);
+        auto cv_map = syn_mech_map[syni.second];
+        size_type n_indices = size(cv_map);
 
         // sort indices but keep track of their original order for assigning
         // target handles
```
bcumming: `std::pair` now?

src/fvm_multicell.hpp
bcumming: fixed.

src/fvm_multicell.hpp
bcumming: comments added.

src/fvm_multicell.hpp
bcumming: renamed.

src/fvm_multicell.hpp
bcumming: reverted.

Issue #87, opened 2016-11-17
Early GPU Performance Analysis [5]


bcumming:
The GPU backend is has been validated against the multicore backend for all of the validation tests and the miniapp. However, these tests take much longer to run on the GPU than the CPU.

The main reason for this is many small copies, typically of a single floating point value, between host and device memory whenever
  * a probe sample is taken
  * spike detection is performed
  * an iclamp contribution is added to the current
  * an event is delivered to a synapse 

These issues should be addressed before we start to benchmark and optimize the core GPU algorithms

This task has two steps
  1. make a detailed list of all locations where these transfers are affecting performance [2]
  2. propose some designs for removing the memcopies [3]

This task will not involve implementation of the design changes, which will be done in a later task after this analysis is complete.

bcumming:
Step 1: run the miniapp with nvprof.

bcumming:
## 1: Taxonomy of host-device copies

### Sampling

The sampling interface allows the user to record voltage and current values at a user-specified location at a user-specified frequency. The call to `cell_.probe(...)` below copies a single value from device to host memory each time it is called:

https://github.com/eth-cscs/nestmc-proto/blob/97e17b187e68a3253fde6c668918605dea2f5d3b/src/cell_group.hpp#L96

### Events

The event system has two calls that use host-device copies:

#### Spike detection

The `test()` call copies a single voltage value to host:

https://github.com/eth-cscs/nestmc-proto/blob/97e17b187e68a3253fde6c668918605dea2f5d3b/src/cell_group.hpp#L121

#### Event delivery

https://github.com/eth-cscs/nestmc-proto/blob/97e17b187e68a3253fde6c668918605dea2f5d3b/src/cell_group.hpp#L129

When an event is delivered the `net_receive` method on the target synapse is called. Any access to the synapse state in the `net_receive` will involve a H2D, D2H or both, depending on whether the operation is read, write or update.

### Testing for physically realistic solution

This was a quick hack that tests whether the voltage at the soma of the first cell in each cell group is within a "reasonable" range after each internal time step of the `cell_group`. This can be disabled without much fuss (or replaced with a better sanity test)

https://github.com/eth-cscs/nestmc-proto/blob/97e17b187e68a3253fde6c668918605dea2f5d3b/src/cell_group.hpp#L93

bcumming:
## Proposed solution : Spike Detection

[...]

Issue #125, opened 2016-12-13
Fix `AtomicAdd` in CUDA


bcumming:
Before CUDA 8 the `AtomicAdd` intrinsic was only provided for 32-bit floats, and not 64-bit doubles.

CUDA 8 added support Pascal architecture, which supports atomic operations on doubles (at least for Tesla devices). As such it added support for `AtomicAdd(double *, double)`. Unfortunately, they made a mess so that user-supplied `AtomicAdd(double*, double)` that use atomic CAS conflict with the new signature, even when compiling for non Pascal targets.

Currently, this is implemented in a very messy way by "hiding" the user-supplied implementations inside a namespace, which is pretty ugly.

We should have
* clean implementation of CAS hack for pre-Pascal (at least, as clean as the aforementioned CUDA compiler bug allows us)
* specialization that uses the optimized `AtomicAdd` when targetting Pascal (i.e. `sm_60`).

PR #126, opened 2016-12-13
generalised stimulus

### LINKED ISSUE #104: Implement GPU stimulii [4]
user points estimate: 4
feature: #67 (GPU Support)

Implement stimulii as part of the mechanism framework, as described here:
https://github.com/eth-cscs/nestmc-proto/issues/87#issuecomment-262263930

Keep it simple: implement a hand written specialization of `mechanism` for each of the multicore and gpu back ends that injects current via the `nrn_current` interface used by all mechanisms.
* implement stimulii as mechanisms and update "mechanism catalogue interface for each backend [2]
* update `cell_group` and `fvm_multicore` initialization [1]
* update relevant unit tests and test [1]

bcumming:
feature: #67 (GPU Support)

Implement stimulii as part of the mechanism framework, as described here:
#87 (comment)

* A hand-written stimulus point process derived from `mechanisms::mechanism` was written for each back end. 
* The lowered `fvm_multicell` type explicitly constructs a stimulus mechanism if there are any stimuli attached to the its cells.
* This mechanism is added to the other mechanisms in the lowered cell, so that the update of current is performed in the current update loop (i.e. via the `nrn_current()`) method.

This isn't an ideal solution: we still have a hard coded stimulus type in the lowered cell, however the stimulus is now "in the right spot", and we can refine this better when we work on a better design for generalised mechanisms (i.e. when we have figured out what we going to do).

fixes #104.


src/backends/fvm_multicore.hpp
```diff
@@ -117,6 +119,8 @@ struct backend {
 
     using mechanism = mechanisms::mechanism_ptr<backend>;
 
+    using stimulus = mechanisms::multicore::stimulus<backend>;
```
bcumming: The namespaces are a bit of a mess here. To be fixed in "the big namespace rationanlization".

src/backends/stimulus_gpu.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <util/pprintf.hpp>
+
+namespace nest{
+namespace mc{
+namespace mechanisms {
+namespace gpu {
+
+namespace kernels {
+    __device__
+    inline double atomicAdd(double* address, double val) {
```
halfflat: issue #125 noted

src/fvm_multicell.hpp
```diff
@@ -170,12 +171,24 @@ class fvm_multicell {
     }
 
     /// return reference to the stimuli
-    stimulus_store_type& stimuli() {
-        return stimuli_;
+    util::optional<mechanism&> find_mechanism(const std::string& name) {
```
halfflat: From discussions earlier, I think it's best not to have this considered as part of the 'normal' API for a solver, but can certainly see the utility as a testing interface.

If you agree, I think it would be good to comment the method to indicate that it is exposing mechanims for testing purposes. Also, if we can get away with just the `const` version below, that would be even safer.

src/fvm_multicell.hpp
```diff
@@ -518,12 +529,39 @@ void fvm_multicell<Backend>::initialize(
...
+        //         parameters
+        // NOTE: the indexes and associated metadata (durations, delays,
+        //       amplitudes) have not been permuted to ascending cv index order,
+        //       as is the case with other point processes.
+        //       This is because the hard-coded stimulus mechanism makes no
+        //       optimizations that rely on this assumption (they are not
+        //       performance critical).
+        if (stim_index.size()) {
+            auto stim = new stimulus(
+                voltage_, current_, memory::make_const_view(stim_index));
```
halfflat: I don't think there is a guarantee that the `stim_index` will be sorted. Perhaps it would be better to generalize the `syn_mech_indices` machinery? The problem of tracking targets and making target handles is also analogous to the problem of tracking entries in the `c.stimuli()` and assigning the appropriate stimulus parameters.

tests/unit/test_fvm_multi.cpp
```diff
@@ -191,18 +194,47 @@ TEST(fvm_multi, stimulus)
     fvm_cell fvcell;
     fvcell.initialize(singleton_view(cell), detectors, targets, probes);
 
-    auto& stim = fvcell.stimuli();
-    EXPECT_EQ(stim.size(), 2u);
+    auto ref = fvcell.find_mechanism("stimulus");
+    if (ref) {
+        auto& stims = ref.get();
```
halfflat: Simpler just to fail early if `!ref`, i.e.
```
ASSERT_TRUE(ref) << "no stimuli retrieved ...";
```
and then no need for `if (ref) { ...`

src/fvm_multicell.hpp
```diff
@@ -170,12 +171,24 @@ class fvm_multicell {
...
-    stimulus_store_type& stimuli() {
-        return stimuli_;
+    util::optional<mechanism&> find_mechanism(const std::string& name) {
+        auto it = std::find_if(
+            std::begin(mechanisms_), std::end(mechanisms_),
+            [&name](mechanism& m) {return m->name()==name;});
+        if (it!=mechanisms_.end()) {
+            return util::optional<mechanism&>(*it);
+        }
+        return {};
```
halfflat: This is a bit unidiomatic for `optional` returns; consider
```
return it==mechanisms_.end()? util::nothing: util::just(*it);
```
or
```
if (it!=mechanisms_.end()) {
    return *it;
}
return util::nothing;
```
or
```
return util::provided(it!=mechanisms_.end()) & *it;
```

src/fvm_multicell.hpp
bcumming: How about having the comment, and making it `const` only?

halfflat: If it's compatible with the unit test, that would be great.

src/backends/stimulus_gpu.hpp
bcumming: yes! This needs to be fixed properly at the same time that we adress the issue in the code that is generated by modcc.

src/fvm_multicell.hpp
bcumming: Done.

src/fvm_multicell.hpp
bcumming: I went with
```
return it==mechanisms_.end() ? util::nothing: util::just(*it);
```

src/fvm_multicell.hpp
bcumming: There really is no need to generalize the `syn_mech_indices` machinery for now. Later, if more use cases pop up, we could use it. I leave this the way that it is. I also think that the comment about sorting of stimulus indices is good where it is.

bcumming:
Ready for another look @halfflat.

PR #127, opened 2016-12-13
Fix modcc precedence parsing bug

### LINKED ISSUE #94: Bug: modcc: incorrect parsing of negative terms in arithmetic expression

halfflat:
* Modify `parse_expression` to take a controlling (parent) precedence.
* `parse_expression` folds left over sequences of sub-expressions with decreasing operator precedence (accumulates in `lhs`).
* Use recursion rather than accumulator for left fold in `parse_binop` to simplify code logic.
* Extend parser unit test to cover more complicated, multi-level expression.
* Remove (now) redundant parenthesis from derivative check block in kinetic rewriter test.

Fixes #94

### DIFF
--- a/modcc/parser.cpp
+++ b/modcc/parser.cpp
@@ -1095,25 +1095,32 @@ expression_ptr Parser::parse_conserve_expression() {
     return make_expression<ConserveExpression>(here, std::move(lhs), std::move(rhs));
 }
 
-expression_ptr Parser::parse_expression() {
+expression_ptr Parser::parse_expression(int prec) {
     auto lhs = parse_unaryop();
+    if(lhs==nullptr) return nullptr;
 
-    if(lhs==nullptr) { // error
-        return nullptr;
-    }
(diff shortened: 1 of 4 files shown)


modcc/parser.cpp
```diff
@@ -1215,44 +1222,35 @@ expression_ptr Parser::parse_integer() {
 }
 
 expression_ptr Parser::parse_binop(expression_ptr&& lhs, Token op_left) {
```
bcumming: The `Token`type isn't trivially copyable because `std::string`. It is being passed by value here: would passing by const reference also work From a cursory glance at the code, it looks like it would.

PR #128, opened 2016-12-14
Symbolic gaussian elimination prototype

halfflat:
Adds a stand-alone prototype application `symge-demo` with source under `proto/symge`.
Addresses issue #99.

`symge-demo` depends upon util/optional.hpp but is otherwise stand-alone.

Demo application:
* Validates symbolic GE across a range of strictly positive definite matrices (by default, up to 10x10).
* Reports (with -v) number of required extra single-assignment symbols and numeric operations.
* With -d, print detailed information on test matrices and symbolic computations to standard error.

Incorporates comments from PR #118 review:
* Rename `mrow` to `row`, `symmrow` to `sym_row`, `symmatrix` to `sym_matrix`, `sepval_t` to `sepval_proxy`.
* Replace entry alias with struct with meaningful field names, underscore data private variable.
* Use `impl` namespace for templated symbolic helper classes.
* Document identifier generator code.
* Simplify Gaussian elimination to pivot only on diagonal elements.
* Add simple (greedy) cost estimator to inform pivot choices.
* Expose more of the parameters (matrix size, number of checks etc.) through command line interface.
* Bring diagonal elements close to one to reduce risk of overflow in numeric checks on large matrices.


halfflat:
Components of this prototype have been incorporated in the kinetic schemes PR #145, but that PR does not retain the more extensive numeric validation code nor does it perform the operation or fill in accounting that is reported in the demo for the evaluation of various pivot choice schemes.

I propose that we keep this demo application for its utility and code base when we revisit this problem, particularly for the investigation of:
* the numerical behaviour for larger matrices (limited dynamic range will likely force periodic normalization of temporaries);
* better pivot selection schemes for reduced fill-in;
* symbolic computation that will arise in higher order implicit schemes.

PR #136, opened 2016-12-22
Fix a few compilation problems that occur when using c++14 mode

biddisco:
make_unique has been moved into nest:mc::util namespace to avoid a conflict with std::make_unique
util:: cbegin and cend clash with std:: equivalents, so we call them by namespace explicitly.

### DIFF
--- a/modcc/functionexpander.cpp
+++ b/modcc/functionexpander.cpp
@@ -5,6 +5,8 @@
 #include "functionexpander.hpp"
 #include "modccutil.hpp"
 
+using namespace nest::mc;
+
 expression_ptr insert_unique_local_assignment(call_list_type& stmts, Expression* e) {
     auto exprs = make_unique_local_assign(e->scope(), e);
     stmts.push_front(std::move(exprs.local_decl));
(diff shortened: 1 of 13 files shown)


miniapp/miniapp.cpp
```diff
@@ -224,7 +224,7 @@ std::unique_ptr<sample_trace_type> make_trace(cell_member_type probe_id, probe_s
     }
     name += probe.location.segment? "dend" : "soma";
 
-    return util::make_unique<sample_trace_type>(probe_id, name, units);
+    return nest::mc::util::make_unique<sample_trace_type>(probe_id, name, units);
```
bcumming: It isn't necessary to use the full namespace `nest::mc::` is it? This is very verbose.

bcumming:
I think that these changes are good, because they disambiguate some things, and will be useful when we finally move to C++14. I would like to see the short form, i.e. `util::xxx` instead of `nest::mc::util::xxx` used wherever possible however.

biddisco:
I did a search/replace and I'll put things back to just util:: namespace and push a new version.

The reason or c++14 is simply that in my CMake Superproject, I'm using some c++ 14 features in hpx and to compile everything in one go, I am using the same flags everywhere. I would not put a dependency on c++14 into nestmc.

biddisco:
I've force pushed a new version without the nest::mc::util

bcumming:
looks good to me.

Issue #134, opened 2016-12-21
Improve nestmc cmake variable names


biddisco:
It is fairly common practice to build several cmake based projects in one go by creating a top level CMakeLists.txt and populating it with 
  add_subdir(project1)
  add_subdir(project2)
etc. 

I use this pattern frequently when project2 depends on project1 and I am developing both projects side by side. Changes in p1 (example HPX) propagate rebuilds to p2 (nestmc+hpx extensions) etc and you always have the same compiler flags/settings for both.
p2 will find_package(p1) and providing p1 has configured a p1-config.cmake file in the build dir, everything works very smoothly. (@halfflat may be aware that I convinced the BBP team to ditch buildyard in favour of this simplified cmake based approach and it has paid off well).

unfortunately, nestmc uses cmake options such as `WITH_STUFF` in a number of places (`WITH_CUDA` for example) and when you have 5-6 projects built at the same time it becomes very hard to remember which option comes from which project and it is convenient/helpful to prefix all user facing options with `HPX_WITH_STUFF`, `DCA_WITH_STUFF`, and even `NMC_WITH_STUFF` so it becomes easy to find options for a particular subproject. @bcumming suggested the prefix `NMC` for nestmc options.

If it's ok with the nestmc team I will undertake to clean up the NMC cmake options.

Issue #130, opened 2016-12-18
master branch does not build


biddisco:
OSX Xcode 8, clang llvm, default options in cmake

```
/Users/biddisco/src/nestmc/mechanisms/multicore/pas.hpp:115:17: error: no member named 'vec_area_' in
      'nest::mc::mechanisms::mechanism<nest::mc::multicore::backend>'
    using base::vec_area_;
          ~~~~~~^
In file included from /Users/biddisco/src/nestmc/src/backends/fvm_multicore.cpp:1:
In file included from /Users/biddisco/src/nestmc/src/backends/fvm_multicore.hpp:6:
/Users/biddisco/src/nestmc/src/mechanism.hpp:80:14: note: in instantiation of template class
      'nest::mc::mechanisms::pas::mechanism_pas<nest::mc::multicore::backend>' requested here
    typename M::view  vec_v,
             ^
In file included from /Users/biddisco/src/nestmc/src/backends/fvm_multicore.cpp:1:
/Users/biddisco/src/nestmc/src/backends/fvm_multicore.hpp:148:16: note: while substituting explicitly-specified template arguments into function template
      'make_mechanism'
        return mechanisms::make_mechanism<Mech<backend>>
               ^
/Users/biddisco/src/nestmc/src/backends/fvm_multicore.cpp:14:31: note: in instantiation of function template specialization
      'nest::mc::multicore::backend::maker<mechanisms::pas::mechanism_pas>' requested here
    { std::string("pas"),     maker<mechanisms::pas::mechanism_pas> },
                              ^
In file included from /Users/biddisco/src/nestmc/src/backends/fvm_multicore.cpp:1:
/Users/biddisco/src/nestmc/src/backends/fvm_multicore.hpp:148:16: error: no matching function for call to 'make_mechanism'
        return mechanisms::make_mechanism<Mech<backend>>
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/biddisco/src/nestmc/src/backends/fvm_multicore.cpp:14:31: note: in instantiation of function template specialization
      'nest::mc::multicore::backend::maker<mechanisms::pas::mechanism_pas>' requested here
    { std::string("pas"),     maker<mechanisms::pas::mechanism_pas> },
                              ^
In file included from /Users/biddisco/src/nestmc/src/backends/fvm_multicore.cpp:1:
[...]

PR #137, opened 2016-12-22
CMake fixes

biddisco:
Here are some cmake related fixes that address #130 (rerunning modcc) and #134 (improving cmake names) as well as a few other cleanups and tweaks. 

I expect some pushback on the changing of #ifdef var names!

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -36,7 +36,7 @@ import ycm_core
 # CHANGE THIS LIST OF FLAGS. YES, THIS IS THE DROID YOU HAVE BEEN LOOKING FOR.
 flags = [
     '-DNDEBUG',
-    '-DWITH_TBB',
+    '-DNMC_HAVE_TBB',
     '-std=c++11',
     '-x',
     'c++',
@@ -54,7 +54,7 @@ flags = [
     'modcc',
     '-I',
     '/cm/shared/apps/cuda/8.0.44/include',
-    '-DWITH_CUDA'
+    '-DNMC_HAVE_CUDA'
 ]
 
 # Set this to the absolute path to the folder (NOT the file!) containing the
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,6 +31,12 @@ if(WITH_TRACE)
     add_definitions("-DWITH_TRACE")
 endif()
 
+#----------------------------------------------------------
+# Option to disable auto running of modcc compiler
+#----------------------------------------------------------
+option(NMC_AUTO_RUN_MODCC_ON_CHANGES 
+  "Rerun modcc compiler whenever *.mod file or modcc compiler change" ON)
+
 # list of libraries to be linked against targets
 set(EXTERNAL_LIBRARIES "")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ enable_language(CXX)
 set(SAVED_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 
 # compilation flags
-set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
+set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake")
 include("CompilerOptions")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXXOPT_DEBUG} ${CXXOPT_CXX11} ${CXXOPT_PTHREAD} ${CXXOPT_WALL}")
 
@@ -147,7 +147,7 @@ endif()
 # Validation data generation
 
 # destination directory for generated data
-set(VALIDATION_DATA_DIR "${CMAKE_SOURCE_DIR}/validation/data" CACHE PATH "location of generated validation data")
+set(VALIDATION_DATA_DIR "${PROJECT_SOURCE_DIR}/validation/data" CACHE PATH "location of generated validation data")
 
 # Whether to build validation data at all
 set(BUILD_VALIDATION_DATA ON CACHE BOOL "generate validation data")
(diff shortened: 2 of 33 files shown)


.ycm_extra_conf.py
```diff
@@ -36,7 +36,7 @@
 # CHANGE THIS LIST OF FLAGS. YES, THIS IS THE DROID YOU HAVE BEEN LOOKING FOR.
 flags = [
     '-DNDEBUG',
-    '-DWITH_TBB',
+    '-DNMC_HAVE_TBB',
```
bcumming: Why have you changed the `WITH` to `HAVE`?

CMakeLists.txt
```diff
@@ -106,83 +145,94 @@ if(WITH_MPI)
...
 # Validation data generation
-
+#----------------------------------------------------------
 # destination directory for generated data
-set(VALIDATION_DATA_DIR "${CMAKE_SOURCE_DIR}/validation/data" CACHE PATH "location of generated validation data")
+set(NMC_VALIDATION_DATA_DIR "${PROJECT_SOURCE_DIR}/validation/data" CACHE PATH "location of generated validation data")
 
 # Whether to build validation data at all
-set(BUILD_VALIDATION_DATA ON CACHE BOOL "generate validation data")
+set(NMC_BUILD_VALIDATION_DATA ON CACHE BOOL "generate validation data")
```
bcumming: I see that you have changed some of these variables to `option` like:
```
option(NMC_WITH_PROFILING "use built-in profiling of miniapp" OFF)
```
Why not for `NMC_BUILD_VALIDATION_DATA`? You have a good reason to be sure: I am looking for some CMake education.

src/CMakeLists.txt
```diff
@@ -16,14 +16,17 @@ set(CUDA_SOURCES
...
-if(${WITH_MPI})
+if(NMC_WITH_MPI)
     set(BASE_SOURCES ${BASE_SOURCES} communication/mpi.cpp)
 endif()
 
 add_library(nestmc ${BASE_SOURCES} ${HEADERS})
 
-add_dependencies(nestmc build_all_mods)
-if(WITH_CUDA)
+if (NMC_AUTO_RUN_MODCC_ON_CHANGES)
```
bcumming: Will this work the first time we check the code out and make? The first time we build the mechanism files have to be generated from the `mod` files by modcc.

src/backends/fvm.hpp
```diff
@@ -2,6 +2,6 @@
 
 #include "fvm_multicore.hpp"
 
-#ifdef WITH_CUDA
+#ifdef NMC_HAVE_CUDA
```
bcumming: Again, how about `NMC_WITH_CUDA`, otherwise we will be getting confused and wondering when to use `HAVE` and when to use `WITH`.

bcumming:
Also, I assume that you have tested this with the different threading backends, MPI and CUDA? :)

.ycm_extra_conf.py
biddisco: The explanation goes like this:
A user says I want Option X (say cuda or openMP) so in CMake they say NMC_WITH_CUDA or NMC_WITH_OPENMP or something of that kind. (ignoring the fact that you have a threading model var for that). It might be that OpenMP is not found (OSX) but it isn't fatal, so you don't mind. Elsewhere in the code there will be checks for if(openmp) etc, which should of course test negative, so a common pattern is to have user facing options WITH_XXX and the actual vars that code checks to be HAVE_XXX so that if(WITH_openmp) passes add_definition(HAVE_openmp) is added. 
User facing options use WITH, actual ifdefs in the code use HAVE to perform the checks (and if cmake needs it, it can also have HAVE_option vars set for various checks).
The commits themselves have quite long log messages with some explanations too.

If you don't like using WITH and HAVE, then a search/replace can change all the _HAVE_back to _WITH_ in a jiffy - (but I'd like to convince you to keep it)

CMakeLists.txt
biddisco: I forgot a couple of options. I'll add them too.

src/CMakeLists.txt
biddisco: good catch. I'd better reconsider that and check if it works (you are probably right and it won't)

src/backends/fvm.hpp
biddisco: WITH for user options, HAVE in the code. not very difficult to remember, but admittedly a pita for you to change.

.ycm_extra_conf.py
bcumming: I see your reasoning. However, currently, these two naming schemes are also used in different "namespaces". You are using `NMC_HAVE_X` for C++ preprocessor variables, and `NMC_WITH_X` for CMake variables. This is because we have a policy of having a hard error if the user requests a feature, say CUDA or OpenMP, instead of "silently" building with the feature disabled. This saves us from lots of silent "MPI isn't working when I compile the MPI target" errors.

bcumming: Yeah, OK, I am convinced after thinking some more. Carry on.

bcumming: We can still have hard errors for most features like CUDA etc, but there will be cases when this might not be the case.

src/CMakeLists.txt
biddisco: Well actually thinking about it, the default value for `AUTO_GENERATE` is ON, so on first checkout, the mod files will be created. Only when an 'advanced' user disables it (presumably on purpose) will the generation be turned off, so it should not be a problem.

biddisco: I just checked and everything works fine on first checkout.

CMakeLists.txt
biddisco: I've updated this PR with the missing OPTIONs so they are consistent.

bcumming:
A question for a later date is whether we should change all of the internal `#define` variables from `NMC_HAVE_XXX` to just `NMC_XXX`. Would be less verbose, and now that we have the `NMC_` prefix, the chances of collisions are low.

biddisco:
You could do, it's a very simple sed operation to change them from _HAVE_ to anything.

Issue #98, opened 2016-11-25
Reorganise source files and namespaces in nestmc project


halfflat:
Point estimate: 4
Feature: #65 

The current state of the code base has some inconsistencies with regards to paths to header files, and the implicit dependencies across different sub-sections of the code are not clear.

The goal is to:
* clarify or adjust the namespaces used for various sub-components,
* split nestmc-specific library source from more generic utility source,
* adopt a sub-directory layout and naming that is consistent across sub-components and which (ideally) reflects namespace names,
* add documentation (e.g. on project wiki) that describes the sub-component dependencies,
* fixes up all the include paths to suit.

The first part of the task will be to identify the sub-components and propose a corresponding organization of the code.

halfflat:
Drawing a line under current work, and leaving issues #538 and #539 to fend for themselves in the backlog.

PR #138, opened 2017-01-12
Compilation fix

vkarak:
Used to fail with Clang 3.9 and GCC 5.3.

### DIFF
--- a/tests/modcc/test_kinetic_rewriter.cpp
+++ b/tests/modcc/test_kinetic_rewriter.cpp
@@ -9,6 +9,7 @@
 #include "expr_expand.hpp"
 #include "test.hpp"
 
+using namespace nest::mc;
 
 stmt_list_type& proc_statements(Expression *e) {
     if (!e || !e->is_symbol() || ! e->is_symbol()->is_procedure()) {
@@ -95,4 +96,3 @@ TEST(KineticRewriter, equiv) {
     EXPECT_EQ(deriv_map["b'"], kin_map["b'"]);
     EXPECT_EQ(deriv_map["c'"], kin_map["c'"]);
 }
-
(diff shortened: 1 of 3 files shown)


halfflat:
Hmm. It seems we're only using `nest::mc` to define within `modccutil.hpp` a `make_unique` function. There are no other dependencies on this namespace, or on the code elsewhere in the source tree.

I'd really prefer to either make `modcc` use the external utility code, if we tidy that up appropriately, or else remain self-contained. If it is self-contained, there is no need to make a three-level deep namespace for a single utility method, and it makes the code look like it depends on stuff that it really doesn't.

vkarak:
I would vote for using the external utility code. That's one of the reasons we merged the two codes. I'll prepare another PR for that.

halfflat:
I think if we do this, we should first address the namespace/source separation issues (as pertaining to the `util::` namespace) described in #98 — either way, I think it would be best to make an individual issue/task for this first in the interest of maintaining a clear association between PRs and tasks which include a technical outline of the proposed changes.

vkarak:
I've opened #139 for refactoring the `modccutil.hpp`. Feel free to enhance it accordingly or move it to a project task.

bcumming:
Personally, I am happy to merge the code so that modcc uses the main util code.
And I am also happy to see fixes like this for now. This is the prototype after all, and we are well aware that there is a "big refactor" around the corner.

halfflat:
But we should keep things as simple as possible before this refactor. With the work I've been doing on `modcc` so far, I've been taking pains to keep it self-contained, so that the later refactoring work will be easier.

vkarak:
Let's keep then all these refactoring issues for (near) future reference. I will go on with the vectorisation backend, which is more urgent for the moment.

bcumming:
Agreed. I merged this PR because it was very simple.
Now, I am getting out of the bike shed (this PR is proving the rule about small PR's generating the most discussion per line).

PR #140, opened 2017-01-18
Bgq

apeyser:
Adapt MPI cmake section to turn off the mpi package for BGQ.
Make profstats.py a module, so that it can be called from within other projects.
Added -z, --profile_only_zero flag for turning off all profile output files except rank zero so that very large node number simulations are practical.


apeyser:
Went with just the option.

PR #144, opened 2017-01-27
Threading pool

apeyser:
Here's a threading pool that implements a subset of tbb used in NestMC.
Goal is to remove omp.


apeyser:
Cleaned up to meet coding standards and pushed

src/threading/cthread.cpp
```diff
@@ -0,0 +1,191 @@
...
+// queue is cleared
+void task_pool::wait(task_group* g) {
+    run_tasks_while(g);
+}
+
+class NMC_NUM_THREADS_ERROR {};
+static auto NMC_NUM_THREADS_VAR{"NMC_NUM_THREADS_VAR"};
+static auto NMC_NUM_THREADS{"NMC_NUM_THREADS"};
+static auto OMP_NUM_THREADS{"OMP_NUM_THREADS"};
+
```
halfflat: I admit, I find this more confusing than just using the C string constants directly below in the calls to `std::getenv`.

src/threading/cthread.cpp
```diff
@@ -0,0 +1,191 @@
...
+// call on main thread
+// uses this thread to run tasks
+// and waits until the entire task
+// queue is cleared
+void task_pool::wait(task_group* g) {
+    run_tasks_while(g);
+}
+
+class NMC_NUM_THREADS_ERROR {};
```
halfflat: Given that an exception of this class will never be caught, what do you think about just calling `std::terminate()` directly in `global_get_num_threads()` below instead? An alternative would be to just return 1 as a fallback, and allow the application to check if there are fewer than expected threads running.

src/threading/cthread.cpp
```diff
@@ -0,0 +1,191 @@
...
+        }
+        if (_quit || ! g->in_flight) {
+            return;
+        }
+
+        run_task run{*this, lck};
+        run.tsk.first();
+    }
+}
+
```
halfflat: Would you consider combining these two methods, in the interests of reducing duplicated code?
e.g. something like
```
void task_pool::run_tasks_while(task_group* g=nullptr) {
    // ...
    while (!finished(g) && _ttasks.empty()) {
        // ...
}

void task_pool::finished(task_group* g) {
    return _quit || (g && !g->in_flight);
}
```

CMakeLists.txt
```diff
@@ -69,6 +69,15 @@ elseif(NMC_THREADING_MODEL MATCHES "omp")
     add_definitions(-DNMC_HAVE_OMP)
     set(NMC_HAVE_OMP TRUE)
 
+elseif(NMC_THREADING_MODEL MATCHES "cthread")
+    find_package(Threads REQUIRED)
+    add_definitions(-DNMC_HAVE_CTHREAD)
+    set(NMC_HAVE_CTHREAD TRUE)
+    
+    if(CMAKE_USE_PTHREADS_INIT)
+      set(CMAKE_C_FLAGS ${CMAKE_C_FLAGS} "-pthread")
```
apeyser: Will do on the first part. I'm not sure on the rest -- I haven't found any clear documentation on how c++ threads "should" be handled with cmake, rather than c pthreads. Just the usuals rumors on stack overflow.

miniapp/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ else()
     add_executable(miniapp.exe ${MINIAPP_SOURCES} ${HEADERS})
 endif()
 
+target_link_libraries(miniapp.exe LINK_PUBLIC ${CMAKE_THREAD_LIBS_INIT} )
```
apeyser: I'll see whether this works throughout the tree, rather than having these target links everywhere

src/threading/cthread.hpp
```diff
@@ -0,0 +1,303 @@
...
+using task_queue = std::deque<task>;
+
+using thread_list = std::vector<std::thread>;
+using thread_map = std::unordered_map<std::thread::id, std::size_t>;
+
+class task_pool {
+private:
+    // lock and signal on task availability change
+    // this is the crucial bit
+    mutex _tasks_mutex;
```
apeyser: Ach, I misread the coding standard. Will do.

apeyser: Even though it hurts my eyes.

src/threading/cthread.hpp
```diff
@@ -0,0 +1,303 @@
...
+    // lock the parallel_vector to update
+    impl::mutex mutex;
+
+    // call a function of type X f() in a lock
+    template<typename F>
+    auto critical(F f) -> decltype(f()) {
+        impl::lock lock{mutex};
+        return f();
+    }
+
```
apeyser: It could be a bit more general, but it's only used in one place, and lambdas anyway are an easy way to get around the void parameter.

src/threading/cthread.hpp
```diff
@@ -0,0 +1,303 @@
...
+    const_iterator end()   const { return data_.end(); }
+
+    const_iterator cbegin() const { return data_.cbegin(); }
+    const_iterator cend()   const { return data_.cend(); }
+
+    void push_back (value_type&& val) {
+        critical([&] {
+            data_.push_back(std::move(val));
+        });
+    }
```
apeyser: ✅

src/threading/cthread.hpp
```diff
@@ -0,0 +1,303 @@
...
+
+    static inline double toc(time_point t) {
+        return std::chrono::duration<double>{tic() - t}.count();
+    }
+
+    static inline double difference(time_point b, time_point e) {
+        return std::chrono::duration<double>{e-b}.count();
+    }
+};
```
apeyser: ✅

src/threading/cthread.hpp
```diff
@@ -0,0 +1,303 @@
...
+
+} // threading
+} // mc
+} // nest
+
+// paralele stable sort uses threading
+#include "cthread_parallel_stable_sort.h"
+
```
apeyser: Suggestion? It is confusing, but it is dependent on the definitions in cthread.hpp, and the sorts depend on it. So the other way is to split all of this up into cthread_impl, cthread_sort and cthread_parallel_stable_sort, which maybe what should be done... but on the other hand, it adds a different type of complexity.

tests/unit/CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ add_definitions("-DDATADIR=\"${PROJECT_SOURCE_DIR}/data\"")
 set(TARGETS test.exe)
 
 add_executable(test.exe ${TEST_SOURCES} ${HEADERS})
+target_link_libraries( test.exe LINK_PUBLIC ${CMAKE_THREAD_LIBS_INIT} )
```
apeyser: ✅

src/threading/cthread.cpp
apeyser: Here is why I was complaining about RAII. I've refactored this a number of times trying to find a decent way of doing this without spreading code over the entire module.

apeyser: Why nullptr? This can't be run without a task_group.

halfflat: It would make (if I understand the code correctly) `run_task_loops()` equivalent to `run_tasks_while(nullptr)`; they are both doing the same thing modulo a different exit condition, which can be wrapped up in a method like `finished()`.

src/threading/cthread.cpp
halfflat: @bcumming has some code already written for asking the OS for hardware thread counts, which could also be incorporated for a possible fallback.

bcumming: The code is here.
https://github.com/bcumming/affinity
For some unknown reason I put a GPL license on it... that can be changed to BSD no worries.

apeyser: std::terminate + an error message? That's why I threw a class to terminate -- at least the words NMC_NUM_THREADS_ERROR comes out to the screen, I believe.

src/threading/cthread.cpp
apeyser: Yeah, I agree. That's how it was and then I started playing with it to see what looked better.

src/threading/cthread.hpp
apeyser: Split into cthread_impl.hpp, cthread_sort.hpp which includes the parallel sort

PR #148, opened 2017-02-01
Extra profiling

apeyser:
Add some finer grained profiling to track exactly what time is spent inside of mpi calls.


bcumming:
Looks good.

PR #149, opened 2017-02-01
dependencies in docs/model/makefile to building images

apeyser:
Build images.dir when building report.pdf
Add outputs to .gitignore


halfflat:
Hi!

I submitted a PR to your branch for this — the existing makefile was poorly behaved (unnecessary rebuilds etc.) and adding the images didn't really help. If you're happy to have `latexmk` in there, then I think this PR (on your branch) will be more robust.

PR #151, opened 2017-02-09
Add dry run feature

bcumming:
Add a dry run mode, inspired by the dry run mode implemented in NEST. A dry run 
of a model simulates running a large distributed model by running only the work
of one of the ranks, with artificial spike input from the other "dummy" ranks.                           

This is implemented as a new global communication back end, `dryrun_global_policy`,
the implementation of which is straightforward:
  - a new implementation of `gather_spikes` that takes the local spikes and
    replicates them `n` times where `n` is the total number of simulated
    ranks.
  - the `global_policy::size()` method returns the number of ranks in the
    simulated run
  - the new back end has to store some state that records the number
    of simulated ranks and cells per rank, which are set using the new
    `global_policy::set_sizes()` method

Some CMake modificatins were required:
  - make the selection of the global communication backend have the same
    interface as that for selecting the threading back end.
  - small improvements to the selection of the threading back end to
    make the `cthread` option visible in ccmake, and have consistent
    CMake variable naming.

Command line options were also extended:
  - a `--dry-run-size` or `-D` option can be used to supple the number
    of dry run ranks on the command line.
  - the miniapp driver was updated to set the dry run size and cell
    count via the new `global_policy::set_sizes()` interface.

### DIFF
--- a/miniapp/miniapp.cpp
+++ b/miniapp/miniapp.cpp
@@ -51,15 +51,26 @@ int main(int argc, char** argv) {
 
     try {
         std::cout << util::mask_stream(global_policy::id()==0);
-        banner();
-
         // read parameters
         io::cl_options options = io::read_options(argc, argv, global_policy::id()==0);
-        std::cout << options << "\n";
-        std::cout << "\n";
-        std::cout << ":: simulation to " << options.tfinal << " ms in "
-                  << std::ceil(options.tfinal / options.dt) << " steps of "
-                  << options.dt << " ms" << std::endl;
+
+        // If compiled in dry run mode we have to set up the dry run
+        // communicator to simulate the number of ranks that may have been set
+        // as a command line parameter (if not, it is 1 rank by default)
+        if (global_policy::kind() == communication::global_policy_kind::dryrun) {
+            // Dry run mode requires that each rank has the same number of cells.
+            // Here we increase the total number of cells if required to ensure
+            // that this condition is satisfied.
+            auto cells_per_rank = options.cells/options.dry_run_ranks;
+            if (options.cells % options.dry_run_ranks) {
+                ++cells_per_rank;
+                options.cells = cells_per_rank*options.dry_run_ranks;
+            }
+
+            global_policy::set_sizes(options.dry_run_ranks, cells_per_rank);
+        }
+
+        banner();
 
         // determine what to attach probes to
         probe_distribution pdist;
@@ -180,7 +191,7 @@ void banner() {
     std::cout << "====================\n";
     std::cout << "  starting miniapp\n";
     std::cout << "  - " << threading::description() << " threading support\n";
-    std::cout << "  - communication policy: " << global_policy::name() << "\n";
+    std::cout << "  - communication policy: " << std::to_string(global_policy::kind()) << " (" << global_policy::size() << ")\n";
 #ifdef NMC_HAVE_CUDA
     std::cout << "  - gpu support: on\n";
 #else
--- a/src/communication/dryrun_global_policy.hpp
+++ b/src/communication/dryrun_global_policy.hpp
@@ -36,7 +36,9 @@ struct dryrun_global_policy {
                 auto& s = local_spikes[i];
                 // the new global spike is the same as the local spike, with
                 // its source index shifted to the dummy domain
-                global_spikes[first_spike+i] = {s.source+first_cell, s.time};
+                auto source = s.source;
+                source.gid += first_cell;
+                global_spikes[first_spike+i] = {source, s.time};
             }
         }
 
@@ -73,7 +75,8 @@ struct dryrun_global_policy {
 
     static void setup(int& argc, char**& argv) {}
     static void teardown() {}
-    static const char* name() { return "dry run"; }
+
+    static global_policy_kind kind() { return global_policy_kind::dryrun; };
 };
 
 using global_policy = dryrun_global_policy;
--- a/src/communication/dryrun_global_policy.hpp
+++ b/src/communication/dryrun_global_policy.hpp
@@ -1,0 +1,83 @@
+#pragma once
+
+#include <cstdint>
+#include <type_traits>
+#include <vector>
+
+#include <communication/gathered_vector.hpp>
+#include <util/span.hpp>
+#include <spike.hpp>
+
+namespace nest {
+namespace mc {
+namespace communication {
+
+extern int dryrun_num_local_cells;
+extern int dryrun_communicator_size;
+
+struct dryrun_global_policy {
+    template <typename Spike>
+    static gathered_vector<Spike>
+    gather_spikes(const std::vector<Spike>& local_spikes) {
+        using util::make_span;
+        using count_type = typename gathered_vector<Spike>::count_type;
+
+        // Build the global spike list by replicating the local spikes for each
+        // "dummy" domain.
+        auto num_spikes_local  = local_spikes.size();
+        auto num_spikes_global = size()*num_spikes_local;
+        std::vector<Spike> global_spikes(num_spikes_global);
+        std::vector<count_type> partition(size()+1);
+
+        for (auto rank: make_span(0u, size())) {
+            auto first_cell = rank*dryrun_num_local_cells;
+            auto first_spike = rank*num_spikes_local;
+            for (auto i: make_span(0, num_spikes_local)) {
+                auto& s = local_spikes[i];
+                // the new global spike is the same as the local spike, with
+                // its source index shifted to the dummy domain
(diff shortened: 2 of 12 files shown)


miniapp/miniapp.cpp
```diff
@@ -51,15 +51,26 @@ int main(int argc, char** argv) {
...
+
+        // If compiled in dry run mode we have to set up the dry run
+        // communicator to simulate the number of ranks that may have been set
+        // as a command line parameter (if not, it is 1 rank by default)
+        if (global_policy::kind() == communication::global_policy_kind::dryrun) {
+            // Dry run mode requires that each rank has the same number of cells.
+            // Here we increase the total number of cells if required to ensure
+            // that this condition is satisfied.
+            auto cells_per_rank = options.cells/options.dry_run_ranks;
+            if (options.cells % options.dry_run_ranks) {
```
apeyser: Maybe I should make on issue of this for further improvement? I'd be interested also in testing imbalances, I think.

src/communication/dryrun_global_policy.hpp
```diff
@@ -0,0 +1,86 @@
...
+        // "dummy" domain.
+        auto num_spikes_local  = local_spikes.size();
+        auto num_spikes_global = size()*num_spikes_local;
+        std::vector<Spike> global_spikes(num_spikes_global);
+        std::vector<count_type> partition(size()+1);
+
+        for (auto rank: make_span(0u, size())) {
+            auto first_cell = rank*dryrun_num_local_cells;
+            auto first_spike = rank*num_spikes_local;
+            for (auto i: make_span(0, num_spikes_local)) {
```
apeyser: Shouldn't this b const auto&& ? Unless you know it collapses to a primitive -- and even then, it should be general.

apeyser: I see above --- looks like it is a primitive.

src/communication/dryrun_global_policy.hpp
```diff
@@ -0,0 +1,86 @@
...
+    static int size() {
+        return dryrun_communicator_size;
+    }
+
+    static void set_sizes(int comm_size, int num_local_cells) {
+        dryrun_communicator_size = comm_size;
+        dryrun_num_local_cells = num_local_cells;
+    }
+
+    template <typename T>
```
apeyser: Some of the following make no sense to me --- I don't know the libraries, so I don't know why min, max should be identity functions.

src/communication/global_policy.hpp
```diff
@@ -1,21 +1,37 @@
 #pragma once
 
-#ifdef NMC_HAVE_MPI
-    #include "communication/mpi_global_policy.hpp"
+#include <string>
+
+namespace nest { namespace mc { namespace communication {
+    enum class global_policy_kind {serial, mpi, dryrun};
```
apeyser: I'd probably go here with a parallel: `auto global_policy_kind {"serial", "mpi", "dryrun"}`
and then make the `to_string` just an array deref.

src/communication/mpi_global_policy.hpp
```diff
@@ -44,14 +51,6 @@ struct mpi_global_policy {
         return nest::mc::mpi::reduce(value, MPI_SUM);
     }
 
-    template <
```
apeyser: Why is this going away? Just confirm that it's supposed to be part of the set.

src/model.hpp
```diff
@@ -159,7 +159,7 @@ class model {
             // events that must be delivered at the start of the next
             // integration period at the latest.
             auto exchange = [&] () {
-                PE("stepping", "communciation");
+                PE("stepping", "communication");
```
apeyser: Arrrgh -- must now fix my scripts! I hadn't fixed it for that reason.
I guess it's small enough to be part of this patch set.

miniapp/miniapp.cpp
bcumming: There would not be much imbalance in this case, if we assume that there are a reasonable number of cells per node (i.e. at least a hundred), because this only increases the number of cells on some of the nodes by 1 to get even load balance.

src/communication/dryrun_global_policy.hpp
bcumming: it should be `const auto&`, I will fix it.

src/communication/dryrun_global_policy.hpp
bcumming: We have to find min and max over all "domains" at some points (e.g. when finding the minimal delay of the network). 
The idea here is that all domains are the same, so that a collective call to min or max should be passed the same argument by all ranks, hence the identity.
Finding the `min_delay` is actually the only use for these functions currently, and the assumptions holds for that use case.

src/communication/global_policy.hpp
bcumming: the `global_policy_kind` is used in the miniapp to test if the application is running in dry run mode at run time. If we used `const char*`, we would have to use `strcmp` to make this test.

src/communication/mpi_global_policy.hpp
bcumming: It is dead code: it was replaced with something cleaner a while back and we forgot to remove it at the time.
I took the opportunity to delete it now.

src/model.hpp
bcumming: I realised that you must have worked around it, because I was doing the same in my own scripts! But it had to get fixed some time.

miniapp/miniapp.cpp
bcumming: One way to mimic imbalances is to add performance models into the spike gather part of the communicator. That is where all of the MPI ranks synchronize in our current event-driven model.
However, the simplicity of the dry run implementation comes from the assumption that all ranks have the same number of cells.

src/communication/dryrun_global_policy.hpp
```diff
@@ -0,0 +1,86 @@
...
+    gather_spikes(const std::vector<Spike>& local_spikes) {
+        using util::make_span;
+        using count_type = typename gathered_vector<Spike>::count_type;
+
+        // Build the global spike list by replicating the local spikes for each
+        // "dummy" domain.
+        auto num_spikes_local  = local_spikes.size();
+        auto num_spikes_global = size()*num_spikes_local;
+        std::vector<Spike> global_spikes(num_spikes_global);
+        std::vector<count_type> partition(size()+1);
```
apeyser: It looks to me like the partition isn't being set: It needs `partition[rank] = first_spike` and then at the end `partition.back() = num_spikes_global`.

src/communication/dryrun_global_policy.hpp
bcumming: I simplified this a bit, so there is no need for the `const` variable.

src/communication/dryrun_global_policy.hpp
bcumming: fixed.

src/communication/dryrun_global_policy.hpp
```diff
@@ -24,24 +24,25 @@ struct dryrun_global_policy {
...
                 // the new global spike is the same as the local spike, with
                 // its source index shifted to the dummy domain
-                auto source = s.source;
-                source.gid += first_cell;
-                global_spikes[first_spike+i] = {source, s.time};
+                auto s = local_spikes[i];
+                s.source.gid += first_cell;
+                global_spikes[first_spike+i] = s;
             }
+            partition[rank+1] = partition[rank]+num_spikes_local;
```
apeyser: So, we're depending here on auto arrays/vectors/etc to be zero initialized? 
Could also be coded after line 34:
```
partition[rank] = first_spike
```
and at the end
```
partition.back() = num_spikes_global
```

but that's just a question of taste.

bcumming: The standard states that `std::vector` with the `std::allocator<T>` will be default initialized to `T()`, which is zero for integer types.

apeyser: By the way, is that true for heap allocated data? I know they generalized the stack allocated and dynamically allocated rules so that C zero-initialization looks like C++ default initialization for those cases - but I'm not sure whether that's also true for null initialization of static objects in C being zero initialization for STLs; I know that for arbitrary C++ objects, there's a default initializer that has the same zero versus null initializer distinction as in C.

apeyser:
dry-run-size should be changed to dry-run-ranks to match the json parameter (or the json parameter should be changed to match the command line).

apeyser:
Dry run mode works, once I identified the problem with my json file: the parameter for json was dry_run_ranks, and I had read the command line parameter instead. Pushed an update to make both match.

PR #153, opened 2017-02-16
Fix modcc optimized code generation

bcumming:
modcc was generating invalid code when generating optimized kernels.
The optimized kernels use raw pointers instead of `view`s, and the generated code was using `view` semantics.
The fix is to conditionally generate pointer or `view` semantics, depending on the optimization level.


halfflat:
Looks correct plus tested.

PR #160, opened 2017-02-20
Some fixes for running modcc when the backend and frontend are dfifferent

apeyser:
Includes a fix to the readme which applies to any external modcc, and a way to keep from rebuilding locally an external modcc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ cmake <path to CMakeLists.txt> -DCMAKE_BUILD_TYPE=release
 make -j8
 
 # set path and test that you can see modcc
-export PATH=`pwd`/bin:$PATH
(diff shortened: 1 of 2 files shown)


halfflat:
Looks fine! Will merge as is, but would you consider the possibility of making this a CMake `option` that is initialized with `ON` or `OFF` based on the presence of `modcc` in the path? This would allow us to force the use of the in-tree modcc even if another version exists in path.

bcumming:
Hello @apeyser and @halfflat, what is happening with this review? As I see it, we have two choices
  1. merge now and make a ticket to improve CMake handling of "cross-compilation"
  2. make @halfflat's suggested modification, then merge.

PR #161, opened 2017-02-21
fix bug introduced when global policy was refactored

bcumming:
The MPI global gather_all unit test now compiles.


w-klijn:
I checked and the build now works on the knl

PR #162, opened 2017-03-02
Morphology generation with L-systems

halfflat:
Adds a stand-alone program for the generation of random morphologies form a L-system description. The algorithm is that of Burke (1992), with some of the extensions provided by Ascoli et al. (2001).

Two sets of L-system parameters have been included, corresponding to alpha motoneurons and Purkinje cells, but there is certainly something wrong with the data for the latter, and more correct numbers will probably need to be synthesized from existing Purkinje cell morphological information.

Documentation for `lmorpho` is incomplete, but the command line help (`--help`) goes some way to explain the usage. In order to get output, one must specify `--swc` or `--pvec` (or both) to emit SWC files or the structural parent vectors. Coarser discretization can be obtained with the `--segment` option.

Some minor modifications have been included in other parts of the source repo:
* Added copy constructor for `TextBuffer` in `modcc/textbuffer.hpp`, required to keep clang++ 3.5 happy.
* Disabled 'maybe-uninitialized' warnings from gcc, as these get raised improperly with some uses of `util::optional`. (Seems that this is also an issue for `boost::optional`.)
* Moved `tinyopt.hpp` into the `src/` directory, and extended the functionality a bit to support keyword arguments.
* `validate.cpp` modified slightly to accommodate new `tinyopt`.


lmorpho/lmorpho.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include "morphio.hpp"
+#include "lsystem.hpp"
+#include "lsys_models.hpp"
+
+namespace to = nest::mc::to;
+using nest::mc::util::optional;
+using nest::mc::util::nothing;
+using nest::mc::util::just;
+
+const char* usage_str =
```
w-klijn: I would add a little bit about what the program does, how is it used?

lmorpho/lsys_models.cpp
```diff
@@ -0,0 +1,139 @@
...
+static constexpr double inf = nest::mc::math::infinity<double>();
+
+// Predefined parameters for two classes of neurons. Numbers taken primarily
+// from Ascoli et al. 2001, but some details (soma diameters for example)
+// taken from earlier literature.
+//
+// Without mixture distribution support, these aren't entirely faithful to
+// the original sources.
+//
+// References:
```
w-klijn: Nice!! references in source code

lmorpho/lsystem.cpp
```diff
@@ -0,0 +1,306 @@
...
+#include <iostream>
+#include <random>
+#include <stack>
+#include <vector>
+
+#include "quaternion.hpp"
+#include "lsystem.hpp"
+#include "morphology.hpp"
+
+// L-system implementation.
```
w-klijn: Add the first paragraph of the wiki entry for l-systems?

lmorpho/lsystem.cpp
```diff
@@ -0,0 +1,306 @@
...
+            t2.rotation *= rotation_y(deg_to_rad*branch_theta2);
+
+            result.children = {t1, t2};
+            return result;
+        }
+
+        if (S.term_test(dl, tip.p.r, g)) {
+            return result;
+        }
+    }
```
w-klijn: Is it better to have some test to prevent never ending loops? To make it saver for future changes?

lmorpho/lsys_models.cpp
```diff
@@ -0,0 +1,139 @@
...
+    L.n_tree = { 8.0, 16.0 };
+
+    // Initial dendrite diameter [µm]. (Dstem)
+    L.diam_initial = { 3.0, 12.0 };
+
+    // Dendrite step length [µm]. (ΔL)
+    L.length_step = { 25 };
+
+    // Initial roll (intrinsic rotation about x-axis) [degrees].
+    L.roll_initial = { -180.0, 180.0 };
```
w-klijn: The choice to make this degrees is to mirror the articles? Most operations you want to do appear to be in radians, so why not use that internally.

lmorpho/morphio.cpp
```diff
@@ -0,0 +1,173 @@
...
+        const auto& points = sec.points;
+        auto n = points.size();
+        if (n<2) {
+            throw std::runtime_error(strprintf("surprisingly short cable: id=%d, size=%ul", sec.id, n));
+        }
+
+        // include first point only for dendrites segments attached to soma.
+        if (sec.parent_id==0) {
+            const auto& p = points[0];
+            swc.emplace_back(kind::fork_point /*dendrite*/, ++id, p.x, p.y, p.z, p.r, parent);
```
w-klijn: Its completely valid and it saves a line. But I would put the ++id on its own line, it feels like a side effect written like this, and they get forgotten

lmorpho/morphio.cpp
```diff
@@ -0,0 +1,173 @@
...
+            parent = id;
+        }
+
+        for (unsigned i = 1; i<n-1; ++i) {
+            const auto& p = points[i];
+            swc.emplace_back(kind::dendrite, ++id, p.x, p.y, p.z, p.r, parent);
+            parent = id;
+        }
+        const auto& p = points.back();
+        swc.emplace_back(sec.terminal? kind::end_point: kind::fork_point, ++id, p.x, p.y, p.z, p.r, parent);
```
w-klijn: A little comment what is going on here?

lmorpho/morphology.cpp
```diff
@@ -0,0 +1,56 @@
...
+    double dx = b.x-a.x;
+    double dy = b.y-a.y;
+    double dz = b.z-a.z;
+
+    return std::sqrt(dx*dx+dy*dy+dz*dz);
+}
+
+void section_geometry::segment(double dx) {
+    unsigned npoint = points.size();
+    if (dx<=0 || npoint<2) return;
```
w-klijn: A comment with why this test?

lmorpho/morphology.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+    std::vector<section_point> sampled;
+    sampled.push_back(points.front());
+    double sampled_length = 0;
+
+    double left = 0;
+    double right = left+distance(points[1], points[0]);
+    double x = length/nseg;
+
+    for (unsigned i = 1; i<npoint;) {
```
w-klijn: looks like a core part of the algo, the variable names do not help. so comments what is going on.

lmorpho/quaternion.hpp
```diff
@@ -0,0 +1,105 @@
+#pragma once
+
+#include <cmath>
+
+struct quaternion {
+    double w, x, y, z;
```
w-klijn: This needs a link to the notation that is being used. I know they are black boxes, but to validate what is going on.

lmorpho/quaternion.hpp
```diff
@@ -0,0 +1,105 @@
...
+    constexpr quaternion(double x, double y, double z): w(0), x(x), y(y), z(z) {}
+
+    // all 4-components
+    constexpr quaternion(double w, double x, double y, double z): w(w), x(x), y(y), z(z) {}
+
+    quaternion conj() const {
+        return {w, -x, -y, -z};
+    }
+
+    quaternion operator*(quaternion q) const {
```
w-klijn: Looking at wikipedia (so that might be wrong) I do not understand the multiplication here.

lmorpho/quaternion.hpp
```diff
@@ -0,0 +1,105 @@
+#pragma once
+
+#include <cmath>
+
+struct quaternion {
```
w-klijn: This would be a great class to add unit tests to?

lmorpho/lsystem.cpp
```diff
@@ -0,0 +1,306 @@
...
+    std::vector<section_point>& points = result.points;
+
+    points.push_back(tip.p);
+    for (;;) {
+        quaternion step = xaxis^tip.rotation;
+        double dl = S.length_step(g);
+        tip.p.x += dl*step.x;
+        tip.p.y += dl*step.y;
+        tip.p.z += dl*step.z;
+        tip.p.r += dl*0.5*S.taper(g);
```
w-klijn: Why the 0.5 is the taper a diameter?

lmorpho/lsystem.cpp
```diff
@@ -0,0 +1,306 @@
...
+    lsys_distribution roll_at_branch;
+    lsys_distribution branch_angle;
+    burke_correlated_r diam_child;
+    burke_exp_test term_test;
+    burke_exp2_test branch_test;
+    double max_extent;
+};
+
+struct section_tip {
+    section_point p = {0., 0., 0., 0.};
```
w-klijn: I would rename to point, to be more alike to the rotation member

lmorpho/morphio.hpp
```diff
@@ -0,0 +1,85 @@
...
+    return n? 1+digits(n/10): 0;
+}
+
+// Write a sequence of morphologies to one or more files
+// as given my `pattern`.
+
+class swc_emitter {
+    multi_file file_;
+
+public:
```
w-klijn: Missing public member descriptions

lmorpho/morphio.hpp
```diff
@@ -0,0 +1,85 @@
...
+// as given my `pattern`. Coalesce the vectors if writing to a single
+// file/stream.
+
+class pvector_emitter {
+    multi_file file_;
+    bool coalesce_ = false;
+    unsigned offset_ = 0;
+
+public:
+    explicit pvector_emitter(std::string pattern, unsigned n=0):
```
w-klijn: public member description

lmorpho/morphology.cpp
```diff
@@ -0,0 +1,56 @@
+#include <cmath>
+#include <vector>
+
+#include <math.hpp>
+
+#include "morphology.hpp"
+
+using nest::mc::math::lerp;
+
+static section_point lerp(const section_point& a, const section_point& b, double u) {
```
w-klijn: What does lerp do?

w-klijn:
I like the additions of the references in the source code

I would add a unit-test suite for the Quaternion struct, or, borrow something from the internet. The first hit on google is a MIT licenced version with existing tests

A number of public member functions are undocumented.

On some places there is magic with single letter variables going on. I'd add comments or name the variables to help the reader.

lmorpho/lsys_models.cpp
halfflat: Yes, exactly. I've used degrees in the parameter sets to make it clearer to a reader, and to match the conventions used in the literature. This is also the reason why parameters that relate to a diameter have not been scaled to relate to the radii we use internally.

lmorpho/lsys_models.cpp
halfflat: I've moved these to the `README.md` now.

lmorpho/lmorpho.cpp
halfflat: Added a `README.md` with a lot more detail, but there is still room there for detailed usage information and e.g. sample output.

lmorpho/lsystem.cpp
halfflat: I think here it is clear from the type.

lmorpho/lsystem.cpp
halfflat: Diameter vs. radius convention; I will clarify in the comments on `lsys_param::taper`.

lmorpho/lsystem.cpp
halfflat: This is covered by the `lsys_distibution` parameters `max_extent` (stops them growing too long) and `max_sections` (stops excessive branching). A bad parameter set could still potentially lead to non-termination though.

lmorpho/morphology.cpp
halfflat: 'Lerp' stands for linear interpolation; there is a bit more detail in `src/math.hpp`, but it's a fairly standard term.

lmorpho/quaternion.hpp
halfflat: Will move into math.h and do so.

lmorpho/quaternion.hpp
halfflat: It follows from the relations:
```
1 = + 1·1 = - i·i = - j·j = - k·k
i = + 1.i = + i.1 = + j·k = - k·j
j = + 1·j = - i·k = + j·1 = + k·i
k = + 1·k = + i·j = - j·i = + k·1
```
and then grouping coefficients.

lmorpho/morphology.cpp
halfflat: I think this is self evident: the segmentation process makes no sense for non-positive `dx` or for a section without any length (i.e. comprising one or fewer points).

lmorpho/morphology.cpp
halfflat: Adding comments.

halfflat:
Made three new commits to address review comments, plus fixed up (somewhat?) the Purkinje cell model parameters.

lmorpho/lsystem.cpp
bcumming: It is clear when we look at the definition, but when the user of this type will see something like
```
// prototype
section_tip foo();

// ...
auto tip = foo();
bar(t.p); // what is p?
bar(t.point); // much clearer
```

w-klijn:
Great change! I'm pro merging. But Ben also commented on the point regarding the point.

bcumming:
Thanks for the review @w-klijn, it is appreciated! I am also satisfied that @halfflat addressed everything (except that one small quibble).
I will merge.

PR #163, opened 2017-03-03
Feature/ospike

apeyser:
This code sorts spikes on the sending node, and then uses the globally sorted spikes on the receiving end to search for node-relevant spikes using C++11 standard binary searches.

Performance characteristics have been reported in the slack channel.


src/communication/communicator.hpp
```diff
@@ -101,17 +102,54 @@ class communicator {
...
+            bool operator() (const spike_type& lhs, const id_type& rhs) const
+            {return lhs.source < rhs;}
+            bool operator() (const id_type& lhs, const spike_type& rhs) const
+            {return lhs < rhs.source;}
+        };
+        // turn pair<it1, it2> into a class with begin()/end()
+        using nest::mc::util::make_range;
+
+        // queues to return
+        auto&& queues = std::vector<event_queue>(num_groups_local());
```
apeyser: Check this.

src/thread_private_spike_store.hpp
```diff
@@ -36,7 +36,7 @@ public :
             spikes.insert(spikes.begin(), b.begin(), b.end());
         }
 
-        return spikes;
+        return std::move(spikes);
```
apeyser: check this

src/communication/communicator.hpp
bcumming: do this by value, and rely on NRVO
```
auto queues = std::vector<event_queue>(num_groups_local());
```

src/thread_private_spike_store.hpp
bcumming: Revert this change.

When returning by value, the return value will be used as an rvalue at the calling site. Though in this case NRVO will lead to copy elision.

src/communication/communicator.hpp
```diff
@@ -101,17 +102,54 @@ class communicator {
...
-                auto gidx = cell_group_index(it->destination().gid);
-                queues[gidx].push_back(it->make_event(spike));
+    /// group as a result of the global spike exchange.    
+    std::vector<event_queue> make_event_queues(const gathered_vector<spike_type>& global_spikes)
+    {
+        // Comparator operator between a spike and a spike source for equal_range
+        struct comp {
+            using id_type = typename spike_type::id_type;
+            bool operator() (const spike_type& lhs, const id_type& rhs) const
+            {return lhs.source < rhs;}
```
w-klijn: If i remember correctly the style guide says
function() {
  body 
  }

apeyser: That would be awful for this case. Maybe better would be
```
function()
    {body();}
```
The declaration part is to long for a one liner, but the body is really trivial.

src/thread_private_spike_store.hpp
apeyser: ✅

src/communication/communicator.hpp
apeyser: ✅

src/communication/communicator.hpp
w-klijn: That would be my alternative also: now it looks like its on the same level as the function. Just look strange.

bcumming:
```
bool operator() (const spike_type& lhs, const id_type& rhs) const {
    return lhs.source < rhs;
}
```

apeyser: That's two extra lines for a trivial operator. In context, which is a trivial struct inside a function to just pass a two-faced lambda, any noise is extra noise. Maybe I should think about putting a "lt" template into util, which will take some really trivial trait classes. Then the whole thing collapses away into two lines defining that 'spike_type x -> x.source' and 'id_type x -> x'.

bcumming: Well, it is one extra line compared to your version. And we have been through this before. You get a lot of noise if it is all on one line and the reader has to parse it.

src/communication/communicator.hpp
```diff
@@ -105,16 +106,15 @@ class communicator {
     /// group as a result of the global spike exchange.    
     std::vector<event_queue> make_event_queues(const gathered_vector<spike_type>& global_spikes)
     {
+        // turn pair<it1, it2> into a class with begin()/end()
+        using nest::mc::util::make_range;
+        using nest::mc::util::lessthan;
         // Comparator operator between a spike and a spike source for equal_range
-        struct comp {
+        struct extractor {
```
apeyser: Here's the other approach: I added a lessthan utility class that makes the comparisons, and then you just need to define extractor to return what you actually want to compare: here, the id_type.

src/util/compare.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+namespace nest {
+namespace mc {
+namespace util {
+
+template<typename E>
+struct lessthan {
```
apeyser: and this is the templated functor, which is templated on another templated functor.

src/communication/communicator.hpp
apeyser: Two extra lines -- it's twice.

I don't want to parse it at all... the whole thing is too damn complex for simply saying "compare x and y.x". See my lessthan class, so I can say what I actually want to say -- but more painful than necessary.

apeyser:
I've got a branch that unifies this with the original backend, plus a new O(connections/node) backend:
https://github.com/apeyser/nestmc-proto/tree/feature/nspike

src/communication/communicator.hpp
```diff
@@ -99,17 +101,54 @@ class communicator {
...
+        // but that buys us very little for large, randomly distributed networks.)
+        auto con_it = connections_.cbegin();
+        const auto con_end = connections_.cend();
+
+        const auto& spikes = global_spikes.values();
+        auto spikes_it = spikes.cbegin();
+        const auto spikes_end = spikes.cend();
+
+        // Search for next block of spikes and connections with the same sender
+        while (con_it != con_end && spikes_it != spikes.end()) {
```
apeyser: This should be spikes_end

apeyser:
Okay, the PR is redone to have four backends:
* linear_spike_queue_policy: the original communicator, walk spikes and search connections
* linear_communication_queue_policy: the inverse of the original, walk connections and search spikes
* global_search_queue_policy: search for blocks of connections, and search for spikes associated with all
* domain_search_queue_policy: as the previous, but instead of using the binary search for spikes, keep the domains for the source of each connection, and search inside that domain block

I've tested them for 50 cells, 10 synapses, 5 compartments, dt = 0.025, tfinal = 50.0, with 1, 2 and 3 dry run ranks.

Backends are set at cmake time with -DNMC_USE_QUEUE=connection/spike/global/domain

src/communication/base_communicator.hpp
```diff
@@ -0,0 +1,149 @@
...
+// target id.
+//
+// The user can add connections to an existing communicator object, where
+// each connection is between any global cell and any local target.
+//
+// Once all connections have been specified, the construct() method can be used
+// to build the data structures required for efficient spike communication and
+// event generation.
+template <typename CommunicationPolicy>
+class base_communicator {
```
bcumming: The classes that we derive from `communicator_base` add just one method: `make_event_queues`.
This implementation could be simplified significantly with type mixins, instead of the full
derivation.                                                                                                                                                                                                         

I can't see any heavy type dependencies, and indeed some of the types can come from the mixin type:
`base::event_queue` is only used inside the specialization, so define it there.
`base::lt_spike_source`: we only do comparisons inside the specialization, so define it there (or use a lambda)

The information you get from: `base::num_groups_local`, `base::cell_group_index`, and `base::connections_` could be passed directly. Here is some pseudo-code:

```
template <typename CommunicationPolicy, typename EventQueueImpl>
class base_communicator {
    using queue_impl = EventQueueImpl;
    queue_impl queue_impl_;
public:
    using event_queue = typename queue_impl::event_queue;

    // define make_event_queues here
    std::vector<event_queue> make_event_queues(const gathered_vector<spike>& global_spikes) {
        return queue_impl_(global_spikes_, connections_, cell_gid_partition_);
    }   
};

// then pick your specialization as follows:
template <typename CommunicationPolicy>
using communicator = communication_base<CommunicationPolicy, global_search_policy>;

File: src/communication/base_communicator.hpp [hunk 8]
```diff
@@ -0,0 +1,149 @@
...
+    using id_type = cell_gid_type;
+
+    /// per-cell group lists of events to be delivered
+    using event_queue =
+        std::vector<postsynaptic_spike_event<time_type>>;
+
+    using gid_partition_type =
+        util::partition_range<std::vector<id_type>::const_iterator>;
+
+    struct spike_extractor {
```
bcumming: See my comment about about how this should be moved to the mixin implementation.

src/communication/domain_search_queue_policy.hpp
```diff
@@ -0,0 +1,80 @@
...
+        // the same source
+        auto con_next = connections_.cbegin();
+        const auto con_end = connections_.cend();
+
+        // Search for next block of spikes and connections with the same sender
+        while (con_next != con_end) {
+            // we grab the next block of connections from the same sender
+            const auto src = con_next->source();
+            const auto domain = con_next->domain();
+            const auto targets = std::equal_range(con_next, con_end, src);
```
bcumming: `std::equal_range` does two binary searches, one to get the lower bound and one to get the upper bound.
We know that `con_next` is the lower bound, so in practice you only need the one call to `std::upper_bound` to find the last connection with the same source.

src/communication/global_search_queue_policy.hpp
```diff
@@ -0,0 +1,85 @@
...
+        // the same source
+        const auto& spikes = global_spikes.values();
+        auto spikes_next = spikes.cbegin();
+        const auto spikes_end = spikes.cend();
+
+        // Search for next block of spikes and connections with the same sender
+        while (con_next != con_end) {
+            // we grab the next block of connections from the same sender
+            const auto src = con_next->source();
+            const auto targets = std::equal_range(con_next, con_end, src);
```
bcumming: `std::equal_range` does two binary searches, one to get the lower bound and one to get the upper bound.
We know that `con_next` is the lower bound, so in practice you only need the one call to `std::upper_bound` to find the last connection with the same source.

src/communication/linear_connection_queue_policy.hpp
```diff
@@ -0,0 +1,61 @@
...
+    // go over all connections, search for spikes with the same source as the connection
+    // and push an event on the connection for each of those spikes
+    // O(connections/node * log(spikes) * spikes/connection)
+    std::vector<event_queue> make_event_queues(const gathered_vector<spike>& global_spikes)
+    {
+        auto queues = std::vector<event_queue>(num_groups_local());
+
+        const auto& spikes = global_spikes.values();
+        const auto spikes_begin = spikes.cbegin();
+        const auto spikes_end = spikes.cend();
```
bcumming: You don't need the `const` in the type qualifiers above.
* `global_spikes` is a const ref, so it `global_spikes.values()` will return a const ref by default, and `auto` will pick this up.
* likewise `cbegin()` and `cend()` return const references, which will be in the `auto` type.

src/communication/linear_spike_queue_policy.hpp
```diff
@@ -0,0 +1,56 @@
...
+    using base::connections_;
+
+public:
+    // go over all spikes, search for connections with the same source as the spike
+    // and then push an event to each connection
+    // O(spikes * log (connections) * connections/spike)
+    std::vector<event_queue> make_event_queues(const gathered_vector<spike>& global_spikes)
+    {
+        auto queues = std::vector<event_queue>(num_groups_local());
+
```
bcumming: You are now sorting the spikes before sending regardless of the `make_event_queues` implementation.
So in this one, you could reduce the set of connections to search in (because connections are also sorted by source index):

```C++
auto conn_begin = connections_.begin();                                                                  
for (const auto spike: global_spikes.values()) {
    auto targets = std::equal_range(conn_begin, connections_.end(), spike.source);
    conn = targets.first;
    for (const auto con: make_range(targets)) {
        const auto gidx = cell_group_index(con.destination().gid);
        queues[gidx].push_back(con.make_event(spike));
    }   
}
```

apeyser:
Reminder to myself - the README needs to be updated with the new options.

src/communication/global_search_queue_policy.hpp
```diff
@@ -0,0 +1,85 @@
...
+            con_next = targets.second; // next iteration, next conn block
+            
+            // and the associated block of spikes
+            const auto sources = std::equal_range(spikes_next, spikes_end,
+                                                  src, lt_spike_src());
+            spikes_next = sources.second; //next block that is > src
+
+            if (sources.first == sources.second) {
+                continue; // skip if no spikes == to source
```
apeyser: Okay, I mean if no spikes in the gathered set are == to the source of connections.
Is that clearer?

CMakeLists.txt
```diff
@@ -184,6 +184,31 @@ elseif(NMC_VECTORIZE_TARGET STREQUAL "AVX2")
...
 #----------------------------------------------------------
+# Which communicator algorithm to use for finding
+#  spike:connection matches: linear/global/domain
+#  communication is O(nodes)
+#  spike  is O(log(nodes))
+#  global is O(log(nodes)),
+#  domain is O(log(neurons/nodes))
+# if the number of spikes goes up with number of nodes
+#----------------------------------------------------------
+set(NMC_USE_QUEUE "domain" CACHE STRING
```
apeyser: Good idea.

CMakeLists.txt
```diff
@@ -184,6 +184,31 @@ elseif(NMC_VECTORIZE_TARGET STREQUAL "AVX2")
...
+  "set the communication connection algorithm, one of connection/spike/global/domain")
+set_property(CACHE NMC_USE_QUEUE PROPERTY STRINGS connection spike global domain)
```
apeyser: I thought that's what I was doing here... It looks to me like what you're referencing.

miniapp/miniapp.cpp
```diff
@@ -200,6 +204,26 @@ std::pair<cell_gid_type, cell_gid_type> distribute_cells(cell_size_type num_cell
     return {cell_from, cell_to};
 }
 
+std::vector<cell_gid_type> partition_domains(cell_size_type num_cells) {
+    using nest::mc::util::make_span;
+    using nest::mc::util::partition_functional;
+    
+    const auto num_domains = communication::global_policy::size();
+    const auto cells_per_domain = (cell_gid_type) (num_cells/(double)num_domains);
```
apeyser: Not round robin. Incrementally added and divided to have equal numbers on every node. So 0...4, 5...9, 10..14 on nodes 1, 2, and 3. The domain partition vector assumes that they're ordered, but not that they're equal sized for future, more sophisticated load balancing.

miniapp/miniapp.cpp
```diff
@@ -200,6 +204,26 @@ std::pair<cell_gid_type, cell_gid_type> distribute_cells(cell_size_type num_cell
...
+
+    std::vector<cell_gid_type> partition;
+    make_partition(partition_functional,
+                   partition,
+                   make_span(0, num_domains),
+                   [=] (domain_gid_type) {return cells_per_domain;});
+    // for (auto i: partition) {
```
apeyser: yes -- missed in several places.

miniapp/miniapp_recipes.cpp
```diff
@@ -134,6 +134,7 @@ class basic_cell_recipe: public recipe {
         std::exponential_distribution<float> delay_dist(delay_distribution_param_);
         float delay = param_.min_connection_delay_ms + delay_dist(rng);
         float weight = param_.syn_weight_per_cell/param_.num_synapses;
+        // not a full cell connection: domain must be filled by caller
```
apeyser: Actually, needs to be removed. In an earlier version, cell_connection had a domain member which was set to (unsigned) -1, but now that's gone, so the comment is not only confusing, but wrong.

src/common_types.hpp
```diff
@@ -21,6 +21,10 @@ using cell_gid_type = std::uint32_t;
 
 using cell_size_type = typename std::make_unsigned<cell_gid_type>::type;
 
+// For cluster domain
```
apeyser: Right -- needs more words to explain that this is an id for domains in the sense of node number in a cluster, aka some source that isn't local to the shared memory.

src/communication/base_communicator.hpp
```diff
@@ -0,0 +1,149 @@
+#pragma once
+
+#include <algorithm>
```
apeyser: Forgot we had that rule

src/communication/base_communicator.hpp
```diff
@@ -0,0 +1,149 @@
+#pragma once
```
apeyser: Don't understand what this is referring to. The name of what is "the generic"? You mean that "communicator" itself is too generic and we should name it something like "internode-communicator" or "domain_communicator"?

src/communication/base_communicator.hpp
apeyser: That was the other option I was considering. And I guess I could template on the types that are declared in the base class... I should try to write it and compare the two versions.

apeyser: I wanted to avoid defining the event_queue and the lt_spike 3 or 4 times as copy & paste, though.

src/communication/base_communicator.hpp
apeyser: Yeah -- I had left that as an open question in the earlier revision. On the one hand, the less_than<extractor> is an extra step of templating. On the other, it really should be a generic lambda rather than a struct for extraction + the <. That's unnecessarily noisy, when all I really want to say is "apply < to id_type and spike.source". lt_spike_source, though, should be:
```
struct lt_spike_source {
    using id_type = typename spike::id_type;                                                                                                                                                                        
    bool operator()(const id_type& l, const spike& r) { return l<r.source; }
    bool operator()(const id_type& l, const spike& r) { return l.source<r; }
};
```

C++11 is still missing generic lambdas that could handle this.

src/communication/base_communicator.hpp
```diff
@@ -0,0 +1,149 @@
...
+    /// Takes as input the list of local_spikes that were generated on the calling domain.
+    /// Returns the full global set of vectors, along with meta data about their partition
+    gathered_vector<spike> exchange(const std::vector<spike>& local_spikes) {
+        // global all-to-all to gather a local copy of the global spike list on each node.
+        auto global_spikes = communication_policy_.gather_spikes( local_spikes );
+        num_spikes_ += global_spikes.size();
+        return global_spikes;
+    }
+
+    /// Check each global spike in turn to see it generates local events.
```
apeyser: This is out of place.

src/communication/base_communicator.hpp
```diff
@@ -0,0 +1,149 @@
...
+    std::size_t cell_group_index(id_type cell_gid) const {
+        EXPECTS(is_local_cell(cell_gid));
+        return cell_gid_partition_.index(cell_gid);
+    }
+
+    std::vector<connection> connections_;
+
+    communication_policy_type communication_policy_;
+
+    uint64_t num_spikes_ = 0u;
```
apeyser: It's in exchange.

src/communication/domain_search_queue_policy.hpp
```diff
@@ -0,0 +1,80 @@
+#pragma once
+
+#ifndef NMC_USE_DOMAIN_SEARCH_QUEUE
+#error "domain_search_queue_policy.hpp should only be compiled in a NMC_USE_DOMAIN_SEARCH_QUEUE build"
```
apeyser: I don't think you would --- something is really f*** if the error is called, and you'd need to hunt from CMakeLists down to figure it out. I don't think there's a better solution, other than adding a pointer to src/communications/communicator.hpp

src/communication/domain_search_queue_policy.hpp
apeyser: Partly true. In fact, equal_range in g++ (and others I'd expect) does one binary search until it finds either the top or the bottom, then continues that binary search to find the other bound. So calling upper_bound is better, but not as much better as you'd expect. Should fix anyway -- it's confusing this way.

What I'm not sure of is, should we be doing a binary search at all, or should I do a walk forward to find the upper_bound? One is log(connections/node) the other length_of_run. Worst case the first is better, but how many connections do we really have on average to the same source?

src/communication/domain_search_queue_policy.hpp
```diff
@@ -0,0 +1,80 @@
...
+            con_next = targets.second; // next iteration, next conn block
+
+            // we grab the block of spikes (by domain) associated with the connections
+            const auto domain_spikes = global_spikes.values_for_partition(domain);
+            const auto sources = std::equal_range(domain_spikes.first,
+                                                  domain_spikes.second,
+                                                  src, lt_spike_src());
+
+            if (sources.first == sources.second) {
+                continue; // skip if no spikes == to src
```
apeyser: skip if we didn't find any spikes for this connection block

Maybe that's better

src/communication/gathered_vector.hpp
```diff
@@ -28,6 +30,15 @@ class gathered_vector {
         return partition_;
     }
 
+    // iterators to a distribution partition
```
apeyser: It's partially tested by Ben's com test. I should check how well, and think about how to make a better unit test from his for this.

src/communication/global_search_queue_policy.hpp
```diff
@@ -0,0 +1,85 @@
...
+            }
+        }
+
+        return queues;
+    }
+};
+
+template <typename CommunicationPolicy>
+using communicator = global_search_communicator<CommunicationPolicy>;
+
```
apeyser: You mean line 81?

src/communication/linear_connection_queue_policy.hpp
```diff
@@ -0,0 +1,61 @@
...
+    
+protected:
+    using base::cell_group_index;
+    using base::connections_;
+
+public:
+    // go over all connections, search for spikes with the same source as the connection
+    // and push an event on the connection for each of those spikes
+    // O(connections/node * log(spikes) * spikes/connection)
+    std::vector<event_queue> make_event_queues(const gathered_vector<spike>& global_spikes)
```
apeyser: Ben suggests a mix-in approach, which I considered. I'll have to write it and compare the two for simplicity.

src/communication/linear_connection_queue_policy.hpp
apeyser: Do they? I see for cbegin() and cend() being a const_iterator, which is an iterator (non-ref and non-const) which reference a constant value in the container. That's different from a const const_iterator, where the iterator itself can't be changed.

src/communication/linear_spike_queue_policy.hpp
apeyser: Yes -- that's implemented in global_search. This one was done afterwards, to compare just the first spike and first connections effects, distinctly from the clustering effect (which is implemented).

src/model.hpp
```diff
@@ -153,7 +185,11 @@ class model {
             PE("stepping", "communication");
 
             PE("exchange");
-            auto local_spikes = previous_spikes().gather();
+            auto local_spikes = previous_spikes().gather();            
+            std::sort(local_spikes.begin(), local_spikes.end(),
```
apeyser: Could add it to the communicator class. Also the connection sorting could be put together so we don't sort either if we don't need to.

src/util/compare.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+template<typename E>
+struct equalto {
+    template<typename T1, typename T2>
+    bool operator() (const T1& t1, const T2& t2) const {
+        return E()(t1) == E()(t2);
+    }
+};
+
+}
```
apeyser: Good idea.

src/util/partition.hpp
```diff
@@ -171,6 +172,32 @@ make_partition(Part& divisions, const Sizes& sizes, T from=T{}) {
...
+constexpr partition_functional_t partition_functional;
+
+template <
+    typename Part,
+    typename Range,
+    typename Func,
+    typename T = typename sequence_traits<Part>::value_type
+>
+partition_range<typename sequence_traits<Part>::const_iterator>
+make_partition(partition_functional_t,
```
apeyser: Transform_view?

src/util/partition.hpp
```diff
@@ -171,6 +172,32 @@ make_partition(Part& divisions, const Sizes& sizes, T from=T{}) {
...
+               Part& divisions, const Range& r, Func f, T from=T{}) {
+    divisions.resize(size(r)+1);
+
+    auto pi = std::begin(divisions);
+    for (const auto& i: r) {
+        *pi++ = from;
```
apeyser: Normally, I'd agree, but this is a really tight memory copy loop. It does make me look twice to figure it out, but then you should look at code like this 4 times before you think you know what's going on.

src/communication/gathered_vector.hpp
bcumming: The unit test only covers the low level `communication::global_policy::gather_spikes()` method which takes the vector of local spikes, and does whatever communication is required by the global policy to make the global spike list.

We need some generic unit tests to check whether these different `make_event_queues` methods generate the right events in queues given a set of global spikes.

src/util/partition.hpp
bcumming: yes, `transform_view`.

src/util/partition.hpp
bcumming: I know that @halflat would agree with @apeyser, but I agree with @w-klijn here. Adding an additional line with `++pi;` would generate code that is just as fast, and easier to reason about... and less prone to bugs.

CMakeLists.txt
bcumming: Oh, that is what is going on here. Sorry, I was asleep at the wheel.

src/util/partition.hpp
bcumming: Yes, I think `transform_view` is appropriate here.
Specifically, `transform_view(make_span(0, n), ... your lambda/functor here ...)`.

src/communication/base_communicator.hpp
bcumming: You define the `event_queue` in one spot: inside the mixin definition.

bcumming: I don't think that you need inheritance at all.

src/communication/domain_search_queue_policy.hpp
bcumming: I suspected that just doing the `std::upper_bound` probably would make no difference.

I had also thought about the linear walk... and I think it should be faster. It is has a more cache-friendly access pattern in some cases.

src/communication/linear_connection_queue_policy.hpp
bcumming: Yes, good point about the `cbegin` and `cend`.

Now I think about it more, in this case you don't need to use `cbegin` and `cend`, because `spikes` is a `const&`, by virtue of itself coming from a `const&`.

apeyser: Right, it's redundant, but on the other hand it specifies it for readers.

apeyser:
@bcumming : Any reason for me to keep this alive, since this has been merged into #326 ?

bcumming:
Yes, I think that there is value in keeping it around until we have benchmarked and tested the different spike list walking algorithms that you have implemented in this PR.

bcumming:
We kept this PR around as a crib-sheet for communicator refactoring. Now the communicator implementation has moved on, so it is time to close this.

PR #165, opened 2017-03-06
fix #164

### LINKED ISSUE #164: modcc doesn't compile with gcc 5
GCC 5 gives an error when compiling modcc on lines like the following

```
for (auto& e: e->terms())
```

This got checked in because GCC 6 handles this case.

bcumming:
Disambiguate e symbol in statements like the following
  for (auto& e: e->terms())
This caused GCC 5 to give an error.

fixes #164


w-klijn:
Code become more readable for users also!
For the rest only local scope changes

PR #167, opened 2017-03-06
move spiked detection to back end

### LINKED ISSUE #106: Generic spike generation [6]
user points estimate: 6
~feature: #67 (GPU Support)~

Currently the only sources of spikes in the miniapp are
1. spikes generated by threashold crossing (i.e. spike detection)
2. spikes manually inserted into spike buffer (ugly and not thread safe)

This task will focus on implementing a generic interface 
  1. A generic interface for implementing spike chain generators will be available [2]
    * C++ abstract base class that can be extended by users in C++ land
    * add support to the `cell_group` and `model` levels of the library
  2. Implement an example Poisson generator (with validation tests) [2]
  3. Unit tests that test the generic interface [1]
  4. Add interface to miniapp  [1]

bcumming:
This PR makes spike detection part of the back end, which is required to have fast spike detection on the gpu.

fixes #106.

backend:
* moved spike detection into the backend.
* each backend provides a threshold_watcher type that is initialized
  with a reference to the field that it is to watch, along with indexes
  and thresholds for each variable in the field that is being watched.
* the multicore back end is a cut and paste of the original code
* the gpu back end calls a kernel, which uses a new data structure
  called a gpu_stack that lets threads in a kernel conditionally
  push back into a flat array.
* the threshold_watch provides a simple host-side interface:
  * a test(t) method that tests for crossings at time t
  * a crossings() method that returns all threshold corssings
  * a clear_crossings() method for reseting the spike buffer to empty
* via this interface, spikes are copied in bulk from device to host
  during the spike gathering phase.

miniapp:
* The miniapp runs a single-step dummy run of model before starting the 
  profiler. Make this only run if the profiler is on. 
* Initialize the spike output callback functions _after_ the dummy run 
  so that spikes from the dummy step are not output.

cell_group:
* the cell_group no longer explicitly performs spike checking, which has 
  been handed off to the backends. This simplifies the cell_group code.

fvm_multicell:
* fixed a bug in the spike detection code when there are more than one 
  cell in a group. Before a new threshold watch of one cell was initialized
  for each cell, now we initialize one watch for the whole cell group.
* some of the spike detection code has moved from cell_group into the 
  lowered cell, as part of moving it into the back end.

memory:
* added a new allocator for CUDA managed memory
* added a managed_ptr and make_managed_ptr, which are managed memory equivalents
  to std::unique_ptr and std::make_unique_ptr.

tests:
* improved host-side spike detection unit tests
* added device-side spike detection unit tests
* added unit test for gpu_stack


CMakeLists.txt
```diff
@@ -221,27 +226,29 @@ set(NMC_VALIDATION_DATA_DIR "${PROJECT_SOURCE_DIR}/validation/data" CACHE PATH
   "location of generated validation data")
 
 #----------------------------------------------------------
-# Whether to build validation data at all
+# Whether to build validation data
```
w-klijn: How big is the validation data? Would it be possible to have a backup pregenerated set available on the repo?

miniapp/miniapp.cpp
```diff
@@ -137,15 +118,35 @@ int main(int argc, char** argv) {
...
+        // reset and add the source spikes once again
         m.reset();
-        // reset the source spikes
         for (auto source : local_sources) {
             m.add_artificial_spike({source, 0});
         }
+#endif
+
+        // file output depends on the input arguments set the output file after
+        // the dummy run to ensure that the initial "seed spikes" are recorded once
```
w-klijn: It is the miniapp, thus that is probable why its here, but a dummy run to be able to record seed spikes feels clutchy.
I think we should think about a production version for spike insertion.

src/backends/fvm_gpu.hpp
```diff
@@ -51,6 +53,14 @@ __global__ void matrix_solve(matrix_solve_param_pack<T, I> params);
 template <typename T, typename I>
 __global__ void assemble_matrix(matrix_update_param_pack<T, I> params, T dt);
 
+template <typename T, typename I, typename Stack>
+__global__
+void test_thresholds(
+    float t, float t_prev, int size,
```
w-klijn: Sorry its not clear to me what t is here: time, threshold, template.
It appears an important function, maybee add function description

src/backends/fvm_gpu.hpp
```diff
@@ -174,6 +185,129 @@ struct backend {
...
+
+        using stack_type = gpu_stack<threshold_crossing>;
+
+        threshold_watcher() = default;
+
+        threshold_watcher(
+                const_view values,
+                const std::vector<size_type>& index,
+                const std::vector<value_type>& thresh,
+                value_type t=0):
```
w-klijn: Again, is this time or threshold?

src/backends/fvm_gpu.hpp
```diff
@@ -174,6 +185,129 @@ struct backend {
 
     static bool has_mechanism(const std::string& name) { return mech_map_.count(name)>0; }
 
+    /// threshold crossing logic
```
w-klijn: Core functionality with lots of comments.

src/backends/fvm_gpu.hpp
```diff
@@ -174,6 +185,129 @@ struct backend {
...
+        threshold_watcher(
+                const_view values,
+                const std::vector<size_type>& index,
+                const std::vector<value_type>& thresh,
+                value_type t=0):
+            values_(values),
+            index_(memory::make_const_view(index)),
+            thresholds_(memory::make_const_view(thresh)),
+            prev_values_(values),
+            is_spiking_(size()),
```
w-klijn: This size function makes me go huh? every time I see it. Could be left over from Python but it just looks wrong in isolation : D

src/backends/fvm_gpu.hpp
```diff
@@ -174,6 +185,129 @@ struct backend {
...
+
+        /// Tests each target for changed threshold state
+        /// Crossing events are recorded for each threshold that
+        /// is crossed since the last call to test
+        void test(value_type t) {
+            EXPECTS(t_prev_<t);
+
+            constexpr int block_dim = 128;
+            const int grid_dim = (size()+block_dim-1)/block_dim;
+            test_thresholds<<<grid_dim, block_dim>>>(
```
w-klijn: From a previous project I remember that might be smart to have all c++ to CUDA function calls in a single file/location. This makes experimenting with blocksizes and such allot easier.

src/backends/gpu_stack.hpp
```diff
@@ -0,0 +1,120 @@
...
+    __host__
+    void clear() {
+        size_ = 0;
+    }
+
+    // The number of items that have been pushed back on the stack.
+    // size may exceed capacity, which indicates that the caller attempted
+    // to push back more values than there was space to store.
+    __host__ __device__
+    unsigned size() const {
```
w-klijn: Would it be smart to have a size_in_memory accessor?

src/cell_group.hpp
```diff
@@ -127,22 +111,23 @@ class cell_group {
...
             // apply events
+            PE("events");
             if (next) {
                 auto handle = get_target_handle(next->target);
                 cell_.deliver_event(handle, next->weight);
             }
             PL();
         }
 
+        // copy out voltage threshold crossings from the back end, then generate
```
w-klijn: I had to read twice to understand, that you only copy the time for the cells that crossed. Rewording would make it easier.

src/cell_group.hpp
```diff
@@ -127,22 +111,23 @@ class cell_group {
...
+        // spikes with global spike source ids.
+        PE("events");
+        for (auto c: cell_.get_spikes()) {
```
w-klijn: Also here you do a get_spikes() call. Maybee call it spikes in the source comment

src/memory/managed_ptr.hpp
```diff
@@ -0,0 +1,118 @@
...
+    managed_ptr(const managed_ptr& other) = delete;
+
+    // Allocate memory and construct in place using args.
+    // This is an extension over the std::unique_ptr interface, because the
+    // point of the wrapper is to hide the complexity of allocating managed
+    // memory and constructing a type in place.
+    template <typename... Args>
+    managed_ptr(construct_in_place_tag, Args&&... args) {
+        managed_allocator<element_type> allocator;
+        data_ = allocator.allocate(1u);
```
halfflat: Shouldn't this `1u` be `sizeof(T)`?

src/memory/allocator.hpp
```diff
@@ -181,6 +181,37 @@ namespace impl {
             }
         };
 
+        // bare bones implementation of standard compliant allocator for managed memory
+        template <size_type Alignment=1024>
+        struct managed_policy {
+            // Managed memory is aligned on 1024 byte boundaries.
+            // So the Alignment parameter must be a factor of 1024
+            static_assert(1024%Alignment==0, "CUDA managed memory is always aligned on 256 byte");
```
w-klijn: The test output text is not the same as what is tested

CMakeLists.txt
halfflat: This is an ongoing question, with as yet no satisfactory answer :)

src/memory/allocator.hpp
```diff
@@ -181,6 +181,37 @@ namespace impl {
             }
         };
 
+        // bare bones implementation of standard compliant allocator for managed memory
+        template <size_type Alignment=1024>
```
w-klijn: There is a function allignment on this class which gives back a default 256.
And not the value of this template parameter

src/backends/gpu_stack.hpp
bcumming: A `size` method that gives the number of type T is standard on all C++ containers.

src/backends/fvm_gpu.hpp
```diff
@@ -240,6 +374,46 @@ void assemble_matrix(matrix_update_param_pack<T, I> params, T dt) {
...
+        const auto v_prev = prev_values[i];
+        const auto v      = values[index[i]];
+        const auto thresh = thresholds[i];
+
+        if (!is_spiking[i]) {
+            if (v>=thresh) {
+                // The threshold has been passed, so estimate the time using
+                // linear interpolation
+                auto pos = (thresh - v_prev)/(v - v_prev);
+                crossing_time = t_prev + pos*(t - t_prev);
```
bcumming: Yes! But it is called from device code, which means we would have to start decorating things in `math.hpp` with some ugly macro cruft, and I didn't feel like making that decision right now.

src/memory/managed_ptr.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+namespace nest {
+namespace mc {
+namespace memory {
+
+// used to indicate that the type pointed to by the managed_ptr is to be
+// constructed in the managed_ptr constructor
+struct construct_in_place_tag {};
+
+// Like std::unique_ptr, but for CUDA managed memory.
```
w-klijn: Lots of comment/documention

src/backends/fvm_gpu.hpp
```diff
@@ -240,6 +374,46 @@ void assemble_matrix(matrix_update_param_pack<T, I> params, T dt) {
...
+            }
+        }
+        else if (v<thresh) {
+            is_spiking[i]=0;
+        }
+
+        prev_values[i] = v;
+    }
+
+    stack.push_back({I(i), crossing_time}, crossed);
```
bcumming: If we use atomics to implement the `push_back`, as is currently the case, then the binary parameter is not needed.

However, if we extend the stack to use methods like scan, then all threads must participate. This code is a simplication from a more sophisticated version that used scan.

Should we modify the stack interface, knowing that we might have to change it back again if a more optimized scan based version is required in the future?

src/backends/gpu_stack.hpp
```diff
@@ -0,0 +1,120 @@
...
+    ~gpu_stack() {
+        allocator().deallocate(data_, capacity_);
+    }
+
+    // Append a new value to the stack.
+    // The value will only be appended if do_push is true.
+    // The do_push parameter is required because all threads in a CUDA thread
+    // block call the push_back method, regardless of whether they have a value
+    // to append.
+    __device__
```
bcumming: see my response above (with atomics no, with something more sophisticated, certainly yest)

src/fvm_multicell.hpp
```diff
@@ -184,7 +178,32 @@ class fvm_multicell {
...
+
+    /// types defined by the back end for threshold detection
+    using threshold_watcher = typename backend::threshold_watcher;
+    using crossing_list     = typename backend::threshold_watcher::crossing_list;
+
+    /// Forward the list of threshold crossings from the back end.
+    /// The list is passed by value, because we don't want the calling code
+    /// to depend on references to internal state of the solver, and because
+    /// for some backends the results might have to be collated before returning.
+    crossing_list get_spikes() const {
```
bcumming: Agreed, I will make the change.

src/memory/managed_ptr.hpp
```diff
@@ -0,0 +1,118 @@
...
+    void swap(managed_ptr&& other) {
+        std::swap(other.data_, data_);
+    }
+
+    __host__ __device__
+    operator bool() const {
+        return is_allocated();
+    }
+
+    void synchronize() const {
```
w-klijn: Is this const ness correct? because it has side effects? Or is that only for the local memory?

src/memory/managed_ptr.hpp
bcumming: Hmmmm, maybe I have deviated from the C++ standard allocator interface. The allocator is templated on type `T`, so asking for `1u` is the same as `1u*sizeof(T)`.
So, no, it is golden. But, might want to standardise allocator interfaces inside the memory namespace.

src/backends/gpu_stack.hpp
```diff
@@ -0,0 +1,120 @@
...
+//  * resetting the stack to an empty state
+//  * querying the size and capacity of the stack
+// Provides a device side interface for
+//  * cooperative grid level push_back
+//  * querying the size and capacity of the stack
+//
+// It is designed to be initialized empty with a given capacity on the host,
+// updated by device kernels, and periodically read and reset from the host side.
+template <typename T>
+class gpu_stack {
```
w-klijn: Does this stack also work in opencl? Otherwise, I propose cuda_device_stack

src/backends/fvm_gpu.hpp
halfflat: I can see the argument either way, though if we keep the conditional push, can we call it `push_conditional`? My personal preference is to keep it simple and then change, but it's only a preference.

src/backends/gpu_stack.hpp
bcumming: We use gpu as shorthand for CUDA/nvidia throughout the code.

src/backends/fvm_gpu.hpp
halfflat: Sorry, just teasing!

CMakeLists.txt
```diff
@@ -4,6 +4,11 @@ cmake_minimum_required(VERSION 2.8)
 project(cell_algorithms)
 enable_language(CXX)
 
+# Hide warnings about mixing old and new signatures for target_link_libraries.
+# These can't be avoided, because the FindCUDA packed provided by CMake uses
+# the old signature, while other packages use the new signature.
+cmake_policy(SET CMP0023 OLD)
```
w-klijn: FYI: The Nvidia ppl are working on native Cuda support in cmake. The current prerelease already supports this. They will not fix the issue in the find_cuda.

bcumming: This is good news, I think.

src/backends/fvm_gpu.hpp
bcumming: Well, I saw that it was a lerp, and I thought about using your lerp code. Then I went to look at your lerp code, and noticed `signum`. Then I decided to do the interpolation myself ;)

halfflat: It's a fair cop.

src/memory/managed_ptr.hpp
halfflat: Cool beans. We can make that another ticket.

miniapp/miniapp.cpp
halfflat: Artificial spike inputs (algorithmic and pre-defined for replay behaviour) is definitely on the desired feature list. Not sure if we have a ticket for them yet, but we can have a look post-hackathon.

src/backends/fvm_gpu.hpp
halfflat: using `t` and `t_<suffix>` for time values is pretty common across the code; I don't think we need to qualify it further. Using `t` for something not a time would be a problem though.

src/memory/managed_ptr.hpp
halfflat: Nope, my misreading of allocator interface. All good.

src/memory/managed_ptr.hpp
halfflat: `const` methods can have side-effects and modify memory, but the unenforced promise is that they behave as if they have no mutable effect.

miniapp/miniapp.cpp
bcumming: This is not a dummy run to seed the spikes. It is a dummy run to initialize the profiler (makes a difference).

The reason that the io code was moved here was that if it is set before the dummy run, the spikes from the dummy run will be recorded and saved to disk... and then will be written again for the real run. So we hook up the io after the dummy run.

bcumming: I agree, running some dummy steps to set up the profiler is a bit cludgy, and not something we would do in production, but a key use of the miniapp is benchmarking, and I want to get good benchmark results for small numbers of steps, in which case initialization overheads of the profiler will affect profiling results.

src/backends/fvm_gpu.hpp
bcumming: `t` is time. We have had this discussion before...

src/backends/fvm_gpu.hpp
bcumming: again, it is time.

src/backends/fvm_gpu.hpp
bcumming: The idea about using the `size()` function is that if you change the underlying data structure, or the way in which the "size" is computed in the future, you don't have to hunt through your code for all the locations where you used the old ad-hoc size calculation. Not that I have done that perfectly well everywhere...

bcumming: But I get your point, it does look a bit "odd".

src/backends/fvm_gpu.hpp
bcumming: We have half this solution here, because all of the CUDA kernel calls in the code, outside of those that are generated for the NMODL files, are defined and called in this file!

Explaining the block dimensions is a good point. This GPU backend stuff is not super clean or structured yet, which it needs to be to handle different GPU versions which might require different configurations.

For this kernel specifically: this will take less than 1% of time to solution: the only important thing is that this work is done on the GPU... we don't need to sweat whether we should use 64, 128 or however many threads!

bcumming: @halfflat and I have discussed what to do with the CUDA quite a bit, and we are leaning towards compiling all of the CUDA code separately (or at runtime, like you have done in past projects).
That would also save us the slow and buggy compilation of the main application with nvcc.

src/backends/gpu_stack.hpp
bcumming: I did have size_in_memory accessor at some point. However, it starts to get hard interpretting this information it the memory is managed by CUDA, and when allocation occurs on 4k page boundaries (sure you might only ask for 20 bytes, but malloc gave you a 4k page).

CMakeLists.txt
bcumming: I have a strong opposition to just storing pre-generated data in the repository. The discussion continues.

miniapp/miniapp.cpp
bcumming: You are right @w-klijn, the comment was a bit ambiguous. I have replaced it with a better description (well, I hope it is clearer).

src/backends/fvm_gpu.hpp
bcumming: Easily fixed by putting a comment above the function that defines what `t` is!

src/cell_group.hpp
bcumming: I don't know: did you look at the code? It clearly creates a new spike for each crossing, using the `index` and the `time` for the crossing.
You have to read the code with the comment.

src/cell_group.hpp
bcumming: I will update the comment.

src/backends/fvm_gpu.hpp
```diff
@@ -174,6 +185,129 @@ struct backend {
...
+
+    private:
+
+        const_view values_;         // values to watch: on gpu
+        iarray index_;              // indexes of values to watch: on gpu
+
+        array thresholds_;          // threshold for each watch: on gpu
+        value_type t_prev_;         // time of previous sample: on host
+        array prev_values_;         // values at previous sample time: on host
+        crossing_list crossings_;   // buffer of crossings: on host
```
bcumming: Right you are. Consider it deleted.

src/backends/fvm_gpu.hpp
bcumming: I have removed the bool flag, and kept the `push_back` name. We can decide what to do when we need to use a method that needs full thread cooperation.

src/backends/gpu_stack.hpp
bcumming: method updated, and comment simplified somewhat ;)

src/fvm_multicell.hpp
bcumming: This is a bit awkward, because it is spike-specific code pretending to be generic! I went with calling them "crossings" in the watcher, to make generalizing the watcher simpler in the future.

bcumming: Actually, I thought that this comment was on a different bit of code.
By this point they are spikes, but with local indexes, not global indexes. I think that `spike` is the clearest term.

src/memory/managed_ptr.hpp
bcumming: It is a good question, but in this case, as @halfflat pointed out, we are promising that we won't mutate the state of the `managed_ptr` contents.

src/memory/managed_ptr.hpp
bcumming: Agree, the comments are a bit thin. All this memory stuff is due a big refactor, and it has to be said that the documentation for `std::unique_ptr` is pretty good.

src/fvm_multicell.hpp
halfflat: I admit to some unease here. We can bikeshed it later.

src/memory/allocator.hpp
bcumming: The `alignment` member returns `Alignement`. I think that the method you are referring to is the one on the `device_allocator` below the `managed_allocator`.

To be honest, this is a bit messy, and needs one hell of a refactoring. It is pretty simple to abuse this interface and violate the alignment guarentees, but that is a job for another PR.

halfflat:
Looks good to me!

CMakeLists.txt
```diff
@@ -221,27 +221,29 @@ set(NMC_VALIDATION_DATA_DIR "${PROJECT_SOURCE_DIR}/validation/data" CACHE PATH
...
-if(JULIA_BIN STREQUAL "JULIA_BIN-NOTFOUND")
-    message(STATUS "julia not found; will not automatically build validation data sets from julia scripts")
-    set(NMC_BUILD_JULIA_VALIDATION_DATA FALSE)
-else()
-    set(NMC_BUILD_JULIA_VALIDATION_DATA TRUE)
-endif()
+# Whether to build validation data
+#----------------------------------------------------------
+# turn off by default
+option(NMC_BUILD_VALIDATION_DATA "generate validation data" OFF)
```
halfflat: Still going to merge the PR, but why turn off validation data generation by default? As things stand, I think it is good to generate it by default if we can.

PR #170, opened 2017-03-07
Bugfixes for `lmorpho`

halfflat:
* Ignore dendrite branches with negative radii arising from correlated child diameter distribution.
* Fix fencepost errors in morphology discretization.
* Rename `tip.p` to `tip.point`.


lmorpho/morphology.cpp
```diff
@@ -39,16 +39,19 @@ void section_geometry::segment(double dx) {
     double x = length/nseg;
 
     // Scan segments for sample points.
-    for (unsigned i = 1; i<npoint;) {
+    unsigned i = 1;
+    for (;;) {
         if (right>x) {
             double u = (x-left)/(right-left);
             sampled.push_back(lerp(points[i-1], points[i], u));
```
w-klijn: What if we have only one point?

lmorpho/morphology.cpp
```diff
@@ -39,16 +39,19 @@ void section_geometry::segment(double dx) {
...
             unsigned k = sampled.size();
-            sampled_length += distance(sampled[k-2], sampled[k]);
+            sampled_length += distance(sampled[k-2], sampled[k-1]);
```
w-klijn: What if we have only 1 sample? Dunno if possible

lmorpho/morphology.cpp
halfflat: We already have a check for npoint minimum 2.

lmorpho/morphology.cpp
halfflat: We have already pushed one point back on line 30, and another on line 46.

PR #174, opened 2017-03-08
native cuda atomicAdd support on Pascal

### LINKED ISSUE #125: Fix `AtomicAdd` in CUDA

bcumming:
Fixes #125

* add cuda_atomic_add and cuda_atomic_sub wrappers for atomic addition
* choose native atomic add for Pascal and later architectures
* choose CAS workaround for devices earlier than Pascal
* add unit test for wrappers
* change default CUDA architecture target to sm_60 in CMakeLists.txt


bcumming:
Look at all those old commits! I have to clean up my fork.

PR #175, opened 2017-03-08
remove small cudaMemcpys

bcumming:
This patch removes the many small `cudaMemcpy` calls for single values, except for those from calling `net_receive` in event delivery.

The small copies during initialization were from when the upper diagonal and time invariant component of the diagonal were computed on the host. There were many small reads/writes to device memory accessing the `p` and `u` vectors.
  * removed by making host-side copies, and copying the result to the device en masse where appropriate

I also added a `constexpr` function for testing if compiled in debug mode, and use that to only perform the `is_physical_solution` test when compiled in debug mode. This test causes a single memcopy on each time step to test whether the voltage has exceeded some "reasonable" beounds.

### DIFF
--- a/src/util/debug.hpp
+++ b/src/util/debug.hpp
@@ -11,6 +11,13 @@ namespace nest {
 namespace mc {
 namespace util {
 
+constexpr inline bool is_debug_mode() {
+#ifndef NDEBUG
+    return true;
+#else
+    return false;
+#endif
+}
 using failed_assertion_handler_t =
     bool (*)(const char* assertion, const char* file, int line, const char* func);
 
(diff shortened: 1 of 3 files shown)


src/util/debug.hpp
```diff
@@ -11,6 +11,13 @@ namespace nest {
 namespace mc {
 namespace util {
 
+constexpr inline bool is_debug_mode() {
+#ifndef NDEBUG
+    return true;
+#else
+    return false;
+#endif
+}
```
halfflat: For another PR: we are going to have to think carefully about run-time checking; it should be independent of what CMake wants to do with Release vs. Debug default flags.

PR #176, opened 2017-03-08
Add flat morphology representation to `nestmc` lib.

halfflat:
This PR is a prelude to closer integration of the random morphology generation with the miniapp, with the first step being support for recipes that create cells from morphologies generated off-line. It aims to use `nest::mc::morphology` as the flat morphology-only representation that can be used to construct `nest::mc::cell` objects and which can exist as a target for SWC conversion and random morphology generation.

* Simplify swc io implementation:
   * Avoid throwing exceptions in `istream` parsing and `swc_record` constructors — only throw when explicitly checking consistency, or when parsing a full sequence of records.
   * Allow direct access to record members.
   * Separate parsing considerations from canonicalization (renumbering, sorting) of a sequence of records.
* Move lmorpho morphology classes into src/
* Add invariant check procedure for morphology.
* Make cells via swc -> morphology -> cell building, rather than direct swc -> cell.
* Allow option to use 'natural' discretization in morphology to determine number of compartments in built cell.
* Update `test_swcio.cpp` to accommodate new API.


src/cell.cpp
```diff
@@ -235,5 +236,49 @@ bool cell_basic_equality(cell const& lhs, cell const& rhs)
...
+        return newcell;
+    }
+
+    EXPECTS(morph.check_invariants());
+
+    // (not supporting soma-less cells yet)
+    newcell.add_soma(morph.soma.r, point3d(morph.soma.x, morph.soma.y, morph.soma.z));
+
+    for (const section_geometry& section: morph.sections) {
+        segmentKind kind = segmentKind::none;
```
bcumming: could this be a bit easier to understand as follows?
```
auto kind = section.kind;
if (kind==segmentKind::none) {
  kind = segmentKind::dendrite;
}
if (kind==segmentKind::soma) {
  throw std::invalid_argument("no support for complex somata");
}
```
This makes it clearer that kind is being assigned `section.kind`, then being filtered.

src/morphology.cpp
```diff
@@ -63,3 +66,39 @@ void section_geometry::segment(double dx) {
     points = std::move(sampled);
     length = sampled_length;
 }
+
+void section_geometry::compute_length() {
```
bcumming: This section_geometry type is getting a bit "big" for the simple struct semantics.

It isn't immediately apparent that `length` is a member of `section_point`. And I am uncomfortable taht it is the user's responsibility to call `compute_length()`.

It might be time to turn this into a proper class.

src/morphology.hpp
```diff
@@ -0,0 +1,68 @@
...
+};
+
+enum class section_kind {
+    soma,
+    dendrite,
+    axon,
+    none
+};
+
+struct section_geometry {
```
bcumming: see my comments above. If this were a class, you could have an `add_point` or `add_points` member that ensures that `length` is always up to date.

src/morphology.hpp
```diff
@@ -0,0 +1,68 @@
...
+    section_kind kind = section_kind::none;
+
+    // Re-discretize the section into ceil(length/dx) segments.
+    void segment(double dx);
+
+    // (Re-)compute length.
+    void compute_length();
+};
+
+struct morphology {
```
bcumming: see above comments.

src/swcio.hpp
```diff
@@ -30,384 +34,213 @@ class swc_record {
         custom
     };
 
+    kind type = kind::undefined; // record type
```
bcumming: You have changed from member variables with trailing underscore to no underscore.

src/cell.cpp
halfflat: The problem here is that `section_kind` (from the morphology) and `segmentKind` (from the cell representation) are different enumerations; we can't assign directly safely.

Using morphology `section_kind` for `segmentKind` in cells would be a cleaner way of doing things, side-stepping the issue; I only refrained from doing that to make the changes (slightly) less invasive.

src/morphology.cpp
halfflat: I'm writing 'right now' a method that sits on the `morphology` class that constructs `section_geometry` structs with the correct fields from e.g. a list of `section_point`s.

I could then remove the `compute_length()` method, and leave it to `morphology` to compute this. Users then are not expected to fill in all the bits of a `section_geometry`, but if they start fiddling with the innards, then on their heads be it. This would be my preferred solution.

src/morphology.hpp
halfflat: Yup. This functionality is currently being added to the `morphology` class. I would then like to keep the `segment_geometry` as a struct, with the intent that external consumers of the object look, but only touch if they really want to take responsibility for what breaks.

src/swcio.hpp
halfflat: Yes; `swc_record` now is a record-like struct with full r/w access by users of the struct. It can represent a record in a sequence of SWC lines from a file even if those records in the file are strictly speaking ill-formed.

Validation/consistency checking is performed separately.

halfflat:
PR updated according to my suggested responses above.

PR #178, opened 2017-03-09
Morphologies in miniapp

halfflat:
* Fix morphology section ctor bug.
* Add morphology pools for miniapp from which morphologies are drawn in the recipe.
* Add command-line options to expose above.
* Add option `--report-compartments` to (slowly) check the min, mean, and max number of compartments in the generated cells across the simulation.
* Use `morphology::add_section` to do all the heavy lifting; no need to make `section_geometry` objects by hand, unless you really, really want to.
* Add more utilities:
   * `util::glob()`: file name globbing; wrapper around POSIX `glob()`.
   * `util::scope_exit()`: RAII guard utility class (as seen previously in our slack channel.)
* Fix `is_sequence<T>` test — it was until now saying everything was a sequence.


src/morphology.cpp
```diff
@@ -106,9 +109,9 @@ static const char* morphology_invariant_violation(const morphology& m) {
         if (id!=i+1) return "section id does not correspond to index";
         if (parent_id>=id) return "section parent id not less than section id";
         if (parent_id>0) {
-            auto parent_index = parent_id-1;
-            terminal[parent_index] = false;
+            terminal[parent_id-1] = false;
         }
+        if (parent_id==0 && !m.has_soma()) return "section has parent 0 but morphology has no (spherical) soma";
```
bcumming: no curly braces.

bcumming:
Looks good for the most part: consider this an expedited merge!

halfflat:
Do leave any less-expedited comments though, so I can patch it up later :)

PR #180, opened 2017-03-09
Bugfix for issue #166: backwards `math::signum`.

### LINKED ISSUE #166: error in math::signum
The math helper function `math::signum` returns +/-1 with the wrong sign. It is only used in one location `tests/validation/trace_analysis.cpp`, so the error hasn't been noticed.

The error can be illustrated by adding the following test to `tests/unit/test_math.cpp`
```
TEST(math, signum) {
    EXPECT_EQ(0, signum(int(0)));
    EXPECT_EQ(0, signum(float(0)));
    EXPECT_EQ(0, signum(unsigned(0)));
    
    EXPECT_EQ(-1, signum(int(-10)));
    EXPECT_EQ(-1, signum(float(-10)));

    EXPECT_EQ(1, signum(int(10)));
    EXPECT_EQ(1, signum(float(10)));
    EXPECT_EQ(1, signum(unsigned(10)));
    
    EXPECT_EQ(1, signum(std::numeric_limits<float>::infinity()));
    EXPECT_EQ(-1, signum(-std::numeric_limits<float>::infinity()));
}
```

halfflat:
Fixes #166 

* Debackwardsate `math::signum()`.
* Unit tests in `test_math.cpp` for same.


halfflat:
Pretty sure those lines are okay, and that validation test was (incorrectly) comparing difference in local _minima_ before instead of local maxima. No change in validation test outcomes observed, and the 'peak's still converge where the l-infinity error converges, by eye-balling the verbose output.

bcumming:
OK!

You should add debackwardsate to the "NestMC style guide", because it is a word now.

PR #195, opened 2017-03-15
Fix that ensures all spikes are saved to disk

### LINKED ISSUE #194: Bug: final set of spikes not always written to file with `miniapp -f`
Running the `miniapp` with `-f` for spike recording may fail to record some of the last spikes that were generated.

bcumming:
fixes #194 

* add an additional exchange at the end of the time
  stepping loop to ensure that any spikes still in
  spike buffers are exchanged and saved to disk
* remove the parallel sort from the cthread back end
  to fix compilation errors with CUDA and cthreads
* fix warning about -1 -> std::size_t conversion with
  CUDA compiler


miniapp/miniapp.cpp
```diff
@@ -276,7 +276,7 @@ void write_trace_json(const sample_trace_type& trace, const std::string& prefix)
 void report_compartment_stats(const recipe& rec) {
 std::size_t ncell = rec.num_cells();
     std::size_t ncomp_total = 0;
-    std::size_t ncomp_min = -1;
+    std::size_t ncomp_min = std::numeric_limits<std::size_t>::max();
```
halfflat: Bit long, innit?

bcumming: warnings: 0
intent: 1

halfflat: If we're working around nvcc, perhaps we should have `compat::minus_one<size_t>::value`?

bcumming: That seems like overkill. And, the intention here is to initialize `ncomp_min` with a very big value. If anything, using `minus_one` for a big variable is even more terrible than -1.

halfflat: `(std::size_t)-1` also works fine to suppress the spurious warning. But I'd rather leave it until nvcc sucks less.

bcumming: I can understand using -1 for `nop`, however for "big unsigned number"... that just feels obtuse.

halfflat: So `~0` would be even worse?

bcumming:
You realise that I have been silently "fixing" those `(unsigned)-1`  that you are so fond of putting in the code?

PR #198, opened 2017-03-16
More robust NMC_NUM_THREADS parsing

### LINKED ISSUE #197: cthread back end quits if NMC_NUM_THREADS not set
This is very annoying, and differs from the other threading models that do not require that an environment variable be set.

I propose:
  * setting the default number of threads to be 1
  * printing a warning at startup if the environment variable has not been set

A more advanced option would be to detect how many core the process has affinity with, and automatically set the number of threads to use all available cores, unless the user explicitly specifies a core count via `NMC_NUM_THREADS`.

bcumming:
fixes #197

*  add functions that can find thread affinity and the number of available
   cores on linux systems via `sched_getaffinity`.
* on other systems they default to "unknown affinity" and return 0 to indicate
  that the number of cores is unknown.
*  set the default number of threads according to the new functions above
   if no environment variable explicitly setting the number of threads is set.
*  use std::stoi to convert string to unsigned to work around intel
   compiler bug, and potential bug if the user requests a negative
   number of threads.
* terminates if no number of threads is provided and the library
  is unable to determine a sensible number automatically.
* uses a simpler regex-based approach to validate the user-
  supplied NUM_NUM_THREADS string.

### DIFF
--- a/src/threading/affinity.cpp
+++ b/src/threading/affinity.cpp
@@ -1,0 +1,63 @@
+#include <vector>
+
+#include <cstdlib>
+
+#ifdef __linux__
+
+    #ifndef _GNU_SOURCE
+        #define _GNU_SOURCE
+    #endif
+
+    extern "C" {
+        #include <sched.h>
+    }
+
+#endif
+
+namespace nest {
+namespace mc {
+namespace threading {
+
+#ifdef __linux__
+std::vector<int> get_affinity() {
+    cpu_set_t cpu_set_mask;
+
+    auto status = sched_getaffinity(0, sizeof(cpu_set_t), &cpu_set_mask);
+
+    if(status==-1) {
+        return {};
+    }
+
+    auto cpu_count = CPU_COUNT(&cpu_set_mask);
+
+    std::vector<int> cores;
+    for(auto i=0; i<CPU_SETSIZE && cores.size()<cpu_count; ++i) {
+        if(CPU_ISSET(i, &cpu_set_mask)) {
+            cores.push_back(i);
+        }
+    }
+
+    if(cores.size() != cpu_count) {
+        return {};
+    }
+
+    return cores;
+}
+#else
+
+// No support for non-linux systems
+std::vector<int> get_affinity() {
+    return {};
+}
+#endif
+
+unsigned count_available_cores() {
+    auto n = get_affinity().size();
+
+    // Assume that there must be at least 1 core if an error was encountered.
+    return n? n: 1;
+}
+
+} // namespace threading
+} // namespace mc
+} // namespace nest
--- a/src/threading/affinity.cpp
+++ b/src/threading/affinity.cpp
@@ -52,10 +52,7 @@ std::vector<int> get_affinity() {
 #endif
 
 unsigned count_available_cores() {
-    auto n = get_affinity().size();
-
-    // Assume that there must be at least 1 core if an error was encountered.
-    return n? n: 1;
+    return get_affinity().size();
 }
 
 } // namespace threading
--- a/src/threading/cthread.cpp
+++ b/src/threading/cthread.cpp
@@ -173,7 +173,7 @@ static size_t global_get_num_threads() {
         terminate("The requested number of threads \""+std::string(str)
             +"\" is not a positive integer");
     }
-    auto num_threads = std::stoi(str);
+
     return std::stoi(str);
 }
 
--- a/src/threading/cthread.cpp
+++ b/src/threading/cthread.cpp
@@ -1,4 +1,5 @@
 #include <cassert>
+#include <cstring>
 #include <exception>
 #include <iostream>
 
@@ -27,7 +28,7 @@ task_pool::run_task::run_task(task_pool& pool, lock& lck):
 {
     std::swap(tsk, pool.tasks_.front());
     pool.tasks_.pop_front();
-    
+
     lck.unlock();
     pool.tasks_available_.notify_all();
 }
@@ -37,7 +38,7 @@ task_pool::run_task::run_task(task_pool& pool, lock& lck):
 task_pool::run_task::~run_task() {
     lck.lock();
     tsk.second->in_flight--;
-    
+
     lck.unlock();
     pool.tasks_available_.notify_all();
 }
@@ -45,7 +46,7 @@ task_pool::run_task::~run_task() {
 template<typename B>
 void task_pool::run_tasks_loop(B finished) {
     lock lck{tasks_mutex_, std::defer_lock};
-    while (true) {  
+    while (true) {
         lck.lock();
 
         while (! quit_ && tasks_.empty() && ! finished()) {
@@ -57,7 +58,7 @@ void task_pool::run_tasks_loop(B finished) {
 
         run_task run{*this, lck};
         run.tsk.first();
-    }    
+    }
 }
 
 // runs forever until quit is true
@@ -79,11 +80,11 @@ task_pool::task_pool(std::size_t nthreads):
     threads_{}
 {
     assert(nthreads > 0);
-  
+
     // now for the main thread
     auto tid = std::this_thread::get_id();
     thread_ids_[tid] = 0;
-  
+
     // and go from there
     for (std::size_t i = 1; i < nthreads; i++) {
         threads_.emplace_back([this]{run_tasks_forever();});
@@ -98,7 +99,7 @@ task_pool::~task_pool() {
         quit_ = true;
     }
     tasks_available_.notify_all();
-    
+
     for (auto& thread: threads_) {
         thread.join();
     }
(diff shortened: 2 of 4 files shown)


src/threading/cthread.cpp
```diff
@@ -132,54 +133,53 @@ void task_pool::wait(task_group* g) {
...
-        nthreads_str = std::getenv(nthreads_var_name);
+        str = std::getenv(nthreads_var_name);
     }
-    else if (! (nthreads_str = std::getenv("NMC_NUM_THREADS"))) {
-        nthreads_str = std::getenv("OMP_NUM_THREADS");
+    else if (! (str = std::getenv("NMC_NUM_THREADS"))) {
+        str = std::getenv("OMP_NUM_THREADS");
     }
 
     // If the selected var is unset,
```
apeyser: comment should be fixed -- it returns now `1` if the conditions are hit.

src/threading/cthread.cpp
```diff
@@ -132,54 +133,53 @@ void task_pool::wait(task_group* g) {
...
-        terminate("Num threads is not a non-zero number");
+    try {
+        // The following with throw if:
+        // * the call to stoul() can't perform string->unsigned conversion;
+        // * the parsed number is followed by any non-whitespace characters.
+        auto end = std::size_t();
+        auto nthreads = std::stoul(str, &end);
+        while (end<std::strlen(str)) {
+            if (!std::isspace(str[end])) {
+                throw int();
```
apeyser: What is the int error class?

src/threading/cthread.cpp
```diff
@@ -132,54 +133,53 @@ void task_pool::wait(task_group* g) {
...
 
-    // and it's got a single non-zero value
-    auto nthreads{std::atoi(nthreads_str)};
-    if (! nthreads) {
-        terminate("Num threads is not a non-zero number");
+    try {
+        // The following with throw if:
+        // * the call to stoul() can't perform string->unsigned conversion;
+        // * the parsed number is followed by any non-whitespace characters.
+        auto end = std::size_t();
```
apeyser: So, this takes either +/-digits..., negates -digits, and then tries to make it an unsigned long - the original only took digits and failed on negative (or +). If we wanted to be totally safe, we would need to check for range errors (but the original didn't)

apeyser: Why `size_t end = 0;` ? It's only used to return a value -- it needn't (and thus I'd argue shouldn't) be set.

src/threading/cthread.cpp
```diff
@@ -132,54 +133,53 @@ void task_pool::wait(task_group* g) {
...
+        auto nthreads = std::stoul(str, &end);
+        while (end<std::strlen(str)) {
+            if (!std::isspace(str[end])) {
+                throw int();
+            }
+            ++end;
+        }
+        return nthreads;
+    }
+    catch(...) {
```
apeyser: Shouldn't this catch the int() specifically? There shouldn't be any other exceptions, and if there are there's something we missed which should be allowed to escape.

src/threading/cthread.cpp
bcumming: Because I am using `auto` on the lhs, so I need a "value" on the rhs.
Could just as easily be `std::size_t end;`

src/threading/cthread.cpp
bcumming: `std::stoi` throws an exception if it can't perform the conversion to unsigned int.
This is just a catch all for "something went wrong in parsing value".

src/threading/cthread.cpp
bcumming: This was just a lazy way of indicating an error that should be handled by the `catch` (because `std::stoi` also throws if it fails).

Should I throw something like:
````
throw std::invalid_argument("trailing non-numeric values on thread count parameter");
```

src/threading/cthread.cpp
```diff
@@ -132,54 +133,53 @@ void task_pool::wait(task_group* g) {
...
-    // and it's got a single non-zero value
-    auto nthreads{std::atoi(nthreads_str)};
-    if (! nthreads) {
-        terminate("Num threads is not a non-zero number");
+    try {
+        // The following with throw if:
+        // * the call to stoul() can't perform string->unsigned conversion;
+        // * the parsed number is followed by any non-whitespace characters.
+        auto end = std::size_t();
+        auto nthreads = std::stoul(str, &end);
```
halfflat: If we're attempting to catch the negative value case, `std::stoul` won't really help us. I'd go with `std::stoi` and check for negative, or (simpler IMO) test the string against a regex:
  1. Use `regex_match` to ensure string has the no minus sign, digits surrounded by only whitespace etc.
  2. Use `strtol` or `strtoul` which don't throw (simplifies logic).
  3. Terminate if no regex match or if `strtol` has returned zero or has set `errno` to `ERANGE`.

If the regex passe, and then if that succeeds, use `std::strtoul` (which doesn't throw)

bcumming: I have a shameful secret: I can't speak regex beyond very basic search. It might as well be written in Cantonese...

halfflat: The simplest version (doesn't try and handle hexadecimal) would just be `std::regex("\\s*(\\d+)\\s*")`

bcumming:
On the topic of: should we set a default number of threads or throw an error?

I think it is reasonable to set a default of 1 because
* I find it very annoying every time I use the cthread interface and I get an error because I didn't explicitly set the number of threads.
* OpenMP sets a default (grabbing as many cores as are available).

However, we should certainly throw and error if an invalid string is provided for `NMC_NUM_THREADS`, because not doing so would cause pain if someone thinks that they are explicitly supplying a valid number and get no error message.

These are just opinions: I am open to persuasion.

src/threading/cthread.cpp
apeyser: That's what I was doing in my original code, but by hand. Didn't know that regex was now C++ standard (finally).

apeyser: And I really would stick to that -- why handle hex, oct, binary...
We just want a normal positive integer with no fancy stuff -- it's a command line.

src/threading/affinity.cpp
```diff
@@ -0,0 +1,63 @@
...
+}
+#else
+
+// No support for non-linux systems
+std::vector<int> get_affinity() {
+    return {};
+}
+#endif
+
+unsigned count_available_cores() {
```
apeyser: I'd return zero here, and then throw an error in the code for deciding on the default number. So, if the code works, it's a good default -- but if it doesn't, you don't get a silent failure ala perl, but a loud failure since in fact the code is not guessing your reasonable intent.

src/threading/cthread.cpp
```diff
@@ -153,14 +153,13 @@ static size_t global_get_num_threads() {
...
-    //   or no var is set,
-    //   error
+    // If the selected var is unset set the number of threads to
+    // the hint given by the standard library
     if (!str) {
-        std::cerr << "WARNING: Environment variable describing number of threads not set: using 1 thread.\n";
-        std::cerr << "Set one of NMC_NUM_THREADS_VAR, NMC_NUM_THREADS or OMP_NUM_THREADS to set the\n";
-        std::cerr << "number of threads.\n";
-        return 1;
+        auto nthreads = nest::mc::threading::count_available_cores();
```
apeyser: so here's where I'd throw an error if I got a zero (or just throw an error in the other place, and if someone wants to use it with a default of 1, let them catch intentionally and set to 1)

apeyser:
* I find it very annoying every time I use the cthread interface and I get an error because I didn't explicitly set the number of threads.
* OpenMP sets a default (grabbing as many cores as are available).

I think the openmp default is reasonable, and we should use it with the new affinity code. Otherwise, I think one should be annoyed -- there's no standard default, and all we can do is annoy the user that they probably aren't doing what they meant, and if they do mean it they have to say it.

src/threading/cthread.cpp
bcumming: I strongly agree about hex and friends. Out of interest @halfflat and I checked whether OpenMP supports hex, and it doesn't (a sane decision).

src/threading/affinity.cpp
bcumming: It now returns zero, and there is a comment in the header that states that this means "unknown affinity"

src/threading/cthread.cpp
```diff
@@ -156,31 +157,24 @@ static size_t global_get_num_threads() {
...
     }
-    catch(std::exception& e) {
+
+    // check that the environment variable string describes a non-negative integer
+    if (!std::regex_match(str, std::regex("\\s*\\d*[1-9]\\d*\\s*"))) {
         terminate("The requested number of threads \""+std::string(str)
             +"\" is not a positive integer");
     }
+    auto num_threads = std::stoi(str);
+    return std::stoi(str);
```
apeyser: Looking at stoi, the sequence "010" is the decimal integer 8. Maybe we should require the regex to start with a nonzero?

Also, we call stoi twice here, I think by accident.

src/threading/cthread.cpp
```diff
@@ -173,7 +173,7 @@ static size_t global_get_num_threads() {
         terminate("The requested number of threads \""+std::string(str)
             +"\" is not a positive integer");
     }
-    auto num_threads = std::stoi(str);
+
     return std::stoi(str);
```
apeyser: Ahh -- just part of my comment applies, that "010" == 8l

src/threading/cthread.cpp
bcumming: This has been fixed by the final commit.

src/threading/cthread.cpp
bcumming: By default `std::stoi` uses base10. You have to set explicitly set the third argument to 0 to get auto-detection of base. e.g.:
```
bcumming@arapiles:test > cat stoi.cpp 
#include <iostream>
#include <string>

int main(void) {
    std::cout << std::stoi("010") << "\n";
    std::cout << std::stoi("010", nullptr, 0) << "\n";
}
bcumming@arapiles:test > g++ stoi.cpp && ./a.out
10
8
```
the docs are here:
http://en.cppreference.com/w/cpp/string/basic_string/stol

apeyser: I misread the docs.. Good to know, and thus ignore my comment.

PR #201, opened 2017-03-21
Feature/comm tests

### LINKED ISSUE #200: Communicator Unit Tests
Is the communicator functioning correctly? Good question!

Let's write some unit tests to find out
  * test spike exchange
  * handle the special case of dry run mode

bcumming:
Add unit tests for communicator

fixes #200

* update global_communication test driver to initialize correctly in
  dry run mode
* add unit tests that test the communication::global_policy:
    * basic initialization
    * global spike exchange (just the spike gather step, not the
      event delivery).
* improve the formatting of the reporting from the MPI GTest wrapper
  to make it easy to see if tests have failed.


tests/global_communication/test.cpp
```diff
@@ -28,12 +38,44 @@ int main(int argc, char **argv) {
...
-    //      0 : success
-    //      1 : failure
-    auto result = RUN_ALL_TESTS();
+    int return_value = 0;
+    try {
+        auto arg = argv+1;
+        while (*arg) {
+            if (auto comm_size = to::parse_opt<unsigned>(arg, 'd', "dryrun")) {
+                if (*comm_size==0) {
+                    throw to::parse_opt_error(*arg, "must be positive integer");
```
w-klijn: comm_size is tested but the error says something about *arg

tests/global_communication/test.cpp
```diff
@@ -28,12 +38,44 @@ int main(int argc, char **argv) {
...
+            }
+        }
+
+        // record the local return value for tests run on this mpi rank
+        //      0 : success
+        //      1 : failure
+        return_value = RUN_ALL_TESTS();
+    }
+
+    catch (to::parse_opt_error& e) {
```
w-klijn: I do not like to use exceptions for 'standard' behavior: Incorrect input argument are at this stage expected! So this should be solved with normal control flow. 
Why not do the usage imediately and then return.

tests/global_communication/test_communicator.cpp
```diff
@@ -14,22 +14,126 @@ using namespace nest::mc;
...
+bool operator==(source_proxy s, int other) {return s.gid==other;};
+
+struct spike_proxy {
+    spike_proxy() = default;
+    spike_proxy(int s, int d): source(s), domain(d) {}
+    source_proxy source = 0;
+    int domain = 0;
+};
+
+TEST(communicator, gather_spikes) {
```
w-klijn: I would split this test into two cases: One specific for dryrun and one for the other: 70% of the body if the test is different. So there is no real need to share this code. You could then skip the test earlier in the chain.

tests/global_communication/test_communicator.cpp
```diff
@@ -14,22 +14,126 @@ using namespace nest::mc;
...
+
+        // test that partition information is correct
+        const auto& part =global_spikes.partition();
+        EXPECT_EQ(unsigned(num_domains+1), part.size());
+        EXPECT_EQ(0, (int)part[0]);
+        for (auto i=1u; i<part.size(); ++i) {
+            EXPECT_EQ(sumn(i-1), (int)part[i]);
+        }
+
+        // test that spikes were correctly exchanged
```
w-klijn: Make this a function?

tests/global_communication/test_communicator.cpp
```diff
@@ -14,22 +14,126 @@ using namespace nest::mc;
...
+        auto sumn = [scale](int n) {return scale*n*(n+1)/2;};
+        const auto n_local_spikes = scale*rank;
+
+        // create local spikes for communication
+        // the ranks generate different numbers of spikes, with the ranks
+        // generating the following number of spikes
+        //      [ 0, scale, 2*scale, 3*scale, ..., (num_domains-1)*scale ]
+        // i.e. 0 spikes on the first rank, scale spikes on the second, and so on.
+        std::vector<spike_proxy> local_spikes;
+        const auto local_start_id = sumn(rank-1);
```
w-klijn: What happens if the rank is 0?

tests/global_communication/test_communicator.cpp
```diff
@@ -14,22 +14,126 @@ using namespace nest::mc;
...
+        }
+
+        // perform exchange
+        const auto global_spikes = policy::gather_spikes(local_spikes);
+
+        // test that partition information is correct
+        const auto& part =global_spikes.partition();
+        EXPECT_EQ(unsigned(num_domains+1), part.size());
+        EXPECT_EQ(0, (int)part[0]);
+        for (auto i=1u; i<part.size(); ++i) {
```
w-klijn: An index starting at 1. Comment why?

tests/global_communication/test_communicator.cpp
```diff
@@ -14,22 +14,126 @@ using namespace nest::mc;
...
+        // contiguous source gid
+        const auto& spikes = global_spikes.values();
+        for (auto i=0u; i<spikes.size(); ++i) {
+            const auto s = spikes[i];
+            EXPECT_EQ(i, unsigned(s.source.gid));
+            EXPECT_EQ(0, s.domain);
+        }
+    }
+    else {
+        const auto scale = 10;
```
w-klijn: Scale is a very generic name. Some comment what is meant

tests/global_communication/test_communicator.cpp
apeyser: Because partition maps are [0, length1, length2, length3] so that you can also treat them as bound pairs [0, length1), [length1, length2)... The other choice is to go from zero, but have the EXPECT_EQ(sumn(i), (int) part[i+1]).

tests/global_communication/test_communicator.cpp
bcumming: That was the intention: have one rank that doesn't generate any spikes to see if that causes any problems. Not terribly scientific.

tests/global_communication/mpi_listener.hpp
```diff
@@ -97,24 +97,30 @@ class mpi_listener : public testing::EmptyTestEventListener {
...
     virtual void OnTestCaseEnd(const TestCase& test_case) override {
         printf_helper(
-            "[PASSED %3d; FAILED %3d] of %3d tests in %s\n\n",
+            "    PASSED %d of %d tests in %s\n",
             test_case_tests_-test_case_failures_,
-            test_case_failures_,
             test_case_tests_,
             test_case.name()
         );
+        if (test_case_failures_>0) {
```
bcumming: Only print FAILURE information if there was a failure. The previous version always printed `PASSED xx; FAILED yy` for each test case... and it always looked like there had been a failure. When people see `FAILED` they automatically think that something failed... so I just put the `FAILED` where there are failures. 
Believe me, the reports are much easier to read.

tests/global_communication/test.cpp
bcumming: `arg` is a `util::optional`, and the way that we take the underlying value inside an `optional` is via the `*` operator.

@halfflat, should the interface for `to::parse_opt_error()` also take an `optional` instead of the underlying value?

halfflat: `arg` is a `char **`; `*arg` is a `char *` that's the currently examined command line argument. `arg` is going to be `argv+n` for some small `n`.

`to::parse_opt<T>(...)` returns an `optional<T>`: if the parse is successful, it holds the value; if it is unsuccessful is holds `nothing`.

The error here is thrown because the value was successfully parsed (`comm_size` is not `nothing`), but the value is not sensible. `*arg` has the C string with the argument that was parsed.

bcumming: `arg` is passed to `to::parse_opt` as `char**&`, and `parse_arg` will increment it 1 or 2 steps, depending on whether the a flag or an parameter is being parsed.

By using `*arg`, you are effectively saying "the string that describes the next argument to be processed".

tests/global_communication/test.cpp
bcumming: If these were the only exceptions that were being thrown, I would generally agree, even though I think it is six of one, half a dozen of the other.

However, the calls to `to::parse_opt` can also throw a `to::parse_opt_error` exception. So throwing the same exception type actually simplifies the error handling here, because you need the try-catch block anyway.

tests/global_communication/test_communicator.cpp
```diff
@@ -14,22 +14,126 @@ using namespace nest::mc;
...
+        auto sumn = [scale](int n) {return scale*n*(n+1)/2;};
```
bcumming: With just one or two variables I think being explicit is clearer. So, when reading the body of the lambda, all symbols should be either arguments or in the capture list.

tests/global_communication/test_communicator.cpp
bcumming: I am not inclined to make it a function because
* it is used in this one location
* you would have to pass in a lot of arguments (`num_domains`, `global_spikes` and the `sumn` lambda
* when inside a function call from a test, you have to take extra steps to ensure that any test failures in the inner loop can be traced back to the test that caused them

tests/global_communication/test_communicator.cpp
halfflat: Just randomly piping in: explicit is usually nicer because it makes the interdependencies clear. Especially if the use of the lambda leaves the scope in which it is defined.

Sometimes though a lambda is used in the same way that you would use any other code within a scope, and there e.g. `[&]` would make perfect sense. Depends on context as always.

tests/global_communication/test_communicator.cpp
bcumming: I have split it into two tests.
The version that was being used for dry run mode was actually perfectly valid in normal MPI mode too: it handles the case where there were the same number of cells on each rank, and the local cells all generate one spike each.
The other test has a test that returns without running the test if we are in dry run mode.

tests/global_communication/test_communicator.cpp
bcumming: I think that you have to read the code. This is a very common pattern when writing loops that work with partition maps, and furthermore any attempt to explain "why 1" in a comment will probably just make it even more confusing.

tests/global_communication/test_communicator.cpp
```diff
@@ -42,97 +50,120 @@ struct source_proxy {
...
+    // domain.
+    // We set the model up with n_local_cells==n_local_spikes with
+    // one spike per local cell, so the result of the global exchange
+    // is a list of num_domains*n_local_spikes spikes that have
+    // contiguous source gid
+    const auto& spikes = global_spikes.values();
+    for (auto i=0u; i<spikes.size(); ++i) {
+        const auto s = spikes[i];
+        EXPECT_EQ(i, unsigned(s.source.gid));
+        EXPECT_EQ(0, s.domain);
```
apeyser: Shouldn't this be i/local_spikes.size()?

bcumming: Good spot. In dry run mode it should be `0`, otherwise it should be `i/n_local_spikes`.
The `spike_proxy::domain` is the "payload", and it is not modified by the exchange (the exchange only modifies the source gid). So, `domain` should be 0 everywhere for dry run.

bcumming: I ran the tests and they passed, so give me a moment to investigate further.

bcumming: I have fixed it. It required 2 changes
  1. mpi mode has to explicitly set the source gid to the correct gid (instead of relying on the communicator back end creating the remote gid correctly)
  2. the check must do the test you suggest when in mpi mode.

tests/global_communication/test_communicator.cpp
```diff
@@ -103,10 +103,16 @@ TEST(communicator, gather_spikes_equal) {
     // is a list of num_domains*n_local_spikes spikes that have
     // contiguous source gid
     const auto& spikes = global_spikes.values();
+    EXPECT_EQ(n_local_spikes*policy::size(), int(spikes.size()));
     for (auto i=0u; i<spikes.size(); ++i) {
         const auto s = spikes[i];
         EXPECT_EQ(i, unsigned(s.source.gid));
-        EXPECT_EQ(0, s.domain);
+        if (is_dry_run()) {
```
apeyser: I guess I still don't understand why the dry-run-mode sources will always be domain 0. If dry run mode ranks > 1, then the sources could be from any of those ranks and thus the domains could be anywhere (and that's what I've seen debugging)

apeyser: So the solution offline is just to rename this to "payload" or "data".
I think that's the last thing.

PR #203, opened 2017-03-21
Tame `time_type` proliferation.

halfflat:
Differing classes had their own `time_type` and other classes were parameterized on it in ways that were compatibile, but only by chance.

With these changes, modifying the time type used in `spike` will propagate through to all dependent classes.

* Rename generic `spike<I, T>` as `basic_spike<I, T>`.
* Use `spike = basic_spike<cell_member_type, float>` as the common spike type.
* Replace instances of `spike_type` aliases with just `spike`.
* `time_type` aliases are defined in terms of `spike::time_type`.
* Remove `time_type` parameterization in `connection`.
* Remove `time_type` parameterization in `communicator`.
* Remove `time_type` parameterization in exporter classes.

### DIFF
--- a/src/connection.hpp
+++ b/src/connection.hpp
@@ -57,7 +57,6 @@ static inline bool operator<(cell_member_type lhs, const connection& rhs) {
 } // namespace mc
 } // namespace nest
 
-template <typename T>
 static inline std::ostream& operator<<(std::ostream& o, nest::mc::connection const& con) {
     return o << "con [" << con.source() << " -> " << con.destination()
              << " : weight " << con.weight()
--- a/src/connection.hpp
+++ b/src/connection.hpp
@@ -9,15 +9,13 @@
 namespace nest {
 namespace mc {
 
-template <typename Time>
 class connection {
 public:
-    using id_type = cell_member_type;
-    using time_type = Time;
+    using time_type = spike::time_type;
 
-    connection()=default;
+    connection() = default;
 
-    connection(id_type src, id_type dest, float w, time_type d) :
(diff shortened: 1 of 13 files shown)


src/thread_private_spike_store.hpp
```diff
@@ -20,7 +20,7 @@ class thread_private_spike_store {
 public :
     using id_type = cell_gid_type;
     using time_type = Time;
-    using spike_type = spike<cell_member_type, time_type>;
+    using spike_type = basic_spike<cell_member_type, time_type>;
```
bcumming: Why not remove the parameterization on `Time` type for `thread_private_spike_store`, and just use `spike` instead of `basic_spike`?

src/connection.hpp
```diff
@@ -9,61 +9,56 @@
...
+static inline bool operator<(cell_member_type lhs, const connection& rhs) {
     return lhs < rhs.source();
 }
 
 } // namespace mc
 } // namespace nest
 
 template <typename T>
-static inline std::ostream& operator<<(std::ostream& o, nest::mc::connection<T> const& con) {
+static inline std::ostream& operator<<(std::ostream& o, nest::mc::connection const& con) {
```
bcumming: Can we now remove the template parameter on this function?

halfflat:
`cell_group.hpp` already had the `time_type` taken from `spike`, but the lower level representations I think we should leave alone for now — there may well be situations where the backend representations and upper level representations of what a time is differ.

bcumming:
looks good to me.

PR #204, opened 2017-03-22
Add `event_binner` class for explicit binning.

halfflat:
* Add class for managing state associated with binning event times across integration periods.
* Include support for no or fixed 'regular' binning.
* Add a gtest-assertion compatible test for comparing sequences of floating point numbers: `testing::seq_almost_eq` in `tests/unit/common.hpp`.
* Rename `cell_` in `cell_group` to `lowered_`, to clarify intent (i.e. lowered cell state is very different from a `cell` object, and maintains state for many cells).
* Reformat some comments for consistency.

Note that the `event_binner` class is not used in this commit for actual binning: the original logic is still in place.

### DIFF
--- a/src/cell_group.hpp
+++ b/src/cell_group.hpp
@@ -39,6 +39,11 @@ public:
         policy_(policy), bin_interval_(bin_interval)
     {}
 
+    // Determine binned time for an event based on policy.
+    // If `t_min` is specified, the binned time will be no lower than `t_min`.
+    // Otherwise the returned binned time will be less than or equal to the parameter `t`,
+    // and within `bin_interval_`.
+
     time_type bin(cell_gid_type id, time_type t, time_type t_min = std::numeric_limits<time_type>::lowest()) {
         time_type t_binned = t;
 
--- a/src/cell_group.hpp
+++ b/src/cell_group.hpp
@@ -19,10 +19,74 @@
 namespace nest {
 namespace mc {
 
+enum class binning_kind {
+    none,
+    regular,   // => round time down to multiple of binning interval.
+    following, // => round times down to previous event if within binning interval.
+};
+
+class event_binner {
+public:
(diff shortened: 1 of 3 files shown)


src/cell_group.hpp
```diff
@@ -19,10 +19,74 @@
...
+        last_event_times_.clear();
+    }
+
+    event_binner(): policy_(binning_kind::none), bin_interval_(0) {}
+
+    event_binner(binning_kind policy, time_type bin_interval):
+        policy_(policy), bin_interval_(bin_interval)
+    {}
+
+    time_type bin(cell_gid_type id, time_type t, time_type t_min = std::numeric_limits<time_type>::lowest()) {
```
bcumming: Needs a bit of documentation.

src/cell_group.hpp
```diff
@@ -65,23 +129,24 @@ class cell_group {
...
         reset_samplers();
-        cell_.reset();
+        binner_.reset();
+        lowered_.reset();
     }
 
-    time_type min_step(time_type dt) {
-        return 0.1*dt;
+    void set_binning_policy(binning_kind policy, time_type bin_interval) {
+        binner_ = event_binner(policy, bin_interval);
```
bcumming: where is the `binner_` used?

src/cell_group.hpp
bcumming: just two sentences describing intent.

src/cell_group.hpp
halfflat: `binner_` not used yet... next (rather intrusive) PR will use the binner and expose the policy.

bcumming: gotcha

halfflat:
Small comment added for `event_binner::bin()`.

bcumming:
lgtm

PR #205, opened 2017-03-23
Remove openmp threading back end

### LINKED ISSUE #152: Remove the OpenMP back end
The recently implemented `cthread` gives us a portable implementation threading implementation that supports task-based parallelism.

So the OpenMP back end, which doesn't support asynchronous tasks, isn't needed any more to provide portability to systems like BG/Q.

I am in favor of removing it, to reduce the overhead of maintaining it.

How do others feel about this proposal?

bcumming:
fixes #152

* remove OpenMP back end implementation
* remove CMake options for OpenMP
* simplify the threading code in CMake while I am at it
* remove Extrae benchmarking scripts, which require OpenMP.


halfflat:
lgtm!

src/threading/parallel_stable_sort.h
```diff
@@ -1,127 +0,0 @@
-/*
```
apeyser: cthreads then needs a parallel sort replacement

src/threading/pss_common.h
```diff
@@ -1,112 +0,0 @@
-/*
```
apeyser: And same with this, I think.

bcumming: The Intel implementation that was introduced with the OpenMP back end caused no end of problems for some compilers. For example we couldn't use CUDA with the cthread back end.
Whatever parallel sort we use, it doesn't have to be the best of the best, just good enough to reduce model building time.

halfflat: `cthread` is now using just a simple, serial `std::sort` implementation. Which is not ideal, but then we were having all those nvcc errors trying to compile the PSS code.

We should definitely consider reintroducing a thread-parallel sort, but I think we're best off writing our own, in order to avoid the nvcc issue, and also to have one that is independent of OpenMP.

apeyser: Okay, opened issue #206

PR #208, opened 2017-03-24
block-interleaved gpu matrix solver

### LINKED ISSUE #185: Interleaved matrix solver on GPU [8]
To improve the efficiency of the "one thread per matrix", interleaved
storage of the matrix and rhs/solution vectors helps coalesce memory
accesses. Some tests have been performed that show that this would be
beneficial, particularly for 10,000 or more matrices on P100 GPUs.

This task has the following steps. Note that step 1 has already been
completed.

 1. Matrix setup phase that takes a set of matrices described by `parent_index`
   and `cell_index` index vectors, and generates the storage for the
   interleaved matrix data, along with the information required to quickly
   perform forward and backward permutation operations between `fvm_multicell`
   storage and interleaved storage. []
 2. Write CUDA kernel that fuses interleaving of current and voltage with
   assembly of the diagonal and rhs of the linear system in interleaved
   storage. [2]
 3. Hines solver that uses interleaved storage [2]
 5. method for "de-interleaving" solution for passing back to caller [1]
 6. Test on ranges of morphologies [2]
 7. Experiment with block width and threads-per-matrix during interleave
   operations to find a first set of default values. [2]

Total requirements [8]

bcumming:
Add a new back end gpu Hines matrix solver that uses a block-interleaved storage pattern to improve memory coallescing during the matrix solve.

  * refactor the `src/backends` path into `src/backends/gpu` and `src/backends/multicore` paths that contain gpu and multicore implementations. This simplifies the names of files and, at least to me, is a cleaner structure for organizing the code.
  * refactor the matrix state and threshold detection members that were declared inline in the back end specifications to separate files.
  * add a new interleaved matrix state back end
  * refactor all of the gpu kernels that were originally in the one backend header file into their own header files
  * write more comprehensive unit tests for the gpu matrix solver back end to test the `interleave` and `reverse_interleave` operations in isolation, as well as ensure that the flat and interleaved back ends produce identical results.
  * add the gpu versions of the kinetic scheme validation tests.

fixes #185


src/backends/fvm_multicore.hpp
```diff
@@ -50,30 +50,38 @@ struct backend {
...
-            p(p), cell_index(cell_index),
-            d(size()), u(size()), rhs(size())
-        {}
-
-        matrix_state(const_iview p, const_iview cell_index, array cap, array cond):
-            p(p), cell_index(cell_index),
-            d(size()), u(size()), rhs(size()),
-            cv_capacitance(std::move(cap)),
-            face_conductance(std::move(cond))
+        //matrix_state(const_iview p, const_iview cell_index, array cap, array cond):
```
bcumming: remove coment

src/backends/matrix_state_interleaved.hpp
```diff
@@ -0,0 +1,248 @@
...
+#include <util/span.hpp>
+#include <util/rangeutil.hpp>
+
+#include "gpu_kernels.hpp"
+
+namespace nest {
+namespace mc {
+namespace gpu {
+
+/// matrix state
```
halfflat: Commenting style: here and below is a mix of blank lines for spacing, blank comments for spacing, `///` and `//`, number of spaces before comment text, and comments as phrases and sentences. This is messy and should be made more consistent.

src/backends/matrix_state_interleaved.hpp
```diff
@@ -0,0 +1,248 @@
...
+    matrix_state_interleaved(const std::vector<size_type>& p,
+                 const std::vector<size_type>& cell_index,
+                 const std::vector<value_type>& cv_cap,
+                 const std::vector<value_type>& face_cond)
+    {
+        // Assert that capacitance and conductance satisfie one of two conditions
+        // * both are defined, with one value for each compartment
+        // * both are empty
+        // Note that the number of compartments is equal to p.size()
+        EXPECTS(   cv_cap.size()==face_cond.size()
```
halfflat: I think it is now safe to require that `cv_cap` et al. be filled out when the constructor is called. As mentioned above, the specification for a matrix state implementation class should be made explicitly in a comment or document.

tests/unit/common.hpp
```diff
@@ -126,8 +126,10 @@ ::testing::AssertionResult seq_almost_eq(Seq1&& seq1, Seq2&& seq2) {
     for (std::size_t j = 0; i1!=e1 && i2!=e2; ++i1, ++i2, ++j) {
         using FP = testing::internal::FloatingPoint<FPType>;
 
-        auto v1 = *i1;
-        auto v2 = *i2;
+        // cast to FPType to avoid warnings about lowering conversion
+        // if FPType has lower precision than Seq{12}::value_type
+        auto v1 = FPType(*i1);
+        auto v2 = FPType(*i2);
```
halfflat: I think the warning was actually providing us a good service; using the `float` test to compare `double`-valued sequences was a quick hack to get around not having a `::testing::seq_near` function. Doing a 4 ulp float test on doubles is violating the semantics implicitly promised by the name (analogous to the gtest names), so we should live with the warning until we implement the correct test.

tests/unit/test_matrix.cu
```diff
@@ -2,64 +2,399 @@
...
+using memory::on_host;
+
+using testing::seq_almost_eq;
+
+using std::begin;
+using std::end;
+
+template <typename T>
+bool operator ==(const std::vector<T>& l, const std::vector<T>& r) {
+    if (l.size()!=r.size()) {
```
halfflat: Can't we use the `std::operator==` for this?

tests/unit/test_matrix.cu
```diff
@@ -2,64 +2,399 @@
...
+        return false;
+    }
+
+    for (auto i=0u; i<l.size(); ++i) {
```
halfflat: This is a bug: test will never terminate if `l.size()` > `UINT_MAX`.

tests/unit/test_matrix.cu
```diff
@@ -2,64 +2,399 @@
...
+    // find the reference interleaved values using host side implementation
+    auto baseline = gpu::interleave_host(values, sizes, starts, BlockWidth, num_mtx, padded_size);
+
+    // find the interleaved values on gpu
+    gpu::interleave<T, I, BlockWidth, LoadWidth>(in.data(), forward.data(), sizes_d.data(), starts_d.data(), padded_size, num_mtx);
+
+    std::vector<T> result_f = assign_from(on_host(forward));
+    std::vector<T> expected = gpu::interleave_host(values, sizes, starts, BlockWidth, num_mtx, padded_size);
+    const auto forward_success = (result_f==expected);
+    EXPECT_TRUE(forward_success);
```
halfflat: `test_interleave` should return a `::testing::AssertionResult` that's tested at the caller's end, or else the callers should use `SCOPED_TRACE` to indicate from where and with what this test is being called. Otherwise, it is hard to determine which invocation of the `test_interleave` call failed.

src/backends/fvm_gpu.hpp
```diff
@@ -96,106 +45,9 @@ struct backend {
...
-
-        // Test if the matrix has the full state required to assemble the
-        // matrix in the fvm scheme.
-        bool has_fvm_state() const {
-            return cv_capacitance.size()>0;
-        }
-    };
+    // matrix back end implementation
+    //using matrix_state = matrix_state_interleaved<value_type, size_type>;
+    using matrix_state = matrix_state_flat<value_type, size_type>;
```
bcumming: Yes, it should.

src/backends/fvm_multicore.hpp
bcumming: good spot Ben! Fixed.

src/backends/fvm_multicore.hpp
```diff
@@ -50,30 +50,38 @@ struct backend {
...
         {
-            auto n = d.size();
-            invariant_d = array(n, 0);
-            for (auto i: util::make_span(1u, n)) {
-                auto gij = face_conductance[i];
-
-                u[i] = -gij;
-                invariant_d[i] += gij;
-                invariant_d[p[i]] += gij;
+            if (has_fvm_state()) {
```
bcumming: I removed the `has_fvm_state` completely, and enforce that capacitance and conductance are provided for each compartment/face.
Ditto for the gpu implementations, and the changes this requires in the unit tests.

src/backends/gpu_kernels.hpp
```diff
@@ -0,0 +1,434 @@
...
+    // The minimum number of bins required to store n values where the bins have
+    // dimension of block_size.
+    __host__ __device__
+    constexpr inline int block_count(int n, int block_size) {
+        return (n+block_size-1)/block_size;
+    }
+
+    // The smallest size of a buffer required to store n items in such that the
+    // buffer has size that is a multiple of block_dim.
+    inline int padded_size (int n, int block_dim) {
```
bcumming: I had it that way to avoid computing `n%block_dim` twice in the same expression. Not for performance reasons: just because it makes it easier to understand the code (in my opinion).
I changed it to `constexpr`... maybe when we move to C++14 we won't need to have this discussion.

src/backends/gpu_kernels.hpp
```diff
@@ -0,0 +1,434 @@
...
+        const int over = n%block_dim;
+        return over ? n+block_dim-over: n;
+    }
+
+    // Placeholders to use for mark padded locations in data structures that use
+    // padding. Using such markers makes it easier to test that padding is
+    // performed correctly.
+    template <typename T> __host__ __device__ constexpr T npos() { return T(); }
```
bcumming: consider it omitted.

src/backends/gpu_kernels.hpp
```diff
@@ -0,0 +1,434 @@
...
+    template <> __host__ __device__ constexpr char npos<char>() { return CHAR_MAX; }
+    template <> __host__ __device__ constexpr unsigned char npos<unsigned char>() { return UCHAR_MAX; }
+    template <> __host__ __device__ constexpr short npos<short>() { return SHRT_MAX; }
+    template <> __host__ __device__ constexpr int npos<int>() { return INT_MAX; }
+    template <> __host__ __device__ constexpr long npos<long>() { return LONG_MAX; }
+    template <> __host__ __device__ constexpr float npos<float>() { return FLT_MAX; }
+    template <> __host__ __device__ constexpr double npos<double>() { return -1.; }
```
bcumming: It is a terrible dummy value. I wish I could provide evidence that -1 never occurs in simulations, but the real reason for this was "because debugging".
It is now `DBL_MAX`, as it should be.

src/backends/gpu_kernels.hpp
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// Data is to be interleaved into blocks of width BlockWidth.
+// The kernel assigns LoadWidth threads to each lane in the block.
+// Hence each thread block is responsible for loading a single block
+// of interleaved matrices.
+
+template <typename T, typename I, int BlockWidth, int LoadWidth, int Threads>
+__global__
+void interleave(
```
bcumming: How about some Fortran style nameing? `flt2intld` anyone? no? OK then.

bcumming: * renamed to `flat_to_interleaved` etc.
* I removed the `_host` suffix. Now gpu and host versions have the same name (but different signatures)
* `num_mtx` makes sense in some contexts (i.e. matrix assembly and solution) and not in others (i.e. interleave operations on vectors). I use `num_vec` where appropriate

src/backends/gpu_kernels.hpp
```diff
@@ -0,0 +1,434 @@
...
+    const T* in, T* out, const I* sizes, const I* starts, int padded_size, int num_mtx)
+{
```
bcumming: And he said "replace all `int` and `auto` with `unsigned`". It compiled! Then he rested.

src/backends/gpu_kernels.hpp
```diff
@@ -0,0 +1,434 @@
...
+}
+
+// A helper that performs the interleave operation on host memory.
+template <typename T, typename I>
+std::vector<T> interleave_host(
+        const std::vector<T>& in,
+        const std::vector<I>& sizes,
+        const std::vector<I>& starts,
+        int block_width, int num_mtx, int padded_length)
+{
```
bcumming: It is only used internally inside the interlave matrix constructor, and as a helper in one of the unit tests. So, I would rather not have a uniform interface + wrapper for such limited (and non-user facing) uses.

src/backends/gpu_kernels.hpp
```diff
@@ -0,0 +1,434 @@
...
+        <<<blocks, Threads>>> (in, out, sizes, starts, padded_size, num_mtx);
+}
+
+/// GPU implementatin of Hines matrix assembly
+/// Flat layout
+/// For a given time step size dt
+///     - use the precomputed alpha and alpha_d values to construct the diagonal
+///       and off diagonal of the symmetric Hines matrix.
+///     - compute the RHS of the linear system to solve
+template <typename T, typename I>
```
bcumming: It is a floating point type. This is consistent throughout pretty much the whole of NestMC. So it is consistent.

That isn't to say that it is necessarily a great naming scheme, however if we rename, we should do so everywhere for consistency. Which is a topic for another PR.

src/backends/gpu_kernels.hpp
```diff
@@ -0,0 +1,434 @@
...
+__global__
+void assemble_matrix_flat(
+        T* d, T* rhs, const T* invariant_d,
+        const T* voltage, const T* current, const T* cv_capacitance,
+        T dt, int n)
+{
+    auto tid = threadIdx.x + blockDim.x*blockIdx.x;
+
+    T factor = 1e-3/dt;
```
bcumming: how about `// 'tis magic`?

halfflat: `// metric system: deep wizardry`

bcumming: Comment added with reference to model documentation.

src/backends/gpu_kernels.hpp
```diff
@@ -0,0 +1,434 @@
...
+
+    const bool do_load  = mtx_id<num_mtx;
+
+    auto load_pos  = do_load? starts[mtx_id] + mtx_lane     : 0;
+    const auto end = do_load? starts[mtx_id] + sizes[mtx_id]: 0;
+    auto store_pos = blk_id*BlockWidth*padded_size + (blk_row*BlockWidth + blk_lane);
+
+    const auto max_size = sizes[0];
+
+    T factor = 1e-3/dt;
```
bcumming: Comment added with reference to model documentation.

src/backends/matrix_state_flat.hpp
```diff
@@ -0,0 +1,121 @@
...
+                 const std::vector<size_type>& cell_index,
+                 const std::vector<value_type>& cv_cap,
+                 const std::vector<value_type>& face_cond):
+        parent_index(memory::make_const_view(p)),
+        cell_index(memory::make_const_view(cell_index)),
+        d(p.size(), impl::npos<T>()),
+        u(p.size()),
+        rhs(p.size(), impl::npos<T>()),
+        cv_capacitance(memory::make_const_view(cv_cap))
+    {
```
bcumming: It is perfectly reasonable.

bcumming: But you won't like the solution! Instead of having all the files with `_gpu` and `_multicore` in their name, I made separate paths. Things were already messy before this suggested refactoring.

Now all the GPU-specific code is in `backends/gpu`. This doesn't exactly reflect the `nest::mc::gpu` and `nest::mc::multicore` namespace structure... maybe we could remove the `multicore` path?

bcumming: I also refactor the `matrix_state` and `threshold_watcher` implementations into their own files, because as you note, the `backend_{gpu, multicore}` definition had too many inline definitions.

halfflat: Regarding namespaces and paths, let's go with anything that is consistent for now, and consider a possible rearrangement in a separate PR.

tests/unit/test_matrix.cu
bcumming: Well, that is a thing! Sure, why not?

tests/unit/test_matrix.cu
bcumming: no longer applicable, because `std::operator==`

tests/unit/test_matrix.cu
bcumming: It now returns `::testing::Assertionresult`

src/backends/gpu/matrix_state_flat.hpp
```diff
@@ -0,0 +1,116 @@
...
+            invariant_d_tmp[i] += gij;
+            invariant_d_tmp[p[i]] += gij;
+        }
+        invariant_d = make_const_view(invariant_d_tmp);
+        u = make_const_view(u_tmp);
+
+        solution = rhs;
+    }
+
+    int num_matrices() const {
```
halfflat: This still has the potential problem of reducing the range; safer to return the same type as `cell_index.size()` (with a check for zero size) or have a separate size field that is properly initialized in the constructor.

src/backends/gpu/matrix_state_interleaved.hpp
```diff
@@ -0,0 +1,255 @@
...
+
+        //
+        //  Calculate the invariant part of the matrix diagonal and the
+        //  upper diagonal on the host, then copy to the device.
+        //
+
+        std::vector<value_type> invariant_d_tmp(p.size(), 0);
+        std::vector<value_type> u_tmp(p.size(), 0);
+        auto face_conductance_tmp = memory::on_host(face_cond);
+        for(auto i: util::make_span(1u, p.size())) {
```
halfflat: Space after `for`.

src/backends/gpu/matrix_state_interleaved.hpp
```diff
@@ -0,0 +1,255 @@
+#pragma once
```
halfflat: Inconsistent commenting style in the source (capitalization, sentence fragments, blank lines etc.).

src/backends/gpu/matrix_state_interleaved.hpp
```diff
@@ -0,0 +1,255 @@
...
+
+        // helper that converts to interleaved format on the host, then copies to device
+        // memory, for use as an rvalue in an assignemt to a device vector.
+        auto interleave = [&] (std::vector<T>const& x) {
+            return memory::on_gpu(
+                flat_to_interleaved(x, sizes_p, cell_index_p, block_dim(), num_mtx, padded_size));
+        };
+        u           = interleave(u_tmp);
+        invariant_d = interleave(invariant_d_tmp);
+        cv_capacitance = interleave(cv_cap);
```
halfflat: Inconsistent spacing/alignment above.

src/backends/gpu/matrix_state_interleaved.hpp
```diff
@@ -0,0 +1,255 @@
...
+        constexpr auto bd = impl::block_dim();
+        constexpr auto lw = impl::load_width();
+        constexpr auto block_dim = bd*lw;
+
+        // the number of threads is threads_per_matrix*num_mtx
+        const auto num_blocks = impl::block_count(num_matrices()*lw, block_dim);
+
+        assemble_matrix_interleaved<value_type, size_type, bd, lw, block_dim>
+            <<<num_blocks, block_dim>>>
+            ( d.data(), rhs.data(), invariant_d.data(),
```
halfflat: Code style: inconsistent spacing. (Space after opening parenthesis, no space before closing; elsewhere we have no space after opening parentheses.)

src/backends/multicore/matrix_state.hpp
```diff
@@ -0,0 +1,113 @@
...
+    array face_conductance;    // [μS]
+
+    // the invariant part of the matrix diagonal
+    array invariant_d;         // [μS]
+
+    const_view solution;
+
+    matrix_state() = default;
+
+    matrix_state( const std::vector<size_type>& p,
```
halfflat: Code style: no space after open paren.

src/backends/gpu/matrix_state_interleaved.hpp
```diff
@@ -0,0 +1,255 @@
...
+
+        matrix_sizes = memory::make_const_view(sizes_p);
+        matrix_index = memory::make_const_view(cell_index_p);
+
+        // allocate space for storing the un-interleaved solution
+        solution = array(p.size());
+    }
+
+    // the number of matrices stored in the matrix state
+    int num_matrices() const {
```
halfflat: Implicit narrowing (see comment about `num_matrices()` above.)

src/backends/multicore/matrix_state.hpp
```diff
@@ -0,0 +1,113 @@
...
+            rhs[first] /= d[first];
+
+            // forward sweep
+            for(auto i=first+1; i<last; ++i) {
+                rhs[i] -= u[i] * rhs[parent_index[i]];
+                rhs[i] /= d[i];
+            }
+        }
+    }
+};
```
halfflat: Unlike other matrix state implementations, this one does not have a `num_matrices()` method. Should it?

tests/unit/common.hpp
halfflat: I still think that the casts here should be reverted.

tests/unit/test_matrix.cu
```diff
@@ -2,64 +2,386 @@
...
+
+    // backward will hold the result of reverse interleave on the GPU
+    auto backward = memory::device_vector<T>(values.size(), npos<T>());
+    gpu::interleaved_to_flat<T, I, BlockWidth, LoadWidth>(forward.data(), backward.data(), sizes_d.data(), starts_d.data(), padded_size, num_mtx);
+
+    std::vector<T> result_b = assign_from(on_host(backward));
+
+    // we expect that the result of the reverse permutation is the original input vector
+    const auto backward_success = (result_b==values);
+    EXPECT_TRUE(backward_success);
```
halfflat: This also should be returning a `::testing::AssertionResult` similar to line 63.

tests/unit/test_matrix.cu
```diff
@@ -2,64 +2,386 @@
...
+        // find the start position of each matrix in the flat storage
+        // we are assuming that the matrices are unpermuted
+        ivec starts(num_mtx, 0);
+        std::partial_sum(begin(sizes), end(sizes)-1, begin(starts)+1);
+
+        tvec values(padded_size*num_mtx);
+        std::iota(values.begin(), values.end(), 0);
 
-    using nest::mc::util::make_span;
+        test_interleave<T, I, 1, 1>(sizes, starts, values, padded_size);
```
halfflat: The assertion results from `test_interleave` calls need to be tested with `EXPECT_TRUE` or similar.

src/backends/fvm.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
-#include "fvm_multicore.hpp"
+#include "./multicore/fvm.hpp"
```
bcumming: agreed

src/backends/gpu/kernels/solve_matrix.hpp
```diff
@@ -0,0 +1,78 @@
...
+{
+    auto tid = threadIdx.x + blockDim.x*blockIdx.x;
+
+    if (tid<num_mtx) {
+        // get range of this thread's cell matrix
+        const auto first = cell_index[tid];
+        const auto last  = cell_index[tid+1];
+
+        // backward sweep
+        for(auto i=last-1; i>first; --i) {
```
bcumming: agreed

src/backends/gpu/matrix_state_flat.hpp
bcumming: I have updated the precondition such that `cell_index.size() > 1u`, which is reasonable. And I also return `unsigned` (if we ever have more than 2^32-1 matrices in a cell group...)

src/backends/gpu/matrix_state_interleaved.hpp
bcumming: Yes.

I have cleaned it up, but I do like the style

```
//
// Stage X
///
```
to mark out regions in longer functions.

src/backends/multicore/matrix_state.hpp
bcumming: I will make the other `num_matrices` methods private, because they are used internally by the other implementations.

src/backends/gpu/matrix_state_flat.hpp
halfflat: If `cell_index.size()>1` is being explicitly checked as precondition, might as well add a check also for `cell_index.size()<=UINT_MAX` (or `UINT_MAX+1`, I guess.)

Note that the `matrix_state_flat::size()` method returns `std::size_t` though. Is there a good reason not to have `num_matrices()` return `std::size_t`?

src/backends/multicore/matrix_state.hpp
bcumming: privatised

bcumming:
I don't think that we have a problem with the relative paths being relative to anything other than the file in which the `#include` statement is made.

```
======== ./include/detail/dog.hpp ========
#pragma once

template <typename T>
T dog() {
    return T();
}

======== ./include/detail/fish.hpp ========
#pragma once

#include "dog.hpp"

======== ./include/foo.hpp ========
#pragma once

#include "detail/fish.hpp"

======== ./main.cpp ========
#include <foo.hpp>

int main(void) {
    return dog<int>();
}
bcumming@arapiles:headers > g++ -Iinclude main.cpp && ./a.out
```

Issue #184, opened 2017-03-14
Implement GPU events II [16]


halfflat:
Following the work of #183, move event queue behaviour to device side for large-step (mindelay/2) periods, allowing desynchronized simulation time across cells.

1. Add infrastructure for per-cell simulation time and time step.
2. Implement device-side per-cell event queues.
3. Implement `cell_group` plumbing and device-side delivery.

## Implementation plan.

The goal is to have event queuing and sampler data delivery performed on the `cell_group` side, and event delivery and probe data capture performed on the lowered cell side. Probe data vs. sampler delivery will be handled in another ticket, but the implementation of this work should anticipate this future work.

There are four proposed components of the implementation:

1. Implement explicit event binning.

    This is required for the split in event handling responsibility between the `cell_group` and the lowered cell. This also addresses the problem of events modifying the time for other events on different cells in the same group. The structure for performing the binning was added in PR #204.

2. Refactor interface between `cell_group` and lowered cell for asyncronous cell state.

    While leaving the underlying behaviour and semantics intact, modify the API presented to `cell_group` by the lowered cell implementation to allow for the fact that outside of a large-step integration period, the time associated with the state of each cell in the cell group might differ.

    This API change will also implement the division in responsibility for event handling outlined above.

3. Push time and integration step period state to mechanisms.

    Modify the `mechanism` API and `modcc`-generated instances to use a vector of time and _dt_ (or rather, of current time and integration step stop time). These vectors will have one element per cell in the cell group.

    Use this desynchronized state to perform per-cell time-stepping and event delivery. Integration across a step or large-step will be over the half-open time interval; events at the end point of the interval will effect the discontinuous state changes in the following integration step.

4. Move event delivery state to the back-end.

    Perform event delivery as the first item of work in `nrn_state` mechanism integration. For the GPU back-end in particular, this will involve the use and implementation of a GPU-specific event list data structure.

halfflat:
This constitutes PR #281 up for review on the `devel/async-integration` branch.

halfflat:
PR #281 merged into `devel/async-integration` branch; merge of most recent master into that branch is PR #284.

PR #210, opened 2017-03-27
Implementation of asynchronous integration API.

halfflat:
* Stage events for next integration interval on lowered cell.
* Use explicit binning for event coalescence.
* Extend `event_queue` to allow checking top of queue against arbitrary predicates.
* Add `--bin-dt` and `--bin-regular` options to miniapp (disable binning with `--bin-dt 0`).
* Tidy up miniapp option settings class.

Integration in lowered cell over multiple steps is deferred until samplers can be set up with back-end polling.

Asynchronous integration itself is not yet implemented.

### DIFF
--- a/miniapp/io.cpp
+++ b/miniapp/io.cpp
@@ -111,46 +111,13 @@ static void update_option(util::optional<T>& opt, const nlohmann::json& j, const
 
 // Read options from (optional) json file and command line arguments.
 cl_options read_options(int argc, char** argv, bool allow_write) {
-
-    // Default options:
-    const cl_options defopts{
-        1000,       // number of cells
-        500,        // synapses_per_cell
-        "expsyn",   // synapse type
-        100,        // compartments_per_segment
-        100.,       // tfinal
-        0.025,      // dt
-        false,      // all_to_all
-        false,      // ring
-        1,          // group_size
-        false,      // probe_soma_only
-        0.0,        // probe_ratio
-        "trace_",   // trace_prefix
-        util::nothing,  // trace_max_gid
-        util::nothing,  // morphologies
-        false,      // morph_rr;
-        false,      // report_compartments;
-
-        // spike_output_parameters:
-        false,      // spike output
-        false,      // single_file_per_simulation
-        true,       // Overwrite outputfile if exists
-        "./",       // output path
-        "spikes",   // file name
-        "gdf",      // file extension
-
-        // dry run parameters:
-        1,          // default dry run size
-
-        // Turn on/off profiling output for all ranks
-        false
-    };
-
     cl_options options;
     std::string save_file = "";
 
     // Parse command line arguments.
     try {
+        cl_options defopts;
+
         CustomCmdLine cmd("nest mc miniapp harness", "0.1");
 
         TCLAP::ValueArg<std::string> ifile_arg(
@@ -179,6 +146,11 @@ cl_options read_options(int argc, char** argv, bool allow_write) {
         TCLAP::ValueArg<double> dt_arg(
             "d", "dt", "set simulation time step to <time> ms",
             false, defopts.dt, "time", cmd);
+        TCLAP::ValueArg<double> bin_dt_arg(
+            "", "bin-dt", "set event binning interval to <time> ms",
+            false, defopts.bin_dt, "time", cmd);
+        TCLAP::SwitchArg bin_regular_arg(
+            "","bin-regular","use 'regular' binning policy instead of 'following'", cmd, false);
         TCLAP::SwitchArg all_to_all_arg(
             "m","alltoall","all to all network", cmd, false);
         TCLAP::SwitchArg ring_arg(
@@ -215,8 +187,6 @@ cl_options read_options(int argc, char** argv, bool allow_write) {
         cmd.reorder_arguments();
         cmd.parse(argc, argv);
 
-        options = defopts;
-
         std::string ifile_name = ifile_arg.getValue();
         if (ifile_name != "") {
             // Read parameters from specified JSON file first, to allow
@@ -232,6 +202,8 @@ cl_options read_options(int argc, char** argv, bool allow_write) {
                     update_option(options.syn_type, fopts, "syn_type");
                     update_option(options.compartments_per_segment, fopts, "compartments");
                     update_option(options.dt, fopts, "dt");
+                    update_option(options.bin_dt, fopts, "bin_dt");
+                    update_option(options.bin_regular, fopts, "bin_regular");
                     update_option(options.tfinal, fopts, "tfinal");
                     update_option(options.all_to_all, fopts, "all_to_all");
                     update_option(options.ring, fopts, "ring");
@@ -275,6 +247,8 @@ cl_options read_options(int argc, char** argv, bool allow_write) {
         update_option(options.compartments_per_segment, ncompartments_arg);
         update_option(options.tfinal, tfinal_arg);
         update_option(options.dt, dt_arg);
+        update_option(options.bin_dt, bin_dt_arg);
+        update_option(options.bin_regular, bin_regular_arg);
         update_option(options.all_to_all, all_to_all_arg);
         update_option(options.ring, ring_arg);
         update_option(options.group_size, group_size_arg);
@@ -315,6 +289,8 @@ cl_options read_options(int argc, char** argv, bool allow_write) {
                 fopts["syn_type"] = options.syn_type;
                 fopts["compartments"] = options.compartments_per_segment;
                 fopts["dt"] = options.dt;
+                fopts["bin_dt"] = options.bin_dt;
+                fopts["bin_regular"] = options.bin_regular;
                 fopts["tfinal"] = options.tfinal;
                 fopts["all_to_all"] = options.all_to_all;
                 fopts["ring"] = options.ring;
@@ -358,6 +334,9 @@ std::ostream& operator<<(std::ostream& o, const cl_options& options) {
     o << "  synapses/cell        : " << options.synapses_per_cell << "\n";
     o << "  simulation time      : " << options.tfinal << "\n";
     o << "  dt                   : " << options.dt << "\n";
+    o << "  binning dt           : " << options.bin_dt << "\n";
+    o << "  binning policy       : " <<
+        (options.bin_dt==0? "none": options.bin_regular? "regular": "following") << "\n";
     o << "  all to all network   : " << (options.all_to_all ? "yes" : "no") << "\n";
     o << "  ring network         : " << (options.ring ? "yes" : "no") << "\n";
     o << "  group size           : " << options.group_size << "\n";
--- a/src/cell_group.hpp
+++ b/src/cell_group.hpp
@@ -153,7 +153,7 @@ public:
             lowered_.add_event(binned_ev_time, handle, ev->weight);
         }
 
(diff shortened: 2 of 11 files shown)


w-klijn:
Would it be smart to rename the merge? something like prepare for asynch..

Just because the last line of the description invalidates the title completely.

src/cell_group.hpp
```diff
@@ -143,51 +143,61 @@ class cell_group {
     }
 
     void advance(time_type tfinal, time_type dt) {
-        while (lowered_.time()<tfinal) {
-            // take any pending samples
-            time_type cell_time = lowered_.time();
+        EXPECTS(lowered_.state_synchronized());
```
w-klijn: This function is getting very long: Would it be smart to split it in multiple steps to improve readability?

src/cell_group.hpp
```diff
@@ -231,7 +241,7 @@ class cell_group {
         auto handle = get_probe_handle(probe_id);
 
         auto sampler_index = uint32_t(samplers_.size());
-        samplers_.push_back({handle, s});
+        samplers_.push_back({handle, probe_id.gid, s});
```
w-klijn: The interface of this constructor would be nicer if only the probe_id would be given (instead of the gid).

src/event_queue.hpp
```diff
@@ -85,6 +103,22 @@ public :
...
+    // Pop and return top event `ev` of queue if `t_until` > `event_time(ev)`.
+    util::optional<value_type> pop_if_before(const time_type& t_until) {
+        using ::nest::mc::event_time;
+        return pop_if(
+            [&t_until](const value_type& ev) { return t_until > event_time(ev); }
+        );
+    }
+
+    // Pop and return top event `ev` of queue unless `event_time(ev)` > `t_until`
+    util::optional<value_type> pop_if_not_after(const time_type& t_until) {
```
w-klijn: negation in function names can be confusing. Is there a way to rename it?

halfflat: Here I think it is unavoidable; because we're limiting ourselves to the use of a strict inequality `>` on time types, we're making a distinction between 'before' (i.e. time is less than top of queue) and 'not after' (i.e. time is not greater than top of queue).

An alternative, `pop_if_before_or_same`, would imply we're testing for equality or greater-than-equal-to, when in truth we're still just using greater-than.

src/cell_group.hpp
halfflat: Ignoring the extra comments, the function is pretty much the same length as before.

That said, it will be much simpler when sampling is also pushed to the back-end, so I would defer a refactoring of this method until then.

src/cell_group.hpp
halfflat: We only need the gid though…

w-klijn: With this implementation the cell_group knows 'that the probe has a gid'.  tight coupling

src/cell_group.hpp
```diff
@@ -143,51 +143,61 @@ class cell_group {
...
+
+        // Bin pending events and enqueue on lowered state.
+        time_type ev_min_time = lowered_.max_time(); // (but we're synchronized here)
+        while (auto ev = events_.pop_if_before(tfinal)) {
+            auto handle = get_target_handle(ev->target);
+            auto binned_ev_time = binner_.bin(ev->target.gid, ev->time, ev_min_time);
+            lowered_.add_event(binned_ev_time, handle, ev->weight);
+        }
+
+        lowered_.start_integration(tfinal, dt);
```
bcumming: The name is a bit misleading, maybe `prepare_integration` or `setup_integration` are more appropriate.

If I understand the comments and your plans correctly, the event delivery and sampler polling will both be moved to the lowered cells, and thus be able to perform time stepping to `t_final` with one call?

In that case, would we need to have a separate `start_integration` and `step_integration` methods?

src/cell_group.hpp
bcumming: @w-klijn is right, in that this function is a bit hard to follow. Sometimes a longer function can make complicated code easier to understand, but in this case things like the sampling logic could be refactored.

But this is a work in progress, it should be cleaner once much of the event and sampling logic has been moved to the back end.

src/cell_group.hpp
```diff
@@ -143,51 +143,61 @@ class cell_group {
     }
 
     void advance(time_type tfinal, time_type dt) {
-        while (lowered_.time()<tfinal) {
-            // take any pending samples
-            time_type cell_time = lowered_.time();
+        EXPECTS(lowered_.state_synchronized());
+
+        // Bin pending events and enqueue on lowered state.
```
bcumming: Is this covered by the profiler? Try to make sure that the profiler is catching everything once the refactoring is complete.

src/fvm_multicell.hpp
```diff
@@ -85,8 +85,56 @@ class fvm_multicell {
         return (this->*h.first)[h.second];
     }
 
-    /// integrate all cell state forward in time
-    void advance(double dt);
+    /// Initialize state prior to a sequence of integration steps.
+    void start_integration(value_type tfinal, value_type dt_max) {
```
bcumming: The name `start_integration` gives the impression that integration will be performed when called.

src/fvm_multicell.hpp
```diff
@@ -85,8 +85,56 @@ class fvm_multicell {
...
+
+    /// Advance one integration step.
+    void step_integration();
+
+    /// Query integration completion state.
+    bool integration_complete() const {
+        return !integration_running_;
+    }
+
+    /// Query per-cell time state. (Placeholders)
```
bcumming: Placeholders for what? Now would be a good time to have more detailed comments that explain the meaning of each returned value, to make it easier for me to see where this is going.

src/fvm_multicell.hpp
```diff
@@ -85,8 +85,56 @@ class fvm_multicell {
...
+        EXPECTS(events_.empty());
+        for (auto& ev: staged_events_) {
+            EXPECTS(ev.time<tfinal);
+            events_.push(std::move(ev));
+        }
+
+        staged_events_.clear();
+    }
+
+    /// Advance one integration step.
```
bcumming: `// this is a loop`
sorry.

src/fvm_multicell.hpp
```diff
@@ -85,8 +85,56 @@ class fvm_multicell {
...
+
+    value_type max_time() const {
+        return t_;
+    }
+
+    bool state_synchronized() const {
+        return true;
+    }
+
+    /// Add an event for processing in next integration stage.
```
bcumming: Do the events have to be processed in the next integration state? Is it true that an event that will occur in a later integration stage could be added, and will be persistent in the queue over multiple stages, until it is time to deliver it?

src/fvm_multicell.hpp
```diff
@@ -194,15 +240,38 @@ class fvm_multicell {
...
-    value_type t_ = value_type{0};
+    value_type t_ = 0;
 
     /// resting potential (initial voltage condition)
     value_type resting_potential_ = -65;
 
+    /// final time in integration round [ms]
+    value_type tfinal_ = 0;
+
+    /// max time step for integration [ms]
```
bcumming: A quibble about naming: this could also be referred to as the `target_dt`, i.e. the time step size that we would like to take, events and `t_final` permitting. The comment might want to mention this.

src/fvm_multicell.hpp
```diff
@@ -194,15 +240,38 @@ class fvm_multicell {
...
+    /// once integration to `tfinal_` is complete.
+    bool integration_running_ = false;
+
+    struct deliverable_event {
+        double time;
+        target_handle handle;
+        double weight;
+    };
+
+    /// events staged for upcoming integration stage
```
bcumming: see comment about: choose to insert either `an` or `the` in front of `upcoming` in this comment.

src/cell_group.hpp
bcumming: Why the additional parameter?

halfflat: The cell group needs to know the gid associated with the probe, so it can test the sampler's requested time against the time for that gid. It is a consequence of the asynchronous behaviour.

miniapp/io.cpp
```diff
@@ -179,6 +146,11 @@ cl_options read_options(int argc, char** argv, bool allow_write) {
         TCLAP::ValueArg<double> dt_arg(
             "d", "dt", "set simulation time step to <time> ms",
             false, defopts.dt, "time", cmd);
+        TCLAP::ValueArg<double> bin_dt_arg(
+            "", "bin-dt", "set event binning interval to <time> ms",
+            false, defopts.bin_dt, "time", cmd);
+        TCLAP::SwitchArg bin_regular_arg(
+            "","bin-regular","use 'regular' binning policy instead of 'following'", cmd, false);
```
halfflat: Totally want to rip out our option handling infrastructure and replace. _I have a plan_.

In the meantime, the hypens-in-long-options style is pretty common (it is certainly GNU style). Google test's underscore use is not standard practice. I think we should stick with it. We could modify the JSON keys to match, when we get around to this clean up.

src/cell_group.hpp
halfflat: Happy to change to `init_integration` or similar. It does, in some sense, get things going: 'prepare' would be a bit misleading.

Either way though, indeed once sampling is also performed at the lowered level there will be no need to have separate methods here.

src/fvm_multicell.hpp
halfflat: I didn't think it was really that transparent — didn't want to use `advance` in the method name to avoid confusion with the previous method and semantics, and I thought the name in isolation was a little ambiguous.

src/fvm_multicell.hpp
halfflat: The assumption here is no: all the events for one 'big' interval and no others are expected to be passed on to the lowered cell. It may be possible to relax this assumption, but I wouldn't do so unless there is a clear gain.

The `cell_group` is responsible for queuing, and holds the priority queue data structure. The lowered implementation now has responsibility for delivering, and the API is meant to enforce a sequence of:
1. `cell_group` provides all the events, associated with their gid.
2. `cell_group` instructs lowered implementation to bulk-load/prepare the provided events for delivery within the integration interval.
3. The lowered implementation uses a pop/top-only queue for handling the discrete deliveries internally over the course of processing an integration interval.

src/fvm_multicell.hpp
halfflat: `dt_max_` is closer to the meaning associated with the interface (i.e. the `cell_group` requests a maximum _dt_). A `target_dt` is closer to the meaning as provided by this lowered cell implementation, as you describe.

A comment on how `dt_max_` is satisfied is more appropriate for the integration code, and I can add one there.

halfflat:
I will update the corresponding ticket to outline the plan of PRs, with the hope of making the goal of these patches more transparent.

halfflat:
Made a small update with a change of `start_integration` to `setup_integration`, and a comment regarding `dt_max_`.

I've updated issue #184 with a more detailed implementation plan.

miniapp/io.cpp
bcumming: I prefer the hyphen approach too, and second the motion to use hyphens in the json files.

bcumming:
Looks good to me. On the condition that the "placeholder" comments turn into documentation in the final version.

PR #217, opened 2017-04-03
Metering support with time meter

bcumming:
* An abstract `meter` class that defines interface for taking a reading, and returning the meter results as a json object.
  * A `time_meter` implwementation of the `meter` that measures wall time.
  * To generate metering reports with global information, the global communication policy interfaces were extended to support `gather` and `barrier` operations. These are trivial for serial and dry run policies, and wrap the appropriate MPI calls for the MPI policy.
  * a `meter_manager` type that stores a list of meters was created
    * will also have memory and power meters soon.
  * a meter manager was added to the miniapp and now records startup, model initialization, time stepping and final file io times.


src/profiling/meter_manager.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // TODO
+};
+
+void meter_manager::checkpoint(std::string name) {
+    checkpoint_names_.push_back(std::move(name));
+
+    // Enforce a global synchronization point the first time that the meters
+    // are used, to ensure that times measured across all domains are
+    // synchronised.
+    if (meters_.size()==0) {
```
w-klijn: I dont get this test: Should you no do the barrier when you have meters?

bcumming: erm, good catch. That should be `if (checkpoint_names_.size()==1)`, because we only want to do the synchronization if this is the first checkpoint.

miniapp/miniapp.cpp
```diff
@@ -101,15 +107,13 @@ int main(int argc, char** argv) {
             report_compartment_stats(*recipe);
         }
 
-        // inject some artificial spikes, 1 per 20 neurons.
-        std::vector<cell_gid_type> local_sources;
+        // Inject some artificial spikes, 1 per 20 neurons.
         cell_gid_type first_spike_cell = 20*((cell_range.first+19)/20);
         for (auto c=first_spike_cell; c<cell_range.second; c+=20) {
-            local_sources.push_back(c);
```
w-klijn: How is the deletion of this change related to the time meter? Would it be better to do this in a separate pr.

src/communication/serial_global_policy.hpp
```diff
@@ -52,6 +52,13 @@ struct serial_global_policy {
         return value;
     }
 
+    template <typename T>
```
w-klijn: Api documentation?

miniapp/miniapp.cpp
bcumming: To get reliable results from the metering the "dummy" step of the model to initialize the profiler had to be removed. This simplified the code a bit, and it was no longer necessary to record the seed spikes in a `local_sources` vector for re-insertion after the dummy step.

bcumming: So removing this was directly related to adding the meters.

miniapp/miniapp.cpp
```diff
@@ -50,6 +51,9 @@ int main(int argc, char** argv) {
     nest::mc::communication::global_policy_guard global_guard(argc, argv);
 
     try {
+        nest::mc::util::meter_manager meters;
+        meters.checkpoint("start");
```
w-klijn: Do I understand that in the current implementation we always measure these timings?

src/communication/dryrun_global_policy.hpp
```diff
@@ -74,6 +74,13 @@ struct dryrun_global_policy {
         return size()*value;
     }
 
+    template <typename T>
```
w-klijn: SHould this also be added to the Global_policy interface.
It also might need some api description/comment

src/profiling/meter_manager.cpp
w-klijn: I might mis understand but, you only have a barrier here when there are NO timers

src/profiling/meter_manager.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    // add memory consumption meter
+    // TODO
+
+    // add energy consumption meter
+    // TODO
+};
+
+void meter_manager::checkpoint(std::string name) {
+    checkpoint_names_.push_back(std::move(name));
```
w-klijn: THe checkpoint name vector is shared between the different meters and you can have duplicate names. Might this cause problems later on?

src/profiling/profiler.cpp
```diff
@@ -51,9 +47,9 @@ namespace gpu {
 }
 #endif
 
-/////////////////////////////////////////////////////////
-// profiler_node
-/////////////////////////////////////////////////////////
+//
+// profiler_node implementation
```
w-klijn: Could use a more descriptive comment: Now its not saying anything

src/profiling/profiler.cpp
```diff
@@ -234,9 +230,9 @@ profiler_node region_type::populate_performance_tree() const {
     return tree;
 }
 
-/////////////////////////////////////////////////////////
-// region_type
-/////////////////////////////////////////////////////////
+//
+// profiler implementation
```
w-klijn: Idem

miniapp/miniapp.cpp
bcumming: No, we don't. The current implementation just measures the time spent in the different parts of the time stepping.
This is higher level: providing time spent in initialization, model setup, model simulation, and shutdown.

src/communication/dryrun_global_policy.hpp
bcumming: it has been added to the `global_policy` interface. Each of the policy implementations (dry run, serial and mpi) expose a similar call.
We don't have any documentation for the other calls in the API. The whole thing should be documented properly.
I will add a comment that explains what this call does.

src/profiling/meter_manager.cpp
bcumming: The meters do need not know about checkpoint names; they simply take measurements when requested.
I don't know if it is a good or bad thing that names could be repeated. We would have to think about how such an infrastructure could be used.

src/profiling/profiler.cpp
bcumming: it is the implementation of the `profiler_node`. What more is their to say?

The header file where the type interface is defined is the appropriate spot for documentation (not that there is much). The `profiler_node` is actually a low level, strictly back end thing that doesn't need full API-style documentation.

miniapp/miniapp.cpp
```diff
@@ -163,7 +171,7 @@ int main(int argc, char** argv) {
 
         meters.checkpoint("output");
 
-        meters.save_to_file("meters.json");
+        util::save_to_file(meters, "meters.json");
```
w-klijn: Nice, should indeed not be the responsibility of the meter!

src/profiling/meter_manager.cpp
```diff
@@ -18,32 +16,34 @@ meter_manager::meter_manager() {
...
 
 void meter_manager::checkpoint(std::string name) {
-    checkpoint_names_.push_back(std::move(name));
+    checkpoint_names.push_back(std::move(name));
 
     // Enforce a global synchronization point the first time that the meters
     // are used, to ensure that times measured across all domains are
     // synchronised.
-    if (meters_.size()==0) {
+    if (meters.size()==0) {
```
w-klijn: Am I correct that this is not meters_ because it is a public member?

bcumming: Yes. And you would also be correct to point out that I still have not fixed the bug you found in the first review!

bcumming:
Adding some unit tests that test the meters is a good idea. It is also hard to do right. For example, we could allocate some memory and test if the memory meter correctly detects the allocation. However, we don't have any control over how `malloc` performs the allocation: it could grab a large chunk of memory, and pass us a part of that memory, keeping the rest of the allocation for later allocations. The memory meter will detect the allocation of the large chunk, not just the small chunk. So, you have to take quite some time to come up with unit tests that account for such things.

PR #220, opened 2017-04-04
Feature/memory meter

### LINKED ISSUE #189: Support memory usage reporting [4]
Add simple checkpointing for memory usage to the profiler.

 1. Implement low-level memory consumption testing code.
 2. Global data structure for recording checkpoint information.
 3. Simple interface (see `memory_checkpoint()` in pseudo-code below.
 4. Communicator support for collecting global memory usage information.
 5. Output of memory consumption report to file.

Basic workflow pseudo code:

```
// start of application

memory_checkpoint("start");
model.initialize();
memory_checkpoint("initialized");
model.run();
memory_checkpoint("run");

// shutdown
```

bcumming:
Add automatic memory usage meter.

When compiled for the GPU the amount of memory used on the GPU is also reported.

fixes #189


src/profiling/memory_meter.cpp
```diff
@@ -0,0 +1,64 @@
...
+    measurement collate(const std::vector<memory_size_type>& readings, std::string name) {
+        using gcom = communication::global_policy;
+
+        // Calculate the local change in allocated memory for each interval.
+        std::vector<memory_size_type> allocated;
+        allocated.push_back(0);
+        for (auto i=1u; i<readings.size(); ++i) {
+            allocated.push_back(readings[i] - readings[i-1]);
+        }
+
```
halfflat: In this particular application not a problem, but I think it's unsafe practice to use an `unsigned` to count up to a `size_t` value. A safe alternative would be to use `std::adjacent_difference` here and then assign zero to `allocated[0]`.

src/profiling/memory_meter.cpp
```diff
@@ -0,0 +1,64 @@
...
+                "the number of checkpoints in the \"memory\" meter do not match across domains");
+        }
+
+        // Gather allocations from across all of the domains onto the root domain.
+        // Note: results are only valid on the root domain on completion.
+        measurement results;
+        results.name = std::move(name);
+        results.units = "kB";
+        for (auto m: allocated) {
+            results.measurements.push_back(gcom::gather(std::round(m/1e3), 0));
```
halfflat: Why round here? We have (up to the resolution of `double`) exact measurements of bytes at this point. Rounding to nearest kilobyte seems like a presentation issue.

src/profiling/meter_manager.cpp
```diff
@@ -9,22 +9,23 @@ meter_manager::meter_manager() {
     meters.emplace_back(new time_meter());
 
     // add memory consumption meter
-    // TODO
+    if (has_memory_metering) {
+        meters.emplace_back(new memory_meter());
+    }
```
halfflat: If the `meter_manager` constructor has responsibility for setting up the `meters` vector, and these `meters` are not meant for public consumption, then `meters` should be private.

src/profiling/meter_manager.cpp
```diff
@@ -9,22 +9,23 @@ meter_manager::meter_manager() {
...
 };
 
 void meter_manager::checkpoint(std::string name) {
-    checkpoint_names.push_back(std::move(name));
-
     // Enforce a global synchronization point the first time that the meters
     // are used, to ensure that times measured across all domains are
     // synchronised.
-    if (meters.size()==0) {
+    if (checkpoint_names.size()==0) {
```
halfflat: Correct behaviour of this method relies upon the public state of the struct members (i.e. `checkpoint_names`) having not been modified. If `checkpoint_names` remains public, then a private initialized flag of some sort should be added to keep this safe.

src/profiling/memory_meter.cpp
bcumming: I see your point. This block of code will be abstracted out in the following PR, and I will fix it there.

src/profiling/memory_meter.cpp
bcumming: Next PR gives the result in bytes.

src/profiling/meter_manager.cpp
bcumming: fixed.

PR #221, opened 2017-04-04
Add `cc-filter` line-based filtering script

halfflat:
* Add perl program `cc-filter`, a general purpose by-line text filter with built-in default rules for filtering text containing C++ types and expressions.
* Add documentation for the tool to the scripts `README.md` file.
* Add demonstration table `filters/massif-strip-cxx` for using `cc-filter` with valgrind massif output.


scripts/README.md
```diff
@@ -90,10 +91,143 @@ are reported instead.
...
+    # a comment comprises a # and any following characters, plus any
+    # preceding whitespace.
+    macro non-comment (^.*?)(?=\s*(?:#|$))
+    rule rev-text s/%non-comment%/$1=~s,[[:punct:]]+,,gr/e
+    rule rev-text s/%non-comment%/reverse(lc($1))/e
+
+This defines one rule, `rev-text` which will remove punctuation in the
+text preceding a possible comment, and then lower-case and reverse it.
+
+    $ echo 'What, you egg!  # ?!' | cc-filter -n --table example
```
bcumming: I would rather that the `--table` argument took the full table file name `example.tbl`, instead of implicitly adding it.

scripts/README.md
```diff
@@ -90,10 +91,143 @@ are reported instead.
...
+    gge uoy tahw  # ?!
+
```
bcumming: This file needs some more examples of using the script for its intended purpose: filtering C++ gunk from files.
Lots of examples.
I am not a perl guy, and I am not a regex guru either. So if I am ever going to use this I want recipes for filtering my files front and center.

scripts/README.md
halfflat: It does; this is in fact a documentation bug. Will fix.

bcumming:
lgtm

PR #222, opened 2017-04-06
Feature/power meter

### LINKED ISSUE #190: Support Power consumption reporting [2]
Add reporting for power consumption on Cray systems.
Leverage the infrastructure from #189.

bcumming:
The final piece in the metering features.

* add a `power_meter` which currently records energy used on each node of a Cray XC{30,40,50} systems, which all have built in `pm_counters` interface to power measurement.
* add information about which node each MPI rank runs on to the metering output in `meters.json`, which is needed to analyse energy recordings, which are per node, not per MPI rank.
* refactor meters to not 
* refactor collation of measurements to the meter manager.
* add support for `gather` with `std::string` to the global communication policy, which required a back end MPI implementation and corresponding unit test.
* add `src/util/config.hpp` that populate the `nest::mc::config` namespace with `consexpr bool` values that reflect the configuration.

fixes #190


src/util/hostname.cpp
```diff
@@ -0,0 +1,30 @@
...
+
+#ifdef __linux__
+std::string hostname() {
+    char name[128];
+    auto result = gethostname(name, sizeof(name));
+    return result? "unknown": name;
+}
+#else
+std::string hostname() {
+    return "unknown";
```
halfflat: It's perverse, but someone could name their machine 'unknown'. Perhaps have this function return an empty string or other form of marked error response, and have the caller handle the user-visible signal for that error?

src/communication/mpi.hpp
```diff
@@ -65,10 +65,6 @@ namespace mpi {
     // T must be trivially copyable
     template<typename T>
     std::vector<T> gather(T value, int root) {
-        static_assert(
-            true,//std::is_trivially_copyable<T>::value,
-            "gather can only be performed on trivally copyable types");
-
```
halfflat: For the public record: it was apparently me who originally commented this out. I have no recollection as to why, though I'm hoping it was just some debugging test that slipped inadvertently into a commit.

src/util/hostname.cpp
halfflat: I still think we need a better signal for failure to determine hostname, e.g. empty string or a string with invalid hostname characters in it (e.g. "<unknown>".)

src/profiling/meter_manager.cpp
```diff
@@ -1,34 +1,79 @@
...
+    // Assert that the same number of readings were taken on every domain.
+    const auto num_readings = readings.size();
+    if (gcom::min(num_readings)!=gcom::max(num_readings)) {
+        throw std::out_of_range(
+            "the number of checkpoints in the \""+name+"\" meter do not match across domains");
+    }
+
+    // Gather across all of the domains onto the root domain.
+    for (auto r: readings) {
+        measurements.push_back(gcom::gather(r, 0));
```
halfflat: I think global communication in a constructor is surprising in a bad way, but I will submit a counter-proposal in a subsequent PR.

src/util/power.cpp
```diff
@@ -0,0 +1,32 @@
...
+namespace mc {
+namespace util {
+
+#ifdef NMC_HAVE_CRAY
+
+energy_size_type energy() {
+    energy_size_type result = -1;
+
+    std::ifstream fid("/sys/cray/pm_counters/energy");
+    if (fid.good()) {
```
halfflat: This test is redundant; if `fid` is not in a good state, `fid >> result` below won't write to `result`, and we're not doing anything special if `fid` isn't in a good state.

src/profiling/power_meter.cpp
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    std::string units() override {
+        return "J";
+    }
+
+    std::vector<double> measurements() override {
+        std::vector<double> diffs;
+
+        for (auto i=1ul; i<readings_.size(); ++i) {
```
halfflat: That `std::size_t` is no larger than an unsigned long is an unwarranted assumption (e.g. unsigned long is only 4 bytes under windows.) Why not just use `std::size_t`?

src/util/hostname.cpp
bcumming: now it uses `util::optional` to only return a string on success.

src/util/power.cpp
bcumming:
```
If extraction fails, zero is written to value and failbit is set. If extraction results in the value too large or too small to fit in value, std::numeric_limits<T>::max() or std::numeric_limits<T>::min()  is written and failbit flag is set. (since C++11)
```

PR #223, opened 2017-04-06
Implicit move ctor fix for the Intel compiler

vkarak:
Intel compiler (at least v16) seems not to generate an implicit move constructor.

### DIFF
--- a/src/util/indirect.hpp
+++ b/src/util/indirect.hpp
@@ -38,7 +38,9 @@ template <typename RASeq, typename Seq>
 auto indirect_view(RASeq& data, const Seq& index_map)
 DEDUCED_RETURN_TYPE(transform_view(index_map, impl::indirect_accessor<RASeq&>(data)));
 
-template <typename RASeq, typename Seq>
+// icpc 17 fails to disambiguate without further qualification, so
+// we replace `template <typename RASeq, typename Seq>` with the following:
+template <typename RASeq, typename Seq, typename = util::enable_if_t<!std::is_reference<RASeq>::value>>
 auto indirect_view(RASeq&& data, const Seq& index_map)
 DEDUCED_RETURN_TYPE(transform_view(index_map, impl::indirect_accessor<RASeq>(std::move(data))));
 
--- a/src/util/iterutil.hpp
+++ b/src/util/iterutil.hpp
@@ -89,8 +89,8 @@ auto back(Seq& seq) -> decltype(*std::begin(seq)) {
  */
 template <typename V>
 struct pointer_proxy: public V {
-    pointer_proxy(const V& v): V{v} {}
-    pointer_proxy(V&& v): V{std::move(v)} {}
+    pointer_proxy(const V& v): V(v) {}
+    pointer_proxy(V&& v): V(std::move(v)) {}
     const V* operator->() const { return this; }
 };
 
--- a/tests/unit/test_transform.cpp
+++ b/tests/unit/test_transform.cpp
@@ -79,7 +79,6 @@ TEST(transform, pointer_access) {
 TEST(transform, pointer_proxy) {
     struct item {
         int x;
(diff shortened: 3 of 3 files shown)


halfflat:
We have more serious problems with the Intel compiler :(

```
$ CC --version
icpc (ICC) 17.0.1 20161005
Copyright (C) 1985-2016 Intel Corporation.  All rights reserved.
$ cat foo.cc
#include <iostream>

template <typename A> void f(A&)  { std::cout << "lvalue\n"; }
template <typename A> void f(A&&) { std::cout << "rvalue\n"; }

int main() {
    struct A {} a;
    f(a);
    f(A{});
}
$ CC -std=c++11 foo.cc
foo.cc(8): error: more than one instance of overloaded function "f" matches the argument list:
            function template "void f(A &)"
            function template "void f(A &&)"
            argument types are: (A)
      f(a);
      ^

compilation aborted for foo.cc (code 2)
```

Which breaks everything that uses `indirect_view`. For the record, icpc's interpretation is incorrect: §14.8.2.4/9 of the (draft) C++11 standard states that the lvalue version should be considered as more specialized than the rvalue version, and thus should be selected unambiguously in the call `f(a)`.

The problem that with the `strict item` corresponds to [DR#1467](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1467); recent clang and gcc adopt the corrected behaviour; icpc 17 at least does not.

Ideally we would just not use `icpc` except for kernels, but that's not yet feasible.

I propose adding an `enable_if` qualification to the `indirect_view` overload to fix the first problem, and changing the brace initialization to parenthesis initialization for the parameter type `V` in `util::pointer_proxy`.

halfflat:
Proposed patch to master branch:
* Qualify template in `indirect_view` to accommodate incorrect function template specialization determination (icpc ignores §14.8.2.4/9 rule on lvalue versus rvalue template arguments).
* Use parenthesis constructors for type parameter in `pointer_proxy`, as icpc has not adopted the corrected behaviour for DR#1467 [http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1467].

```
diff --git a/src/util/indirect.hpp b/src/util/indirect.hpp
index 1e4dbba..30501e3 100644
--- a/src/util/indirect.hpp
+++ b/src/util/indirect.hpp
@@ -38,7 +38,9 @@ template <typename RASeq, typename Seq>
 auto indirect_view(RASeq& data, const Seq& index_map)
 DEDUCED_RETURN_TYPE(transform_view(index_map, impl::indirect_accessor<RASeq&>(data)));
 
-template <typename RASeq, typename Seq>
+// icpc 17 fails to disambiguate without further qualification, so
+// we replace `template <typename RASeq, typename Seq>` with the following:
+template <typename RASeq, typename Seq, typename = util::enable_if_t<!std::is_reference<RASeq>::value>>
 auto indirect_view(RASeq&& data, const Seq& index_map)
 DEDUCED_RETURN_TYPE(transform_view(index_map, impl::indirect_accessor<RASeq>(std::move(data))));
 
diff --git a/src/util/iterutil.hpp b/src/util/iterutil.hpp
index 00e52d6..6c8e135 100644
--- a/src/util/iterutil.hpp
+++ b/src/util/iterutil.hpp
@@ -89,8 +89,8 @@ auto back(Seq& seq) -> decltype(*std::begin(seq)) {
  */
 template <typename V>
 struct pointer_proxy: public V {
-    pointer_proxy(const V& v): V{v} {}
-    pointer_proxy(V&& v): V{std::move(v)} {}
+    pointer_proxy(const V& v): V(v) {}
+    pointer_proxy(V&& v): V(std::move(v)) {}
     const V* operator->() const { return this; }
 };

```

halfflat:
Looks good to me! But then I would say that…

PR #227, opened 2017-04-07
Initial microbenchmark builds.

halfflat:
* Use git submodule for incorporating Google benchmark library.
* Add one microbenchmark for comparing `util::transform_view` performance.

Note that the microbenchmarks are not built by default; they can be built with `make ubenches`, and then run individually. The microbenchmarks will be built in `tests/ubench/`, relative to the build directory.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "tests/ubench/google-benchmark"]
+	path = tests/ubench/google-benchmark
+	url = https://github.com/google/benchmark
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -13,6 +13,10 @@ add_subdirectory(global_communication)
 # Tests for performance: This could include stand alone tests. These do not necessarily be run automatically
 add_subdirectory(performance)
 
+# Microbenchmarks.
+add_subdirectory(ubench)
+
+
 # modcc tests
 if(NOT use_external_modcc)
     add_subdirectory(modcc)
--- a/tests/ubench/CMakeLists.txt
+++ b/tests/ubench/CMakeLists.txt
@@ -1,23 +1,40 @@
 include(ExternalProject)
 
-set(gbench_install_dir "${PROJECT_BINARY_DIR}/gbench")
-
-set(gbench_cmake_args
-    -DCMAKE_BUILD_TYPE=release
-    -DCMAKE_INSTALL_PREFIX=${gbench_install_dir}
-    -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
-    -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER})
-
-ExternalProject_Add(gbench
-    SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/google-benchmark"
-    CMAKE_ARGS "${gbench_cmake_args}"
-#    INSTALL_DIR "${gbench_install_dir}"
-)
-
-message(STATUS "gbench_install_dir:  ${gbench_install_dir}")
+# List of micro benchmarks to build.
 
 set(bench_sources
     accumulate_functor_values.cpp)
+
+# Set up google benchmark as an external project.
+
+set(gbench_src_dir "${CMAKE_CURRENT_SOURCE_DIR}/google-benchmark")
+set(gbench_install_dir "${PROJECT_BINARY_DIR}/gbench")
+
+set(gbench_cmake_args
+    "-DCMAKE_BUILD_TYPE=release"
+    "-DCMAKE_INSTALL_PREFIX=${gbench_install_dir}"
+    "-DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}"
+    "-DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}")
+
+
+# Attempt to update git submodule if required.
+find_package(Git)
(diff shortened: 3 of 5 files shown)


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tests/ubench/google-benchmark"]
```
bcumming: We removed all of the submodules because they were overcomplicating things.

There are two approaches that I would prefer:
- include a copy of Google Benchmark in the source
- make it a requirement that the user must install.

tests/ubench/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+set(gbench_cmake_args
+    -DCMAKE_BUILD_TYPE=release
+    -DCMAKE_INSTALL_PREFIX=${gbench_install_dir}
+    -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+    -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER})
+
+ExternalProject_Add(gbench
+    SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/google-benchmark"
+    CMAKE_ARGS "${gbench_cmake_args}"
+#    INSTALL_DIR "${gbench_install_dir}"
```
bcumming: remove?

tests/ubench/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+)
+
+message(STATUS "gbench_install_dir:  ${gbench_install_dir}")
+
+set(bench_sources
+    accumulate_functor_values.cpp)
+
+foreach(bench_src ${bench_sources})
+    string(REGEX REPLACE "\\.[^.]*$" "" bench_exe "${bench_src}")
+    add_executable("${bench_exe}" "${bench_src}")
```
bcumming: This is getting inconsistent. All of the executables have an `.exe` extension, except for `lmorpho`. Let's go with `bench.exe` for consistency. Whether using an `.exe` extension is a good thing is a separate discussion.

halfflat: We've actually got some others too: `modcc`, `test_modcc`. The `.exe` extension is a bit weird, so given that we're not already consistent on this, I'd prefer to avoid it. (And change the others, too.)

tests/ubench/CMakeLists.txt
halfflat: Or perhaps put back in. I'll do one or the other! `INSTALL_DIR` in `ExternalProject_Add` does almost, but not quite, nothing.

.gitmodules
halfflat: In this instance, I think that these alternatives would be worse than using a git submodule:
* the source comprises a considerable number of files; incorporating then would add a lot to our tree while (in principle) we would not be updating them with new local revisions;
* ideally the benchmark library and benchmarked code would be built with the sample compiler and options, which would generally not be the case for an externally built component.

The 'standard' CMake approach would be to clone the google benchmark repository at CMake configure time, but this leads to a lot of redundant downloads and access to github not only at download time, but also configure/build time.

On the other hand, most of the issues we had with git submodules in the past arose due to our needing to update the submodules both from without the project and within. For this application, the gooogle benchmark code would correspond to a single revision that would need to be updated rarely, if ever, and we would not be pushing changes back upstream from the submodule.

I propose adding a test to the CMake script that checks if the submodule has been properly instantiated (or rather, test that there is a `.git` subdirectory in the subfolder) and either warns that the submodule needs to be updated or attempts to do so itself.

halfflat:
As per our discussion offline, I will add a `README.md` with an entry for each of the micro-benchmarks (i.e. the one benchmark at this stage) describing the motivation for the test and summarizing the results on at least one platform.

tests/ubench/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+
+# Attempt to update git submodule if required.
+find_package(Git)
+if(NOT EXISTS "${gbench_src_dir}/.git")
+    if(GIT_FOUND)
+        exec_program("${GIT_EXECUTABLE}" "${CMAKE_CURRENT_SOURCE_DIR}"
+            ARGS submodule update --init google-benchmark)
+    else()
+        message(WARNING "google-benchmark submodule apparently not updated, but cannot find git")
```
bcumming: How about the following message: `Unable to update the google-benchmark submodule: git not found`

What happens if git is not found? From what I can tell the message is printed, and we continue trying to build the external project.

What happens if there is an error running the `git submodule update --init google-benchmark` command?

tests/CMakeLists.txt
```diff
@@ -13,6 +13,10 @@ add_subdirectory(global_communication)
 # Tests for performance: This could include stand alone tests. These do not necessarily be run automatically
 add_subdirectory(performance)
 
+# Microbenchmarks.
+add_subdirectory(ubench)
```
bcumming: Isn't this gong to attempt to build the micro benchmarks by default?

tests/ubench/CMakeLists.txt
halfflat: If git is not found, or git submodule fails, it will attempt to build anyway. I don't think this is a bad fallback behaviour though; it allows for someone to clone the repo on one machine, but attempt to configure and build on another that might have a poor tools setup.

halfflat:
Updated `tests/ubench/CMakeLists.txt` with various `EXCLUDE_FROM_ALL`s added.

PR #228, opened 2017-04-11
Move string-printf utility from lmorpho into util.

halfflat:
* Add `util::strprintf` printf-alike function that returns a `std::string` result.
* Include simple adaptors for `std::string` and standard smart pointer arguments fot `util::strprintf` (i.e. `std::string` arguments can be used with `%s`, smart pointers with `%p`).
* Add unit tests to suit.

### DIFF
--- a/src/util/strprintf.hpp
+++ b/src/util/strprintf.hpp
@@ -27,7 +27,7 @@ namespace impl {
 }
 
 template <typename... Args>
-static std::string strprintf(const char* fmt, Args&&... args) {
+std::string strprintf(const char* fmt, Args&&... args) {
     thread_local static std::vector<char> buffer(1024);
 
     for (;;) {
@@ -36,14 +36,14 @@ static std::string strprintf(const char* fmt, Args&&... args) {
             throw std::system_error(errno, std::generic_category());
         }
         else if ((unsigned)n<buffer.size()) {
-            return std::string(buffer.data());
+            return std::string(buffer.data(), n);
         }
         buffer.resize(2*n);
     }
 }
 
 template <typename... Args>
-static std::string strprintf(std::string fmt, Args&&... args) {
+std::string strprintf(const std::string& fmt, Args&&... args) {
     return strprintf(fmt.c_str(), std::forward<Args>(args)...);
 }
 
--- a/src/util/strprintf.hpp
+++ b/src/util/strprintf.hpp
@@ -1,0 +1,52 @@
+#pragma once
+
+/* Thin wrapper around std::snprintf for sprintf-like formatting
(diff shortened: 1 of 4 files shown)


halfflat:
I was using this printf wrapper while testing code in the event delivery work, so felt it was time to move it into general utilities from lmorpho.

src/util/strprintf.hpp
```diff
@@ -0,0 +1,52 @@
...
+static std::string strprintf(const char* fmt, Args&&... args) {
+    thread_local static std::vector<char> buffer(1024);
+
+    for (;;) {
+        int n = std::snprintf(buffer.data(), buffer.size(), fmt, impl::sprintf_arg_translate(std::forward<Args>(args))...);
+        if (n<0) {
+            throw std::system_error(errno, std::generic_category());
+        }
+        else if ((unsigned)n<buffer.size()) {
+            return std::string(buffer.data());
```
bcumming: This might be faster: `return std::string(buffer.data(), n);`

src/util/strprintf.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+    template <typename T, typename Deleter>
+    T* sprintf_arg_translate(const std::unique_ptr<T, Deleter>& x) { return x.get(); }
+
+    template <typename T>
+    T* sprintf_arg_translate(const std::shared_ptr<T>& x) { return x.get(); }
+}
+
+template <typename... Args>
+static std::string strprintf(const char* fmt, Args&&... args) {
```
bcumming: Does this function need to be static?

src/util/strprintf.hpp
```diff
@@ -0,0 +1,52 @@
...
+        }
+        else if ((unsigned)n<buffer.size()) {
+            return std::string(buffer.data());
+        }
+        buffer.resize(2*n);
+    }
+}
+
+template <typename... Args>
+static std::string strprintf(std::string fmt, Args&&... args) {
```
bcumming: Two comments:
* is static needed?
* this isn't a sink for the `fmt` string, so should it be passed as `const std::string&`?

src/util/strprintf.hpp
halfflat: It would be faster, but would potentially allow NUL characters in the string. On balance, I think it is better to use the method you suggest.

src/util/strprintf.hpp
halfflat: Nope!

src/util/strprintf.hpp
halfflat: Definitely should be `const std::string&`.

PR #229, opened 2017-04-12
refactor event_binner

bcumming:
* event_binner definition removed from cell_group.hpp header
* event_binner is not templated, so implementation placed in cpp file


bcumming:
I will make separate unit tests for the `event_binner`

PR #230, opened 2017-04-12
Feature/cellgroup cleanup

bcumming:
Some small changes to remove redundant and stale features from the interface for cell_group and model, as a first step towards "virtualization" of the cell_group interaface.

* Remove value_type from model.
* Remove probe interface from cell_group because probes communicate via the sampler interface on cell groups.
* depricate cell_group::remove_samplers()
* move cell_group::reset_samplers() to private
* depricate cell_group::clear_events()
  * was only used by cell_group::reset()

### DIFF
--- a/src/cell_group.hpp
+++ b/src/cell_group.hpp
@@ -63,7 +63,7 @@ public:
 
     void reset() {
         spikes_.clear();
-        clear_events();
+        events_.clear();
         reset_samplers();
         binner_.reset();
         lowered_.reset();
@@ -164,10 +164,6 @@ public:
         return spike_sources_;
     }
 
-    void clear_events() {
-        events_.clear();
-    }
-
     void add_sampler(cell_member_type probe_id, sampler_function s, time_type start_time = 0) {
         auto handle = get_probe_handle(probe_id);
 
@@ -177,25 +173,8 @@ public:
         sample_events_.push({sampler_index, start_time});
     }
 
-    void remove_samplers() {
-        sample_events_.clear();
-        samplers_.clear();
-        sampler_start_times_.clear();
-    }
-
-    void reset_samplers() {
-        // clear all pending sample events and reset to start at time 0
-        sample_events_.clear();
-        for(uint32_t i=0u; i<samplers_.size(); ++i) {
-            sample_events_.push({i, sampler_start_times_[i]});
-        }
-    }
-
-    value_type probe(cell_member_type probe_id) const {
-        return lowered_.probe(get_probe_handle(probe_id));
-    }
-
 private:
+
     // gid of first cell in group.
     cell_gid_type gid_base_;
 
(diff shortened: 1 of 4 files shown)


src/model.hpp
```diff
@@ -208,7 +208,8 @@ class model {
         current_spikes().get().push_back({source, tspike});
     }
 
-    void attach_sampler(cell_member_type probe_id, sampler_function f, time_type tfrom = 0) {
+    template <typename F>
+    void attach_sampler(cell_member_type probe_id, F&& f, time_type tfrom = 0) {
```
halfflat: If we're perfectly forwarding `F` to the (implicit) `std::function` constructor through the second argument of `cell_group::add_sampler`, I'd use `std::forward<F>(f)` below.

Otherwise, how about moving the definition of `sampler_function` outside of `cell_group` (but in the same header, or a common header) and use that?

src/cell_group.hpp
```diff
@@ -277,6 +258,14 @@ class cell_group {
     target_handle get_target_handle(cell_member_type target_id) const {
         return target_handles_[handle_partition_lookup(target_handle_divisions_, target_id)];
     }
+
+    void reset_samplers() {
+        // clear all pending sample events and reset to start at time 0
+        sample_events_.clear();
+        for(uint32_t i=0u; i<samplers_.size(); ++i) {
```
halfflat: While we're here, we should make this `for` consistent/safe (standard integer size vs. `std::size_t` argument.) Happy to do the edit in-place if you've no objection.

bcumming: There was actually a reason that a 32 bit unsigned type was used explicitly here. Using `std::size_t` instead and you get the following warning:
```
/users/bcumming/software/nestmc-proto/src/cell_group.hpp:266:32: warning: narrowing conversion of ‘i’ from ‘std::size_t {aka long unsigned int}’ to ‘uint32_t {aka unsigned int}’ inside { } [-Wnarrowing]
             sample_events_.push({i, sampler_start_times_[i]});
```
To keep the `size_t` index, we have to cast `i` to a narrower type inside the loop.
Which is implicitly saying "there are less than `2^32` samplers".

My fix is to have a `sample_event::size_type`, and explicitly use that for the index.

src/model.hpp
bcumming: I am using `std::forward` because I just can't be bothered reverse engineering how samplers work another time. When these are lowered into the back end, they need to be documented.

src/cell_group.hpp
```diff
@@ -164,38 +164,18 @@ class cell_group {
...
-        events_.clear();
-    }
-
     void add_sampler(cell_member_type probe_id, sampler_function s, time_type start_time = 0) {
         auto handle = get_probe_handle(probe_id);
 
-        auto sampler_index = uint32_t(samplers_.size());
+        using size_type = sample_event<time_type>::size_type;
+        auto sampler_index = size_type(samplers_.size());
         samplers_.push_back({handle, probe_id.gid, s});
```
halfflat: I know we're just pushing around the type here. Let's take this, but I think we can improve on this later.

PR #233, opened 2017-04-13
Try more places for validation data; workaround CMake FindCUDA bug.

### LINKED ISSUE #232: CMake 3.7/3.8, FindCUDA and CPP defines with quotes
Building with CUDA enabled breaks when using CMake 3.7 through 3.8, due to broken quote handling somewhere in the bowels of `FindCUDA`. Fixing this properly probably involves purging CMake from the surface of the planet and sending Kitware developers to re-education camps, but this is out of scope for the project.

As a work-around, I propose disabling the `-DDATADIR="some/path"` define when the CMake version is as above, and putting a bit more smarts in the validation test code for trying to find the data:
1. If a path is given on the command line, use that, otherwise:
2. if the `NMC_DATADIR` environment variable is set, use that, otherwise:
3. if the macro `NMC_DATADIR` was defined at compile time, and points to a directory, use that, otherwise:
4. try `./validation/data` or `../validation/data` in turn.

halfflat:
Fixes #232.

* Try `NMC_DATADIR` environment variable for validation data path, or else if the #defined `NMC_DATADIR` does not point to a directory, try `./validation/data` and `../validation/data`.
* Don't define `NMC_DATADIR` if CMake version 3.7 or 3.8.
* Extend C++17 filesystem emulation with (POSIX) implementation of `is_directory` and supporting classes, functions and enums `filesystem_error`, `file_status`, `status(..)`, `file_type` and `perms`.

### DIFF
--- a/src/util/path.cpp
+++ b/src/util/path.cpp
@@ -1,12 +1,18 @@
+#include <cerrno>
+
+// POSIX headers
+extern "C" {
+#include <glob.h>
+#include <sys/stat.h>
+}
+
 #include <util/scope_exit.hpp>
 #include <util/path.hpp>
-
-// POSIX header
-#include <glob.h>
 
 namespace nest {
 namespace mc {
 namespace util {
+namespace posix {
 
 std::vector<path> glob(const std::string& pattern) {
     std::vector<path> paths;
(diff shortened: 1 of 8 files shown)


halfflat:
The meat (i.e. interface to the OS function) is required anyway; the extra lines are interface fluff to make it align with the C++17 way of doing things. Should we ever want to support Windows, we get to keep the interface, but will have to substitute a Windows-specific implementation regardless.

The `stat` equivalents for Windows will likely be the straightforward ones, while making a `path` class for Windows will be more work, especially if we want to be good citizens and handle UTF-16 paths correctly.

halfflat:
The compilation error on OS X is due to my misuse of `_POSIX_C_SOURCE`. Will fix.

halfflat:
Updated with commit that omits preprocessor test and fixes `_POSIX_C_SOURCE` usage.

halfflat:
Additional fix to make `test_cuda.exe` build with CMake 3.7/3.8.

src/util/path.cpp
```diff
@@ -1,12 +1,19 @@
...
-// POSIX header
-#include <glob.h>
-
 namespace nest {
 namespace mc {
 namespace util {
+namespace posix {
 
 std::vector<path> glob(const std::string& pattern) {
     std::vector<path> paths;
```
bcumming: I can't leave a comment on line 19 below, however, something is fishy about `GLOB_TILDE` being used in both `#ifdef` statements.

src/util/path.cpp
```diff
@@ -1,12 +1,19 @@
+// POSIX headers
```
bcumming: Elsewhere we use guards like `#ifdef __linux__`. Here you have opted to make it something that is configured in the build system.
We should come up with one policy, and follow it.

tests/validation/validation_data.cpp
```diff
@@ -15,6 +16,38 @@ namespace mc {
 
 trace_io g_trace_io;
 
+util::path trace_io::find_datadir() {
+    // If environment variable is set, use that in preference.
+    if (const char* env_path = std::getenv("NMC_DATADIR")) {
+        return util::path(env_path);
```
bcumming: Should we test `util::is_directory(env_path)` before returning?

halfflat: There's not much we can do if it's not a directory; second-guessing the user here and substituting something else would lead to obscure issues. Better that it just not find the data IMO.

src/util/path.cpp
halfflat: Yup that's definitely fishy. I'll fix (or confirm that despite how it looks it is correct) and update. This is from a previous commit/PR.

tests/validation/validation_data.cpp
bcumming: It isn't clear whether the path returned from this function is meant to be valid or not. The later tests test if paths exist.

halfflat: It's best-effort plus override behaviour. Where we draw a line around 'best-effort' is certainly debatable, but I think we should defer to the user (i.e. environment variable or command-line parameter) if they provide a path explicitly at run time.

bcumming: I think that is reasonable, there just needs to be a little bit of documentation that explains that the path that is returned may not be valid.

halfflat: Will add!

halfflat:
Updated with comments + clearer implementation for `find_datadir()`.

PR #236, opened 2017-04-18
Virtualize cell_group

bcumming:
* The `cell_group` type was templated on the back end FVM cell implementation,
    and modelled a multi-compartment cell.
    * It is now an abstract base class that models a group of cells.
    * The old `cell_group` is renamed `mc_cell_group`, where the `mc` stands for 
      "multi-compartment".
  * Remove template parameter `Cell` from `model` type
    * By virtue of making the `cell_group`s in model an abstract base class, a
      model can have different `cell_group` types.
  * Extend `model` constructor to take a hint about where to construct cells
    * Currently two options: multicore or gpu.
    * This is a placeholder for a more sophisticated implementation once we have
      the requirements for a richer "ecosystem" of cell types.
  * Simplified some types to remove template dependencies between front
    and back ends.
    * Define a global `using time_type=float` in `common_types.hpp`
    * Define a `sampler_funtion` type that standardises sample functions


tests/validation/validate_ball_and_stick.cpp
```diff
@@ -3,24 +3,24 @@
 
 #include <fvm_multicell.hpp>
 
-using lowered_cell = nest::mc::fvm::fvm_multicell<nest::mc::multicore::backend>;
+const auto backend = nest::mc::backend_policy::use_multicore;
```
w-klijn: This name is allot more self explanatory!

tests/validation/validate_kinetic.hpp
```diff
@@ -23,11 +28,11 @@ void run_kinetic_dt(nest::mc::cell& c, float t_end, nlohmann::json meta, const s
     };
 
     meta["sim"] = "nestmc";
-    meta["backend"] = LoweredCell::backend::name();
+    meta["backend"] = backend==backend_policy::use_multicore? "cpu" : "gpu";
```
w-klijn: This feels fragile, but it is probably due to the transitional nature of the change. I would add an TODO?

.ycm_extra_conf.py
```diff
@@ -36,7 +36,7 @@
 # CHANGE THIS LIST OF FLAGS. YES, THIS IS THE DROID YOU HAVE BEEN LOOKING FOR.
 flags = [
     '-DNDEBUG',
-    '-DNMC_HAVE_TBB',
+    '-DNMC_HAVE_CTHREAD',
```
bcumming: It was to fix a long standing issue with clang code completion in vim... something that only I use. I could revert it.

miniapp/miniapp.cpp
```diff
@@ -202,11 +198,7 @@ void banner() {
     std::cout << "  starting miniapp\n";
     std::cout << "  - " << threading::description() << " threading support\n";
     std::cout << "  - communication policy: " << std::to_string(global_policy::kind()) << " (" << global_policy::size() << ")\n";
-#ifdef NMC_HAVE_CUDA
-    std::cout << "  - gpu support: on\n";
-#else
-    std::cout << "  - gpu support: off\n";
-#endif
+    std::cout << "  - gpu support: " << (config::has_cuda? "on": "off") << "\n";
```
bcumming: This parameter is a constant, that is set to `true` if the GPU back end was enabled at compilation, `false` otherwise.
The reasoning behind the parameters (which you can find in `src/util/config.hpp`) is to remove preprocessor cruft like `#ifdef NMC_HAVE_CUDA` from most of the code. They  are `constexpr bool` values, that can be used in `if` statements, and as template parameters, etc, unllke preprocessor macros.

src/mc_cell_group.hpp
```diff
@@ -0,0 +1,272 @@
+#pragma once
+
+#include <cstdint>
+#include <functional>
+#include <iterator>
+#include <vector>
+
+#include <algorithms.hpp>
+#include <cell_group.hpp> // CHANGE
```
bcumming: Good spot. The `CHANGE` comments were placeholders that I was using during development, which I forgot to remove. I will remove them.

tests/validation/validate_kinetic.hpp
bcumming: Agreed, I was thinking that it was a bit fragile when I implemented it. But, as you say, this is placeholder code, and a proper solution would have been overkill.

I will add a `FIXME` to this line.

.ycm_extra_conf.py
w-klijn: If it is meant to be in then leave it. Just wanted to make sure you were aware

miniapp/miniapp.cpp
w-klijn: Ok. But should it be part of the config then? Because that suggests something to be changed outside of compile time.

bcumming: The parameter should stay where it is in `nest::mc::config`.
I think that its use is appropriate here, because we are printing to the output whether the miniapp has gpu support, which it does by virtue of being compiled with the CUDA back end.

The `config::has_cuda` value is also used below to choose at run time whether to select the multicore or gpu back end for the model. This was not an option before, and it could be made a runtime choice. But I am inclined to make the rule `by default miniapp is multicore, and if you compile with CUDA it uses the GPU`, because that is what we had before.

tests/validation/validate_ball_and_stick.hpp
```diff
@@ -37,7 +35,7 @@ void run_ncomp_convergence_test(
         {"dt", dt},
         {"sim", "nestmc"},
         {"units", "mV"},
-        {"backend", LoweredCell::backend::name()}
+        {"backend", backend==backend_policy::use_multicore? "cpu" : "gpu"}
```
halfflat: We should either change the key to "backend_policy" (or similar) and record a representation of the policy directly, or add an interface to `model` that allows us to query the lowered cell name and use that to set the field below (line 53, or perhaps more cleanly, in a once off call before the loop, e.g. `model{singleton_recipe{c}, backend}.cell_implementation(0)`).

tests/validation/validate_kinetic.hpp
```diff
@@ -23,11 +28,11 @@ void run_kinetic_dt(nest::mc::cell& c, float t_end, nlohmann::json meta, const s
     };
 
     meta["sim"] = "nestmc";
-    meta["backend"] = LoweredCell::backend::name();
+    meta["backend"] = backend==backend_policy::use_multicore? "cpu" : "gpu"; // FIXME: fragile
```
halfflat: Yup, it is. We should either just change the key to `backend_policy` and defer the problem, or add an interface to `model` that allows querying the name of an implementation corresponding to a gid.

tests/validation/validate_synapses.hpp
```diff
@@ -30,7 +30,7 @@ void run_synapse_test(
         {"model", syn_type},
         {"sim", "nestmc"},
         {"units", "mV"},
-        {"backend", LoweredCell::backend::name()}
+        {"backend", backend==backend_policy::use_multicore? "cpu" : "gpu"} // FIXME: fragile
```
halfflat: See comment on `validate_kinetic.hpp`

tests/validation/validate_ball_and_stick.hpp
```diff
@@ -37,7 +35,7 @@ void run_ncomp_convergence_test(
         {"dt", dt},
         {"sim", "nestmc"},
         {"units", "mV"},
-        {"backend", LoweredCell::backend::name()}
+        {"backend", backend==backend_policy::use_multicore? "cpu" : "gpu"} // FIXME: fragile
```
halfflat: See comment on `validate_kinetic.hpp`

src/sampler_function.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+#include <functional>
+
+#include <common_types.hpp>
+#include <util/optional.hpp>
+
+namespace nest {
+namespace mc {
+
+using sampler_function = std::function<util::optional<time_type>(time_type, float)>;
```
bcumming: changed.

src/simple_sampler.hpp
```diff
@@ -55,7 +56,7 @@ class simple_sampler {
     }
 
     template <typename Time = float, typename Value = double>
```
bcumming: indeed.

src/backends/fvm.hpp
```diff
@@ -4,4 +4,14 @@
...
     #include <backends/gpu/fvm.hpp>
+#else
+// FIXME: intermediate fix as part of virtualization of cell_group.
+// This requires that a runtime choice can be made during model set up
+// about which backend is to be used group by group. We use the multicore
+// backend as a dummy GPU back end when compiled without gpu support so that
+// this decision does not require preprocessor directives in model.hpp.
+namespace nest { namespace mc { namespace gpu {
+    using backend = nest::mc::multicore::backend;
+}}} // namespace nest::nc::gpu
```
bcumming: a `null_backend` has been added
  * derives from multicore back end
  * always returns `false` when queried if it implements a specific mechanism
  * throws if the user attempts to make a mechanism instance
  * new `static bool is_supported()` method returns `false`
Also added a simple unit test to check that the `null_backend` is used when not compiled with gpu support.

tests/unit/test_mc_cell_group.cpp
```diff
@@ -23,7 +23,7 @@ nest::mc::cell make_cell() {
 }
 
 TEST(cell_group, test) {
-    cell_group<fvm_cell> group{0, util::singleton_view(make_cell())};
+    mc_cell_group<fvm_cell> group{0, util::singleton_view(make_cell())};
```
bcumming: That was a copy and paste oversight.

tests/validation/validate_kinetic.hpp
bcumming: I am going to defer the problem, because I don't know what the future holds.

src/model.hpp
```diff
@@ -89,14 +110,14 @@ class model {
     }
 
     // one cell per group:
-    model(const recipe& rec):
-        model(rec, util::partition_view(util::make_span(0, rec.num_cells()+1)))
+    model(const recipe& rec, backend_policy policy):
+        model(rec, util::partition_view(util::make_span(0, rec.num_cells()+1)), policy)
```
bcumming: I removed the default value from the other constructor: now both require that the policy is passed explicitly.

tests/validation/validate_kinetic.hpp
```diff
@@ -28,7 +28,7 @@ void run_kinetic_dt(
     };
 
     meta["sim"] = "nestmc";
-    meta["backend"] = backend==backend_policy::use_multicore? "cpu" : "gpu"; // FIXME: fragile
+    meta["backend"] = to_string(backend);
```
halfflat: This is the only remaining thing: for consistency, this key should also be "backend_policy".

PR #237, opened 2017-04-20
On-going implementation of asynchronous integration/back-end event delivery

halfflat:
This branch is to track progress on issue #184. The PR corresponds to step 3 of the implementation plan.

* Add three new views in mechanism state:
   * `vec_ci_`: cv index to cell index (w.r.t. one lowered multicell instance) map.
   * `vec_t_`: current or integration starting time, indexed by cell.
   * `vec_t_to`: integration stopping time, indexed by cell.
* Extend notion of indexed view in `modcc` to encompass `CellIndexedVariable` variables, which have a per-cell value.
* Add/change 'built-in' variables `t`, `t_to`, and (for `BREAKPOINT`/`nrn_state()` blocks only) `dt` for NMODL.
* Additional utility functions:
   * `util::minmax_value`: return one-pass reduction for minimum and maximum.
   * `util::fill`: range-compatible wrapper for `std::fill`.
   * `util::empty` now works across arbitrary sequences (defaults to calling `empty()` method if it exists).
   * Extends C++17-like functionality for path testing and manipulation (c.f. PR #233).
* Modifies back-end matrix assembly, threshold detection and stimulus implementations to be asynchronous-friendly.
* Rename some back-end mechanism variables for clarity, now that there are multiple sorts of cell index vectors present. `cell_index` for example has been renamed to `cell_cv_divs` where it acts as a partition division of cv indices (length 1+#cells), and `cell_to_cv` where it acts as a map from cell to index of first cv for that cell (length #cells).
* Add `multi_event_queue` data structure, which handles multiple streams of events, bulk loaded but individually popped.
* Perform event delivery in `fvm_multicell::step_integration()`, setting unsynchronized integration stop times accordingly.


scripts/cc-filter
```diff
@@ -226,7 +226,7 @@ rule cxx:unsigned-int      s/\bunsigned\s+int\b/unsigned/g
 group cxx:tidy cxx:rm-template-space cxx:unsigned-int
 
 rule cxx:strip-qualified   s/%cxx:qualified%//g
-rule cxx:strip-args        s/(%cxx:identifier%%cxx:template-args%?)%cxx:paren-args%/$1(...)/g
+rule cxx:strip-args        s/(%cxx:identifier%%cxx:template-args%?)%cxx:paren-args%(?!:)/$1(...)/g
```
halfflat: This was added to prevent the rule matching the line numbers in the diagnostic output of `nvcc`. which reports errors by line in the style: 
> filename.cpp(123): error: ...

as opposed to the gcc/clang style:
> filename.cpp:123:45: error: ...

src/fvm_multicell.hpp
```diff
@@ -501,6 +525,17 @@ void fvm_multicell<Backend>::initialize(
...
     voltage_ = array(ncomp, resting_potential_);
+    cv_to_cell_ = iarray(ncomp, 0);
+    time_ = array(ncell_, 0);
+    time_to_ = array(ncell_, 0);
+
+    // initialize cv_to_cell_ values from compartment partition
+    std::vector<size_type> cv_to_cell_tmp(ncomp);
+    for (size_type i = 0; i<ncell_; ++i) {
+        auto comps = cell_comp_part[i];
+        std::fill(cv_to_cell_tmp.begin()+comps.first, cv_to_cell_tmp.begin()+comps.second, i);
```
halfflat: Elsewhere, this same operation is done with the construction `util::fill(util::subrange_view(vec, span), value)`, which IMO is cleaner and should be used here too.

src/fvm_multicell.hpp
```diff
@@ -265,19 +275,33 @@ class fvm_multicell {
...
+    std::vector<std::vector<deliverable_event>> staged_events_;
 
     /// event queue for integration period
     /// TODO: Placeholder; this will change when event lists are moved to
     /// a backend structure.
-    event_queue<deliverable_event> events_;
+    using cell_event_queue = multi_event_stream<deliverable_event>;
+    cell_event_queue events_;
+
+    bool no_pending_events() const {
```
bcumming: I would rename this `has_pending_events()` and use `!events_.has_pending_events()`.

Because I prefer `true` to indicate presence, and `false` to indicate absence (I am easily confused).

Discuss.

src/fvm_multicell.hpp
bcumming: That would look like: `fill(subrange_view(cv_to_cell_tmp, comps), i);`?

modcc/cprinter.cpp
```diff
@@ -584,20 +589,25 @@ void CPrinter::visit(APIMethod *e) {
...
+
+            auto external = var->external_variable();
+            auto const& name = var->name();
+            auto const& index_name = external->index_name();
+
+            text_.add_gutter();
+            text_ << "auto " + index_name + " = ";
+
+            if(external->is_cell_indexed_variable()) {
+                text_ << "util::indirect_view(util::indirect_view(" + index_name + "_, vec_ci_), node_index_);\n";
```
bcumming: Take care with KNL/vectorized back end.

modcc/expression.hpp
```diff
@@ -389,6 +393,9 @@ class LocalDeclaration : public Expression {
     LocalDeclaration(Location loc, std::string const& name)
     :   Expression(loc)
     {
+        if (name=="v") {
+            std::cerr << "got a live one\n";
```
halfflat: did indeed! Nice catch :)

src/backends/gpu/kernels/assemble_matrix.hpp
```diff
@@ -76,10 +90,19 @@ void assemble_matrix_interleaved(
...
 
-    // The 1e-3 is a constant of proportionality required to ensure that the
-    // conductance (gi) values have units μS (micro-Siemens).
-    // See the model documentation in docs/model for more information.
-    T factor = 1e-3/dt;
+    T factor = 0;
+    const unsigned permuted_cid = blk_id*BlockWidth + blk_lane;
+
+    if (permuted_cid<num_mtx) {
+        // The 1e-3 is a constant of proportionality required to ensure that the
```
halfflat: Will move.

src/backends/gpu/kernels/test_thresholds.hpp
```diff
@@ -4,6 +4,13 @@ namespace nest {
 namespace mc {
 namespace gpu {
 
+/// Cuda lerp by u on [a,b]: (1-u)*a + u*b.
```
halfflat: Will pop into `impl::`; as per convo there is a justification for the routine.

src/event_queue.hpp
```diff
@@ -135,6 +136,137 @@ public :
...
+class multi_event_stream {
+public:
+    using value_type = Event;
+    using time_type = impl::event_time_type<Event>;
+    using size_type = unsigned;
+
+    multi_event_stream() {}
+
+    explicit multi_event_stream(size_type n_stream):
+       span_(n_stream, {0u, 0u}) {}
```
halfflat: Just wanted to be explicit for clarity, but yes, not necessary.

src/event_queue.hpp
```diff
@@ -135,6 +136,137 @@ public :
...
+        auto evi = begin(events);
+        for (size_type i = 0; i<n_streams(); ++i) {
+            if (evi != end(events)) {
+                span_[i].first = size_type(ev_.size());
+                ev_.insert(ev_.end(), begin(*evi), end(*evi));
+                span_[i].second = size_type(ev_.size());
+                ++evi;
+
+                // check size for wrapping!
+                if (ev_.size()>std::numeric_limits<size_type>::max()) {
```
halfflat: Will add.

src/event_queue.hpp
```diff
@@ -135,6 +136,137 @@ public :
...
+    }
+
+    void clear() {
+        ev_.clear();
+        remaining_ = 0;
+        span_.assign(n_streams(), {0u, 0u});
+    }
+
+    // Load events from a sequence (of length at least `size()`) of
+    // sequences of events.
```
halfflat: Let's go with `EXPECTS` in both cases, so that we keep the constraint loose but can tighten the noose at configuration time if we like.

src/event_queue.hpp
```diff
@@ -135,6 +136,137 @@ public :
...
+
+        using ::nest::mc::event_time;
+        auto t = event_time(top_unsafe(i));
+        return t_until > t? t: t_until;
+    }
+
+    // Generic conditional pop: pop and return head of `i`th stream if
+    // non-empty and the head satisfies predicate.
+    template <typename Pred>
+    util::optional<value_type> pop_if(size_type i, Pred&& pred) {
```
halfflat: The interface will be reduced when we move this to back-end (multicore and definitely for gpu).

src/fvm_multicell.hpp
halfflat: Happy to go with `has_` methods in preference to `no_` methods, but have no real preference provided they have the right prefixes.

src/fvm_multicell.hpp
halfflat: Yup.

modcc/cprinter.cpp
halfflat: Fair cop. Will add impl.

src/backends/multicore/threshold_watcher.hpp
```diff
@@ -52,37 +59,34 @@ class threshold_watcher {
...
-        return t_prev_;
-    }
-
     /// Tests each target for changed threshold state
     /// Crossing events are recorded for each threshold that
     /// is crossed since the last call to test
-    void test(value_type t) {
+    void test() {
         for (auto i=0u; i<size(); ++i) {
+            auto cv     = cv_index_[i];
```
halfflat: We could, but given the shortness of the method, it seems to me to be an unnecessary qualification.

src/event_queue.hpp
halfflat: Regarding its genericity: doesn't hurt and is fairly clear, so I'd prefer to leave it in for now. Class will become more concrete when it is specialized for back-end implementation.

bcumming: The reason I bought up the template parameter, is that it was not immediately clear to me that the type was a "list of lists", and it took a bit before I figured it out.
The generic parameter isn't needed, so i would still err on the side of removing it until such an interface is required.

halfflat: It's um, right there in the comment.

halfflat:
PR updated, but results not yet validated on GPU or KNL, so not yet ready for merge.

halfflat:
GPU side seems fine; still issues with KNL.

halfflat:
KNL code was actually fine; issue was #247 — incorporated compiler bug patch and tests, validations pass. Ready for re-review.

bcumming:
lgtm.

PR #240, opened 2017-04-26
Issue/238 get cell kind

w-klijn:
Add a get_cell_kind function returning a enum with cell types to both the cell and the cell_group.

This allows grouping of recipe created cells into like kind groups of cells at a later stage.


src/cell.hpp
```diff
@@ -99,6 +99,11 @@ class cell {
          }
      }
 
+    /// Return the kind of cell, used for grouping into cell_groups
+    cell_kind const get_cell_kind() const  {
```
bcumming: Should we make an abstract base class called `cell`, and make our current `cell` implementation derive from this, and have a more descriptive name like `mc_cell`, for "multicompartment cell"?

halfflat: Renaming `cell` to `mc_cell` (or even `mc_cell_description`) I think is sound, and other sorts of cell or fake cell will need different description objects. But is there anything really in common across them? I don't think we need an abstract base class.

When a recipe says it has a multi-compartment cell, we can ask it for the corresponding `mc_cell` description, either via a dedicated method that can return e.g. an `optional` value, or via a common method that returns a variant- or type-erased type.

I think this is actually a good use case for `std::any` (or more precisely, our implementation of it, given that that is C++17 and/or Boost).

w-klijn: Check, I was thinking this also, but that should be done in a seperate change i guess?

src/common_types.hpp
```diff
@@ -53,6 +53,18 @@ DEFINE_LEXICOGRAPHIC_ORDERING(cell_member_type,(a.gid,a.index),(b.gid,b.index))
 
 using time_type = float;
 
+// Enumeration used to indentify the cell type/kind, used by the model to
+// group equal kinds in the same cell group.
+
+enum cell_kind {
+    multicompartment,
+    poisson,
+    from_file,
```
w-klijn: I was thinking about a neuron injecting spikes from file. This would allow deterministic testing. It is a usable cell type on the same level as the Poisson spike, I should be more clear in the enum naming? 

From slack: sam [4:35 PM] 
Just thinking about names. 'Multicompartment' is a bit of a misnomer, because we're really describing a cable-model neuron which is continuous in space; the compartmentalization is distinct from every other aspect of its description. 'Poisson' is also a bit of a misnomer, because as I understand it, these spike sources will produce a number of different sorts of random spike chains, only one of which is actually Poisson.

We have to use the term 'multicompartment' when describing NestMC because that is the term that describes this class of simulator. Internally though, it would be nice to be able to make a distinction between the continuous description and the discretization.

[4:38] 
I'd lean more towards names such as `point_neuron` (or perhaps `single_compartment`); `cable1d_neuron`; `stochastic_spike_source`; etc.

w.klijn [4:39 PM] 
Good call. It is taken from NEST but we can go our own 'logical' way in naming things.

[4:39] 
Ow: the from file neuron would replay spikes from a file on disk.

sam [4:39 PM] 
Then `spike_replay` :smile:

src/cell.hpp
bcumming: Good point.
With that approach, this method should be removed.

src/common_types.hpp
bcumming: I would leave it out until you are ready to include it in the PR.
Indeed, if we are going to make a minimal PR, then there should just be the one `cell_kind`, i.e. `multicompartment`.

Regarding the semantics, @halfflat has a point. The multicompartment solution discretization is on the level of the `fvm_multicell`.

For the "poisson" cell types, it depends how specific we make the cell groups. We could have a cell group just for a generalised Poisson cell type, and cell group specializations for  other spike train generating cell types (e.g. cell groups that output pre recorded cells from file).

src/common_types.hpp
```diff
@@ -57,12 +57,12 @@ using time_type = float;
 // group equal kinds in the same cell group.
 
 enum cell_kind {
```
bcumming: I would remove the unimplemented kinds, i.e. leave just `cable1d_neuron`.
Others can be added as they are implemented.

PR #244, opened 2017-05-01
modularize domain decomposition

### LINKED ISSUE #242: move domain decomposition into model construction
Currently domain decomposition is performed before model construction, and a simple partition of the cell ids across domains is passed into the model constructor.

To handle different cell types, internal to external gid mappings, and load balancing of cells with different computational weights, the model constructor will have to handle domain decomposition.

This task performs the first step: moving the domain decomposition logic that is currently implemented (a naive partition of gid across the domains) into the model constructor.

bcumming:
The domain decomposition, whereby cells were partitioned across domains/MPI
ranks and then grouped together, was performed in an ad-hoc manner.
This PR modularizes the domain decomposition.

* A `domain_decomposition` class performs the cell partitioning given the recipe
  and some parameters.
* This implementation has a single `domain_decomposition` class and flags are 
  provided in a `group_rules` struct
  * The `domain_decomposition` could be specialized, or the simple rules struct
    could provide some sort of policy implementation in the future when the 
    need arises
* The domain_decomposition class is initialized inside the model constructor
  and is maintained as state of the model.
* The cell model constructor interface has been simplified to a single constructor
  that takes a recipe and a reference to a `domain_decomposition`.
  * In the future we might pass in only a moveable subset of the 
    `domain_decomposition` information required to build the model.                                      

fixes #242.

### DIFF
--- a/miniapp/miniapp.cpp
+++ b/miniapp/miniapp.cpp
@@ -84,10 +84,12 @@ int main(int argc, char** argv) {
                     options.file_extension, options.over_write);
         };
 
-        auto backend = config::has_cuda?
+        group_rules rules;
+        rules.policy = config::has_cuda?
             backend_policy::prefer_gpu: backend_policy::use_multicore;
+        rules.target_group_size = options.group_size;
 
-        model m(*recipe, backend);
+        model m(*recipe, rules);
 
         if (options.report_compartments) {
             report_compartment_stats(*recipe);
@@ -102,8 +104,7 @@ int main(int argc, char** argv) {
         m.set_binning_policy(binning_policy, options.bin_dt);
 
         // Inject some artificial spikes, 1 per 20 neurons.
-        cell_gid_type first_spike_cell = 20*((cell_range.first+19)/20);
-        for (auto c=first_spike_cell; c<cell_range.second; c+=20) {
+        for (cell_gid_type c=0; c<recipe->num_cells(); c+=20) {
             m.add_artificial_spike({c, 0});
         }
 
--- a/miniapp/miniapp.cpp
+++ b/miniapp/miniapp.cpp
@@ -34,7 +34,6 @@ using file_export_type = io::exporter_spike_file<global_policy>;
 void banner();
 std::unique_ptr<recipe> make_recipe(const io::cl_options&, const probe_distribution&);
 std::unique_ptr<sample_trace_type> make_trace(cell_member_type probe_id, probe_spec probe);
-std::pair<cell_gid_type, cell_gid_type> distribute_cells(cell_size_type ncells);
 using communicator_type = communication::communicator<communication::global_policy>;
 
 void write_trace_json(const sample_trace_type& trace, const std::string& prefix = "trace_");
@@ -77,16 +76,6 @@ int main(int argc, char** argv) {
         pdist.all_segments = !options.probe_soma_only;
 
         auto recipe = make_recipe(options, pdist);
-        auto cell_range = distribute_cells(recipe->num_cells());
-
-        std::vector<cell_gid_type> group_divisions;
-        for (auto i = cell_range.first; i<cell_range.second; i+=options.group_size) {
-            group_divisions.push_back(i);
-        }
-        group_divisions.push_back(cell_range.second);
-
-        EXPECTS(group_divisions.front() == cell_range.first);
-        EXPECTS(group_divisions.back() == cell_range.second);
 
         auto register_exporter = [] (const io::cl_options& options) {
             return
@@ -95,9 +84,11 @@ int main(int argc, char** argv) {
                     options.file_extension, options.over_write);
         };
 
-        model m(*recipe,
-                util::partition_view(group_divisions),
-                config::has_cuda? backend_policy::prefer_gpu: backend_policy::use_multicore);
+        auto backend = config::has_cuda?
+            backend_policy::prefer_gpu: backend_policy::use_multicore;
+
+        model m(*recipe, backend);
+
         if (options.report_compartments) {
             report_compartment_stats(*recipe);
         }
@@ -177,20 +168,6 @@ int main(int argc, char** argv) {
         return 2;
     }
     return 0;
-}
-
-std::pair<cell_gid_type, cell_gid_type> distribute_cells(cell_size_type num_cells) {
-    // Crude load balancing:
-    // divide [0, num_cells) into num_domains non-overlapping, contiguous blocks
-    // of size as close to equal as possible.
-
-    auto num_domains = communication::global_policy::size();
-    auto domain_id = communication::global_policy::id();
-
-    cell_gid_type cell_from = (cell_gid_type)(num_cells*(domain_id/(double)num_domains));
-    cell_gid_type cell_to = (cell_gid_type)(num_cells*((domain_id+1)/(double)num_domains));
-
-    return {cell_from, cell_to};
 }
 
 void banner() {
--- a/miniapp/miniapp.cpp
+++ b/miniapp/miniapp.cpp
@@ -88,8 +88,9 @@ int main(int argc, char** argv) {
         rules.policy = config::has_cuda?
             backend_policy::prefer_gpu: backend_policy::use_multicore;
         rules.target_group_size = options.group_size;
+        auto decomp = domain_decomposition(*recipe, rules);
 
-        model m(*recipe, rules);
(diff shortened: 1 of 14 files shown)


bcumming:
I am looking for some feedback here.
It has unit tests for using without MPI, however I have only got placeholders for the MPI tests.
All these tests will become much richer when we have multiple cell types and gid renumbering.

halfflat:
Currently debating design with @bcumming IRL ...

halfflat:
In terms of the design, I am counter-proposing the following:
* The `domain_decomposition` object is instantiated at the application level.
* `model` gets passed only the `recipe` object and a `group_allocation` object which comprises
  * A cell gid partition (representing map from local group index to cell gid interval).
  * A map from group index to (suggested) hardware back-end.

The `model` code then remains very nearly unchanged; it gets the `cell_kind` for any particular group from the recipe directly.

Why keep them separate? I believe it is a cleaner, tighter design, for the following reasons:
* The domain decomposition can be queried, or potentially tweaked, at the application level independently of model creation.
* Querying the decomposition gives the option of the application-level code pre-allocating or configuring structures, i/o etc. just for the cells that are process-local, again without needing to first instantiate the model.
* Keeping decomposition information queries outside of the `model` object reduces the size of the `model` interface. 
* The model takes a more concrete data structure for the decomposition, instead of wrapping a more opaque object whose behaviour needs to be understood in conjunction with the model initialization code.
* If a more complicated domain decomposition process requires global communication, then that communication point would be explicit in the application code rather than implicit in the model constructor.
* It makes it easy to instantiate multiple models over arbitrary and explicit ranges of cells, simplifying testing code.

miniapp/miniapp.cpp
```diff
@@ -111,8 +104,7 @@ int main(int argc, char** argv) {
         m.set_binning_policy(binning_policy, options.bin_dt);
 
         // Inject some artificial spikes, 1 per 20 neurons.
-        cell_gid_type first_spike_cell = 20*((cell_range.first+19)/20);
-        for (auto c=first_spike_cell; c<cell_range.second; c+=20) {
+        for (cell_gid_type c=0; c<recipe->num_cells(); c+=20) {
```
bcumming: Yes. This interface will vanish, because it was just a hack to run benchmarks.

src/domain_decomposition.hpp
```diff
@@ -0,0 +1,116 @@
...
+#include <util/optional.hpp>
+#include <util/partition.hpp>
+#include <util/transform.hpp>
+
+namespace nest {
+namespace mc {
+
+// Meta data used to guide the domain_decomposition in distributing
+// and grouping cells.
+struct group_rules {
```
bcumming: This contains information required to perform the domain decomposition. It doesn't require cell type information: such information is needed in the output of the domain decompositon, where it is used to build the concrete cell groups.

src/domain_decomposition.hpp
```diff
@@ -0,0 +1,116 @@
...
+        auto num_domains = communication::global_policy::size();
+        auto domain_id = communication::global_policy::id();
+
+        // partition the cells globally across the domains
+        num_global_cells_ = rec.num_cells();
+        first_cell_ = (gid_type)(num_global_cells_*(domain_id/(double)num_domains));
+        last_cell_ = (gid_type)(num_global_cells_*((domain_id+1)/(double)num_domains));
+
+        // partition the local cells into cell groups
+        if (num_local_cells()>0) {
```
bcumming: I have added a detailed comment that describes the algorithm.

src/domain_decomposition.hpp
```diff
@@ -0,0 +1,116 @@
...
+            gid_type group_size = 1;
+            group_starts_.push_back(first_cell_);
+            auto group_kind = rec.get_cell_kind(first_cell_);
+            group_kinds_.push_back(group_kind);
+            gid_type gid = first_cell_+1;
```
bcumming: Is +1 really a magic number in this context? It is simply the next index.
I have added a comment that describes the algorithm in more detail, which hopefully makes these steps clearer.

src/domain_decomposition.hpp
```diff
@@ -0,0 +1,116 @@
...
+            while (gid<last_cell_) {
+                auto kind = rec.get_cell_kind(gid);
+                if (kind!=group_kind || group_size>=rules.target_group_size) {
```
bcumming: see new comment documentation.

src/model.hpp
```diff
@@ -220,16 +213,18 @@ class model {
 
     // only thread safe if called outside the run() method
     void add_artificial_spike(cell_member_type source, time_type tspike) {
-        current_spikes().get().push_back({source, tspike});
+        if (domain_.is_local_gid(source.gid)) {
```
w-klijn: Should we add a TODO to remove this function when we have 'real' artificial cells. To prevent that we have non-threadsave code laying about?

src/model.hpp
```diff
@@ -43,30 +44,27 @@ class model {
...
-           const util::partition_range<Iter>& groups,
-           backend_policy policy):
-        cell_group_divisions_(groups.divisions().begin(), groups.divisions().end()),
-        backend_policy_(policy)
+    model(const recipe& rec, const domain_decomposition& decomp):
+        domain_(decomp)
     {
         // set up communicator based on partition
-        communicator_ = communicator_type(gid_partition());
+        communicator_ = communicator_type(domain_.gid_group_partition());
```
w-klijn: Very clean and clear implementation!

tests/global_communication/test_domain_decomposition.cpp
```diff
@@ -0,0 +1,30 @@
...
+using namespace nest::mc;
+
+using communicator_type = communication::communicator<communication::global_policy>;
+
+static bool is_dry_run() {
+    return communication::global_policy::kind() ==
+        communication::global_policy_kind::dryrun;
+}
+
+TEST(domain_decomp, basic) {
```
w-klijn: Could you give a hint what type of tests you would expect here?
As a source code comment

src/domain_decomposition.hpp
w-klijn: I could imagine that this rules object in the future will be expanded with function like:
cell_size get_size_for_type(cell_kind):

src/domain_decomposition.hpp
```diff
@@ -0,0 +1,153 @@
...
+
+        auto num_domains = communication::global_policy::size();
+        auto domain_id = communication::global_policy::id();
+
+        // Partition the cells globally across the domains.
+        num_global_cells_ = rec.num_cells();
+        cell_begin_ = (cell_gid_type)(num_global_cells_*(domain_id/(double)num_domains));
+        cell_end_ = (cell_gid_type)(num_global_cells_*((domain_id+1)/(double)num_domains));
+
+        // Partition the local cells into cell groups that satisfy three
```
w-klijn: I completely understand this now: thanx!

It even hints at possible refactorings

src/domain_decomposition.hpp
```diff
@@ -0,0 +1,153 @@
...
+    group_range_type get_group(cell_size_type i) const {
+        return {group_starts_[i], group_starts_[i+1], group_kinds_[i]};
+    }
+
+    /// Tests whether a gid is on the local domain.
+    bool is_local_gid(cell_gid_type i) const {
+        return i>=cell_begin_ && i<cell_end_;
+    }
+
+    gid_partition_type gid_group_partition() const {
```
w-klijn: This function is not documented but it is maybe the most important one!

tests/unit/test_domain_decomposition.cpp
```diff
@@ -0,0 +1,112 @@
...
+        EXPECT_TRUE(decomp.is_local_gid(i));
+    }
+    EXPECT_FALSE(decomp.is_local_gid(num_cells));
+}
+
+TEST(domain_decomposition, multi_cell_groups)
+{
+    unsigned num_cells = 10;
+
+    // test group sizes from 1 up to 1 more than the total number of cells
```
w-klijn: Would it be smart to test the beginning also? So add 2 and the first and last should not be in the group.
That way all edge cases are covered

src/model.hpp
halfflat: I think there is no need at this point — it does what it says on the tin. It's likely this will go away when we have artificial spike sources in the model, but maybe it'd be useful to keep for tests. Let's not anticipate that yet :)

PR #248, opened 2017-05-03
Work-around for inlining bug in icc 16.0.3

### LINKED ISSUE #247: Unit tests fail on KNL with AVX512 vector back-end.
Steps to repeat:
1. Build and compile for KNL:
```
% mkdir b; cd b
% CC=icc CXX=icpc cmake -DNMC_VECTORIZE_TARGET=KNL -DNMC_WITH_CUDA=OFF -DNMC_THREADING_MODEL=cthread -DNMC_WITH_ASSERTIONS=OFF -DCMAKE_BUILD_TYPE=debug ..
…
% make -j test.exe
…
```
2. Run unit tests `fvm_multi.init`, `probe.fvm_multicell`, or `mc_cell_group.test`.
```
% ./tests/test.exe --gtest_filter=fvm_multi.init
…
[ RUN      ] fvm_multi.init
…/tests/unit/test_fvm_multi.cpp:89: Failure
Value of: test_nan(mat.u(1, J.size()))
  Actual: false
Expected: true
…/tests/unit/test_fvm_multi.cpp:90: Failure
Value of: test_nan(mat.d)
  Actual: false
Expected: true
…/tests/unit/test_fvm_multi.cpp:91: Failure
Value of: test_nan(J.solution())
  Actual: false
Expected: true
[  FAILED  ] fvm_multi.init (1 ms)
…
```
```
% ./tests/test.exe --gtest_filter=probe.fvm_multicell
…
[ RUN      ] probe.fvm_multicell
…/tests/unit/test_probe.cpp:80: Failure
Value of: lcell.probe(probes[0])
  Actual: -nan
Expected: lcell.voltage()[probes[0].second]
Which is: -nan
…/tests/unit/test_probe.cpp:81: Failure
Value of: lcell.probe(probes[1])
  Actual: -nan
Expected: lcell.voltage()[probes[1].second]
Which is: -nan
[  FAILED  ] probe.fvm_multicell (1 ms)
…
```
```
% /tests/test.exe --gtest_filter=mc_cell_group.test
…
[ RUN      ] mc_cell_group.test
warning: solution out of bounds for cell 0 at (max) t 0.01 ms
warning: solution out of bounds for cell 0 at (max) t 0.02 ms
warning: solution out of bounds for cell 0 at (max) t 0.03 ms
…
warning: solution out of bounds for cell 0 at (max) t 50 ms
warning: solution out of bounds for cell 0 at (max) t 50 ms
…/tests/unit/test_mc_cell_group.cpp:39: Failure
Value of: group.spikes().size()
  Actual: 0
Expected: 4u
Which is: 4
[  FAILED  ] mc_cell_group.test (308 ms)
…
```

halfflat:
Fixes #247.

Forces evaluation of partition bounds call within the `div_component_sampler` constructor, which for subtle and obscure reasons is mis- or un-computed with icpc and the `-xMIC-AVX512` target option.

* Add `compat::sink()` function that forces evaluation of the argument.
* Add `compat::sink_if_icc_leq(version, const X&)` function that performs this only when run with the Intel compiler subject to version bound.
* Use the `compat::sink_if_icc_leq` function in the `div_component_sampler` constructor.

### DIFF
--- a/src/util/compat.hpp
+++ b/src/util/compat.hpp
@@ -34,4 +34,27 @@ inline void compiler_barrier_if_xlc_leq(unsigned ver) {
 template <typename X>
 inline constexpr bool isinf(X x) { return std::isinf(x); }
 
+// Work around a bad inlining-related optimization with icpc 16.0.3 and -xMIC-AVX512,
+// by forcing a computation.
+
+template <typename X>
+inline void sink(const X& x) {
+    char buf[sizeof x];
+    volatile char* bufptr = buf;
+    const char* xptr = reinterpret_cast<const char*>(&x);
+
+    for (std::size_t i = 0; i<sizeof buf; ++i) {
+        *bufptr++ = *xptr++;
+    }
 }
+
+template <typename X>
+inline void sink_if_icc_leq(unsigned ver, const X& x) {
+#if defined(__INTEL_COMPILER_BUILD_DATE)
+    if (__INTEL_COMPILER_BUILD_DATE<=ver) {
+        sink(x);
+    }
+#endif
+}
+
+} // namespace compat
--- a/src/cell.hpp
+++ b/src/cell.hpp
@@ -100,7 +100,7 @@ public:
      }
 
     /// Return the kind of cell, used for grouping into cell_groups
-    cell_kind const get_cell_kind() const  {
+    cell_kind get_cell_kind() const  {
         return cell_kind::cable1d_neuron;
     }
 
--- a/src/cell_group.hpp
+++ b/src/cell_group.hpp
@@ -18,7 +18,7 @@ class cell_group {
 public:
     virtual ~cell_group() = default;
 
-    virtual cell_kind const get_cell_kind() const = 0;
+    virtual cell_kind get_cell_kind() const = 0;
 
     virtual void reset() = 0;
     virtual void set_binning_policy(binning_kind policy, time_type bin_interval) = 0;
--- a/src/compartment.hpp
+++ b/src/compartment.hpp
@@ -5,11 +5,12 @@
 
 #include <common_types.hpp>
 #include <math.hpp>
+#include <util/compat.hpp>
 #include <util/counter.hpp>
 #include <util/iterutil.hpp>
 #include <util/partition.hpp>
 #include <util/span.hpp>
-#include "util/rangeutil.hpp"
+#include <util/rangeutil.hpp>
 #include <util/transform.hpp>
 
 namespace nest {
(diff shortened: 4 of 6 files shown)


src/util/compat.hpp
```diff
@@ -34,4 +34,27 @@ inline void compiler_barrier_if_xlc_leq(unsigned ver) {
 template <typename X>
 inline constexpr bool isinf(X x) { return std::isinf(x); }
 
+// Work around a bad inlining-related optimization with icpc 16.0.3 and -xMIC-AVX512,
+// by forcing a computation.
+
+template <typename X>
+inline void sink(const X& x) {
```
w-klijn: This look like a lot of compute. Is this the easiest way?
http://stackoverflow.com/questions/2219829/how-to-prevent-gcc-optimizing-some-statements-in-c
I'm way out of my depth here, so please ignore.

Just to be sure how often is this called?

vkarak: This was my concern as well about the volatile. I'm afraid that every byte access there will go through memory.

If that's not in a critical path, I'm fine.

bcumming: So, I think that the question here is why cast to `char*` instead of using `X` directly as follows:
```
template <typename X>
inline void sink(const X& x) {
    X buf;
    volatile X* bufptr = &buf;
    bufptr[0] = x;
}
```

The main reason I can see is that this requires `X` to be copyable, which might not always be the case. By casting to `char*` Sam is forcing the compiler to copy the bytes. Maybe overkill because we use this in just one spot to sink a `std::pair`, which is copyable. But it is more generic.

vkarak: I think the key here is the `volatile` keyword, which is just there to prevent compiler optimisations. The `buf` doesn't have any real meaning, it's an automatic variable, initialised but not used, which would have been otherwise optimised out. In fact, we could fool the compiler with just a single-char buffer avoiding the full copy of `x`:

```C++
char buf;
volatile char *bufptr = &buf;
const char* xptr = reinterpret_cast<const char*>(&x);
*bufptr = *xptr;
```

I tried this with Intel 16 on Tave and worked fine.

halfflat: Indeed! I'll patch and resubmit.

halfflat: Actually, I'm not sure about the reliability of just copying one byte from `x`; while it works in this particular instance, the bug presented itself as in fact getting one field of the pair correct and not the other.

But I will replace the `char` array with a single `char`.

halfflat: Wow. Depending where I do the volatile cast, the compiler does or does not generate correct code. It is a delicate, insane, snowflake.

halfflat: What does work is in fact to place an explicit barrier after the `make_partition`. I'll resubmit without the sink, and instead with a `compiler_barrier_if_icc_leq` which is directly akin to `compiler_barrier_if_xlc_leq`.

PR #255, opened 2017-05-05
Feature/any

bcumming:
Partial implementation of `std::any` from C++17 standard.
     http://en.cppreference.com/w/cpp/utility/any

The implementation is in the `util` library as `util::any`.

Implements a standard-compliant subset of the full interface.

  * Does not avoid dynamic allocation of small objects.
  * Does not implement the `in_place_type<T>` constructors from the standard.
  * Does not implement the `emplace` modifier from the standard.

### DIFF
--- a/src/util/any.hpp
+++ b/src/util/any.hpp
@@ -4,8 +4,6 @@
 #include <typeinfo>
 #include <type_traits>
 
-#include <iostream> // TODO: remove
-
 #include <util/meta.hpp>
 
 // Partial implementation of std::any from C++17 standard.
@@ -16,8 +14,6 @@
 // - Does not avoid dynamic allocation of small objects.
 // - Does not implement the in_place_type<T> constructors from the standard.
 // - Does not implement the emplace modifier from the standard.
-// - Does not implement the make_any non-member function from the standard.
-//
 
 namespace nest {
 namespace mc {
@@ -220,6 +216,14 @@ T any_cast(any&& operand) {
 
 // Constructs an any object containing an object of type T, passing the
 // provided arguments to T's constructor.
+//
+// This does not exactly follow the standard, which states that
+// make_any is equivalent to
+//   return std::any(std::in_place_type<T>, std::forward<Args>(args)...);
+// i.e. that the contained object should be constructed in place, whereas
+// this implementation constructs the object, then moves it into the
+// contained object.
+// FIXME: rewrite with in_place_type when available.
 template <class T, class... Args>
 any make_any(Args&&... args) {
     return any(T(std::forward<Args>(args) ...));
--- a/src/util/any.hpp
+++ b/src/util/any.hpp
@@ -1,0 +1,110 @@
+#pragma once
+
+#include <memory>
+#include <typeinfo>
+#include <type_traits>
+
+#include <util/meta.hpp>
+
+// Partial implementation of std::any from C++17 standard.
+//      http://en.cppreference.com/w/cpp/utility/any
+//
+// Implements a standard-compliant subset of the full interface.
+//
+// Does not attempt to avoid dynamic allocation of small objects.
+
+
+namespace nest {
+namespace mc {
+namespace util {
+
+class any {
+public:
+    constexpr any() = default;
+
+    any(const any& other):
+        state_(other.state_->copy())
+    {}
+
+    any(any&& other) {
+        std::swap(other.state_, state_);
+    }
+
+    // any uses value semantics
+    template <
+        typename T,
+        typename = typename
+            util::enable_if_t<!std::is_same<util::decay_t<T>, any>::value>
+    >
+    any(T&& other):
+        state_(new model<util::decay_t<T>>(std::forward<T>(other)))
+    {}
+
+    bool has_value() const {
+        return (bool)state_;
+    }
+
+    const std::type_info& type() const {
+        return state_->type();
+    }
+
+private:
+
+    struct concept {
+        virtual ~concept() = default;
+
+        // TODO: comments
+        virtual const std::type_info& type() = 0;
+        virtual concept* copy() = 0;
+        //virtual void swap(concept& other) = 0;
+        //void(*move)(storage_union& src, storage_union& dest) noexcept;
+    };
+
+    template <typename T>
+    struct model: public concept {
+        ~model() = default;
+
+        concept* copy() override {
+            return new model<T>(*this);
+        }
+
+        const std::type_info& type() override {
+            return typeid(T);
+        }
+
+        model(const T& other):
+            value(other)
+        {}
+
+        model(T&& other):
+            value(std::move(other))
+        {}
+
+        T value;
+    };
+
+    std::unique_ptr<concept> state_;
+};
+
+template<class T>
+T any_cast(const any& operand) {
+
+}
+
+/*
+template<class T>
+T any_cast(any& operand);
+
+template<class T>
+T any_cast(any&& operand);
+
+template<class T>
+const T* any_cast(const any* operand);
+
+template<class T>
+T* any_cast(any* operand);
+*/
+
+} // namespace util
+} // namespace mc
+} // namespace nest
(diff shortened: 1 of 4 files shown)


src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+    template <typename ValueType, typename U=any_cast_remove_qual<ValueType>>
+    using test_any_cast_constructable = typename
+        std::conditional<
+               std::is_constructible<ValueType, const U&>::value
+            && std::is_constructible<ValueType, U&>::value
+            && std::is_constructible<ValueType, U>::value,
+            std::true_type,
+            std::false_type>::type;
+
```
halfflat: That said, this test seems too specific. Looking at the reference, the requirements on `ValueType` depend on the particular overload of `any_cast`.

src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+        model(const T& other):
+            value(other)
+        {}
+
+        model(T&& other):
+            value(std::move(other))
+        {}
+
+        void* pointer() {
```
halfflat: missing `override`

src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+            return &value;
+        }
+
+        const void* pointer() const {
```
halfflat: missing `override`

src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+        impl::test_any_cast_constructable<T>::value,
+        "any_cast requires ");
+
+    if (operand==nullptr || !operand->has_value()) {
+        return nullptr;
+    }
+
+    if (operand->type() != typeid(T)) {
+        throw bad_any_cast();
+    }
```
halfflat: The `any_cast<T>(const any*)` overload shouldn't throw (in fact, should be no-except). When `typeid(T)` is not the same as `operand->type()`, it should return `nullptr`.

As an implementation note, you could potentially replace the private `unsafe_cast<T>` method with a 'safe' version that did the `typeid(T)` test and returned `nullptr` on failure.

This applies of course to the non-const version below.

src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+        std::conditional<
+            std::is_move_constructible<T>::value
+            & !std::is_lvalue_reference<T>::value,
+            std::true_type, std::false_type>::type;
+
+    auto ptr = any_cast<U>(&operand);
+    if (ptr==nullptr) {
+        throw bad_any_cast();
+    }
+    return impl::move_else_copy<T>(static_cast<T*>(ptr), do_move());
```
halfflat: This looks fairly complicated (in terms of conditional moving etc.). Does `return std::move(*any_cast<T>(&operand));` not work? If not, an explicatory comment would help.

src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+    void swap(any& other) noexcept {
+        std::swap(other.state_, state_);
+    }
+
+    bool has_value() const noexcept {
+        return (bool)state_;
+    }
+
+    const std::type_info& type() const noexcept {
+        return state_->type();
```
halfflat: Should return `typeid(void)` if `!has_value()`. (This may simplify the `any_cast` implementations below, too.)

tests/unit/test_any.cpp
```diff
@@ -0,0 +1,207 @@
...
+    util::any ai(42);
+    auto i = util::any_cast<int>(ai);
+    EXPECT_EQ(typeid(i), typeid(int));
+    EXPECT_EQ(i, 42);
+}
+
+// test any_cast(any&&)
+TEST(any, any_cast_rvalue) {
+    auto moved = util::any_cast<moveable>(util::any(moveable()));
+    std::cout << "moves and copies: " << moved.moves << " " << moved.copies  << "\n";
```
halfflat: This test looks like it might be still being prototyped.

src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+    template <typename T>
+    friend const T* any_cast(const any* operand);
+
+    template <typename T>
+    friend T* any_cast(any* operand);
+
+    template <typename T>
+    T* unsafe_cast() {
+        return reinterpret_cast<T*>(state_->pointer());
```
bcumming: You are right. This was a hang over from a slightly different earlier implementation which required a more "industrial" solution.

src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+} // namespace impl
+
+// If operand is not a null pointer, and the typeid of the requested T matches
+// that of the contents of operand, a pointer to the value contained by operand,
+// otherwise a null pointer.
+template<class T>
+const T* any_cast(const any* operand) {
+    static_assert(
+        impl::test_any_cast_constructable<T>::value,
+        "any_cast requires ");
```
bcumming: The standard does not say that it is a strict compile time requirement, but it does state that the program is ill-formed if this condition is violated, so I figured that I could just catch the problem at compile time.

src/util/any.hpp
bcumming: True that.

src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+template<class T>
+T any_cast(any&& operand) {
+    using U = impl::any_cast_remove_qual<T>;
+    using do_move = typename
+        std::conditional<
+            std::is_move_constructible<T>::value
+            & !std::is_lvalue_reference<T>::value,
+            std::true_type, std::false_type>::type;
+
```
bcumming: You learn something new every day. Will do.

src/util/any.hpp
bcumming: good catch.

bcumming: Fixed, and a unit test for `any::type` was added.

src/util/any.hpp
bcumming: Right you are. I got a bit confused when reading the specs, and thought that it was an enumerated list of requirements that applied to all cases. I will narrow the scope of the tests.

src/util/any.hpp
bcumming: It does work... I was making life more complicated than it had to be.

src/util/any.hpp
bcumming: I have kept the `unsafe_cast`. Moving the `typeid(T)` makes another `if` statement, because we still have to test if `operand` is not null.

src/util/any.hpp
```diff
@@ -0,0 +1,269 @@
...
+    }
+};
+
+class any {
+public:
+    constexpr any() = default;
+
+    any(const any& other):
+        state_(other.state_->copy())
+    {}
```
bcumming: I have tried to make things more compact.

bcumming:
I have not explicitly implemented `std::swap(util::any)`, because we get it for free with the appropriate move assignment.

However the standard explicitly states that it should be overloaded to call the `any::swap` member modifier as follows `lhs.swap(rhs)`

http://en.cppreference.com/w/cpp/utility/any/swap2

So, do I actually have to explicitly overload it?

bcumming:
I also prefer not to have friend functions defined inline inside a class. No strong reason, I just find it easier to read that way.

src/util/any.hpp
```diff
@@ -0,0 +1,230 @@
...
+    }
+    return static_cast<T>(std::move(*ptr));
+}
+
+// Constructs an any object containing an object of type T, passing the
+// provided arguments to T's constructor.
+template <class T, class... Args>
+any make_any(Args&&... args) {
+    return any(T(std::forward<Args>(args) ...));
+}
```
halfflat: I believe this is not quite the same as what the standard promises, in that this will construct the `T` object and then (generally speaking) use `T`'s move constructor to place it in the wrapped store, while the standard description indicates that the wrapped `T` be constructed in-place.

Solution would be to add comment documenting the difference (or verifying that there is in fact no difference, but I don't think this is true) or implementing the `in_place_t` constructors and invoking them with `make_any` via perfect forwarding.

bcumming: I added a comment. Let's wait until #256.

PR #257, opened 2017-05-09
util::unique_any: a non-copyable util::any 

bcumming:
A non copyable variant of `util::any`.                                                                    
The two main use cases for such a container are                                                           
 1. storing types that are not copyable.                                                                  
 2. ensuring that no copies are made of copyable types that have to be stored                             
    in a type-erased container.                                                                           
                                                                                                          
`unique_any` has the same semantics as any with the execption of copy and copy                            
assignment, which are explicitly forbidden for all contained types.                                       
The requirement that the contained type be copy constructable has also been                               
relaxed.                                                                                                  
                                                                                                          
The `any_cast` non-member functions have been overridden for `unique_any`, with                           
the same semantics as for `any`.

### DIFF
--- a/src/util/unique_any.hpp
+++ b/src/util/unique_any.hpp
@@ -1,0 +1,197 @@
+#pragma once
+
+#include <memory>
+#include <typeinfo>
+#include <type_traits>
+
+#include <util/any.hpp>
+#include <util/meta.hpp>
+
+// A version of util::any that is not copyable.
+// The two main use cases for such a container are
+//  1. for storing types that are not copyable.
+//  2. for ensuring that no copies are made of types that are copyable.
+//     e.g. in performance critical code.
+
+namespace nest {
+namespace mc {
+namespace util {
+
+class unique_any {
+public:
+    constexpr unique_any() = default;
+
+    unique_any(unique_any&& other) noexcept {
+        std::swap(other.state_, state_);
+    }
+
+    template <
+        typename T,
+        typename = typename util::enable_if_t<!std::is_same<util::decay_t<T>, unique_any>::value>
+    >
+    unique_any(T&& other) {
+        using contained_type = util::decay_t<T>;
+        state_.reset(new model<contained_type>(std::forward<T>(other)));
+    }
+
+    unique_any& operator=(unique_any&& other) noexcept {
+        swap(other);
+        return *this;
+    }
+
+    template <
+        typename T,
+        typename = typename util::enable_if_t<!std::is_same<util::decay_t<T>, unique_any>::value>
+    >
+    unique_any& operator=(T&& other) {
+        using contained_type = util::decay_t<T>;
+        state_.reset(new model<contained_type>(std::forward<T>(other)));
+        return *this;
+    }
+
+    void reset() noexcept {
+        state_.reset(nullptr);
+    }
+
+    void swap(unique_any& other) noexcept {
+        std::swap(other.state_, state_);
+    }
+
+    bool has_value() const noexcept {
+        return (bool)state_;
+    }
+
+    const std::type_info& type() const noexcept {
+        return has_value()? state_->type(): typeid(void);
+    }
+
+private:
+    struct interface {
+        virtual ~interface() = default;
+        virtual const std::type_info& type() = 0;
+        virtual void* pointer() = 0;
+        virtual const void* pointer() const = 0;
+    };
+
+    template <typename T>
+    struct model: public interface {
+        ~model() = default;
+
+        model(const T& other): value(other) {}
+
+        model(T&& other): value(std::move(other)) {}
+
+        const std::type_info& type() override {
+            return typeid(T);
+        }
+
+        void* pointer() override {
+            return &value;
+        }
+
+        const void* pointer() const override {
+            return &value;
+        }
+
+        T value;
+    };
+
+    std::unique_ptr<interface> state_;
+
+protected:
+
+    template <typename T>
+    friend const T* any_cast(const unique_any* operand);
+
+    template <typename T>
+    friend T* any_cast(unique_any* operand);
+
+    template <typename T>
+    T* unsafe_cast() {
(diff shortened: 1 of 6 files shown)


tests/unit/test_unique_any.cpp
```diff
@@ -0,0 +1,394 @@
...
+    // be in a moved from state, and enforcing std::move(a) it
+    // is made clearer in the calling code that a has been invalidated.
+    //     util::any_cast<T&&>(a)
+    T val(util::any_cast<T&&>(std::move(a)));
+    EXPECT_EQ(val.value, 100);
+}
+
+// test any_cast(unique_any*) and any_cast(const unique_any*)
+//   - these have different behavior to any_cast on reference types
+//   - are used by the any_cast on refernce types
```
halfflat: small ytpo 'refernce'

tests/unit/test_unique_any.cpp
```diff
@@ -0,0 +1,394 @@
...
+    unique_any empty;
+    EXPECT_EQ(util::any_cast<int>(&empty), nullptr);
+    EXPECT_EQ(util::any_cast<int>((util::unique_any*)nullptr), nullptr);
+
+    // Check that constness of the returned pointer matches that the input.
+    {
+        unique_any a(42);
+        auto p = util::any_cast<int>(&a);
+        static_assert(std::is_same<int*, decltype(p)>::value,
+                "any_cast(any*) should not return const*");
```
halfflat: Why static asserts instead of `EXPECT_TRUE` as used in the reference tests? (This also applies to the corresponding tests in `test_any.cpp`.)

tests/unit/test_unique_any.cpp
```diff
@@ -0,0 +1,394 @@
...
+    // test that values were swapped
+    EXPECT_EQ(util::any_cast<int>(a2), 42);
+    EXPECT_EQ(util::any_cast<double>(a1), 3.14);
+
+    // test that underlying pointers did not change
+    EXPECT_EQ(pi, util::any_cast<int>(&a2));
+    EXPECT_EQ(pd, util::any_cast<double>(&a1));
+}
+
+// test operator=(any&&)
```
halfflat: Test comment out of date: this is testing `unique_any::operator=(unique_any&&)`.

tests/unit/test_unique_any.cpp
```diff
@@ -0,0 +1,394 @@
...
+    // that the value was moved, and not copied.
+    auto ptr = tmp.data();
+
+    unique_any a;
+    a = std::move(tmp);
+
+    auto vec = util::any_cast<std::vector<int>>(&a);
+
+    // ensure the value was moved
+    EXPECT_EQ(ptr, vec->data());
```
halfflat: This guarantee is a little hard to validate against the standard: it _is_ guaranteed, but requires knowing that the standard allocator has `propagate_on_container_move_assignment::value` true and that this in turn indicates that pointers to elements are not invalidated on move (clarified in http://www.open-std.org/JTC1/SC22/WG21/docs/lwg-active.html#2321).

I think it would be clearer to use e.g. `testing::nocopy<T>` and check the number of move-assign and move-ctor invokations.

tests/unit/test_unique_any.cpp
```diff
@@ -0,0 +1,394 @@
...
+    // check casting
+    {
+        auto a = make_unique_any<double>(42u);
+
+        EXPECT_EQ(typeid(double), a.type());
+        EXPECT_EQ(42.0, any_cast<double>(a));
+    }
+
+    // check forwarding of parameters to constructor
+    {
```
halfflat: For thoroughness, checking a 0-parameter and 2- (or more) parameter constructor would help ensure that the correct variadic implementation of `make_unique_any` was deployed. (Also applies to `make_any`).

tests/unit/test_unique_any.cpp
```diff
@@ -0,0 +1,394 @@
...
+        // take a pointer to the orignal data to later verify
+        // that the value was moved, and not copied.
+        auto ptr = tmp.data();
+
+        auto a = make_unique_any<std::vector<int>>(std::move(tmp));
+
+        auto vec = any_cast<std::vector<int>>(&a);
+
+        // ensure the value was moved
+        EXPECT_EQ(ptr, vec->data());
```
halfflat: Same issue as above regarding how this perfectly correct test is possibly not easily seen to be correct.

tests/unit/test_unique_any.cpp
bcumming: Carelessness.
More generally, which is the best approach?
Using `EXPECT_TRUE` has the benefit of not stopping compilation and running of the other tests if we run into a buggy compiler or if someone makes a breaking change.
But compile time checks give earlier and harder diagnosis of a problem.

bcumming: Changed to a run time check, i.e. `EXPECT_TRUE`

src/util/unique_any.hpp
```diff
@@ -0,0 +1,211 @@
...
+    unique_any(T&& other) {
+        using contained_type = util::decay_t<T>;
+        state_.reset(new model<contained_type>(std::forward<T>(other)));
+    }
+
+    unique_any& operator=(unique_any&& other) noexcept {
+        swap(other);
+        return *this;
+    }
+
```
bcumming: Yes. Deferring.

tests/unit/test_unique_any.cpp
bcumming: It is guaranteed, and quite a clear test on a type that is not trivially copyable. If I don't get the expected behavior with `std::vector`, I would like to know about it.

I extend by adding an additional test with `testing::no_copy`.

tests/unit/test_unique_any.cpp
bcumming: added some tests for 0 and more than 1 argument.

PR #259, opened 2017-05-10
Feature/generic cell groups

bcumming:
Refactor `model` and `recipe` to build models that have different cell types.

* Refactor `recipe::get_cell` to return `unique_any` so that.
  * All recipe definitions in tests and miniap had to be updated to use 
    the new interface.
* Make a `cell_group_factory` that forwards arguments for building a
  cell group to the appropriate `cell_group` constructor.
* Refactor `model` to use generic cell types
  * Constructor now delegates `cell_group` generation to the 
    `cell_group_factory`.
  * Add an implementation file `model.cpp` for model to reduce compilation
    times (by 2-7 seconds on my desktop).
* Refactor probe enumeration code in `model` and `cell_group`
  * add interface to `cell_group` for querying enumeration of probes
    in a `cell_group`
  * use this interface instead of directly computing enumeration in
    `model` constructor, which no longer has easy access to probe
    information.

### DIFF
--- a/miniapp/miniapp.cpp
+++ b/miniapp/miniapp.cpp
@@ -249,15 +249,17 @@ void write_trace_json(const sample_trace_type& trace, const std::string& prefix)
 }
 
 void report_compartment_stats(const recipe& rec) {
-}
-/*
-std::size_t ncell = rec.num_cells();
+    std::size_t ncell = rec.num_cells();
     std::size_t ncomp_total = 0;
     std::size_t ncomp_min = std::numeric_limits<std::size_t>::max();
     std::size_t ncomp_max = 0;
 
     for (std::size_t i = 0; i<ncell; ++i) {
-        std::size_t ncomp = rec.get_cell(i).num_compartments();
+        std::size_t ncomp = 0;
+        auto c = rec.get_cell(i);
+        if (auto ptr = util::any_cast<cell>(&c)) {
+            ncomp = ptr->num_compartments();
+        }
         ncomp_total += ncomp;
         ncomp_min = std::min(ncomp_min, ncomp);
         ncomp_max = std::max(ncomp_max, ncomp);
@@ -265,4 +267,3 @@ std::size_t ncell = rec.num_cells();
 
     std::cout << "compartments/cell: min=" << ncomp_min <<"; max=" << ncomp_max << "; mean=" << (double)ncomp_total/ncell << "\n";
 }
-*/
--- a/miniapp/miniapp.cpp
+++ b/miniapp/miniapp.cpp
@@ -33,7 +33,7 @@ using sample_trace_type = sample_trace<time_type, double>;
 using file_export_type = io::exporter_spike_file<global_policy>;
 void banner();
 std::unique_ptr<recipe> make_recipe(const io::cl_options&, const probe_distribution&);
-std::unique_ptr<sample_trace_type> make_trace(cell_member_type probe_id, probe_spec probe);
+std::unique_ptr<sample_trace_type> make_trace(probe_record probe);
 using communicator_type = communication::communicator<communication::global_policy>;
 
 void write_trace_json(const sample_trace_type& trace, const std::string& prefix = "trace_");
@@ -117,7 +117,7 @@ int main(int argc, char** argv) {
                 continue;
             }
 
-            traces.push_back(make_trace(probe.id, probe.probe));
+            traces.push_back(make_trace(probe));
             m.attach_sampler(probe.id, make_trace_sampler(traces.back().get(), sample_dt));
         }
 
@@ -207,7 +207,7 @@ std::unique_ptr<recipe> make_recipe(const io::cl_options& options, const probe_d
     }
 }
 
-std::unique_ptr<sample_trace_type> make_trace(cell_member_type probe_id, probe_spec probe) {
+std::unique_ptr<sample_trace_type> make_trace(probe_record probe) {
     std::string name = "";
     std::string units = "";
 
@@ -224,7 +224,7 @@ std::unique_ptr<sample_trace_type> make_trace(cell_member_type probe_id, probe_s
     }
     name += probe.location.segment? "dend" : "soma";
 
-    return util::make_unique<sample_trace_type>(probe_id, name, units);
+    return util::make_unique<sample_trace_type>(probe.id, name, units);
 }
 
 void write_trace_json(const sample_trace_type& trace, const std::string& prefix) {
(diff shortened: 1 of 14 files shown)


miniapp/miniapp_recipes.cpp
```diff
@@ -80,35 +81,36 @@ class basic_cell_recipe: public recipe {
...
         auto gen = std::mt19937(i); // TODO: replace this with hashing generator...
 
         auto cc = get_cell_count_info(i);
         const auto& morph = get_morphology(i);
         unsigned cell_segments = morph.components();
 
-        auto cell = make_basic_cell(morph, param_.num_compartments, cc.num_targets,
+        auto wrapped_cell = make_basic_cell(morph, param_.num_compartments, cc.num_targets,
                         param_.synapse_type, gen);
+        auto& c = util::any_cast<cell&>(wrapped_cell);
```
w-klijn: It was confusing to have cell named cell. In this case the one letter variable is better

miniapp/miniapp.cpp
```diff
@@ -33,7 +33,7 @@ using sample_trace_type = sample_trace<time_type, double>;
 using file_export_type = io::exporter_spike_file<global_policy>;
 void banner();
 std::unique_ptr<recipe> make_recipe(const io::cl_options&, const probe_distribution&);
-std::unique_ptr<sample_trace_type> make_trace(cell_member_type probe_id, probe_spec probe);
+std::unique_ptr<sample_trace_type> make_trace(probe_record probe);
```
w-klijn: Narrow interface is better indeed

src/mc_cell_group.hpp
```diff
@@ -59,8 +60,36 @@ class mc_cell_group: public cell_group {
             ++source_gid;
         }
         EXPECTS(spike_sources_.size()==n_detectors);
+
```
w-klijn: I guess this is from the model section? This is a better place

src/mc_cell_group.hpp
```diff
@@ -59,8 +60,36 @@ class mc_cell_group: public cell_group {
...
+            }
+            ++probe_gid;
+        }
+
     }
 
+    mc_cell_group(cell_gid_type first_gid, const std::vector<util::unique_any>& cells):
+        mc_cell_group(
+            first_gid,
+            util::transform_view(
```
w-klijn: I did not know that we have these functions. And reading a complate code base is not a good idea.

I think we should add a documentation section summarizing all the nice utils you have written. 
Not for this PR just an observation.

src/model.cpp
```diff
@@ -0,0 +1,250 @@
+#include <model.hpp>
+
+namespace nest {
+namespace mc {
+
+// Helper function for building cell groups used by model constructor
+// See bottom of file for definition.
+cell_group_ptr make_cell_group(
```
w-klijn: SHould this be on the cell_group interface. It has nothing to do with the model ( the model uses it.)
You dont expect it on this source location

src/model.cpp
```diff
@@ -0,0 +1,250 @@
...
+
+            auto group = domain_.get_group(i);
+            std::vector<util::unique_any> cells(group.end-group.begin);
+
+            for (auto gid: util::make_span(group.begin, group.end)) {
+                auto i = gid-group.begin;
+                cells[i] = rec.get_cell(gid);
+            }
+
+            cell_groups_[i] = make_cell_group(
```
w-klijn: Aah better then my implementation!

src/model.cpp
```diff
@@ -0,0 +1,250 @@
...
+                    group.kind, group.begin, std::move(cells), domain_.backend());
+            PL(2);
+        });
+
+    // store probes
+    for (const auto& c: cell_groups_) {
+        util::append(probes_, c->probes());
+    }
+
+    // generate the network connections
```
w-klijn: Yay breaking up the code in smaller named parts

src/model.cpp
```diff
@@ -0,0 +1,250 @@
...
+                PL(2);
+            });
+    };
+
+    // task that performs spike exchange with the spikes generated in
+    // the previous integration period, generating the postsynaptic
+    // events that must be delivered at the start of the next
+    // integration period at the latest.
+    auto exchange = [&] () {
+        PE("stepping", "communication");
```
w-klijn: Ok, looking at the future.
I was thinking about NOT delivering poisson spikes to global if they are used as single neuron input:
8000 hz for every neuron in the model would kill all performance. This would have to be done somewhere here.

src/model.cpp
```diff
@@ -0,0 +1,250 @@
...
+        threading::task_group g;
+        g.run(exchange);
+        g.run(update_cells);
+        g.wait();
+
+        t_ = tuntil;
+    }
+
+    // Run the exchange one last time to ensure that all spikes are output
+    // to file.
```
w-klijn: This feels desperate. Why do we need to do this? 
why do we have to run again? What is the exceptional situation?

src/mc_cell_group.hpp
bcumming: We are well aware of this, very well aware.

src/model.cpp
bcumming: Agreed. These heavy duty poisson generators that you are talking about should not be implemented as a cell type in the long run (for now they can).

Sam and I feel that they should be implemented as a synapse type in NMODL. Because the spike train that they generate is consumed by only one synapse, so there is no need to pass the spike information through the spike-event system.

src/model.cpp
bcumming: Spikes are output to file after each exchange.
We perform the exchange every `min_delay`ms.
When time stepping finishes half way through one of these `min_delay` periods, you have to ensure that the spikes that have been stored in local buffers are also output to file, which requires one last spike exchange.
This was a long standing bug, whereby some spikes at the end of a simulation were not being output.

src/model.cpp
bcumming: It is a factory for `cell_groups` (that should actually be in the name).
So it can't be part of the `cell_group` interface.
Logically, you can argue  that it belongs in its own file.
I implemented it in `model.cpp` because the constructor of `model` is the only spot where it is used.

src/mc_cell_group.hpp
bcumming: Yes it is from the model section.
This is a better spot, though we could argue that it still isn't the best spot. I am going to let Sam do the rationalization of this bit of code when he has the time!

miniapp/miniapp_recipes.cpp
bcumming: Hmmmm, I have actually reverted this code! I don't know why it has showed up in your review.
I agree that `cell` is not the best name because it clashes with the type name.

PR #261, opened 2017-05-10
Moving event delivery to back-end: multicore.

halfflat:
Defines the back-end event-delivery abstraction and provides an implementation for `multicore`. Breaks `gpu` back-end.

* Move `multi_event_stream` to `src/backends/multicore/multi_event_stream.hpp`.
* Simplify building process of `multi_event_stream`.
* Move common event-delivery types (`target_handle`, `deliverable_event`) to `src/backends/event.hpp`.
* Add zero-dt check/support to `multicore::matrix_state`.
* Add debugging helper `util::sepval` for printing/tracing container values.
* Fix end-of-integration check in `fvm_multicell`.
* Document event delivery process and abstraction in `src/backends/event_delivery.md`.

### DIFF
--- a/src/backends/multicore/fvm.hpp
+++ b/src/backends/multicore/fvm.hpp
@@ -3,14 +3,18 @@
 #include <map>
 #include <string>
 
+#include <backends/event.hpp>
 #include <common_types.hpp>
+#include <event_queue.hpp>
 #include <mechanism.hpp>
 #include <memory/memory.hpp>
 #include <memory/wrappers.hpp>
+#include <util/meta.hpp>
 #include <util/rangeutil.hpp>
 #include <util/span.hpp>
 
 #include "matrix_state.hpp"
+#include "multi_event_stream.hpp"
 #include "stimulus.hpp"
 #include "threshold_watcher.hpp"
 
@@ -39,9 +43,12 @@ struct backend {
     using host_view   = view;
     using host_iview  = iview;
 
-    /// matrix state
-    using matrix_state =
-        nest::mc::multicore::matrix_state<value_type, size_type>;
+    using matrix_state = nest::mc::multicore::matrix_state<value_type, size_type>;
+    using multi_event_stream = nest::mc::multicore::multi_event_stream;
+
+    // re-expose common backend event types
+    using deliverable_event = nest::mc::deliverable_event;
+    using target_handle = nest::mc::target_handle;
 
     //
     // mechanism infrastructure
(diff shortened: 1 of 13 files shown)


src/mechanism.hpp
```diff
@@ -39,7 +39,8 @@ class mechanism {
 
     using ion_type = ion<backend>;
 
-    mechanism(const_iview vec_ci, const_view vec_t, const_view vec_t_to, view vec_v, view vec_i, iarray&& node_index):
+    mechanism(value_type mech_id, const_iview vec_ci, const_view vec_t, const_view vec_t_to, view vec_v, view vec_i, iarray&& node_index):
```
bcumming: Should the `mech_id` argument have type `size_type`?

src/mechanism.hpp
```diff
@@ -66,10 +67,25 @@ class mechanism {
     virtual bool uses_ion(ionKind) const = 0;
     virtual void set_ion(ionKind k, ion_type& i, const std::vector<size_type>& index) = 0;
 
+    // TODO: virtualize, move to backend
```
bcumming: This will have to be virtualized for the GPU?

src/util/debug.hpp
```diff
@@ -65,6 +66,31 @@ void debug_emit_trace(const char* file, int line, const char* varlist, const Arg
     }
 }
 
+namespace impl {
+    template <typename Seq, typename Separator>
+    struct sepval {
```
bcumming: Needs documentation.

src/backends/multicore/fvm.hpp
```diff
@@ -39,9 +43,12 @@ struct backend {
     using host_view   = view;
     using host_iview  = iview;
 
-    /// matrix state
-    using matrix_state =
-        nest::mc::multicore::matrix_state<value_type, size_type>;
+    using matrix_state = nest::mc::multicore::matrix_state<value_type, size_type>;
+    using multi_event_stream = nest::mc::multicore::multi_event_stream;
+
+    // re-expose common backend event types
```
halfflat: It is part of the shared API between the lowered cell and the back-end event delivery.

The `fvm_multicell` depends on and uses types and methods defined by the `backend` class, but not the other way around. Correspondingly, these have to live in the `backend` side.

The only difference with other `backend`-defined objects is that these structs are in-fact common across the `backend` implementations.

src/mechanism.hpp
halfflat: It should!

src/mechanism.hpp
halfflat: Yup! (Already is virtualized in the ongoing work on another branch.)

bcumming:
lgtm.

PR #268, opened 2017-05-16
pade approximation of exp in cnexp integration

### LINKED ISSUE #265: Pade approximation of exponentials
We should investigate using a lower order approximation to the exponential used for time integration in the _cnexp_ time stepping method in NMODL. The time integration scheme is first order in time, so there might be an opportunity to use a less expensive operation that computing the full exponential while not reducing the accuracy of the solver.

https://en.wikipedia.org/wiki/Pad%C3%A9_table

bcumming:
This PR replaces the expensive call to `exp` when integrating dependent variables in the cnexp method with a 2nd order pade approximation.

  * modcc inserts `exp_pade_11` and `exp_pade_22` functions into every module, which define pade approximations of 2nd and 4th order respectively (m=n=1 and m=n=2).
  * the generated solver code calls the `exp_pade_11` function instead of the built in `exp` unary operator

The validation tests pass for both the 2nd and 4th order approximations, so we we go with the 2nd order.

fixes #265

### DIFF
--- a/modcc/module.cpp
+++ b/modcc/module.cpp
@@ -209,119 +209,21 @@ bool Module::semantic() {
         return true;
     };
 
+    // Add built in function that approximate exp use pade polynomials
+    functions_.push_back(
+        Parser{"FUNCTION exp_pade_11(z) { exp_pade_11=(1+0.5*z)/(1-0.5*z) }"
+        }.parse_function());
+    functions_.push_back(
+        Parser{ "FUNCTION exp_pade_22(z)"
+        "{ exp_pade_22=(1+0.5*z+0.08333333333333333*z*z)/(1-0.5*z+0.08333333333333333*z*z) }"
+        }.parse_function());
+
     // move functions and procedures to the symbol table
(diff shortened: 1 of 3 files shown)


modcc/module.cpp
```diff
@@ -209,119 +209,21 @@ bool Module::semantic() {
...
 
+    // Add built in function that approximate exp use pade polynomials
+    functions_.push_back(
+        Parser{"FUNCTION exp_pade_11(z) { exp_pade_11=(1+0.5*z)/(1-0.5*z) }"
+        }.parse_function());
+    functions_.push_back(
+        Parser{ "FUNCTION exp_pade_22(z)"
+        "{ exp_pade_22=(1+0.5*z+0.08333333333333333*z*z)/(1-0.5*z+0.08333333333333333*z*z) }"
+        }.parse_function());
+
```
halfflat: Indentation, spacing around args here is a bit weird.

bcumming: yeah, I can fix it. It isn't immediately clear to me how to indent code inside strings...

PR #270, opened 2017-05-16
remove ineffectual compiler directives from nmodl generated code

bcumming:
Remove some old compiler directives and properties from modcc-generated kernels that were not used by the intel compiler, and were not recognised by gcc. 

Now the optimized x86 kernels can be compiled by both gcc and intel compilers.

### DIFF
--- a/modcc/cprinter.cpp
+++ b/modcc/cprinter.cpp
@@ -336,9 +336,7 @@ std::string CPrinter::emit_source() {
     text_.add_line("array data_;");
     for(auto var: array_variables) {
         if(optimize_) {
-            text_.add_line(
-                "__declspec(align(array::alignment())) value_type *"
-                + var->name() + ";");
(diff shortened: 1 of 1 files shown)


modcc/cprinter.cpp
```diff
@@ -661,13 +659,7 @@ void CPrinter::emit_api_loop(APIMethod* e,
 }
 
 void CPrinter::print_APIMethod_unoptimized(APIMethod* e) {
-    // there can not be more than 1 instance of a density channel per grid point,
-    // so we can assert that aliasing will not occur.
-    if(optimize_) text_.add_line("#pragma ivdep");
-
```
vkarak: Do we have already AVX2 vectorised backends? If no, I should hurry up then!

bcumming: We do not have an AVX2 backend!

PR #276, opened 2017-05-18
Issue/271 regression tests

w-klijn:
Adds regression test suite

Started using make regression_test
 - Start a full run of nestmc (without need to set environments)
 - Compares output with blessed data using a delta
 - Display type of problem and which line on problems

Includes:
 - compare.py script performing the comparison of the data files
 - validate.sh.in uninstantiated runner script, will be configured by cmake.
 - data/*   Blessed data sets 

After testing the following delta are found:

|machine |rgaph|alltoall|ring|
|---------|------|-------|----|
|my pc   |0.000001  |0.000001   |0.000001   |  
|  julia headnode |  0.000001 |   0.000001| 0.000001  |  
|  Julia compute |0.001   | 0.001  | 0.000001  |  
|juron headnode |0.001 | 0.000001| 0.000001|
|Juron compute | ~0.1 | ~0.1| ~0.1|

I am setting the delta in the tests to 0.001, since we are still working on the cuda implementation


tests/regression/scripts/compare.py
```diff
@@ -0,0 +1,136 @@
...
+            
+            exit(1) #failure
+
+        line_data = (time, line_idx, stripped_line)
+        parsed_data[gid].append(line_data)
+
+    return parsed_data
+
+
+def compare(path1, data1, path2, data2, delta):
```
w-klijn: I adressed this issue and now treat it all the same

tests/regression/scripts/compare.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    compares data1 and data2 on spike time equality.
+    if a problem is found at the delta level, store result
+    if whole spikes are missing just exit!
+
+    print all errors and then exit(1)
+    """
+    combined_data = collections.defaultdict(lambda : [[],[]])
+    
+    for gid, spike_data in data1.items():
```
w-klijn: Combine all the spike data into a single dict with gid keys

tests/regression/scripts/compare.py
```diff
@@ -0,0 +1,143 @@
...
+    different_spikes = []
+    for gid, (data_1, data_2)in combined_data.items():
+        gid_list1 = data_1
+        gid_list2 = data_2
+
+        if len(gid_list1) != len(gid_list2):
+            for idx, (time1, time2) in enumerate(map(None, gid_list1, gid_list2)):
+                # We have to loop all spikes, check here if we have missing spikes 
+                # and treat those different
+                if time1 == None or time2 == None:
```
w-klijn: When there are spikes missing treat as such else test the difference

PR #279, opened 2017-05-19
Issue/241 frequency spiking neuron

w-klijn:
Replace the 20 artificial spikes we inserted until now with spikes generated with a frequency spiking cell.
This is a cell that spikes regularly for a set period.
This connections outgoing from this cell mirror all the outgoing connections of cells with gid % 20 == 0.
This means that 5% extra connections have been added to the model to implement the equivalent behaviour with an 'implemented' neuron.

To implement this behavoir a new cell_group type has been introduced. 
The fs_cell is extremely small and is not implemented on the backends.

Includes:
Unit test for the fs_cell functionality


src/fs_cell_group.hpp
```diff
@@ -0,0 +1,98 @@
...
+        using util::make_span;
+
+        auto source_gid = cell_gid_type{ gid_base_ };
+        for (auto i : make_span(0, cells.size())) {
+            // Copy all the fs_cells
+            cells_.push_back(util::any_cast<fs_cell>(cells[i]));
+
+            // create a lid to gid map
+            spike_sources_.push_back(source_id_type{ source_gid, 0 });
+            ++source_gid;
```
halfflat: Given that we already have an index `i` in this loop, why maintain a separate counter for `source_gid`? I.e., I think it would be simpler to just have
```
    spike_sources_.push_back({gid_base_+i, 0});
```
or equivalent.

tests/CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ add_subdirectory(performance)
 # Microbenchmarks.
 add_subdirectory(ubench)
 
+# regression / delta tests
+# Employing the full simulator. validated using deltas on output data
+add_subdirectory(regression)
```
halfflat: Can you split out the regressions PR contents from this PR? It would make it easier to review…

tests/unit/test_fs_cell.cpp
```diff
@@ -0,0 +1,60 @@
...
+    // Expect 10 excluding the 0.2
+    spikes = sut.spikes_until(0.2);
+    EXPECT_EQ(size_t(10), spikes.size());
+}
+
+TEST(fs_cell, cell_kind_correct)
+{
+    fs_cell sut(0.1, 0.01, 0.2);
+
+    EXPECT_EQ(cell_kind::regular_frequency, sut.get_cell_kind());
```
halfflat: Is this code possibly out of sync? The `cell_kind` enum above has `fs_cell` instead.

miniapp/io.hpp
```diff
@@ -17,7 +17,7 @@ namespace io {
 
 struct cl_options {
     // Cell parameters:
-    uint32_t cells = 1000;
+    uint32_t cells = 1001;  // The last cell is always a regular spiking neuron that fires once at t=0
```
bcumming: I think it is better if the number of cells in the model does not include the artificial spike source cell.
From a user's point of view, the number of cells is the number of "real" cells.
So I would make the default number of cells 1000, and have the recipe insert the additional spike source cell automatically.

miniapp/miniapp.cpp
```diff
@@ -193,7 +188,9 @@ std::unique_ptr<recipe> make_recipe(const io::cl_options& options, const probe_d
     p.morphology_round_robin = options.morph_rr;
 
     p.num_compartments = options.compartments_per_segment;
-    p.num_synapses = options.all_to_all? options.cells-1: options.synapses_per_cell;
+
+    // TODO: Put all recipe parameters in the recipes file
+    p.num_synapses = options.all_to_all? options.cells-2: options.synapses_per_cell;
```
bcumming: The -2 is a bit of a magic number. By not counting the additional spiking cell in `options.cells`, this becomes -1, which makes sense.

miniapp/miniapp_recipes.cpp
```diff
@@ -82,6 +82,11 @@ class basic_cell_recipe: public recipe {
     cell_size_type num_cells() const override { return ncell_; }
 
     util::unique_any get_cell(cell_gid_type i) const override {
+        // The last 'cell' is a fs_cell with one spike at t=0
+        if (i == ncell_-1) {
```
bcumming: see above comments. Not counting the spike source in `ncell_` removes quite a few of these `-1` expressions.

src/fs_cell.hpp
```diff
@@ -0,0 +1,79 @@
...
+    {}
+
+    /// Return the kind of cell, used for grouping into cell_groups
+    cell_kind  get_cell_kind() const  {
+        return cell_kind::fs_neuron;
+    }
+
+    /// Collect all spikes until tfinal.
+    // updates the internal time state to tfinal as a side effect
+    std::vector<time_type> spikes_until(time_type tfinal)
```
bcumming: code style: no new line for opening brace `{`

src/fs_cell_group.hpp
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    virtual ~fs_cell_group() = default;
+
+    cell_kind get_cell_kind() const override
```
bcumming: code style: no new line for opening brace `{`

tests/CMakeLists.txt
bcumming: I agree, the regression tests need to be reviewed separately.

w-klijn:
I implemented all comment EXCEPT the rss_cell as an explicit part of the description: The neuron is part of the model and should therefore not be a surprise . I could split it out into 1000 normal neurons and 1 regular spiking neuron. But the current recipe structure and definition currently is full of magic . 
I would suggest to defer this change until we have had a design session on the model creation API.

w-klijn:
I changed the recipes to add the rss_cell automatically and have the user only define the nr of mc cells.
All review comments have been addresses.

PR #280, opened 2017-05-19
Add required thread synchronization to matrix kernel.

halfflat:
There is a potential data race in the `assemble_matrix_interleaved` kernel, where threads in a different warp can overwrite the `buffer_v` and `buffer_i` values before they are used to update the `d` and `rhs` vectors.

This race has been exercised in the asynchronous event delivery branch.

* Add `__syncthreads()` to assemble matrix interleaved kernel after `d` and `rhs` update.

### DIFF
--- a/src/backends/gpu/kernels/assemble_matrix.hpp
+++ b/src/backends/gpu/kernels/assemble_matrix.hpp
@@ -94,6 +94,8 @@ void assemble_matrix_interleaved(
             rhs[store_pos] = gi*buffer_v[blk_pos] - buffer_i[blk_pos];
         }
 
+        __syncthreads();
+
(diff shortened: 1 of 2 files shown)


halfflat:
It's something that's tricky to set up a unit test for: it requires a particular set of scheduling decisions by the GPU. These were triggered exactly after three integration rounds in the async-integration branch version of the miniapp, but I'm having trouble coming up with a way to trigger it in isolation.

halfflat:
The second `__syncthreads()` could go at the end of the loop instead. I placed it here, because it was directly after the reads to shared memory that required the synchronization — I thought this was clearer.

halfflat:
Turns out I missed a couple of other points where it is required; PR updated.

PR #282, opened 2017-05-23
modcc: AVX2 backend for mechanisms

vkarak:
We're still 3.5% and 5% slower than the icc `#pragma` version on Haswell and Broadwell, respectively. I will have a deeper look later on, but I've put this PR up in case you want to merge anytime.


vkarak:
Still not, due to the SVML dependency...

PR #288, opened 2017-06-02
AVX512 CMake target

vkarak:
Adds a new `AVX512` target for processors supporting only the core AVX512 functionality.


vkarak:
Resolved conflicts with master.

PR #290, opened 2017-06-02
Reduce asynchronous integration overheads

### LINKED ISSUE #285: Investigate and address performance regression on async-integration branch.
Like-for-like comparisons of performance with the miniapp default networks between `devel/async-integration` and `master` indicate that the asynchronous integration code is barely affected by event delivery, but wears a significant overhead in time stepping.

Median time-stepping times over five runs on `greina7` (P100, 8 cores without hyperthreading):

| time (s) | branch | device | t | g | n |
|--:|--:|--:|--:|--:|--:|
|  9.62 | devel | cpu |  100 |    1 |  1000 |
| 95.67 | devel | cpu |  100 |    1 | 10000 |
|  9.30 | devel | cpu |  100 |    4 |  1000 |
|  0.69 | devel | gpu |  15 | 1000 |  1000 |
| 1.44 | devel | gpu |   30 | 1000 |  1000 |
|  6.22 | devel | gpu | 100 | 1000 |  1000 |
| 23.45 | devel | gpu |100 | 10000 | 10000 |
|  7.89 | master | cpu | 100 |    1 |  1000 |
| 78.32 | master | cpu |  100 |    1 | 10000 |
| 12.10 | master | cpu |  100 |    4 |  1000 |
|  0.56 | master | gpu |   15 | 1000 |  1000 |
| 26.75 | master | gpu |  30 | 1000 |  1000 |

Note that in these models, stopping at _t = 15_ avoids any event delivery. Times for the master branch gpu version with the default _t = 100_ took too long to be worth benchmarking.

Based on these figures, without event delivery, integration is taking approximately 22% longer on both CPU and GPU in the development branch:

| device | t | g | n | master (s) | devel (s) | overhead |
|--:|--:|--:|--:|--:|--:|--:|
| cpu | 100 | 1 | 1000 | 7.89 | 9.62 | 21.9% |
| gpu | 15 | 1000 | 1000 | 0.56 | 0.69 | 23.2% |

This overhead is likely larger than it need be: profile the code and reduce this overhead.

halfflat:
*NOTE:* This PR includes the contents of PR #284 — that PR should be merged first.

Fixes #285.

Addresses issue identified in #285, reducing asynchronous integration implementation overhead to approximately 11-12% in the miniapp example tested.

* Explicitly compute and store per-cell and per-compartment `dt` from integration time bounds.
* Determine lower bound on number of integration steps per interval in order to avoid explicit checking of minimum cell times each step.
* Avoid any time value checking for samplers in the integration loop if none of them could be triggered in the interval.

### DIFF
--- a/modcc/sss
+++ b/modcc/sss
@@ -1,0 +1,21 @@
+__inline __m512d
+__attribute__ ((__gnu_inline__, __always_inline__, __artificial__))
+_mm512_i32gather_pd (__m256i __index, double const *__addr, int __scale)
+{
+  __m512d v1_old = _mm512_undefined_pd ();
+  __mmask8 mask = 0xFF;
+
+  return (__m512d) __builtin_ia32_gathersiv8df ((__v8df) v1_old,
+                                                __addr,
+                                                (__v8si) __index, mask,
+                                                __scale);
+}
+
+_mm256_i32gather_pd (double const *base, __m128i index, const int scale)
+
+
+clean up:
+
+_mm512_i32gather_pd(__m256i index, double const *base, int scale)
+_mm256_i32gather_pd(double const *base, __m128i index, int scale)
+
--- a/modcc/cprinter.cpp
+++ b/modcc/cprinter.cpp
@@ -88,8 +88,8 @@ std::string CPrinter::emit_source() {
     //////////////////////////////////////////////
     int num_vars = array_variables.size();
     text_.add_line();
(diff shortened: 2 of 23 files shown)


bcumming:
The conflicts that were introduced when merging #284 have to be resolved before this PR can be reviewed.

modcc/sss
```diff
@@ -0,0 +1,21 @@
+__inline __m512d
```
bcumming: This looks like a scratch file that was accidentally added.

halfflat: Totally is, yes!

PR #296, opened 2017-06-19
Devel/async integration

halfflat:
Merge latest master.


halfflat:
NB: verified unit and validation test on knl and avx2.

PR #301, opened 2017-06-23
Improve reduce by key GPU performance.

bcumming:
Optimized reduce by key used by the GPU back end when accumulating synapse
current contributions to compartment currents. This leads to significant
speedup in the miniapp for cells with few compartments and many synapses.
  * Implemented `gpu::reduce_by_key` device function that uses warp intrinsics
    to perform reduction between threads in a warp before using a global
    atomic update to store the result.
  * Added unit tests for `reduce_by_key` functionality.
  * Added micro benchmarks that compare against using CUDA atomics.
  * Modified `CudaPrinter` modcc class to emit `reduce_by_key` in place
    of `cudaAtomicAdd` functions.

Some improvements to meter reporting
  * shortened names of metering regions in miniapp to make them easier to grep
  * json is no longer used as an intermediate data type when gathering
    distributed meters into a single report, instead conversion to json is
    performed just before writing to file.
  * added a print function for summarizing meter results to a stream.

### DIFF
--- a/src/backends/gpu/kernels/reduce_by_key.hpp
+++ b/src/backends/gpu/kernels/reduce_by_key.hpp
@@ -1,0 +1,132 @@
+#pragma once
+
+#include <cstdint>
+#include "detail.hpp"
+
+namespace nest {
+namespace mc {
+namespace gpu {
+
+namespace impl{
+
+// return the power of 2 that is less than or equal to i
+__device__ __inline__
+unsigned rounddown_power_of_2(unsigned i) {
+    // handle power of 2 and zero
+    if(__popc(i)<2) return i;
+
+    return 1u<<(31u - __clz(i));
+}
+
+// The __shfl warp intrinsic is only implemented for 32 bit values.
+// To shuffle a 64 bit value (usually a double), the value must be copied
+// with two 32 bit shuffles.
+// get_from_lane is a wrapper around __shfl() for both single and double
+// precision.
+
+__device__ __inline__
+double get_from_lane(double x, unsigned lane) {
+    // split the double number into two 32b registers.
+    int lo, hi;
+
+    asm volatile( "mov.b64 { %0, %1 }, %2;" : "=r"(lo), "=r"(hi) : "d"(x) );
+
+    // shuffle the two 32b registers.
+    lo = __shfl(lo, lane);
+    hi = __shfl(hi, lane);
+
+    // return the recombined 64 bit value
+    return __hiloint2double(hi, lo);
+}
+
+__device__ __inline__
+float get_from_lane(float value, unsigned lane) {
+    return __shfl(value, lane);
+}
+
+// run_length_type Stores information about a run length.
+//
+// A run length is a set of identical adjacent indexes in an index array.
+//
+// When doing a parallel reduction by index each thread must know about
+// which of its neighbour threads are contributiing to the same global
+// location (i.e. which neighbours have the same index).
+//
+// The constructor takes the thread id and index of each thread
+// and the threads work cooperatively using warp shuffles to determine
+// their run length information, so that each thread will have unique
+// information that describes its run length and its position therein.
+struct run_length_type {
+    unsigned left;
+    unsigned right;
+    unsigned width;
+    unsigned lane_id;
+
+    __device__ __inline__
+    bool is_root() const {
+        return left == lane_id;
+    }
+
+    __device__ __inline__
+    bool may_cross_warp() const {
+        return left==0 || right==threads_per_warp();
+    }
+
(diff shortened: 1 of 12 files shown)


src/profiling/meter_manager.cpp
```diff
@@ -101,39 +103,54 @@ nlohmann::json to_json(const measurement& mnt) {
     };
 }
 
-nlohmann::json to_json(const meter_manager& manager) {
+// Simple type for gathering distributed meter information
+struct meter_report {
+    std::vector<std::string> checkpoints;
+    unsigned num_domains;
+    std::string global_model;
```
halfflat: `global_model` is an odd name for this field: can we use something like `global_communicator` instead? In fact, this could just hold the enum value, with the conversion to string performed in `to_json()` — this allows the `meter_report` to be the purely programmatic representation, and the `to_json()` output to be the text serialized representation.

src/profiling/meter_manager.cpp
```diff
@@ -142,6 +159,42 @@ void save_to_file(const meter_manager& manager, const std::string& name) {
     }
 }
 
+// Print easy to read report of meters so a stream.
+void print(const meter_manager& manager, std::ostream& o) {
```
halfflat: `util::print` is a very generic name for a function like this, and also conflates the text representation (a purely local operation) with the global communication rank test. I'd prefer that firstly the rank test be removed, and made the responsibility of the caller, and secondly that this be made an `operator<<` overload, or a method on the report struct, or at least called `pretty_print`.

tests/unit/test_reduce_by_key.cu
```diff
@@ -0,0 +1,107 @@
...
+
+    if (tid<n) {
+        gpu::reduce_by_key(src[tid], dst, index[tid]);
+    }
+}
+
+template <typename T>
+void reduce(const std::vector<T>& in, std::vector<T>& out, const std::vector<int>& index, unsigned block_dim=128) {
+    EXPECT_EQ(in.size(), index.size());
+
```
halfflat: Just to make things a bit more robust, would you consider testing for monotonicity of `index`, and automatically creating, zeroing and returning the `out` vector to suit? It would also remove a bit of the code in the test functions below.

src/backends/gpu/kernels/reduce_by_key.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+namespace impl{
+
+// return the power of 2 that is less than or equal to i
+__device__ __inline__
+unsigned rounddown_power_of_2(unsigned i) {
+    // handle power of 2 and zero
+    if(__popc(i)<2) return i;
+
+    return 1u<<(31u - __clz(i));
```
bcumming: Sure, I have made the change.

The API accepting only `int` is an oversight. It behaves as "expected", as the following code illustrates:
```
#include <limits>
#include <cstdio>
#include <iostream>

__global__
void krnl(unsigned x) {
    printf("%-12d : %d\n", x, __clz(x));
}

int main() {
    unsigned r=1;
    krnl<<<1,1>>>(0);
    for (int i=0; i<32; ++i) {
        krnl<<<1,1>>>(r);                                                                                
        r <<= 1;
    }   

    krnl<<<1,1>>>(std::numeric_limits<unsigned>::max());
    cudaDeviceSynchronize();

    return 0;
}

=== output ===
> nvcc clz.cu -arch=sm_35 && ./a.out
0            : 32
1            : 31
2            : 30
4            : 29
8            : 28
16           : 27
32           : 26
64           : 25
128          : 24
256          : 23
512          : 22
1024         : 21
2048         : 20
4096         : 19
8192         : 18
16384        : 17
32768        : 16
65536        : 15
131072       : 14
262144       : 13
524288       : 12
1048576      : 11
2097152      : 10
4194304      : 9
8388608      : 8
16777216     : 7
33554432     : 6
67108864     : 5
134217728    : 4
268435456    : 3
536870912    : 2
1073741824   : 1
-2147483648  : 0
-1           : 0
```

src/profiling/meter_manager.cpp
bcumming: Done.

src/profiling/meter_manager.cpp
bcumming: Agreed, I have improved this based on your suggestions. Now there is an `operator<<`, and the report struct is generated directly by the caller (the minapp in our case), and output from the calling code. The `util::save_to_file` has been removed completely.

tests/unit/test_reduce_by_key.cu
bcumming: done.

PR #305, opened 2017-06-27
LIF neuron implementation with CPU backend

kabicm:
1) Implementation of LIF neuron model with no kernel and no external input (I_e=0).

The input current to each neuron is therefore just the sum of all the weights of incoming spikes

Binning policy and sampling are ignored!

We integrate in jumps dt = min(t_final - t, t_event - t), since we know the exact solution of the differential equal describing the membrane potential.

We use std::queue inside lif_cell_group.hpp for events waiting to be processed in advance, because we don't require the incoming events to be sorted. Instead, as soon as advance is invoked, it redistributed the events into event_queue of each cell. Since event_queue is a priority_queue, it will sort the events on the fly. This is done for efficiency reasons, because it is faster to have many small priority queues (one for each cell), than to have one large priority queue of all the events inside the cell group.

2) Implementation of Poisson cell group where each neuron produces Poisson-distributed spikes independently. These cells should not receive any events, but if it happens, the exception is thrown. The seed is hardcoded at the moment, so the official solution for seeds should be found.

3) Implementation of Brunel networks consisting of 3 neuronal populations: excitatory, inhibitory and Poisson population. Each neuron receives fixed number of connections from all three populations. Due to the refractory period, the activity of the network is mostly driven by Poisson input and not due to the recurrent connections.


tests/unit/test_lif_cell_group.cpp
```diff
@@ -0,0 +1,195 @@
...
+    EXPECT_EQ(3, spikes.size());
+}
+
+TEST(lif_cell_group, spikes_testing) {
+    std::vector<util::unique_any> cells;
+    cells.emplace_back(lif_cell_description());
+    
+  
+    
+    std::unique_ptr<lif_cell_group> group = std::unique_ptr<lif_cell_group>(static_cast<lif_cell_group*>(cell_group_factory(
```
bcumming:
```
auto group = cell_group_factory(cell_kind::lif_neuron, 0, cells, backend_policy::use_multicore);                                                                                                                    
auto gp = dynamic_cast<lif_cell_group*>(group.get());

```

kabicm:
The requested changes should be implemented now!

tests/unit/test_brunel_network.cpp
```diff
@@ -0,0 +1,254 @@
...
+            cell_connection conn;
+            conn.source = {i, 0};
+            conn.dest = {gid, 0};
+            conn.weight = weight_exc_;
+            conn.delay = delay_;
+
+            connections.push_back(conn);
+        }
+
+        // Sample incoming inhibitory connections.
```
bcumming: The for loop can be expressed more succinctly:

```
for (auto i: sample_subset(gid, ncells_exc_, ncells_exc_ + ncells_inh_, in_degree_inh_)) {
   ...
}
```

PR #311, opened 2017-06-29
Fix model ctor exception behaviour.

### LINKED ISSUE #310: Bad behaviour when recipe method throws.
Concrete demo to follow, but @w-klijn has observed non-deterministic segmentation faults when `recipe::get_cell_description` throws an exception.

This seems very likely to be due to a thread in the thread pool propagating the exception up out of `model::model`, where the `model` object is destroyed while other threads are still writing to its members.

halfflat:
Fixes #310.

* Use `std::packaged_task` to wrap recipe description queries in `model` constructor.
* Add unit test that is sensitive to correct exception handling in cell group creation tasks in the `model` constructor.

### DIFF
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -61,6 +61,7 @@ model::model(const recipe& rec, const domain_decomposition& decomp):
             });
     }
 
+    tasks.wait();
     cell_groups_.reserve(n_group);
     for (auto& f: cell_group_future) {
         cell_groups_.push_back(f.get());
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -1,5 +1,7 @@
 #include <model.hpp>
 
+#include <future>
+#include <type_traits>
 #include <vector>
 
 #include <backends.hpp>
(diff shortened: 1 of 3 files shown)


bcumming:
I don't think that an ad-hoc solution is a good thing here. Exception handling should be tackled systematically.

halfflat:
I agree it should be, but in the short-term we do have this bug. I think there is merit in adopting a shorter term solution pending a better, structural fix.

halfflat:
(This PR might not be that best short term solution, though!)

bcumming:
It isn't a big issue for us currently.
Any solution, would have to be compatible with the TBB and serial backends.

halfflat:
Defending this code, it is indeed compatible with any threading back-end, AFAICT. By wrapping it up in a `packaged_task`, it presents exactly the same interface — a `void ()` function — but looks after its own exceptions.

halfflat:
Hmm, this PR seems to be missing a required `wait` on the task group, which I thought I had added on the requisite branch. I'll make this fix.

bcumming:
There was no planning or agreement that this was a priority task before the PR was made.
We have other, much higher priority work to do, so let's make an issue for this, and get to it in time.

bcumming:
Can we close this PR?

halfflat:
We can close it if there's an issue made to cover the lack. Personally I think there's no harm in merging it, given that it fixes a bug and the 'right' fix will be significantly in the future.

bcumming:
This fix just irks me, for some reasons I can put my finger on, and also because it just irks me.

All complexity of a bug in the threading code has been moved forward into user code.

What happens when an exception is thrown with TBB or the serial back ends? Is this something that could be fixed better inside the cthread implementation if this isn't a problem with the other back ends?

Does this affect performance when running with many threads and TBB? I have concerns about serialising pushing the cell group task creation. This is working against the divide-and-conquere approach taken by TBB in a `parallel_for_all` call.

bcumming:
How about this as a proposal: move this infrastructure into the cthread `parallel_for` call? This means we get to keep the `parallel_for` interface, without affecting other back ends (assuming that TBB does a better job of handling exceptions.)

halfflat:
Sounds good to me; I'll leave a note on #310.

bcumming:
This is now being tracked in #310.

PR #314, opened 2017-07-03
306 spike from file

w-klijn:
Import spikes from a file supplied as command line argument

Functionality is contained in dss_cell_group, storing the spikes and performing the correct emitting at timesteps as supplied in its description.
An ostream spike time parser 
A new command line argument -I with a Path.

Includes:
unit test
Doc strings

Produces equal spike times on the regression test if a single spike at 0.0 is inserted.


miniapp/io.cpp
```diff
@@ -389,6 +409,64 @@ std::ostream& operator<<(std::ostream& o, const cl_options& options) {
...
+
+/// Parse spike times from a stream
+/// A single spike per line, white space is ignored
+/// All characters after a parsed floating point nr on a line are ignored
+/// Throws a usage error when parsing fails
+///
+/// Returns a unique_ptr to a vector of time_type
+
+std::unique_ptr<std::vector<time_type> > parse_spike_times_from_stream(
+    std::ifstream & fid) {
```
bcumming: This function is overly complicated, and will actually fail on some cases, e.g. in the following file:
```
1 2.3
3.2 2.3
```
The output will be `[12.3, error]`, when you would expect `[1, 3.2]`.

Here is a simpler and safer approach:
```
std::vector<time_type> times;
std::string line;                                                                                        
while (std::getline(fid, line)) {
    std::stringstream s(line);

    time_type t;
    s >> t >> std::ws;

    if (!s || s.peek()!=EOF) {
        throw std::runtime_error(                                                                        
            util::sprintf(
                "Unable to parse spike file on line %d: \"%s\"\n",
                times.size(), line));
        }
    }

    times.push_back(t);
}
```

miniapp/io.cpp
```diff
@@ -389,6 +409,64 @@ std::ostream& operator<<(std::ostream& o, const cl_options& options) {
...
+
+/// Parse spike times from a file supplied in path
+/// A single spike per line, white space is ignored
+/// All characters after a parsed floating point nr. on a line are ignored
+/// Throws a usage error when opening file or parsing fails
+///
+/// Returns a unique_ptr to a vector of time_type
+
+std::unique_ptr<std::vector<time_type> > get_parsed_spike_times_from_path(
+    const std::string& path) {
```
bcumming: again, this indentation is very misleading. Put the curly brace on a new line.

bcumming: use the `util::path` code for this.

src/dss_cell.hpp
```diff
@@ -0,0 +1,82 @@
...
+        dss_cell_description(std::vector<time_type> &  spike_times):
+            spike_times(spike_times)
+        {}
+    };
+
+    /// Construct a dss cell from its description
+
+    dss_cell(dss_cell_description descr){
+        spike_times_.reserve(descr.spike_times.size());
+        std::copy(descr.spike_times.begin(), descr.spike_times.end(), back_inserter(spike_times_));
```
bcumming:
```
dss_cell(dss_cell_description descr):
    spike_times(descr.spike_times)
{
    if (!std::is_sorted(spike_times_.begin(), spike_tmies_.end())) {
        util::sort(spike_times);
    }
}
```

miniapp/io.cpp
w-klijn: Done, although,

 util::sprintf does not exist, I used util::strprintf. 
I observed that this function eats the first character of the supplied string (to print the ""??). Is this expected behavior?

miniapp/io.cpp
w-klijn: Done

The eating of the first character does not happen when I emit a string with strprintf on this location... interesting

miniapp/miniapp_recipes.cpp
```diff
@@ -84,8 +87,15 @@ class basic_cell_recipe: public recipe {
     }
 
     util::unique_any get_cell_description(cell_gid_type i) const override {
-        // The last 'cell' is a rss_cell with one spike at t=0
+        // The last 'cell' is a spike source cell. Either a regular spiking
+        // or a spikes from file.
         if (i == ncell_) {
+            if (param_.spike_file_input) {
+                auto spike_times = io::get_parsed_spike_times_from_path(param_.input_spike_path);
+                return util::unique_any(std::move(
```
w-klijn: I did the same for the line below

miniapp/io.cpp
bcumming: Do you mean that it eats the first character of `line`? That is not expected, I think.

miniapp/miniapp_recipes.hpp
```diff
@@ -39,6 +39,11 @@ struct basic_recipe_param {
 
     // If true, iterate through morphologies rather than select randomly.
     bool morphology_round_robin = false;
+
+    // If true we are importing the spikes injected in the network from file
+    // instead of a single spike at t==0
```
w-klijn: Done

src/dss_cell.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+        /// The description needs a vector of doubles for the description
+
+        dss_cell_description(std::vector<time_type> &  spike_times):
+            spike_times(spike_times)
+        {}
+    };
+
+    /// Construct a dss cell from its description
+
```
w-klijn: removed

src/dss_cell.hpp
```diff
@@ -0,0 +1,82 @@
...
+namespace nest {
+namespace mc {
+
+/// data spike source: A cell that generates spikes provided as a vector of
+/// floating point valued spiketimes at the start of a run
+
+class dss_cell {
+public:
+    struct dss_cell_description {
```
w-klijn: Indeed. But they are always used together and always change together.
I agreed that you want to have them separate to have the recipe produce description and not objects. What is the added benefit of having it in a separate sourcefile?

src/dss_cell.hpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /// Return the kind of cell, used for grouping into cell_groups
+
+    cell_kind  get_cell_kind() const  {
+        return cell_kind::data_spike_source;
+    }
+
+    /// Get  all spikes until tfinal.
+
```
w-klijn: Done

src/dss_cell.hpp
```diff
@@ -0,0 +1,82 @@
...
+    // updates the internal time state to tfinal as a side effect
+    std::vector<time_type> spikes_until(time_type tfinal) {
```
w-klijn: Done

src/dss_cell_group.hpp
```diff
@@ -0,0 +1,96 @@
...
+        for (auto cell: cells_) {
+            cell.reset();
+        }
+    }
+
+    void set_binning_policy(binning_kind policy, time_type bin_interval) override
+    {} // Nothing to do?
+
+    void advance(time_type tfinal, time_type dt) override {
+        // TODO: Move source information to dss_cell implementation
```
w-klijn: I forgot to put a ? behind it:
Is this something that we want, should the cell implementation know that it has a gid, or is this a property of the group?

src/dss_cell.hpp
bcumming: I don't know if they need to be in separate files for now, however they should be distinct classes. In practice, the description is used in the front end, and directly manipulated/built by the user or application driving the library.

Here the front end description is part of the back end implementation. For a `dss_cell` we will probably have just the one back end implementation, however for many cell types we could have multiple solvers and back end combinations, and you would be forced to separate the definitions. For consistency, they should always be separate.

miniapp/io.cpp
```diff
@@ -389,6 +410,53 @@ std::ostream& operator<<(std::ostream& o, const cl_options& options) {
...
+
+        time_type t;
+        s >> t >> std::ws;
+
+        if (!s || s.peek() != EOF) {
+            throw std::runtime_error( util::strprintf(
+                    "Unable to parse spike file on line %d: \"%s\"\n",
+                    times.size(), line));
+
+            std::cout << t << std::endl;
```
bcumming: This looks like some debug code that wasn't removed.

miniapp/io.cpp
```diff
@@ -389,6 +410,53 @@ std::ostream& operator<<(std::ostream& o, const cl_options& options) {
...
+}
+
+/// Parse spike times from a file supplied in path
+/// A single spike per line, trailing white space is ignored
+/// Throws a usage error when opening file or parsing fails
+///
+/// Returns a vector of time_type
+
+std::vector<time_type> get_parsed_spike_times_from_path(nest::mc::util::path path) {
+    // Open the spike file
```
bcumming: this comment is redundant.

miniapp/io.hpp
```diff
@@ -1,12 +1,15 @@
...
 
-#include <string>
 #include <cstdint>
 #include <iosfwd>
 #include <stdexcept>
+#include <string>
 #include <utility>
+#include <vector>
 
+#include <common_types.hpp>
```
bcumming: Remove the blank line and sort headers in alphabetical order.

src/dss_cell.hpp
bcumming: We can leave it like this for now, and open a ticket to refactor all these at once.

src/dss_cell_group.hpp
bcumming: Do we need the `dss_cell` at all? We are unlikely to have different "back ends" for this, so it could be simpler to put all of the logic in the `cell_group`.

src/dss_cell.hpp
w-klijn: With the refactor asked later in the PR I had to move it outside. For this specific cell the discussion is moot.

src/dss_cell_group.hpp
w-klijn: Refactored all logic into the cell_group

miniapp/io.cpp
w-klijn: Check

miniapp/io.hpp
w-klijn: We group standard and XMC headers always in two groups with a blank line.
I sorted but leave the blank line to confirm to the rest of the code base

PR #324, opened 2017-07-12
Brunel Network miniapp

kabicm:
Miniapp for simulating the Brunel network of LIF neurons. The network consists of 3 populations: excitatory, inhibitory and Poisson population. Each neuron from the network receives a fixed number (proportional to the size of the population) of incoming connections from each of the three populations. 

The parameters that can be passed as command-line arguments are the following:

**-n** (--n_excitatory): total number of cells in the excitatory population. 

**-m** (--n_inhibitory): total number of cells in the inhibitory population.

**-e** (--n_external): total number of cells in the Poisson (external) population.

**-p** (--in_degree_prop): the proportions of connections from each of the 3 populations that each neurons receives.

**-w** (--weight): weight of all excitatory connections.

**-g** (--rel_inh_w): relative strength of inhibitory synapses with respect to the excitatory ones.

**-d** (--delay): the delay of all connections.

**-r** (--rate): rate of Poisson cells [kHz]

**-t** (--tfinal): length of the simulation period [ms]

**-s** (--delta_t): simulation time step [ms] (this parameter is ignored)

**-G** (--group-size): number of cells per cell group

**-f** (--spike-file-output): save spikes to file [Bool]

**-z** (--profile-only-zero): only output profile information for rank 0

**-v** (--verbose): present more verbose information to stdout

For example, we could run the command like:
**./brunel_miniapp.exe -n 400 -m 100 -e 400 -p 0.05 -w 1.2 -d 0.1 -g 1.0 -r 1 -t 100 -s 1 -G 10 -f**

========================================================================

The output of the simulation can be visualised using the provided python script **plot_spikes_brunel.py**. This script expects 3 command line arguments: 

**-n** (--n_exc): total number of cells in the excitatory population. 

**-m** (--n_inh): total number of cells in the inhibitory population.

**-e** (--n_ext): total number of cells in the Poisson (external) population.

which should match the values of the miniapp simulation. For example, if the miniapp was run as in  the previous example, then this script should be invoked as:

**python3 plot_spikes_brunel.py -n 400 -m 100 -e 400**
 
This script will then produce the following plot:
[brunel_spikes.pdf](https://github.com/eth-cscs/nestmc-proto/files/1141480/brunel_spikes.pdf)
where only a half of Poisson spikes are plotted for the sake of clarity.


miniapp/brunel/brunel_miniapp.cpp
```diff
@@ -0,0 +1,326 @@
...
+ All the connections have the same delay. The strenght of excitatory and poisson connections is given by
+ parameter weight, whereas the strength of inhibitory connections is rel_inh_strength * weight.
+ Poisson neurons all spike independently with mean frequency poiss_rate [kHz].
+ Because of the refractory period, the activity is mostly driven by poisson neurons and
+ recurrent connections have just a small effect.
+ */
+class brunel_recipe: public recipe {
+public:
+    brunel_recipe(cell_size_type nexc, cell_size_type ninh, double in_degree_prop, float weight, float delay, float rel_inh_strength, double poiss_rate):
+    ncells_exc_(nexc), ncells_inh_(ninh), delay_(delay), rate_(poiss_rate)
```
bcumming: Indent this line.

miniapp/brunel/brunel_miniapp.cpp
```diff
@@ -0,0 +1,326 @@
...
+ chosen randomly, in_degree_prop * ninh inhibitory connections and in_degree_prop * nexc Poisson connections.
+ All the connections have the same delay. The strenght of excitatory and poisson connections is given by
+ parameter weight, whereas the strength of inhibitory connections is rel_inh_strength * weight.
+ Poisson neurons all spike independently with mean frequency poiss_rate [kHz].
+ Because of the refractory period, the activity is mostly driven by poisson neurons and
+ recurrent connections have just a small effect.
+ */
+class brunel_recipe: public recipe {
+public:
+    brunel_recipe(cell_size_type nexc, cell_size_type ninh, double in_degree_prop, float weight, float delay, float rel_inh_strength, double poiss_rate):
```
bcumming: Split long lines like this over multiple lines.

miniapp/brunel/brunel_miniapp.cpp
```diff
@@ -0,0 +1,326 @@
...
+    {
+        // Make sure that in_degree_prop in the interval (0, 1]
+        if (in_degree_prop <= 0.0 || in_degree_prop > 1.0) {
+            std::out_of_range("The proportion of incoming connections should be in the interval (0, 1].");
+        }
+        
+        // Set up the parameters.
+        ncells_ext_ = nexc;
+        weight_exc_ = weight;
+        weight_inh_ = -1.0 * rel_inh_strength * weight_exc_;
```
bcumming: why not just `weight_inh_ = -rel_inh_strength * weight_exc_`

miniapp/brunel/brunel_miniapp.cpp
```diff
@@ -0,0 +1,326 @@
...
+        weight_ext_ = weight;
+        in_degree_exc_ = (int) (in_degree_prop * nexc);
+        in_degree_inh_ = (int) (in_degree_prop * ninh);
+        in_degree_ext_ = (int) (in_degree_prop * ncells_ext_);
```
bcumming: casting to `int` will always round-down.
Would `std::round()` be more appropriate here?
Should we check that the sum of `in_degree_{exc,inh,ext}_ == in_degree_prop * num_cells()`?

miniapp/brunel/brunel_miniapp.cpp
```diff
@@ -0,0 +1,326 @@
...
+    
+    try {
+        nest::mc::util::meter_manager meters;
+        meters.start();
+        
+        std::cout << util::mask_stream(global_policy::id()==0);
+        // read parameters
+        io::cl_options options = io::read_options(argc, argv, global_policy::id()==0);
+        
+        /*
```
bcumming: Delete the commented code below, or mark it with a TODO.

miniapp/brunel/brunel_miniapp.cpp
```diff
@@ -0,0 +1,326 @@
...
+        cell_size_type group_size = options.group_size;
+        
+        brunel_recipe recipe(nexc, ninh, in_degree_prop, w, d, rel_inh_strength, poiss_rate);
+
+        auto register_exporter = [] (const io::cl_options& options) {
+            return
+            util::make_unique<file_export_type>(
+                                                options.file_name, options.output_path,
+                                                options.file_extension, options.over_write);
+        };
```
bcumming: You can left indent the statement above:
```
util::make_unique<file_export_type>
    (options.file_name, options.output_path,options.file_extension, options.over_write);
```
or use one line less
```
util::make_unique<file_export_type>(options.file_name, options.output_path,
                                    options.file_extension, options.over_write);
```

miniapp/brunel/plot_spikes_brunel.py
```diff
@@ -0,0 +1,62 @@
...
+    file = open(file_name, 'r')
+    
+    for line in file:
+        ids += [float(line.split()[0])]
+        times += [float(line.split()[1])]
+    
+    file.close()
+
+    return (ids, times)
+
```
bcumming: How about making `nexc` and `ninh` command line parameters, so that this can be used to plot the output from arbitrary brunel simulations?

src/model.cpp
```diff
@@ -105,8 +105,9 @@ time_type model::run(time_type tfinal, time_type dt) {
                 group->enqueue_events(current_events()[i]);
                 PL();
 
+                 PE("cells");
```
bcumming: Indentation on this `PE` and the corresponding `PL` below is a bit funny.

miniapp/brunel/brunel_miniapp.cpp
kabicm: I am not sure this is really something we want for the following reason. 

Let us assume that we have only 101 excitatory and 109 inhibitory neuron and no poisson (just hypothetically).  Then, we would get that:

0.05 * 101 rounded to 5 (and we probably want it to be rounded like this)
0.05 * 109 rounded to 5 (and we probably want it to be rounded like this)
which makes in total 10.

However, on the other side 0.05 * (101 + 109) would be rounded to 11. Thus, the program would complain in this case (since 10 != 11), even though I don't see anything wrong here.

bcumming: You have satisfied me that we don't need the additional check.

PR #325, opened 2017-07-17
Feature/node description

bcumming:
Refactor hardware querying code

  * create src/hardware path for hardware-querying code
  * create nest::mc::hw namespace for hardware-querying code
  * move memory, affinity and power code from util and threading code to
  the new diretory and namespace
  * add nest::mc::threading::num_threads function that attempts to set.
  the number of threads according to an environment variable, or by
  using the low level hardware querying if no environment variable is
  set.
  * add hardware-query for counting the number of available gpus

### DIFF
--- a/src/hardware/affinity.cpp
+++ b/src/hardware/affinity.cpp
@@ -52,7 +52,7 @@ std::vector<int> get_affinity() {
 }
 #endif
 
-util::optional<std::size_t> num_available_cores() {
+util::optional<std::size_t> num_cores() {
     auto cores = get_affinity();
     if (cores.size()==0u) {
         return util::nothing;
--- a/src/hardware/affinity.cpp
+++ b/src/hardware/affinity.cpp
@@ -1,6 +1,7 @@
+#include <cstdlib>
 #include <vector>
 
-#include <cstdlib>
+#include <util/optional.hpp>
 
 #ifdef __linux__
 
@@ -16,7 +17,7 @@
 
 namespace nest {
 namespace mc {
-namespace threading {
+namespace hw {
 
 #ifdef __linux__
 std::vector<int> get_affinity() {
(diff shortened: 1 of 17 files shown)


src/hardware/affinity.cpp
```diff
@@ -51,10 +52,14 @@ std::vector<int> get_affinity() {
 }
 #endif
 
-unsigned count_available_cores() {
-    return get_affinity().size();
+util::optional<std::size_t> num_available_cores() {
```
halfflat: Just on naming: namespace `hw` is great, but why not also `src/hw/...` for the code? It's still clear and unambiguous IMO, and has the namespace and directory name line up, which is convenient.

`num_available_cores()`: why not just `num_cores()` or `count_cores()`? The latter might be clearer in that it indicates that it is an active process rather than simply looking something up.

src/hardware/gpu.cpp
```diff
@@ -0,0 +1,23 @@
+#ifdef NMC_HAVE_GPU
+    #include <cuda_runtime.h>
+#endif
+
+namespace nest {
+namespace mc {
+namespace hw {
+
+#ifdef NMC_HAVE_GPU
+unsigned num_available_gpus() {
```
halfflat: Similar to above with naming; `num_available_gpus()` is kinda verbose; we're already under `hw::` and having 'available' in the name doesn't really add anything.

src/hardware/power.hpp
```diff
@@ -4,14 +4,14 @@
 
 namespace nest {
 namespace mc {
-namespace util {
+namespace hw {
 
 // Energy in Joules (J)
 using energy_size_type = std::uint64_t;
 
 // Returns negative value if unable to read energy
```
halfflat: Just noticed comment (I know this is older code): 'negative value' doesn't make sense with `uint64_t` return type.

src/hardware/affinity.cpp
bcumming: I chose `hw` for the namespace because keystrokes.
I chose `hardware` for the path name because it is very clear and unambiguous.

Considerthe name of `num_available_X` shortened.

src/hardware/power.hpp
bcumming: Comment has been updated to say that it returns `energy_size_type(-1)`.

PR #326, opened 2017-07-18
refactor domain decomposition for arbitrary gid distribution.

bcumming:
changes to domain_decomposition:
  * domain_decomposition performs two pass load balancing in constructor
    * first pass performs global load balance
    * second pass distributes cells locally between cpu and gpu cell_groups.
    * current logic for this is very simple and naive.
    * will be replaced with a load_balancer, which returns a lighter
      domain_decomposition description in a follow up pull request.
  * provides a simple group_description type that contains gid, cell_kind and 
    target backend information for cell_group_factory.

changes to communicator
  * constructor takes a domain_decomposition and recipe
  * interface for adding connections and constructing connection table removed
    * constructor performs this task.
    * construction is more complicated, because connections are partitioned
      by source gid which requires multiple passes over connection information
      in the recipe.
  * make_event_queues updated
    * spikes and connections are now partitioned by source domain.
    * optimisation dynamically chooses to iterate over either connection or
      spike list, which ever is shorter.
  * exchange method now sorts local_spikes before global gather to facilitate
    optimised spike/connection searching.

changes to miniapp
  * automatically uses gpu if available and compiled with gpu support.
  * banner prints out useful information about number of cores, gpus and ranks.
  * remove -g cell group size flag

changes to cell_group
  * cell_group interfaces take a list of gid instead of a range
  * updated internal cell_group logic to convert between gid and local id
    * use a vector for lid->gid
    * use hash table for gid->lid in cell_groups that need this lookup

changes to unit tests
  * tests for the domain decomposition
  * tests for the communicator that test ring and all2all networks


src/communication/communicator.hpp
```diff
@@ -45,31 +48,68 @@ class communicator {
...
+        connection_part_ = algorithms::make_index(ps_cnts);
+        auto offsets = connection_part_;
+        std::size_t pos = 0;
+        for (auto gid: gids) {
+            const auto lg = gid_props.get(gid)->local_group;
+            for (auto c: rec.connections_on(gid)) {
+                const auto i = offsets[ps_doms[pos]]++;
+                connections_[i] = {c.source, c.dest, c.weight, c.delay, lg};
+                ++pos;
+            }
```
halfflat: Rather than have both a range-based `gid` loop and a separate `pos` counter, it would be clearer just to have an indexed loop with `pos`, or store all the gid information in the one vector and iterate over that.

src/communication/communicator.hpp
```diff
@@ -45,31 +48,68 @@ class communicator {
...
-    /// builds the optimized data structure
-    /// must be called after all connections have been added
-    void construct() {
-        if (!std::is_sorted(connections_.begin(), connections_.end())) {
-            threading::sort(connections_);
+        // Construct the connections.
+        // The loop above gave the information required to construct in place
+        // the connections as partitioned by the domain of their source gid.
+        connections_.resize(n_cons);
+        connection_part_ = algorithms::make_index(ps_cnts);
```
halfflat: This is another occasion where using `util::partition_view` simplifies some of the code below. e.g.
```
    auto cp = util::make_partition(connection_part_, ps_cnts);
    // ...
    threading::parallel_for::apply(0, num_domains_,
        [&](cell_gid_type i) { util::sort(util::subrange_view(connections_, cp[i])); });
```
and in `make_event_queues`:
```
    auto cp = util::partition_view(connection_part_);
    for (auto dom: make_span(0, num_domains_)) {
        auto cons = subrange_view(connections_, cp[dom]);
        // ...
```

src/communication/communicator.hpp
```diff
@@ -45,31 +48,68 @@ class communicator {
...
+        auto offsets = connection_part_;
+        std::size_t pos = 0;
+        for (auto gid: gids) {
+            const auto lg = gid_props.get(gid)->local_group;
+            for (auto c: rec.connections_on(gid)) {
```
halfflat: We're querting the recipe twice per gid for connections, which might be an expensive operation. Perhaps worth querying once and keeping the data until the end of the construction phase?

src/communication/communicator.hpp
```diff
@@ -101,49 +144,75 @@ class communicator {
...
+                    {return src<spk.source;}
+            };
+            if (cons.size()<spks.size()) {
+                auto sp = spks.begin();
+                auto cn = cons.begin();
+                while (cn!=cons.end() && sp!=spks.end()) {
+                    auto sources =
+                        std::equal_range(sp, spks.end(), cn->source(), spike_pred());
+
+                    for (auto s: make_range(sources.first, sources.second)) {
```
halfflat: In parallel with `make_span` and `subrange_view`, it's probably worth making a convenience `make_range` wrapper that takes a pair.

src/communication/communicator.hpp
```diff
@@ -101,49 +144,75 @@ class communicator {
...
+                bool operator()(const spike& spk, const cell_member_type& src)
+                    {return spk.source<src;}
+                bool operator()(const cell_member_type& src, const spike& spk)
+                    {return src<spk.source;}
+            };
+            if (cons.size()<spks.size()) {
+                auto sp = spks.begin();
+                auto cn = cons.begin();
+                while (cn!=cons.end() && sp!=spks.end()) {
+                    auto sources =
```
halfflat: This'd still be less than 100 chars wide if left on one line...

src/event_queue.hpp
```diff
@@ -27,6 +27,14 @@ struct postsynaptic_spike_event {
...
     time_type time;
     float weight;
+
+    friend bool operator==(postsynaptic_spike_event l, postsynaptic_spike_event r) {
+        return l.target==r.target && l.time==r.time && l.weight==r.weight;
+    }
+
+    friend bool operator<(postsynaptic_spike_event l, postsynaptic_spike_event r) {
+        return l.time<r.time;
+    }
```
halfflat: Do we need these?

miniapp/io.cpp
```diff
@@ -280,7 +276,6 @@ cl_options read_options(int argc, char** argv, bool allow_write) {
         update_option(options.bin_regular, bin_regular_arg);
         update_option(options.all_to_all, all_to_all_arg);
         update_option(options.ring, ring_arg);
-        update_option(options.group_size, group_size_arg);
```
bcumming: Adding support for user control over group sizes, and other parameters, would require some interface for passing policies/hints to the load-balancer/domain decomposition. This isn't too complex, but for the sake of simplicity I didn't include that in this set of changes.
We can do this in a follow up task?

src/backends/event.hpp
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include <iostream>
```
bcumming: Removed.

src/hardware/node.hpp
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+namespace nest {
+namespace mc {
+namespace hw {
+
+// Information about the computational resources available on a compute node.
+// Currently a simple enumeration of the number of cpu cores and gpus, which
+// will become richer.
+struct node {
```
bcumming: I originally had `node_description` which I shortened when it was moved to the `hw` namespace. I will go with `node_info`.

miniapp/miniapp.cpp
```diff
@@ -67,7 +68,9 @@ int main(int argc, char** argv) {
             global_policy::set_sizes(options.dry_run_ranks, cells_per_rank);
         }
 
-        banner();
+        hw::node nd;
+        nd.num_cpu_cores = threading::num_threads();
```
bcumming: I agree more or less with this. Here is what I have done
* `node_info` had its smarts removed: it is just a container.
* added a `get_node_info()` function that will return a `node_info` with the number of cores and gpus using the `hw` namespace
* in the miniapp I set `nd.num_cpu_cores` to the number of threads used by the threading back end: `threading::num_threads()`. Because the threading back end currently chooses the number of threads using its own internal logic from parsing environment variables and the `hw::` interface, I leave it like this... in future this will change when users are responsible for initializing the threading backend.
* If there is more than one gpu available, use one gpu.

src/domain_decomposition.hpp
```diff
@@ -13,97 +16,102 @@
...
+    }
+
+    backend_kind backend() const {
+        return backend_;
+    }
+
+    const std::vector<cell_gid_type>& gids() const {
+        return gids_;
+    }
 };
```
bcumming: Agreed. An earlier version stored the gid information in a compact form (begin and end index pair) and generated the gid indexes on the fly when requested.
Now it is just a dumb wrapper.

src/domain_decomposition.hpp
```diff
@@ -13,97 +16,102 @@
...
-            }
-            group_starts_.push_back(cell_end_);
+        auto b = first_cell_on_domain(domain_id_);
+        auto e = first_cell_on_domain(domain_id_+1);
+
+        // Local load balance
+
+        std::unordered_map<kind_type, std::vector<cell_gid_type>> kind_lists;
+        for (auto gid: make_span(b, e)) {
+            kind_lists[rec.get_cell_kind(gid)].push_back(gid);
```
bcumming: The `partition_view` code really needs documentation. This is the main reason I don't use it as much as I should.

bcumming: I have made the change, but to be honest I don't think it makes the code any clearer. Using the array of divisions directly is quite clear, and I suspect that adding the additional abstraction of a `partition_view` is more confusing.

src/hardware/node.cpp
```diff
@@ -0,0 +1,26 @@
...
+    if (!avail || *avail==0u) {
+        num_cpu_cores = 1;
+    }
+    num_cpu_cores = *avail;
+}
+
+node::node(int c, int g):
+    num_cpu_cores(c), num_gpus(g)
+{}
+
```
bcumming: I changed it to a light weight mutable container: see previous comment.

src/communication/communicator.hpp
```diff
@@ -45,31 +48,68 @@ class communicator {
...
+        // Count the number of local connections (i.e. connections terminating on this domain)
+        //  -> n_cons: scalar
+        // Calculate and store domain id of the presynaptic cell on each local connection
+        //  -> ps_doms: array with one entry for every local connection
+        // Also the count of presynaptic sources from each domain
+        //  -> ps_cnts: array with one entry for each domain
+        cell_local_size_type n_cons = 0;
+        std::vector<unsigned> ps_doms;
+        std::vector<cell_size_type> ps_cnts(num_domains_);
+        for (auto gid: gids) {
```
bcumming: Good catch.

src/communication/communicator.hpp
bcumming: Again, for my tastes using a `partition_view` improves the code in terms of readability, and can actually be distracting because it introduces an additional abstraction.

src/communication/communicator.hpp
bcumming: it is a nested loop, so the `pos` is needed, I think.

src/communication/communicator.hpp
bcumming: just, only just, 4 more characters and it would stay!

src/domain_decomposition.hpp
```diff
@@ -113,42 +121,71 @@ class domain_decomposition {
...
+// gid. The properties are stored in a gid_property struct.
+class gid_prop_map {
+public:
+    struct prop {
+        prop(cell_gid_type gid, cell_gid_type group): gid(gid), local_group(group) {}
+        cell_gid_type gid;
+        cell_gid_type local_group;
+        friend bool operator<(prop l, prop r) {return l.gid<r.gid;};
+    };
+
```
bcumming: You are right, what was I thinking.
This type has been removed completely.
* I used a hash table in the model to map gid to local cell group index, which I have marked with a big TODO to remove when the sampling interface is revamped.
* I use a flat vector in the `communicator` constructor, as suggested in your comment on that part of the code suggested.

bcumming:
I have updated the PR to address comments. I agreed with and made changes for all comments bar one.

I wasn't convinced that using `partition_view` makes the code simpler. I made the requested changes in the `domain_decomposition` constructor, for little benefit and reverted them. You will have to twist my arm to use `partition_view`.

src/domain_decomposition.hpp
halfflat: `is_local_gid` would still have constant complexity: I'm not sure where the `log` originates in your estimate.

For a counter-claim regarding `partition_view` and friends:
* It avoids duplication of the indexing operations involving e.g. `upper_bound` or `lower_bound`, reducing the opportunity for bugs to creep in.
* The `make_partition` code is more flexible than the `make_index` routine, and better named.
* There already exists partition view-using code in the source, so a developer will need to be familiar with it regardless.

src/domain_decomposition.hpp
```diff
@@ -13,97 +16,91 @@
...
+        // listed before the others. This is a very primitive attempt at
+        // scheduling; the cell_groups that run on the GPU will be executed
+        // before other cell_groups, which is likely to be more efficient.
+        std::vector<cell_kind> kinds;
+        for (auto l: kind_lists) {
+            kinds.push_back(cell_kind(l.first));
         }
-        return gid_group_partition().index(gid);
-    }
+        std::partition(kinds.begin(), kinds.end(), has_gpu_backend);
```
halfflat: The comment here is welcome, but this ultimately is introducing an ugly dependency between thread group implementation details and the innards of the domain decomposition that will have to be addressed later on. When this is merged, would you add a ticket for resolving this?

src/event_queue.hpp
```diff
@@ -32,9 +32,11 @@ struct postsynaptic_spike_event {
         return l.target==r.target && l.time==r.time && l.weight==r.weight;
     }
 
+    /*
     friend bool operator<(postsynaptic_spike_event l, postsynaptic_spike_event r) {
         return l.time<r.time;
     }
+    */
```
halfflat: Probably best just to delete them if not used, rather than comment out.

src/hardware/node_info.hpp
```diff
@@ -7,14 +7,16 @@ namespace hw {
 // Information about the computational resources available on a compute node.
 // Currently a simple enumeration of the number of cpu cores and gpus, which
 // will become richer.
-struct node {
-    node();
-    node(int c, int g);
+struct node_info {
+    node_info() = default;
+    node_info(unsigned c, unsigned g);
```
halfflat: I don't think this constructor adds anything that we wouldn't get just by simple aggregate initialization, but if it is kept we could just have the ctor body inline here and save any concern about a potentially complicated implementation that would warrant its source be placed in a separate file. The explicit default constructor could also be removed if the two-parameter ctor is removed.

tests/validation/validate_ball_and_stick.hpp
```diff
@@ -50,7 +50,7 @@ void run_ncomp_convergence_test(
                 seg->set_compartments(ncomp);
             }
         }
-        hw::node nd(1, backend==backend_kind::gpu? 1: 0);
+        hw::node_info nd(1, backend==backend_kind::gpu? 1: 0);
```
halfflat: Just a matter of taste, I guess, but I don't really see `a==b? 1: 0` to be any clearer than just `a==b` in these contexts.

src/communication/communicator.hpp
```diff
@@ -45,31 +48,71 @@ class communicator {
...
-    void add_connection(connection con) {
-        EXPECTS(is_local_cell(con.destination().gid));
-        connections_.push_back(con);
-    }
-
-    /// returns true if the cell with gid is on the domain of the caller
-    bool is_local_cell(cell_gid_type gid) const {
-        return algorithms::in_interval(gid, cell_gid_partition_.bounds());
-    }
+    explicit communicator(const recipe& rec, const domain_decomposition& dom_dec) {
```
apeyser: So, the choices are: recipe, domain -> communicator, domain -> recipe -> communicator, recipe -> domain -> communicator. Any arguments for this approach?

src/communication/communicator.hpp
```diff
@@ -45,31 +48,71 @@ class communicator {
...
+            for (auto c: cell.conns) {
+                const auto i = offsets[src_domains[pos]]++;
+                connections_[i] = {c.source, c.dest, c.weight, c.delay, cell.local_group};
+                ++pos;
+            }
         }
+
+        // Sort the connections for each domain.
+        // This is num_domains_ independent sorts, so it can be parallelized trivially.
+        const auto& cp = connection_part_;
```
apeyser: We should look at all the task allocation parts, to see whether any of them block dependent work.

src/communication/communicator.hpp
```diff
@@ -101,49 +147,74 @@ class communicator {
...
+        using util::make_range;
+
+        auto queues = std::vector<event_queue>(num_local_groups_);
+        const auto& sp = global_spikes.partition();
+        const auto& cp = connection_part_;
+        for (auto dom: make_span(0, num_domains_)) {
+            auto cons = subrange_view(connections_, cp[dom], cp[dom+1]);
+            auto spks = subrange_view(global_spikes.values(), sp[dom], sp[dom+1]);
+
+            struct spike_pred {
```
apeyser: Ah, my idea was accepted!

src/domain_decomposition.hpp
```diff
@@ -13,97 +16,91 @@
 namespace nest {
 namespace mc {
 
-// Meta data used to guide the domain_decomposition in distributing
-// and grouping cells.
-struct group_rules {
-    cell_size_type target_group_size;
-    backend_policy policy;
+inline bool has_gpu_backend(cell_kind k) {
```
apeyser: Isn't this a repeat of earlier code on cell_kind:: ?

src/domain_decomposition.hpp
apeyser: What should be partitioned? It depends on whether it's a one off partition, or a repeated partition.

src/domain_decomposition.hpp
```diff
@@ -13,97 +16,91 @@
...
-        return gid_group_partition().index(gid);
-    }
+        std::partition(kinds.begin(), kinds.end(), has_gpu_backend);
 
-    /// Returns the gid of the first cell on the local domain.
-    cell_gid_type cell_begin() const {
-        return cell_begin_;
+        for (auto k: kinds) {
+            // put all cells into a single cell group on the gpu if possible
+            if (node_.num_gpus && has_gpu_backend(k)) {
```
apeyser: Can't this be generalized so we have one place to handle calls depending on backends? The code has many of these 'ifs' specifically for gpu's, so if we add any more backends, we'll have to hunt throughout the code again. Maybe another policy class?

src/dss_cell_group.hpp
```diff
@@ -56,25 +49,25 @@ class dss_cell_group : public cell_group {
     {}
 
     void advance(time_type tfinal, time_type dt) override {
-        for (auto cell_idx: util::make_span(0, not_emit_it_.size())) {
+        for (auto lid: util::make_span(0, not_emit_it_.size())) {
```
apeyser: Can't this just be a range that returns a reference into not_emit_it?

src/cell_group_factory.cpp
```diff
@@ -14,26 +16,29 @@ namespace mc {
...
+cell_group_ptr cell_group_factory(const recipe& rec, const group_description& group) {
+    // Make a list of all the cell descriptions to be forwarded
+    // to the appropriate cell_group constructor.
+    std::vector<util::unique_any> descriptions;
+    descriptions.reserve(group.gids.size());
+    for (auto gid: group.gids) {
+        descriptions.push_back(rec.get_cell_description(gid));
+    }
+
+    switch (group.kind) {
```
halfflat: If we're going to make good on our modularity promise when it comes to cell groups, we are going to have to find a solution for registering both new cell kinds and new cell group implementations for a cell kind at run- or link-time: the decompositor/scheduler will need to decide which implementation to allocate a cell group to, based on the implementation's estimated performance characteristics (and in the GPU case, hardware availability).

But while worth considering and planning for, I don't think it's something we're going to pursue in the initial version.

src/communication/communicator.hpp
halfflat: I believe the goal is still to have the description of the domain decomposition be a separate type from the object that actually performs the decomposition (certainly this is what I was advocating before, and I _think_ I've convinced @bcumming).

The `communicator` object will still need to get the source gid list per target and will need the gid per group info to make the event queues, so I think we're thereby constrained to an interface which is at least equivalent to _recipe_ × _domain decomposition_ → _communicator_

src/communication/communicator.hpp
```diff
@@ -45,31 +48,71 @@ class communicator {
...
+        cell_local_size_type n_cons = 0;
+        std::vector<unsigned> src_domains;
+        std::vector<cell_size_type> src_counts(num_domains_);
+        for (auto i: make_span(0, num_local_groups_)) {
+            const auto& group = dom_dec.get_group(i);
+            for (auto gid: group.gids) {
+                gid_info info(gid, i, rec.connections_on(gid));
+                n_cons += info.conns.size();
+                for (auto con: info.conns) {
+                    const auto src = dom_dec.gid_domain(con.source.gid);
```
halfflat: We know below it has to be promotable to `size_t` (it's used as an index into `src_counts`); it may well be clearer just to use `std::size_t` for `src` here and be explicit.

src/domain_decomposition.hpp
halfflat: This is all going to have to be replaced (ultimately) with some sort of cost model or estimate that is flexible enough to encompass working-set sizes etc. etc.

Not to say there isn't a cleaner way of doing the current somewhat hacky approach, but we should recognize that it is at least a compromise.

src/dss_cell_group.hpp
halfflat: `lid` is used as an index into other vectors within the look body though. Functional solution would be to have a `zip`-like view…

That said, not convinced `lid` is really a good name for a loop index; elsewhere, we tend to use `lid` to refer to the local index or identifier of a type of thing within one cell. Sometimes `i` for a loop index is all we need.

src/domain_decomposition.hpp
bcumming: Sorry, you are right about the constant complexity.

I implemented both version, and I found the version that uses `partition_view` harder to undestand.

src/event_queue.hpp
bcumming: Of course. I must have commented it out as a hack to check if it was used, then forgot to actually remove the code.

src/hardware/node_info.hpp
bcumming: I moved the constructor inline.

src/communication/communicator.hpp
```diff
@@ -208,8 +211,7 @@ class communicator {
 
 private:
     cell_size_type num_local_groups_;
-    cell_size_type num_domains_;
-    std::vector<connection> connections_;
+    cell_size_type num_domains_; std::vector<connection> connections_;
```
bcumming: fixed.

tests/validation/validate_ball_and_stick.hpp
bcumming: To me it is clearer, because I am explicitly stating that the result of the expression is an integer, whereas `a==b` relies on the implicit conversion of `bool` to unsigned.

miniapp/miniapp.cpp
```diff
@@ -67,7 +69,12 @@ int main(int argc, char** argv) {
             global_policy::set_sizes(options.dry_run_ranks, cells_per_rank);
         }
 
-        banner();
+        // Use a node description that uses the number of threads used by the
+        // threading back end, and 1 gpu if available.
+        hw::node_info nd;
+        nd.num_cpu_cores = threading::num_threads();
+        nd.num_gpus = hw::num_gpus()? 1: 0;
```
bcumming: A `node_description` is a simple description of resources available on a node. This is currently a  count of each resource, currently cpu cores and gpus.
In this context, it is used to tell the `domain_decomposition` how to partition cell groups. We can only support a single GPU per MPI rank at the moment, so we use `min(1, num_gpus);`.

I have changed the statement to
```
nd.num_gpus = hw::num_gpus()>0? 1: 0;
```
make this a little bit clearer.

src/backends.hpp
```diff
@@ -6,16 +6,16 @@
...
-    prefer_gpu          //  use gpu back end when supported by cell_group type
+enum class backend_kind {
+    multicore,      //  use multicore backend for all computation
+    gpu          //  use gpu back end when supported by cell_group type
 };
 
-inline std::string to_string(backend_policy p) {
-    if (p==backend_policy::use_multicore) {
-        return "use_multicore";
+inline std::string to_string(backend_kind p) {
```
bcumming: I changed it to a switch statement, which will create a compiler warning if a case is not set.
```
inline std::string to_string(backend_kind p) {
    switch (p) {
        case backend_kind::multicore:
            return "multicore";
        case backend_kind::gpu:
            return "gpu";
    }   
}
```

src/cell_group_factory.cpp
bcumming: Performance of the factory was not considered when writing this code. It is only used when setting up the model, and the cost of making the decisions inside the factory are very small relative to the construction costs of the cell groups themselves.

As @halfflat points out, this is more like a placeholder, waiting for support for more more flexible user-defined` cell_group`s.

Note that by the time the `cell_group_factory` is invoked, the decision about which hardware back end the `cell_group` will run on will have been made by the load balancing/domain decomposition methods.

src/communication/communicator.hpp
bcumming: What @halfflat said.

src/communication/communicator.hpp
bcumming: It is necessary because `std::equal_range` requires that the predicate can "convert" between the input types `T1` and `T2`.

src/communication/communicator.hpp
bcumming: I think that we are safe here. The communicator is constructed before the cell_groups. I think that it should be possible to construct the communicator at the same time that the cell_groups are being constructed, because the work is independent (they both take the recipe and domain_decomposition as read only input parameters).

bcumming: However, I don't think that this is something we should try to do! A strong case would have to be made to justify the additional complexity that such an approach would entail.

src/domain_decomposition.hpp
bcumming: Here we have a helper that is used by the load balancing code to mark a cell_group for creation on either the host or device, if available.

The earlier `cell_kind::` code used the result of this decision, and generated the concrete instantion of the cell_group for the chosen back end.

This is the load balancing code that would make more nuanced decisions about how to partition work over cpu and gpu according to performance models. However for now, we simply place all work that can run on the gpu on the gpu if a gpu is available.

src/domain_decomposition.hpp
bcumming: You bring up a good point.

This is the only point where the decision about which back end to use for cell groups is made. It uses the simple method:

1. if a gpu is available and if the cell_kind has a gpu implementation, create a single cell_group with all cells of cell_kind to run on on the gpu
2. else create many cell groups, with 1 cell per group to run on the cpu.

This is not great. For one, as @halfflat points out, it does not consider the relative benefits of placing the work on cpu or gpu, group sizes, and so on.

This patch is intended as the smallest possible change that puts these decisions "in the right place", even if the actual decision making logic is too simple.

There will be a follow up patch that will split the load balancing and domain decomposition descriptions. Then we can focus on the load balancer that will make more general decisions based on cost models etc.

src/dss_cell_group.hpp
bcumming: I changed the name `cell_idx` because that sounded more like a `gid`, i.e. cell index, as opposed to and index into a cell local set.

I will change it to `i`.

src/domain_decomposition.hpp
bcumming: I will create a ticket, and add a TODO to the code.

src/domain_decomposition.hpp
```diff
@@ -13,97 +16,94 @@
...
+
+        // Global load balance
+
+        gid_divisions_.reserve(num_domains_+1);
+        for (auto d: make_span(0, num_domains_+1)) {
+            gid_divisions_.push_back(first_cell(d));
         }
-    }
+        gid_part_ = util::partition_view(gid_divisions_);
+
```
halfflat: The `+1` logic can be avoided with `make_partition`, e.g. (omitting `util::`)
```
    auto dom_size = [=](cell_gid_type dom) -> cell_gid_type {
        const cell_gid_type B = num_global_cells/num_domains_;
        const cell_gid_type R = num_global_cells-B*num_domains_;
        return B+(dom < R);
    };

    gid_part_ = make_partition(gid_divisions_, transform_view(make_span(0, num_domains_), dom_size));
```

src/domain_decomposition.hpp
```diff
@@ -113,42 +113,35 @@ class domain_decomposition {
...
-    backend_policy backend_policy_;
-    cell_gid_type cell_begin_;
-    cell_gid_type cell_end_;
+    int num_domains_;
+    int domain_id_;
+    hw::node_info node_;
     cell_size_type num_global_cells_;
-    std::vector<cell_size_type> group_starts_;
+    std::vector<cell_gid_type> gid_divisions_;
+    decltype(util::make_partition(gid_divisions_, gid_divisions_)) gid_part_;
```
halfflat: A more natural `decltype` here would be:
```
    decltype(util::partition_view(gid_divisions_)) gid_part_
```

bcumming: Indeed!

PR #327, opened 2017-07-26
add curly braces to test to silence gcc 7.1 warnings

bcumming:
There was a spurious warning with gcc 7.1.x in `test_algorithms` as a result of using a Google Test macro (`EXPECT_EQ`) inside an `if` block without surrounding `{}`. This fix adds the missing braces to silence the warnings.


halfflat:
Looks fine to me. (What was the warning, out of curiosity?)

PR #328, opened 2017-07-27
Basic Sphinx Documentation

bcumming:
This adds basic Sphinx documentation to the `doc` path.

- uses restructured text for documentation
- no Doxygen support yet: this is straightforward to integrate with the Sphinx docs, once we have some Doxygen in our source code

Uses the ReadTheDocs Sphinx theme:
https://github.com/rtfd/sphinx_rtd_theme
Which is included as a git submodule.

Added support for building the documentation using the CMake build system
- add `docs` target, which will build documentation in `build/docs/html` on `make docs`.
- add `cmake/FindSphinx.cmake` script for finding sphinx.

Tested with the ReadTheDocs infrastructure
- ReadTheDocs correctly checks out the repository and required rtd_theme sub module when master branch is updated
- This was tested on an experimental fork: this patch does not add automatic documentation building for the main repository: this will wait until we are closer to release.


halfflat:
Not super happy about the inclusion of the theme boilerplate and binary fonts, to be honest. Is there a way to avoid this? Perhaps using an approach similar to how we pull in gbench?

bcumming:
I agree completely about the boilerplate and fonts.

To generate the documentation locally I think something along the lines of ubench, whereby we check out the style and font information when `make docs` is run, will work.

How this will work with ReadTheDocs is unclear to me. I was planning to set up the hooks on ReadTheDocs so that our online documentation is automatically rebuilt whenever master is updated. ReadTheDocs will look in the `docs` path and run `conf.py`. We have to look at how to get it to access style and font information that isn't in the repository.

apeyser:
Shouldn't @w-klijn have a look at this, since he has some experience with Sphinx?

docs/_themes/sphinx_rtd_theme/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+"""
+import os
+
+__version__ = '0.2.5b1'
+__version_full__ = __version__
+
+
+def get_html_theme_path():
+    """Return list of HTML theme paths."""
+    cur_dir = os.path.abspath(os.path.dirname(os.path.dirname(__file__)))
```
bcumming: This is part of the rtd theme from github.

docs/conf.py
```diff
@@ -0,0 +1,169 @@
...
+project = 'Arbor'
+copyright = '2017, ETHZ & FZ Julich'
+author = 'ETHZ & FZ Julich'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = '1.0'
```
bcumming: Good point.

docs/install.rst
```diff
@@ -0,0 +1,58 @@
...
+
+.. _install_requirements:
+
+Before starting
+===============
+
+We will require:
+
+  * git
+  * cmake
```
bcumming: This documentation is a place holder.
It has to be fleshed out with such information.

bcumming:
It would be very good to get feedback from @w-klijn!

doc/conf.py
```diff
@@ -0,0 +1,166 @@
...
+# If your documentation needs a minimal Sphinx version, state it here.
+#
+# needs_sphinx = '1.0'
+
+# Add any Sphinx extension module names here, as strings. They can be
+# extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
+# ones.
+extensions = ['sphinx.ext.todo',
+    'sphinx.ext.mathjax',
+    'sphinx.ext.githubpages']
```
halfflat: The `githubpages` extension doesn't come with Sphinx before version 1.4, and this is sufficiently new that it's not available as a package on at least my current distro. I don't think we need it though: I've tried building the docs without this extension and it _looks_ okay to me. Could you confirm?

bcumming: Removed.

cmake/FindSphinx.cmake
```diff
@@ -0,0 +1,21 @@
...
+    SPHINX_EXECUTABLE
+    NAMES sphinx-build sphinx-build2
+    DOC "Path to sphinx-build executable")
+
+# this will also set SPHINX_FOUND to true if SPHINX_EXECUTABLE exists
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(
+    Sphinx
+    "Failed to locate sphinx-build executable"
+    SPHINX_EXECUTABLE)
```
bcumming: I updated the `ATTRIBUTIONS.md` file, and added a comment at the top of the file giving credit where it is due (thanks LLVM!).

PR #331, opened 2017-08-18
Better TBB CMake integration

### LINKED ISSUE #330: Fix TBB Build
Add support for building and linking against TBB to CMake.

The TBB integration has not been tested in a while, and is not working.

The most recent version of TBB has CMake scripts for easy integration of TBB into CMake projects, which is preferable our current in-house `FindTBB.cmake` script.

bcumming:
Better integration of the TBB threading back end into CMake.

Add support for the CMake package introduced in TBB 2017.7 to automatically download and build TBB when the TBB threading back end is selected.

With this patch, the user does not have to build+install TBB to use it in Arbor.

fixes #330


CMakeLists.txt
```diff
@@ -64,13 +64,31 @@ set(NMC_THREADING_MODEL "cthread" CACHE STRING "set the threading model, one of
 set_property(CACHE NMC_THREADING_MODEL PROPERTY STRINGS cthread tbb serial )
 
 if(NMC_THREADING_MODEL MATCHES "tbb")
+    # download and build TBB
+    include(cmake/TBBGet.cmake)
+    include(cmake/TBBBuild.cmake)
+    tbb_get(TBB_ROOT tbb_root SOURCE_CODE)
+    tbb_build(TBB_ROOT ${tbb_root} CONFIG_DIR TBB_DIR)
```
halfflat: I am concerned that this requires that the build environment has access to the wider internet, which historically is not necessarily the case in some hpc environments.

CMakeLists.txt
```diff
@@ -126,7 +144,10 @@ set_property(CACHE NMC_SYSTEM_TYPE PROPERTY STRINGS Generic Cray BGQ )
 
 # Cray specific flags
 if(${NMC_SYSTEM_TYPE} MATCHES "Cray")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -dynamic")
+    # we no longer set the -dynamic flag for the compilere here, instead dynamic
+    # linking should be enabled by the user by setting the environment variable:
+    #   CRAYPE_LINK_TYPE=dynamic
+    #set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -dynamic")
     add_definitions(-DNMC_HAVE_CRAY)
```
halfflat: While understanding that the TBB developers make this extremely awkward, I am again worried about the use of dynamic libraries in environments where the build file system is not available on the compute nodes.

tests/validation/CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ set(VALIDATION_CUDA_SOURCES
 
 if(NMC_VALIDATION_DATA_DIR)
     if ("${CMAKE_VERSION}" MATCHES "^3.[78].")
-        message(WARNING "CMake ${CMAKE_VERSION} has broken FindCUDA; omitting NMC_DATADIR define.")
+        # disable warning
+        #message(WARNING "CMake ${CMAKE_VERSION} has broken FindCUDA; omitting NMC_DATADIR define.")
     else()
```
halfflat: Why is this warning disabled? We still have the problem while we use this version of `FindCUDA`.

bcumming: Sorry, I disable that in nearly every branch, because CMake messages are a big pain if you use `ccmake`.

halfflat: We could move to a purely environment-variable based solution, but this would complicate running the validation tests (I think we'd need a wrapper script that CMake fills out that sets up the environment or we require the user supply the path on the validation test executable command line if the data directory is not found in a 'usual' relative location.)

halfflat:
lgtm, given caveats above.

PR #340, opened 2017-09-01
Basic CI support with TravisCI

bcumming:
Add support for continuous integration with Travis CI. 
This implements bare bones support that can be extended over time.

Travis CI test environments:
- All use gcc 5.
- Test the serial distributed back end with serial and cthread threading backends.
- Test mpi with cthread.
- The tbb test failed sporadically because CMake, so it is disabled for now.

The test script:
- Builds the unit tests, global_communication tests and miniapp.
- Asserts that all unit and global_communication tests pass.
- Asserts that the miniapp runs successfully.
  - does not test miniapp output for now.

There is plenty of scope for improving the tests.
A key improvement will be to use validated output for the validation and miniapp
to provide some validation.

There were some small fixes required to make the tests pass on Travis
- `communication/mpi.hpp` now sets default size and rank values of 1 and 0 respectively
  to allow all unit tests to pass when built with MPI.
- The wrappers around MPI API calls use `const_cast` to support MPI implementations that
  are not "const aware".
- A missing header was added to `tests/unit/test_range` to make `std::unordered_multimap`
  available.`


src/communication/mpi.hpp
```diff
@@ -106,9 +106,11 @@ namespace mpi {
         std::vector<char> buffer(displs.back());
 
         PE("MPI", "Gather");
-        MPI_Gatherv(str.data(), counts[rank()], traits::mpi_type(),                  // send
-                    buffer.data(), counts.data(), displs.data(), traits::mpi_type(), // receive
-                    root, MPI_COMM_WORLD);
+        MPI_Gatherv(
+            // const_cast required for MPI implementations that don't use const* in their interfaces
```
apeyser: Could also #define MPICH2_CONST (See CMake for for BGQ)

src/communication/mpi.cpp
```diff
@@ -8,8 +8,8 @@ namespace mpi {
 
 // global state
 namespace state {
```
apeyser: What do the initial values here mean?

src/communication/mpi.hpp
bcumming: This was to work around an ancient version of OpenMPI provided by Ubuntu/TravisCI.

src/communication/mpi.cpp
bcumming: They are sane defaults (1 mpi rank with rank 0), to make the unit tests pass when compiled with MPI.  The unit tests don't initialize MPI, so the default signed values of -1 were being converted to there unsigned version in the domain_decomposition unit test (the one in `tests.exe`, not the `global_communication.exe` test runner which initializes MPI).

This is not a permanent solution: @halfflat and I plan to use "executors", which will be more flexible than the current hard-coded compile time approach.

apeyser: Ah -- I see, if the init is never called, it pretends to be rank 0 of a world size 1.
Could we just comment that for the time being, so that if someone trips over it, they won't be confused? -1, -1 says to me that if init is never called, mark it with illegal values -- but the new version has unclear, far away implications.

apeyser: Ah -- I see, if the init is never called, it pretends to be rank 0 of a world size 1.
Could we just comment that for the time being, so that if someone trips over it, they won't be confused? -1, -1 says to me that if init is never called, mark it with illegal values -- but the new version has unclear, far away implications.

bcumming: I have added a comment, marking it as a TODO that should be removed when we add the run time executors.

PR #345, opened 2017-09-06
Stand alone CUDA compilation for threshold_watcher in gpu backend

bcumming:
Refactor the threshold_watcher and stack data structures in the gpu backend so that they are amenable to separable compilation. To wit:
- The `gpu::stack<T>` type had both `__host__` and `__device__` member functions, to allow dual use:
    - make `gpu::stack<T>` have a host-only interface that wraps a POD type `gpu::stack_base<T>`
    - implement a `push_back(stack_base, value)` method in `backends/gpu/kernels/stack.hpp` that is visible only to device code.
- The `test_thresholds` kernel was moved from a header file to a .cu file that is , and
    - template parameters were removed and the types provided by `backends/fvm_types.hpp` used
    - a simple c function interface, callable from host side code, is defined in `backends/gpu/threshold_common.hpp`
    - the c function is implemented in the cuda code, and is available at link time.
In this manner header files that define simple types and prototypes are used by both host
and CUDA translation units. Hopefully this will allow the use of compilers that don't support
more recent C++ standards for the low-level cuda, with C++14/17 support in the front end.

Also, some small fixes:
- simplify the `gpu::impl::padded_size` function (both to read and in terms of efficiency)
- use `typeid` as the default for pretty-printing types in the memory back end.
- update `test_gpu_stack` unit test to support new gpu stack interface.
- fix bug in `test_spikes` unit test, which was not running the GPU back end in the cuda unit tests.

### DIFF
--- a/src/backends/gpu/stack.hpp
+++ b/src/backends/gpu/stack.hpp
@@ -1,6 +1,9 @@
 #pragma once
 
+#include <algorithm>
+
 #include <memory/allocator.hpp>
+#include "stack_common.hpp"
 
 namespace nest {
 namespace mc {
@@ -22,91 +25,85 @@ namespace gpu {
 template <typename T>
 class stack {
     using value_type = T;
-    using allocator = memory::managed_allocator<value_type>;
+    template <typename U>
+    using allocator = memory::managed_allocator<U>;
 
-    // The number of items of type value_type that can be stored in the stack
-    unsigned capacity_;
-
-    // The number of items that have been stored
-    unsigned size_;
-
-    // Memory containing the value buffer
-    // Stored in managed memory to facilitate host-side access of values
-    // pushed from kernels on the device.
-    value_type* data_;
+    using base_type = stack_base<value_type>;
+    base_type* base_;
 
 public:
+    stack& operator==(const stack& other) = delete;
+    stack(const stack& other) = delete;
 
-    stack(unsigned capacity):
-        capacity_(capacity), size_(0u)
-    {
-        data_ = allocator().allocate(capacity_);
+    stack(stack&& other) {
+        std::swap(base_, other.base_);
     }
+
+    stack& operator=(stack&& other) {
+        std::swap(base_, other.base_);
+        return *this;
+    }
+
+    stack(unsigned capacity) {
+        base_ = allocator<base_type>().allocate(1);
+        base_->capacity = capacity;
+        base_->size = 0;
+        base_->data = allocator<value_type>().allocate(capacity);
+    }
+
+    stack(): stack(0) {}
 
     ~stack() {
-        allocator().deallocate(data_, capacity_);
-    }
-
-    // Append a new value to the stack.
-    // The value will only be appended if do_push is true.
-    __device__
-    void push_back(const value_type& value) {
-        // Atomically increment the size_ counter. The atomicAdd returns
-        // the value of size_ before the increment, which is the location
-        // at which this thread can store value.
-        unsigned position = atomicAdd(&size_, 1u);
-
-        // It is possible that size_>capacity_. In this case, only capacity_
-        // entries are stored, and additional values are lost. The size_
-        // will contain the total number of attempts to push,
-        if (position<capacity_) {
-            data_[position] = value;
+        if (base_) {
+            allocator<value_type>().deallocate(base_->data, base_->capacity);
+            allocator<base_type>().deallocate(base_, 1);
         }
     }
 
-    __host__
     void clear() {
-        size_ = 0;
+        base_->size = 0;
     }
 
     // The number of items that have been pushed back on the stack.
     // size may exceed capacity, which indicates that the caller attempted
     // to push back more values than there was space to store.
-    __host__ __device__
     unsigned size() const {
-        return size_;
+        return base_->size;
     }
 
     // The maximum number of items that can be stored in the stack.
-    __host__ __device__
     unsigned capacity() const {
-        return capacity_;
+        return base_->capacity;
+    }
+
+    base_type& base() {
+        return *base_;
     }
 
     value_type& operator[](unsigned i) {
-        EXPECTS(i<size_ && i<capacity_);
-        return data_[i];
+        EXPECTS(i<base_->size && i<base_->capacity);
+        return base_->data[i];
     }
 
     value_type& operator[](unsigned i) const {
-        EXPECTS(i<size_ && i<capacity_);
-        return data_[i];
+        EXPECTS(i<base_->size && i<base_->capacity);
+        return base_->data[i];
     }
 
     value_type* begin() {
-        return data_;
+        return base_->data;
     }
     const value_type* begin() const {
-        return data_;
+        return base_->data;
     }
 
     value_type* end() {
-        // Take care of the case where size_>capacity_.
-        return data_ + (size_>capacity_? capacity_: size_);
+        // Take care of the case where size>capacity.
+        return base_->data + std::min(base_->size, base_->capacity);
     }
     const value_type* end() const {
-        // Take care of the case where size_>capacity_.
-        return data_ + (size_>capacity_? capacity_: size_);
+        // Take care of the case where size>capacity.
+        return base_->data + std::min(base_->size, base_->capacity);
     }
 };
 
--- a/src/backends/gpu/stack.hpp
+++ b/src/backends/gpu/stack.hpp
@@ -34,7 +34,7 @@ class stack {
     storage_type* create_storage(unsigned n) {
         auto p = allocator<storage_type>().allocate(1);
         p->capacity = n;
-        p->size = 0;
+        p->stores = 0;
         p->data = allocator<value_type>().allocate(n);
         return p;
     }
@@ -62,23 +62,23 @@ public:
     }
 
     void clear() {
-        storage_->size = 0u;
+        storage_->stores = 0u;
     }
 
     // The number of items that have been pushed back on the stack.
     // This may exceed capacity, which indicates that the caller attempted
     // to push back more values than there was space to store.
     unsigned pushes() const {
-        return storage_->size;
+        return storage_->stores;
     }
 
     bool overflow() const {
-        return storage_->size>capacity();
+        return storage_->stores>capacity();
     }
 
     // The number of values stored in the stack.
     unsigned size() const {
-        return std::min(storage_->size, storage_->capacity);
+        return std::min(storage_->stores, storage_->capacity);
     }
 
     // The maximum number of items that can be stored in the stack.
--- a/src/backends/gpu/stack.hpp
+++ b/src/backends/gpu/stack.hpp
@@ -28,82 +28,92 @@ class stack {
     template <typename U>
     using allocator = memory::managed_allocator<U>;
 
-    using base_type = stack_base<value_type>;
-    base_type* base_;
+    using storage_type = stack_storage<value_type>;
+    storage_type* storage_;
+
+    storage_type* create_storage(unsigned n) {
+        auto p = allocator<storage_type>().allocate(1);
+        p->capacity = n;
+        p->size = 0;
+        p->data = allocator<value_type>().allocate(n);
+        return p;
+    }
 
 public:
-    stack& operator==(const stack& other) = delete;
+    stack& operator=(const stack& other) = delete;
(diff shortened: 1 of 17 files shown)


src/backends/gpu/stack.hpp
```diff
@@ -22,91 +25,85 @@ namespace gpu {
...
-        // at which this thread can store value.
-        unsigned position = atomicAdd(&size_, 1u);
-
-        // It is possible that size_>capacity_. In this case, only capacity_
-        // entries are stored, and additional values are lost. The size_
-        // will contain the total number of attempts to push,
-        if (position<capacity_) {
-            data_[position] = value;
+    stack(): stack(0) {}
+
```
halfflat: With appropriate checks in the accessors, a default ctor which just sets `base_=0` might be preferable, avoiding any allocations in this case. (Nice fix though for zero-sized `cudaManagedMalloc` case below!)

src/backends/gpu/threshold_watcher.hpp
```diff
@@ -89,26 +86,29 @@ class threshold_watcher {
...
 
         // Check that the number of spikes has not exceeded
         // the capacity of the stack.
-        EXPECTS(stack_->size() <= stack_->capacity());
+        // ATTENTION: requires cudaDeviceSynchronize to avoid simultaneous
+        // host-device managed memory access.
+#ifdef NMC_HAVE_ASSERTIONS
+        cudaDeviceSynchronize();
+        EXPECTS(stack_.size() <= stack_.capacity());
+#endif
```
halfflat: Could avoid the `#ifdef` guard with e.g.
```
EXPECTS((cudaDeviceSynchronize(), stack_.size() <= stack_.capacity()));
```

src/backends/gpu/threshold_watcher.hpp
```diff
@@ -51,15 +48,15 @@ class threshold_watcher {
         thresholds_(memory::make_const_view(thresh)),
         prev_values_(values),
         is_crossed_(size()),
-        stack_(make_managed_ptr<stack_type>(10*size()))
+        stack_(10*size())
```
halfflat: Not in scope, but something for later: demystify the magic number `10`?

src/backends/gpu/stack.hpp
```diff
@@ -22,91 +25,85 @@ namespace gpu {
...
-    unsigned size_;
+    template <typename U>
+    using allocator = memory::managed_allocator<U>;
 
-    // Memory containing the value buffer
-    // Stored in managed memory to facilitate host-side access of values
-    // pushed from kernels on the device.
-    value_type* data_;
+    using base_type = stack_base<value_type>;
+    base_type* base_;
```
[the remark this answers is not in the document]
bcumming: Good idea.

src/backends/gpu/stack.hpp
```diff
@@ -22,91 +25,85 @@ namespace gpu {
...
+    stack(stack&& other) {
+        std::swap(base_, other.base_);
     }
 
-    ~stack() {
-        allocator().deallocate(data_, capacity_);
+    stack& operator=(stack&& other) {
+        std::swap(base_, other.base_);
+        return *this;
+    }
```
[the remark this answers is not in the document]
apeyser: You'd have the dealloc at the end of the containing context, correct?

src/backends/gpu/stack.hpp
```diff
@@ -22,91 +25,85 @@ namespace gpu {
...
-    // The number of items that have been stored
-    unsigned size_;
+    template <typename U>
+    using allocator = memory::managed_allocator<U>;
 
-    // Memory containing the value buffer
-    // Stored in managed memory to facilitate host-side access of values
-    // pushed from kernels on the device.
-    value_type* data_;
+    using base_type = stack_base<value_type>;
```
apeyser: Isn't stack_base really stack_storage?

src/backends/gpu/stack_common.hpp
```diff
@@ -0,0 +1,36 @@
...
+    fvm_size_type index;    // index of variable
+    fvm_value_type time;    // time of crossing
+
+    friend bool operator==(threshold_crossing l, threshold_crossing r) {
+        return l.index==r.index && l.time==r.time;
+    }
+};
+
+template <typename T>
+struct stack_base {
```
apeyser: See my comment on stack_base -> stack_storage

src/backends/gpu/threshold_watcher.hpp
apeyser: Yes, but that's much less obvious.

src/backends/gpu/stack.hpp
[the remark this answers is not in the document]
halfflat: Could do it either way, given the lack of exception guarantees. I'd normally do something like:
```
    destroy(); // private method: dealloc our stuff if base_ != 0
    base_ = other.base_;
    other.base_ = nullptr;
```
and the only use `std::swap` if it saved code duplication.

src/backends/gpu/threshold_watcher.hpp
halfflat: Avoids the #ifdef warts though. Also, I think the comma operator is under-rated and under-utilized :)

src/backends/gpu/stack.hpp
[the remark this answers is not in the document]
bcumming: Usuing `std::swap` without thinking is a old habit of mine. I agree that it is for the best to change this code to follow your suggested patter. However I personally don't find deffering the deallocation until the end of the enclosing scope too surprising.

src/backends/gpu/stack.hpp
bcumming: True. You and @halfflat both noticed this. I went for the name `stack_storage`, and added a comment to explain this.

src/backends/gpu/threshold_watcher.hpp
bcumming: Oh @halfflat, if everyone started using it you would stop liking it ;).

Having to choose between `ifdef` warts or comma operator, I would choose the comma.

src/backends/gpu/threshold_watcher.hpp
bcumming: Yeah, that is a hack. It deserves a comment to that effect.

src/backends/gpu/stack.hpp
```diff
@@ -22,91 +25,85 @@ namespace gpu {
...
     }
 
     // The number of items that have been pushed back on the stack.
     // size may exceed capacity, which indicates that the caller attempted
     // to push back more values than there was space to store.
-    __host__ __device__
     unsigned size() const {
-        return size_;
+        return base_->size;
     }
```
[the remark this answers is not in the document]
bcumming: This type is intended for relatively small stacks: it would be performance suicide to use atomic CUDA operations to fill a stack that required something like `2^32` entries.

src/backends/gpu/stack.hpp
```diff
@@ -22,91 +25,85 @@ namespace gpu {
...
-        return data_ + (size_>capacity_? capacity_: size_);
+        // Take care of the case where size>capacity.
+        return base_->data + std::min(base_->size, base_->capacity);
     }
     const value_type* end() const {
-        // Take care of the case where size_>capacity_.
-        return data_ + (size_>capacity_? capacity_: size_);
+        // Take care of the case where size>capacity.
+        return base_->data + std::min(base_->size, base_->capacity);
     }
```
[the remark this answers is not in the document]
bcumming: How about `size()` returning `std::min(size, capacity)`? the `size` member could be renamed to something more descriptive like `pushes`.

src/backends/gpu/stack.hpp
bcumming: I prefer to have "always valid" state, instead of having two cases in accessors.
The one part of the interface that wouldn't work with that approach is the `base()` method, which has been renamed `storage()` in the latest fixes.

src/backends/gpu/stack.hpp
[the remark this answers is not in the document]
halfflat: I think this would be an improvement on the API. Do we care about the number of pushes other than to verify there hasn't been an overflow condition? If not, then I'd go with a boolean `overflow()` method instead, which would allow e.g. the wrap-safe test if that ever became a concern.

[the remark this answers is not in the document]
bcumming: I went ahead and did this (and implemented an `is_overflowed` method).

src/backends/gpu/stack.hpp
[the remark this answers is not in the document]
halfflat: Point taken! I think though a robust solution involving a flag would only have a single write overhead when every thread is attempting to push when the stack is already at capacity, e.g.
```
if (position<s.capacity) {
    s.data[position] = value;
}
else {
    s.overflow_flag = 1; // this flag is initialized to zero on construction, clear()
}
```
What do you think?

src/backends/gpu/stack.hpp
```diff
@@ -22,91 +25,85 @@ namespace gpu {
...
 public:
+    stack& operator==(const stack& other) = delete;
+    stack(const stack& other) = delete;
 
-    stack(unsigned capacity):
-        capacity_(capacity), size_(0u)
-    {
-        data_ = allocator().allocate(capacity_);
+    stack(stack&& other) {
+        std::swap(base_, other.base_);
```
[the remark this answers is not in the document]
bcumming: The irony here is that we are doing this so that we can use C++14, which gives us `std::exchange`.

src/backends/gpu/stack.hpp
halfflat: Having `base_` always be a valid pointer would also mean that the move constructor needs to make sure that the other `base_` pointer ends up pointing to a valid `stack_base` object. There's a touch of complexity either way, but the always-valid `base_` does involve more allocations.

bcumming: I have gone the route of always valid state. To satisfy the requirements that a moved-from object still be valid, I insist that stacks with capacity 0 and default constructed stacks both point to valid state with `storage.capacity=0`.
This might not be completely optimal, however it gives correct behavior as per the standard, and the type is used in just one location in the back end.

src/backends/gpu/stack.hpp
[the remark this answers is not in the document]
bcumming: The `push_back` uses an `atomicAdd()` on `s.size`, and tests the value returned from the atomic operation to decide whether to write. So `size` will get incremented past `capacity` regardless.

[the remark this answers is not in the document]
bcumming: Oh, I see what you are saying. I go back to my point that it is just unreasonable to use a stack for so many items!

bcumming:
I have updated the PR to address the review comments.
I went with always having valid state for the `stack::storage_` (formerly `stack::base_`), with the default constructor also allocating memory for the `storage_`.
As a result, the `std::swap` approach to implementing move semantics is valid, and will give expected heap allocation results for the inteded use case of allowing deferred initialization of stacks inside a constructor.

src/backends/gpu/stack.hpp
```diff
@@ -22,91 +25,95 @@ namespace gpu {
...
     // The number of items that have been pushed back on the stack.
-    // size may exceed capacity, which indicates that the caller attempted
+    // This may exceed capacity, which indicates that the caller attempted
     // to push back more values than there was space to store.
-    __host__ __device__
+    unsigned pushes() const {
+        return storage_->size;
+    }
+
+    bool is_overflowed() const {
```
halfflat: You know I hate this method name :)

src/backends/gpu/stack_common.hpp
```diff
@@ -0,0 +1,40 @@
...
+// gpu code.
+template <typename T>
+struct stack_storage {
+    using value_type = T;
+
+    // The number of items of type value_type that can be stored in the stack
+    unsigned capacity;
+
+    // The number of items that have been stored
+    unsigned size;
```
halfflat: Comment and member name don't quite align with semantics, being the number of attempted stores, or the offset where the store would be performed if within capacity. i.e. `size` is only the size if it is less than or equal to `capacity`.

How about, e.g.
```
    // Offset to next storage slot. Counts number of pushes, and equals size of stack if not greater than capacity.
    unsigned next;
```

PR #353, opened 2017-09-20
Seperable compilation of mechanism kernels on GPU

bcumming:
Seperable compilation of the CUDA kernels generated by modcc from NMODL files.

This is the biggest piece of the puzzle for seperable compilation.

Updates to the CMake workflow
* update the `build_modules()` helper function to cleanly handle calls to modcc
  that generate multiple output files.
* add a new library target `gpu_mechanisms` for the seperately compiled CUDA
  kernels and the implementation of their C wrappers.
* reduce verbosity of compilation messages.

Simplification of namespaces in C++ 
* Removed the `nest::mc::mechanisms` namespace
* All gpu mechanisms and their implementations moved from
  `nest::mc::mechanisms::gpu::mech_name` to `nest::mc::gpu`, and likewise
  for the multicore mechanisms.
* Types like `nest::mc::mechanisms::ion` moved to `nest::mc` namespace.
  - in the future they might want to move to `arb::fvm`.

Removed template parameters for `value_type` and `size_type` from all of the 
mechanism implementations.
* Uses `fvm_value_type` and `fvm_size_type` everywhere instead.

Refactored `CUDAPrinter` in modcc
* The `CUDAPrinter` keeps track of 3 text buffers, one each for 
  "implementation", "interface" and "implementation interface".
* CUDAPrinter generates three outputs for mechanism X
    * `X_impl.hpp` : Definition of `ParamPack` struct used to pass
      function arguments from front end to back CUDA kernels, and the 
      prototypes of the C wrappers around CUDA kernels.
    * `X_impl.cu`  : CUDA kernels and the implementation of their C wrappers.
    * `X.hpp`      : Defines the same interface as before, with calls out to
      the seperately compiled kernels.
* modcc driver had to be refactored slightly to handle multiple outputs from
  the same printer visitor.


bcumming:
I have another PR lined up behind this one, so I can address your nits in the next PR (promise I will).

PR #359, opened 2017-09-26
Bench events

### LINKED ISSUE #357: Document event-setup overheads on GPU [4]
Setting up the data structures that manage events during an integration interval is a bottleneck on the GPU implementation.

Benchmark and analysis of the bottleneck will be collected here.

bcumming:
Add a ubenches benchmark that illustrates scaling issues with event-setup on cell groups with many cells.

The benchmark also illustrates some alternative approaches that are significant optimizations over the current approach.

Fixes #357


.ycm_extra_conf.py
```diff
@@ -53,6 +53,8 @@
     '-I',
     'modcc',
     '-I',
+    'tests/ubench/google-benchmark/include',
+    '-I',
```
bcumming: This is to help YouCompleteMe vim plugin to find the google bench headers.

tests/ubench/event_setup.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include <benchmark/benchmark.h>
+
+using namespace nest::mc;
+
+std::vector<postsynaptic_spike_event> generate_inputs(size_t ncells, size_t ev_per_cell) {
+    std::vector<postsynaptic_spike_event> input_events;
+    std::default_random_engine engine;
+    std::uniform_int_distribution<cell_gid_type>(0u, ncells);
+
+    std::default_random_engine gen;
```
halfflat: Let's not use `std::default_random_engine`: it's implementation defined, and could be particularly poor on one hand, and give a different sequence on different platforms on the other. If quality is not a particular concern, I'd stick with `std::minstd_rand`.

tests/ubench/event_setup.cpp
```diff
@@ -0,0 +1,207 @@
...
+void single_queue(benchmark::State& state) {
+    using pev = postsynaptic_spike_event;
+
+    const std::size_t ncells = state.range(0);
+    const std::size_t ev_per_cell = state.range(1);
+
+    // state
+    std::vector<pev> input_events = generate_inputs(ncells, ev_per_cell);
+
+    //auto binner = event_binner(binning_kind::regular, 0.001);
```
halfflat: We should probably nuke the binner in this benchmark entirely — still not clear if we should even use it by default, given we now have much lower per event-time overhead of delivery.

tests/ubench/event_setup.cpp
```diff
@@ -0,0 +1,207 @@
...
+        staged_events.reserve(input_events.size());
+        size_t i=0;
+        for (auto& lane: event_lanes) {
+            while (auto e = lane.pop_if_before(1.f)) {
+                staged_events.push_back(*e);
+            }
+            part[++i] = staged_events.size();
+        }
+
+        benchmark::ClobberMemory();
```
halfflat: If the events in `input_events` were to have times greater than `1.f` — and the generated event set probably should also include events that are too far in the future to be delivered this round — then the `event_lane` queues should be cleared at the end of this loop, because they won't be empty.

tests/ubench/event_setup.cpp
```diff
@@ -0,0 +1,207 @@
...
+        }
+
+        // remove events that were delivered from the event lanes
+        auto i=0u;
+        for (auto& lane: event_lanes) {
+            lane.erase(lane.begin(), ext[i++]);
+        }
+
+        benchmark::ClobberMemory();
+    }
```
halfflat: For consistency across runs, if the generated event set is extended to include events after time `1.f`, then the lanes will grow in each iteration of the benchmark loop; they should be cleared within the loop.

tests/ubench/event_setup.cpp
```diff
@@ -0,0 +1,207 @@
...
+    using pev = postsynaptic_spike_event;
+    const std::size_t ncells = state.range(0);
+    const std::size_t ev_per_cell = state.range(1);
+
+    auto input_events = generate_inputs(ncells, ev_per_cell);
+
+    // state
+    std::vector<std::vector<pev>> event_lanes(ncells);
+    std::vector<size_t> part(ncells+1);
+    std::vector<decltype(input_events.begin())> ext(ncells);
```
halfflat: Logic is fine, but I reckon it would simplify the code and the reading thereof if `ext` just held a length/offset rather than the iterators.

tests/ubench/event_setup.cpp
bcumming: Done.

tests/ubench/event_setup.cpp
bcumming: using `std::mt19937`

tests/ubench/CMakeLists.txt
```diff
@@ -63,7 +64,8 @@ foreach(bench_src ${bench_sources})
     add_executable("${bench_exe}" EXCLUDE_FROM_ALL "${bench_src}")
     add_dependencies("${bench_exe}" gbench)
     target_include_directories("${bench_exe}" PRIVATE "${gbench_install_dir}/include")
-    target_link_libraries("${bench_exe}" "${gbench_install_dir}/lib/libbenchmark.a")
+    target_link_libraries("${bench_exe}" LINK_PUBLIC "${gbench_install_dir}/lib/libbenchmark.a")
+    target_link_libraries("${bench_exe}" LINK_PUBLIC ${NESTMC_LIBRARIES})
```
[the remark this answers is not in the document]
bcumming: If the binner dependency is removed, this is no longer needed. Consider it nuked.

tests/ubench/event_setup.cpp
bcumming: I made the change. Made some bits simpler, others more verbose.

tests/ubench/event_setup.cpp
bcumming: I add an explicit call to `.clear()` for the queues/lanes in each benchmark.
This didn't impact the vector implementation, but strangely enough this had a small but siginificant increase for the queue based approaches (despite them being based on a `std::vector` under the hood). Strange.

PR #369, opened 2017-10-24
Optimise host-side event wrangling [8]

### LINKED ISSUE #368: Optimise host-side event wrangling with GPU back end [8]
The process that starts with generating events from spikes, then packing them into optimised data structures for processing on GPU or MC back ends, is a serial bottleneck for models with large cell groups.

This was illustrated in a micro benchmark:
https://github.com/eth-cscs/arbor/tree/master/tests/ubench#event_setup

This task should take the lessons from the microbenchmark, and improve performance with large cell groups using two approaches
  * optimization of data structures
  * parallelisation where possible.

bcumming:
This PR optimises the host side generation of post-synaptic spike events from
spikes, and their conversion into deliverable events in optimized data
structures.

The first change is to have the communicator build one list of post-synaptic
spike events for each cell on the local domain, instead of one list for 
each cell group. Binning events from a large cell_group level list into per cell
lists was one of the most time consuming parts of the original approach.
This optimization reduces complexity in two ways:
* time-sorting the events delivered to a cell group is faster, because many small lists are sorted, instead of one large list.
* the step of stable sort or binning by cell index is not required.

This step required minor changes to the `communicator`:
* the gid of cells on the local domain are given a local index, in the range
  [0, num_local_cells);
* the cell_group index in the connection table is replaced with local index;
* a member `domain_id_partition_` stores the partition of gid over cell
  groups in this domain;
* `communicator::make_event_queues` builds and returns a list of events for
  each local cell;
* a `communicator::group_queue_range(i)` function returns the range of local
  cells in cell_group `i`. The calling code can use this to partition the
  list of queues;

Changes in time management
* model keeps track of the current integration "epoch". Typically an
  epoch is an interval of min_delay/2, described by an unsigned value
  starting at zero for the first epoch.
* tfinal and epoch index passed to both the `advance` and `enqueue_events`
  methods of cell_groups.

The main optimization is in the `mc_cell_group::enqueue_events`
* the `mc_cell_group` maintains two vectors of events for each cell, both
  of which are sorted in delivery time of events.
* The first contains events that are to be delivered in the current epoch,
  and potentially events that are to be delivered in a future epoch. This
  is used by the `mc_cell_group::advance()` method
* The second is used by the `mc_cell_group::enqueue_events()` method to
  build a complete list of all events to be delivered after the current
  epoch.
* because events are passed in already binned into per-cell lists, the
  operation of merging the incoming events with events that are already in
  the queue for each cell can be parallelised across cells inside the
  cell_group, which is a key optimization with few cell_groups with many
  cells per group.
* Another key optimization is from using short sorted vectors, which are
  much more efficient than the old approach of pushing to and popping from
  one large priority queue.
* The event_binner data structure is also simplified, because it only
  has to support a single cell, instead of a set of cells with arbitrary
  index, because events for a given gid are processed sequentially.

Small fixes and additions:
* added `event_time_less` helper functor to `generic_event.hpp` for
  comparing generic event times with scalar times;
* added `subrange_view_type` metafunction to `rangutil.hpp`, to simplify
  expressing the type returned by `subrange_view`.

Fixes #368


src/cell_group.hpp
```diff
@@ -21,8 +23,25 @@ class cell_group {
...
+    //    vector, in same order that input gid were provided to the cell_group
+    //    constructor.
+    // tfinal:
+    //    The final time for the current integration epoch. This may be used
+    //    by the cell_group implementation to omptimise event queue wrangling.
+    // epoch:
+    //    The current integration epoch. Events in events are due for delivery
+    //    in epoch+1 and later.
+    virtual void enqueue_events(
+            util::subrange_view_type<std::vector<std::vector<postsynaptic_spike_event>>> events,
```
apeyser: The subrange_view_type must then give us a reference into the `post_synaptic_spike_event list[][]`

apeyser: So the subrange must include all (but may have more) events than events that occur from the beginning of `epoch` until it's end at `tfinal` that must be delivered in `epoch+1` (and|or?) later, correct?

src/communication/communicator.hpp
```diff
@@ -56,11 +56,11 @@ class communicator {
         // For caching information about each cell
         struct gid_info {
             using connection_list = decltype(std::declval<recipe>().connections_on(0));
-            cell_gid_type gid;
-            cell_gid_type local_group;
-            connection_list conns;
-            gid_info(cell_gid_type g, cell_gid_type lg, connection_list c):
-                gid(g), local_group(lg), conns(std::move(c)) {}
+            cell_gid_type gid;       // global identifier of cell
+            cell_gid_type domain_id; // index of cell in this domain
```
apeyser: domain_id is the id inside the domain, not the id of the domain?

src/communication/communicator.hpp
```diff
@@ -75,22 +75,26 @@ class communicator {
         gid_infos.reserve(dom_dec.num_local_cells);
 
         cell_local_size_type n_cons = 0;
+        cell_local_size_type n_gid = 0;
         std::vector<unsigned> src_domains;
         std::vector<cell_size_type> src_counts(num_domains_);
-        for (auto i: make_span(0, num_local_groups_)) {
-            const auto& group = dom_dec.groups[i];
+        for (auto g: make_span(0, num_local_groups_)) {
```
apeyser: `auto&&` but I guess you know the underlying type for make_space

src/communication/communicator.hpp
```diff
@@ -75,22 +75,26 @@ class communicator {
         gid_infos.reserve(dom_dec.num_local_cells);
 
         cell_local_size_type n_cons = 0;
+        cell_local_size_type n_gid = 0;
```
apeyser: Maybe a short end of line comment on what n_gid is

src/communication/communicator.hpp
```diff
@@ -101,11 +105,17 @@ class communicator {
...
             for (auto c: cell.conns) {
                 const auto i = offsets[src_domains[pos]]++;
-                connections_[i] = {c.source, c.dest, c.weight, c.delay, cell.local_group};
+                connections_[i] = {c.source, c.dest, c.weight, c.delay, cell.domain_id};
                 ++pos;
             }
         }
 
+        // Build cell partition by group for passing events to cell groups
+        domain_id_partition_.resize(num_local_groups_+1);
```
apeyser: Is this initialized to `domain_id_partiton_[0] = 0`

src/cell_group.hpp
bcumming: Yes. The events being passed in by this interface are for delivery in any epoch *after* `epoch`, not just the `epoch+1`.

src/communication/communicator.hpp
bcumming: Yes. Not the best name.

src/communication/communicator.hpp
bcumming: `auto` is fine here. It is going to be an integer type.

src/communication/communicator.hpp
bcumming: Yes. See http://en.cppreference.com/w/cpp/container/vector/resize
Specifically: "If the current size is less than count additional default-inserted elements are appended".

src/cell_group.hpp
bcumming: The challenge here is that both the `advance` and `enqueue_events` methods are aware of which epoch they are in, and when that epoch ends. However, we do not have information about when any future epochs will end, so we can only "bin" events into two categories
1. will be delivered in this epch
2. will be deilvered later

bcumming: You will see in the `mc_cell_group` implementation, we maintain two lists:
1. *current events*: contains all events that must be delivered in the current epoch, and potentially events to be delivered in a future epoch.
2. "future events": events to be delivered in a future epoch.

Events to be deilvered in a future epoch can exist in both queues.

src/backends/gpu/multi_event_stream.hpp
```diff
@@ -58,18 +59,20 @@ class multi_event_stream_base {
         n_nonempty_stream_(1)
     {}
 
+    // The list of events must be sorted in ascending order of index,
+    // with events of equal index further time-sorted.
```
halfflat: Can't we just say 'sorted first by index and then by time', as used below? Or is this phrasing not common?

src/backends/gpu/multi_event_stream.hpp
```diff
@@ -58,18 +59,20 @@ class multi_event_stream_base {
...
     template <typename Event>
-    void init(std::vector<Event>& staged) {
+    void init(const std::vector<Event>& staged) {
         using ::arb::event_time;
```
halfflat: Did you have a measure of how much time is spent in the `multi_event_stream::init()` method? Given that the cell group has the partition-by-index information already, there is the possibility of reducing the work this call does. If the time spent here is inconsequential, then of course it is not important.

src/backends/gpu/multi_event_stream.hpp
```diff
@@ -131,7 +135,7 @@ class multi_event_stream: public multi_event_stream_base {
 
     // Initialize event streams from a vector of events, sorted first by index
     // and then by time.
-    void init(std::vector<Event> staged) {
+    void init(const std::vector<Event>& staged) {
         multi_event_stream_base::init(staged); // reorders `staged` in place.
```
halfflat: Comment no longer applicable.

src/mc_cell_group.hpp
```diff
@@ -72,32 +79,41 @@ class mc_cell_group: public cell_group {
...
+            for (auto& lane: lanes) {
+                lane.clear();
+            }
+        }
     }
 
     void set_binning_policy(binning_kind policy, time_type bin_interval) override {
-        binner_ = event_binner(policy, bin_interval);
+        binners_.clear();
+        binners_.resize(gids_.size(), event_binner(policy, bin_interval));
```
halfflat: While the binner memory overhead is going to be similar (smaller but same order) in the `binning_kind::following` case, we're making many more objects in the `binning_kind::none` and `binning_kind::regular` cases. Is the change to make a separate binner for each cell a performance optimization for the `following` case?

src/profiling/meter_manager.cpp
```diff
@@ -141,29 +148,38 @@ nlohmann::json to_json(const meter_report& report) {
...
         if (m.name=="time") {
-            o << strprintf("%16s", "time (s)");
+            o << strprintf("%16s", "time(s)");
         }
         else if (m.name.find("memory")!=std::string::npos) {
-            o << strprintf("%16s", m.name+" (MB)");
+            o << strprintf("%16s", m.name+"(MB)");
+        }
+        else if (m.name.find("energy")!=std::string::npos) {
+            o << strprintf("%16s", m.name+"(kJ)");
```
halfflat: This is getting brittle and obscure: perhaps now is the time to store unit and scale information in the meter data...

src/backends/gpu/multi_event_stream.hpp
bcumming: I had that information, it was non-trivial, but the main bottleneck there was the `stable_sort`. Once that was no longer needed, it becomes very small.

I agree that things could be further optimized, but it isn't really needed at the moment.

src/communication/communicator.hpp
bcumming: Changed to `index_on_domain`. Verbose, but clearer.

src/communication/communicator.hpp
```diff
@@ -101,11 +105,17 @@ class communicator {
...
+        for (auto g: make_span(0, num_local_groups_)) {
+            domain_id_partition_[g+1] = domain_id_partition_[g] + dom_dec.groups[g].gids.size();
+        }
+
```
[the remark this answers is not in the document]
bcumming: Done.

src/mc_cell_group.hpp
bcumming: You are right. I missed this detail when refactoring.
The memory overheads aren't much, with one small object per cell. Given that this type is used only for multi-compartment cells, that is relatively little.

src/mc_cell_group.hpp
```diff
@@ -210,9 +227,52 @@ class mc_cell_group: public cell_group {
...
+            // STEP 1: sort events in place in events[l]
+            util::sort_by(events[l], [](const pse& e){return event_time(e);});
+            PL();
+
+            PE("merge");
+            // STEP 2: clear lf to store merged list
+            lf[l].clear();
+
+            // STEP 3: merge new events and future events from lc into lf
+            auto pos = std::lower_bound(lc[l].begin(), lc[l].end(), tfinal, event_time_less());
```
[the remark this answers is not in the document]
bcumming: Quite often we have to write functors of this type to work with events and spikes. I find it distracting to have to keep creating them when needed.

I prefer to stick with a `std::vector` than go with a specialised data structure. Or are do you mean a specialised data structure for both queues and their associated methods?

src/profiling/meter_manager.cpp
bcumming: That is a PR for another day...

src/connection.hpp
```diff
@@ -15,20 +15,20 @@ class connection {
...
+        domain_index_(didx)
     {}
 
     float weight() const { return weight_; }
     time_type delay() const { return delay_; }
 
     cell_member_type source() const { return source_; }
     cell_member_type destination() const { return destination_; }
-    cell_gid_type group_index() const { return group_index_; }
+    cell_gid_type domain_index() const { return domain_index_; }
```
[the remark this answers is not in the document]
bcumming: Done and done. It isn't really clear why have the distinction. `cell_gid_type` is an index, while `cell_size_type` is explicitly an unsigned type that talks about the size of a collection. In this context, we are assigning an index to cells, so you could argue that `cell_gid_type` is appropriate.

src/mc_cell_group.hpp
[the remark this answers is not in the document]
halfflat: Regarding the comparator, I have no objection to the generic code, just thinking that it could live in the same file as the code that uses it, at least until it becomes useful elsewhere.

I was thinking of the data structure that wraps the per-epoch per-cell event collections; if we move the machinery up to the model level as we were discussing earlier, than that would side-step the issue.

src/mc_cell_group.hpp
halfflat: If there's no significant performance gain from having one per cell, then we could just revert to the original binner, no?

src/cell_group.hpp
```diff
@@ -21,8 +23,25 @@ class cell_group {
...
+            time_type tfinal,
+            std::size_t epoch) = 0;
```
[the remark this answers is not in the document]
bcumming: I made such a type, and put it in its own header.
It makes some things clearer, and makes a little bit more typing in other spots.

src/mc_cell_group.hpp
bcumming: There is a performance penalty to using the old approach with the `following` case: a hash table lookup on every single event.

src/communication/communicator.hpp
```diff
@@ -57,10 +57,10 @@ class communicator {
         struct gid_info {
             using connection_list = decltype(std::declval<recipe>().connections_on(0));
             cell_gid_type gid;       // global identifier of cell
-            cell_gid_type domain_id; // index of cell in this domain
+            cell_gid_type index_on_domain; // index of cell in this domain
```
halfflat: Would still prefer this to be a `cell_size_type` — if we can keep `cell_gid_type` for cell global ids, then we have the benefit of type-driven documentation, which I really, really like.

src/communication/communicator.hpp
```diff
@@ -128,7 +128,7 @@ class communicator {
     /// The range of event queues that belong to cells in group i.
     std::pair<cell_size_type, cell_size_type> group_queue_range(cell_size_type i) {
         EXPECTS(i<num_local_groups_);
-        return {domain_id_partition_[i], domain_id_partition_[i+1]};
+        return index_part_[i];
```
halfflat: Or even just expose (directly or through a method) the partition view itself. Just a matter of taste at this point though.

src/connection.hpp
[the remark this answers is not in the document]
halfflat: Just to document in-person conversation: if we can keep `cell_gid_type` just for GIDs, there is a benefit to code documentation; if we see something of this type, then we can presume there is a potential or concrete cell with that gid, and that that value is not about to change for that cell, etc. etc.

src/communication/communicator.hpp
bcumming: Done.

PR #377, opened 2017-11-01
Add mechanism parameter setting/new implementation.

### LINKED ISSUE #350: Mechanism parameter setting from recipes. [4]
Depends on #349.

* Extend `cell` description for `cable1d_neuron` cells to allow parameter descriptions compatible with the work described in #349.
* Set mechanism values in `fvm_multicell` from recipe accordingly.
* Bonus: extend `cable1d_neuron` probes to allow polling of mechanism state variables.

halfflat:
Fixes #350 

* Replace parameter_list with mechanism_spec.
* Add prototype for mechanism parameter schema checking.
* Allow mechanism weights to be set after construction.
* Combine range parameters on density mechanisms by linear contribution in CVs.
* Cable segment electrical parameters are now member variables.
* Publish mechanism parameter information through new method `mechanism::field_info`; note this will be replaced/improved in upcoming dynamic mechanism catalog work.
* Access mechanism parameter scalars and range data via `mechanism::field_view_ptr` and `mechanism::field_value_ptr` methods.
* Allow mechanism 'global' parameters to be set via a method of specializing mechanisms (and giving them corresponding aliases).
* Extend recipe interface to allow querying of per-cell-kind global information for use by cell group implementations.
* Add unit tests for above - note that linear density mechanism parameter test is tightly coupled with the FVM discretization scheme.


src/recipe.hpp
```diff
@@ -62,6 +62,8 @@ class recipe {
 
     virtual std::vector<cell_connection> connections_on(cell_gid_type) const = 0;
     virtual probe_info get_probe(cell_member_type probe_id) const = 0;
+
+    virtual util::any get_global_properties(cell_kind) const { return util::any{}; };
```
bcumming: Because this returns an `any`, a comment explaining what is expected inside the `any` is needed.

src/segment.hpp
```diff
@@ -87,79 +80,44 @@ class segment {
...
-    {
-        auto it = std::find_if(
-            mechanisms_.begin(), mechanisms_.end(),
-            [&n](parameter_list const& l){return l.name() == n;}
-        );
-        if(it==mechanisms_.end()) {
-            throw std::out_of_range(
-                "attempt to access a parameter that is not defined in a segment"
-            );
+    void add_mechanism(mechanism_spec mech) {
```
bcumming: Has the expected behavior changed when adding a mechanism with the same name as a mechanism that is already present? i.e. from throwing an error to replacing the existing mechanism?

src/util/simple_table.hpp
```diff
@@ -0,0 +1,44 @@
+#pragma once
```
bcumming: Is this needed? Can't the user use a `map` or `unordered_map`? This is used in one location, so I don't know if it is a good idea to make extra library code to support it.

tests/common_cells.hpp
```diff
@@ -279,10 +277,10 @@ inline cell make_cell_simple_cable(bool with_stim = true) {
     I /= f;
 
     for (auto& seg: c.segments()) {
-        seg->add_mechanism(pas_parameters());
-        seg->mechanism("membrane").set("r_L", r_L);
-        seg->mechanism("membrane").set("c_m", c_m);
-        // seg->mechanism("pas").set("g", gbar);
+        seg->add_mechanism(mechanism_spec("pas"));
+        // seg->add_mechanism(mechanism_spec("pas").set("gbar", gbar)
```
bcumming: remove commented code?

tests/unit/test_fvm_multi.cpp
```diff
@@ -265,21 +266,21 @@ TEST(fvm_multi, mechanism_indexes)
...
-    segs[2]->add_mechanism(hh_parameters());
-    segs[3]->add_mechanism(pas_parameters());
+    segs[1]->add_mechanism("pas");
+    segs[2]->add_mechanism("hh");
+    segs[3]->add_mechanism("pas");
 
     for (auto& seg: segs) {
         if (seg->is_dendrite()) {
-            seg->mechanism("membrane").set("r_L", 100);
+            seg->rL = 100;
```
bcumming: The membrane-specific `rL` and `cm` are now set differently to other parameters.
They are not associated with a specific mechanism, but this approach is less "consistent" with the method for setting mechanism parameters.

tests/unit/test_fvm_multi.cpp
```diff
@@ -328,6 +329,337 @@ TEST(fvm_multi, mechanism_indexes)
     }
 }
 
+namespace {
+    double wm_impl(double wa, double xa) {
```
bcumming: what is `wm`?

tests/unit/test_fvm_multi.cpp
```diff
@@ -328,6 +329,337 @@ TEST(fvm_multi, mechanism_indexes)
...
+
+    template <typename... R>
+    double wmean(double w, double x, R... rest) {
+        return wm_impl(w, w*x, rest...);
+    }
+}
+
+// Test area-weighted linear combination of density mechanism parameters.
+
+TEST(fvm_multi, density_weights) {
```
bcumming: Nice test, but you are right, it is a bit fragile.
I suppose it is the level you have to test at for this sort of thing.

modcc/blocks.hpp
```diff
@@ -33,14 +33,31 @@ struct Id {
                        // string == no value
     unit_tokens units;
 
+    std::pair<std::string, std::string> range; // empty component => no range set
```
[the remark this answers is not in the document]
halfflat: Sure!

modcc/cprinter.cpp
```diff
@@ -200,6 +207,16 @@ std::string CPrinter::emit_source() {
     text_.add_line("}");
     text_.add_line();
 
+    // Override `set_weights` method only for density mechanisms.
+    if (module_->kind() == moduleKind::density) {
+        text_.add_line("void set_weights(array&& weights) override {");
```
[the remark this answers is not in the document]
halfflat: I was just following the convention in place for the constructor, but to be honest, I think there shouldn't be any rvalue parameter at all here. Can we postpone that until the mechanism catalogue/prototype stuff coming up next?

src/fvm_multicell.hpp
```diff
@@ -420,6 +421,52 @@ class fvm_multicell {
...
+        }
+
+        size_type mech_id = mechanisms_.size();
+        auto m = backend::make_mechanism(impl_name, mech_id, cv_to_cell_, time_, time_to_, dt_comp_, voltage_, current_, {}, node_indices);
+        if (impl_name!=name) {
+            m->set_alias(name);
+        }
+
+        for (const auto& pv: global_params) {
+            auto field = m->field_value_ptr(pv.first.c_str());
```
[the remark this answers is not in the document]
halfflat: The original choice was due to the use of C strings in the lookup tables and in the code that did the calling of these methods. This is a legacy of that choice, but there's no strong reason not to have a `std::string` interface instead or in addition.

src/fvm_multicell.hpp
```diff
@@ -420,6 +421,52 @@ class fvm_multicell {
...
+            }
+            m.get()->*field = pv.second;
+        }
+
+        mechanisms_.push_back(std::move(m));
+        return *mechanisms_.back();
+    }
+
+    // Throwing-wrapper around mechanism (range) parameter look up.
+    static view& mech_field(mechanism& m, const std::string& param_name) {
```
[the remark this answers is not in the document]
halfflat: Should indeed just be a `view`.

src/recipe.hpp
halfflat: Like the cell descriptions, it will change from cell kind to cell kind. But I can add a comment to that effect.

src/segment.hpp
halfflat: Yes, it has.

src/util/simple_table.hpp
halfflat: It's really low overhead and also very simple. To be honest, I find myself wanting to use something like this very often in code generally, and there may well be uses of `map` in `arbor` that are only there because a simple table lookup did not already exist.

tests/common_cells.hpp
halfflat: Thanks, will remove.

tests/unit/test_fvm_multi.cpp
halfflat: It's true, but I believe that's as it should be: they mean different things and play a different role in the model description, and ultimately, are used in different parts of the solver.

tests/unit/test_fvm_multi.cpp
halfflat: Weighted mean; when it moved above the test body, it lost its context. Will add comment.

modcc/cprinter.cpp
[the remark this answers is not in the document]
bcumming: I am happy to wait.

Issue #386, opened 2017-11-08
Setting temperature in model


bcumming:
The temperature is currently hard coded at 6.3 degrees celcius (279.45 Kelvin).

When the mechanism interface and parameter setting work described in has been finished in #365 and #349, this can be tackled fairly easily.

halfflat:
This is now implemented as a field in `mc_cell_global_properties` for the purpose of Nernst calculations. It's _not_ propagated through the NMODL magic `celsius` parameter yet.

halfflat:
Started an approach to have `celsius` be magic in NMODL, and link it up with a special scalar external data source, but it's pretty invasive (and doesn't work yet).

Simpler solution is to have special-case handling for a global parameter called `celsius` in `fvm_lowered_impl::reset()`. Less invasive, but also much hackier — it treads all over the existing mechanism specialization semantics for global scalar parameters.

halfflat:
Without deeply messing with the way procedures are handled (it's particularly a problem for SIMD printing), the 'clean' way looks like:
* A parameter `celsius` is recognized as an externally-backed indexed variable (but not a range variable unless we want to add that functionality later).
* No longer being a true parameter, using `celsius` in an NMODL procedure requires passing it in as an argument.

It involves making further restrictions on the NMODL we accept, but the alternatives to this are either very hacky or very invasive. This is currently the approach I will take to a PR.

PR #387, opened 2017-11-08
Use Nernst equation to calculate reversal potentials

### LINKED ISSUE #376: Calcluate ion channel reversal potentials using Nernst equation. [3]
Steps:
1. add `valency` field to ion channels
2. add temperature information to a model
3. add interface to `ion_channel` type for calculating reversal potential given temperature, valency and concentrations
4. implement CUDA and MC backend implementations
5. update time stepping loop to perform update

bcumming:
Replace constant values for ion species reversal potentials taken from HH model with values that are updated on each time step via the Nernst equation.

* implement Nernst equation in multicore and gpu back ends
* extend interface of `ion` type to proved a method `update_reversal_potential` which calls the back end nernst routine.
* add `valency` and default concentration values to the ion s[ecies.
* add interface for "resetting" and ion species (for restarting simulations).

Fixes #376


src/backends/gpu/kernels/nernst.cu
```diff
@@ -0,0 +1,36 @@
...
+namespace kernels {
+    template <typename T>
+    __global__ void nernst(std::size_t n, int valency, T temperature, const T* Xo, const T* Xi, T* eX) {
+        auto i = threadIdx.x+blockIdx.x*blockDim.x;
+
+        constexpr T RF = 1e3*8.3144598/96485.33289; // factor 1e3 to scale from V -> mV
+        T factor = RF*temperature/valency;
+        if (i<n) {
+            eX[i] = factor*std::log(Xo[i]/Xi[i]);
+        }
```
halfflat: As we saw in the NEURON code, there the computed potential is clamped (for when one of the concentrations is zero, or (improperly) negative). Should we also?

src/backends/gpu/kernels/nernst.cu
```diff
@@ -0,0 +1,36 @@
...
+            fvm_value_type temperature,
+            const fvm_value_type* Xo,
+            const fvm_value_type* Xi,
+            fvm_value_type* eX)
+{
+    constexpr int block_dim = 128;
+    const int grid_dim = impl::block_count(n, block_dim);
+    kernels::nernst<<<grid_dim, block_dim>>>
+        (n, valency, temperature, Xo, Xi, eX);
+}
```
halfflat: As a micro-optimization, I suspect it would be better to have the CPU compute the `RF*temperature/valency` factor and pass it in as a parameter to the kernel when it is enqueued. Thoughts?

src/ion.hpp
```diff
@@ -71,19 +74,50 @@ public :
...
+
+    void set_default_internal_concentration(value_type c) {
+        EXPECTS(c>=value_type(0));
+        Xi_default_ = c;
+    }
+    void set_default_external_concentration(value_type c) {
+        EXPECTS(c>=value_type(0));
+        Xo_default_ = c;
+    }
+
```
halfflat: Why not just expose the member variables for valency and default concentrations (after a suitable renaming)? This would be my preference as it makes the code smaller and the effect transparent.

src/ion.hpp
```diff
@@ -71,19 +74,50 @@ public :
...
 
+    void reset() {
+        memory::fill(iX_, 0);
+        memory::fill(Xi_, Xi_default_);
+        memory::fill(Xo_, Xo_default_);
+    }
+
+    /// Calculate the reversal potential for all compartments using Nernst equation
+    /// temperature is in degrees Kelvin
+    void update_reversal_potential(value_type temperature) {
```
halfflat: I think the member function name should indicate that it updates the reversal potential according to the Nernst equation.

src/fvm_multicell.hpp
```diff
@@ -1075,13 +1087,25 @@ void fvm_multicell<Backend>::reset() {
     set_time_global(0);
     set_time_to_global(0);
 
+    // Update ion species:
+    //   - clear currents
+    //   - reset concentrations to defaults
+    //   - recalculate reversal potentials
```
halfflat: `ion::reset()` as defined below doesn't actually recalculate the reversal potentials (stale comment?)

src/fvm_multicell.hpp
```diff
@@ -1075,13 +1087,25 @@ void fvm_multicell<Backend>::reset() {
...
-        // TODO : the parameters have to be set before the nrn_init
-        // for now use a dummy value of dt.
         m->set_params();
         m->nrn_init();
     }
 
+    // Updeate reversal potential to account for changes to concentrations made
+    // by calls to nrn_init() in mechansisms.
+    for (auto& i: ions_) {
+        i.second.update_reversal_potential(6.3+273.15); // TODO: use temperature specfied in model
```
halfflat: I'd pop this TODO into a ticket, or extend the `arb::cell_global_properties` (in `cell.hpp`) to include the temperature field and assign this from that.

src/fvm_multicell.hpp
```diff
@@ -1075,13 +1087,25 @@ void fvm_multicell<Backend>::reset() {
...
+    }
+
     for (auto& m : mechanisms_) {
-        // TODO : the parameters have to be set before the nrn_init
-        // for now use a dummy value of dt.
         m->set_params();
         m->nrn_init();
     }
 
+    // Updeate reversal potential to account for changes to concentrations made
```
halfflat: typo

src/fvm_multicell.hpp
bcumming: Already done: #386

src/backends/gpu/kernels/nernst.cu
bcumming: how about just clamping the concentrations, e.g. 
```
constexpr value_type min_conc = 1e-6;
log(max(Xo[i], min_conc)/max(Xi[i], min_conc))
```

src/backends/gpu/kernels/nernst.cu
bcumming: It might work as a micro-optimization, but I doubt that this will be a bottleneck. If it is a problem, then we could make the optimization.

Otherwise I find it clearer to have the Nernst equation spelt out directly inside the kernel. Less room for mistakes that way.

src/fvm_multicell.hpp
bcumming: `ion::reset()` should calculate the reversal potentials, because they depend on the concentration values that are reset. I have fixed its implementation to do this.

src/ion.hpp
bcumming: done

src/backends/gpu/kernels/nernst.cu
bcumming: From private discussion, we have decided to leave it as is, and see how/when the assumptions of the Nernst equation break down, before proposing solutions.

src/ion.hpp
bcumming: done.

src/ion.hpp
```diff
@@ -74,16 +77,35 @@ public :
...
     }
 
+    void reset() {
+        // The Nernst equation uses the assumption of nonzero concentrations:
+        EXPECTS(default_internal_concentration> value_type(0));
+        EXPECTS(default_external_concentration> value_type(0));
+        memory::fill(iX_, 0);
+        memory::fill(Xi_, default_internal_concentration);
+        memory::fill(Xo_, default_external_concentration);
+        nernst_reversal_potential(6.3+273.15); // TODO: use temperature specfied in model
```
halfflat: The 6.3 °C temperature now appears as an explicit number in a few places. Until there is a settable temperature somewhere, this at least should sit in a constant somewhere (e.g. `fvm_multicell`). `reset()` can then take the temperature as a parameter.

I'm also a bit worried that there is no way to explicitly set the reversal potential independent of any concentrations, making it hard to obtain the same behaviour as e.g. NEURON's hh mechanism. If this is regarded as out-of-scope, I'll make a ticket with a proposed API.

validation/ref/neuron/nrn_validation.py
```diff
@@ -129,6 +129,17 @@ def add_soma(self, diam, **kw):
         soma.gl_hh = p['gl_hh']
         soma.el_hh = p['el_hh']
 
+        # Use the following exact values calcualted using
+        # the Nernst equation, because NEURON does not use correct values
+        # for R and F.
+        soma.ena =  63.55148117386
```
halfflat: Comment should include from what concentration values these were derived (i.e. NEURON's default values in `membdef.h`) so that they can be verifiable — otherwise this is just a magic number.

validation/ref/numeric/HHChannels.jl
```diff
@@ -19,12 +19,15 @@ immutable HHParam
...
+        # we use values derived from Nernst equation instead
+        #ena    = 115.0mV + -65.0mV,
+        #ek     = -12.0mV + -65.0mV,
         c_m    = 0.01F*m^-2,
         gnabar = .12S*cm^-2,
-        ena    = 115.0mV + -65.0mV,
+        ena    =  63.55148117386mV,
         gkbar  = .036S*cm^-2,
-        ek     = -12.0mV + -65.0mV,
+        ek     = -74.17164678272mV,
```
halfflat: Similarly, comment here to say that these values correspond to our Nernst-derived values given concentrations of ... etc.

src/backends/multicore/fvm.hpp
```diff
@@ -148,6 +148,21 @@ struct backend {
...
 
+    // Calculate the reversal potential eX (mV) using Nernst equation
+    // eX = RT/zF * ln(Xo/Xi)
+    //      R: universal gas constant 8.3144598 J.K-1.mol-1
+    //      T: temperature in Kelvin
+    //      z: valency of species (K, Na: +1) (Ca: +2)
+    //      F: Faraday's constant 96485.33289 C.mol-1
+    //      Xo/Xi: ratio of out/in concentrations
+    static void nernst(int valency, value_type temperature, const_view Xo, const_view Xi, view eX) {
+        constexpr value_type RF = 1e3*8.3144598/96485.33289; // factor 1e3 to scale from V -> mV
```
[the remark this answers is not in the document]
bcumming: Done. `src/constants.hpp`, in `arb::constant` namespace.

src/ion.hpp
bcumming: I have put it in `constants.hpp` as `squid_temperature` (a ridiculous name, with the understanding that it i will soon be replaced).

validation/ref/neuron/nrn_validation.py
bcumming: Added comment with all values used in the calculation.

validation/ref/numeric/HHChannels.jl
bcumming: Added comment with all values used in the calculation.

PR #390, opened 2017-11-15
Move event wrangling from cell_group to model

### LINKED ISSUE #378: Move event wrangling from `cell_group` to `model`
The `model` passes unsorted events to `cell_group`s via `cell_group::enqueue_events`.
Each `cell_group` implementation is responsible for implementing  sorting, binning and sheduling of events.

There are two drawbacks:
1. Each `cell_group` implementation is responsible for implementing optimized sorting, binning, and scheduling of events. This has to be reimplemented for each new `cell_group`, which is duplicated effort, and imposes a maintenance overhead.
2. There are opportunities to parallelize both calls to `enqueue_events` and the operations on individual cell event lanes inside `enqueue_events`. Having parallelism inside and outside this call makes optimizing the parallelism on a case-by-case basis difficult.

The event sorting method implemented in PR #369 has benefited from significant optimizations, and will be further optimized as need be. This task would share this code between all `cell_group` implementations, and make simplify scheduling of parallel work in the event wranging.

1. Move the per-cell "lanes" implemented in `mc_cell_group` to `model`:
  - ownership of data structures is moved;
  - scheduling of all parallel event-wrangling is in one spot in `model`.
2. Optimze the event binning code, in particular the hash table lookup with `following` style binning.

### LINKED ISSUE #385: Fix reproducibility by sorting events first by time then by weights
In case two events have exactly the same time (which can happen when a very large number of events is produced), there is an ambiguity in the order in which such events are processed which might result in different numberical results when the same simulation is repeated.

To ensure that the order of operations is deterministic, the events that happened at the same time should further be sorted by their weights.

bcumming:
Remove `cell_group::enqueue_events` interface:
* replace by passing a sorted list of events to the `cell_group::advance` method.
* `cell_group` is provided with events to deliver in each epoch: no longer needs to keep track of events due in future epochs. 

Add stand alone `merge_events` method:
* generates a sorted list of events to be delivered after the current epoch is finished.
* takes an unsorted set of events and the sorted set of events already in the model.
* simplifies unit testing.

Implement reproduceable ordering of events:
* order events to be delivered in ascending order of time, then target, then weight.
* fixes very infrequent collisions where two events with different weights are delivered
to the same target simultanously. On different systems the order in which events were delivered in different orders. 
* fixes #385
                                                                                                         
Fixes #378

### DIFF
--- a/src/event_queue.hpp
+++ b/src/event_queue.hpp
@@ -39,7 +39,8 @@ struct postsynaptic_spike_event {
     }
 };
 
-using event_lane_subrange = util::subrange_view_type<std::vector<std::vector<postsynaptic_spike_event>>>;
+using event_vector = std::vector<postsynaptic_spike_event>;
+using event_lane_subrange = util::subrange_view_type<std::vector<event_vector>>;
 
 template <typename Event>
 class event_queue {
--- a/src/event_queue.hpp
+++ b/src/event_queue.hpp
@@ -7,12 +7,13 @@
 #include <type_traits>
 #include <utility>
 
-#include "common_types.hpp"
-#include "generic_event.hpp"
-#include "util/meta.hpp"
-#include "util/optional.hpp"
-#include "util/range.hpp"
-#include "util/strprintf.hpp"
+#include <common_types.hpp>
+#include <generic_event.hpp>
+#include <util/meta.hpp>
+#include <util/optional.hpp>
+#include <util/range.hpp>
+#include <util/rangeutil.hpp>
+#include <util/strprintf.hpp>
 
 namespace arb {
 
@@ -37,6 +38,8 @@ struct postsynaptic_spike_event {
         return o << "E[tgt " << e.target << ", t " << e.time << ", w " << e.weight << "]";
     }
 };
+
+using event_lane_subrange = util::subrange_view_type<std::vector<postsynaptic_spike_event>>;
 
 template <typename Event>
 class event_queue {
--- a/src/event_queue.hpp
+++ b/src/event_queue.hpp
@@ -39,7 +39,7 @@ struct postsynaptic_spike_event {
     }
 };
 
-using event_lane_subrange = util::subrange_view_type<std::vector<postsynaptic_spike_event>>;
+using event_lane_subrange = util::subrange_view_type<std::vector<std::vector<postsynaptic_spike_event>>>;
 
 template <typename Event>
 class event_queue {
--- a/src/event_queue.hpp
+++ b/src/event_queue.hpp
@@ -29,8 +29,12 @@ struct postsynaptic_spike_event {
     time_type time;
     float weight;
 
-    friend bool operator==(postsynaptic_spike_event l, postsynaptic_spike_event r) {
+    friend bool operator==(const postsynaptic_spike_event& l, const postsynaptic_spike_event& r) {
         return l.target==r.target && l.time==r.time && l.weight==r.weight;
+    }
+
+    friend bool operator<(const postsynaptic_spike_event& l, const postsynaptic_spike_event& r) {
+        return std::tie(l.time, l.target, l.weight) < std::tie(r.time, r.target, r.weight);
     }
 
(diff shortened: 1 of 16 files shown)


src/model.cpp
```diff
@@ -189,4 +214,74 @@ void model::set_local_spike_callback(spike_export_function export_callback) {
...
+    // For ordering events by: delivery time, then target index, then weight.
+    auto lex_bind = [](const pse& e){return std::tie(e.time, e.target, e.weight);};
+    auto lex_less = [lex_bind] (const pse& l, const pse& r) {return lex_bind(l)<lex_bind(r);};
+
+    // Merge the incoming events with events in lc that are not to be delivered
+    // in this epoch, and store the result in lf.
+
+    // STEP 1: sort events in place in events[l]
+    PE("sort");
+    util::sort_by(events, lex_bind);
```
halfflat: We've got `lex_bind` here, but `event_bind` in `model::inject_events`. The bind is only used for sorting as far as I can tell. Can we define instead either a `pse_less` that does this comparison, or make it the `operator<` overload for `postsynaptic_spike_event`?

src/event_queue.hpp
```diff
@@ -38,6 +39,9 @@ struct postsynaptic_spike_event {
     }
 };
 
+using event_vector = std::vector<postsynaptic_spike_event>;
```
halfflat: Cosmetic, but: `event_vector` is a bit more specific than the name implies. Consider e.g. `pse_vector`? We've got `pse` as an alias for `postsynaptic_spike_event` elsewhere in the code.

src/mc_cell_group.hpp
```diff
@@ -87,35 +80,31 @@ class mc_cell_group: public cell_group {
...
-                auto h = target_handles_[target_handle_divisions_[lid]+e.target.index];
-                auto ev = deliverable_event(e.time, h, e.weight);
-                staged_events_.push_back(ev);
+        // skip event binning if empty lanes are passed
+        if (event_lanes.size()) {
+            for (auto lid: util::make_span(0, gids_.size())) {
+                auto& lane = event_lanes[lid];
+                for (auto e: lane) {
+                    if (e.time>=ep.tfinal) break;
+                    e.time = binners_[lid].bin(e.time, tstart);
```
halfflat: At one point we were talking about lifting the event binning out of the cell groups too. Is this still a good idea? In scope?

bcumming: Probably a good idea. Would best be done as a seperate PR to keep the scope of this PR well-defined, and because moving binning into model would not just be a simple refactoring.

src/event_queue.hpp
bcumming: Done.

I was in two minds about the name when writing this: `pse` felt "too abbreviated", and `postsynaptic_spike_event` was too long winded.

src/model.cpp
bcumming: I changed the name of `lex_bind` and `lex_less` to `pse_tie` and `pse_less`, and  described the lambda `event_bind` when passing it in `inject_events`.
I still need both the `tie` and the `less` operators for `sort_by` and `merge` respectively, so I would rather define both in line, where needed here.

halfflat: If we haven't already got a range-like sort wrapper, it'd be easy to add one — this could obviate the need for both a bind and a less.

bcumming: I will add one.

src/util/rangeutil.hpp
```diff
@@ -146,13 +146,27 @@ sort(Seq& seq) {
...
     std::sort(std::begin(canon), std::end(canon));
 }
 
+template <typename Seq, typename Less>
+enable_if_t<!std::is_const<typename sequence_traits<Seq>::reference>::value>
+sort(const Seq& seq, const Less& less) {
+    auto canon = canonical_view(seq);
+    std::sort(std::begin(canon), std::end(canon), less);
+}
+
```
halfflat: Looks like this version of `util::sort` has been inserted twice...

PR #391, opened 2017-11-15
 Add custom streambuf for indented output.

halfflat:
These classes aim to provide an alternative to `TextBuffer` in `modcc` printers.

* Implement a streambuf wrapper `prefixbuf` that prepends a prefix string to each line.
* Adds stream manipulators `setprefix`, `indent`, `popindent`, `settab` for manipulating the prefix in a `prefixbuf`.
* Provide a wrapper stream `pfxstringstream` around `std::stringbuf` with `std::ostringstream` semantics.

As implemented, a prefix string set with the `setprefix` is applied to the underlying streambuf, and is not a property of the stream. This behaviour may be surprising, so consider storing this prefix with the stream as a possible enhancement.

### DIFF
--- a/tests/unit/test_prefixbuf.cpp
+++ b/tests/unit/test_prefixbuf.cpp
@@ -1,0 +1,113 @@
+#include "../gtest.h"
+
+#include <cstring>
+#include <iomanip>
+#include <string>
+#include <sstream>
+
+#include <util/prefixbuf.hpp>
+
+using namespace arb::util;
+
+TEST(prefixbuf, prefix) {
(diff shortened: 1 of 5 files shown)


tests/unit/test_prefixbuf.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include <cstring>
+#include <iomanip>
+#include <string>
+#include <sstream>
+
+#include <util/prefixbuf.hpp>
+
+using namespace arb::util;
+
```
bcumming: Comments above each test case to state what is being tested, because the test is the documentation.

PR #394, opened 2017-11-20
Fix GPU event delivery bugs

bcumming:
Fix two issues:

* Not all event queues were being delivered for groups with more than one cell, which affected the GPU version of the miniapp.
* A GPU unit test was updated to use the new `cell_group::advance()` interface.


halfflat:
* Properly iterate over all event queues in the case where a cell group has more than once cell (triggered in e.g. GPU miniapp).
* Update GPU unit test to use the new `cell_group::advance()` interface.

PR #398, opened 2017-11-22
support ion species concentration updates

### LINKED ISSUE #373: Correctly handle ion concentration updates in modcc
modcc crashes for NMODL files that attempt to update the concentration of ionic species by solving an ODE. This isn't a surprise, since no attempt was made to handle the solution of state equations for "external" fields.

Doing this properly would require modifying NMODL quite significantly to handle generic species, which isn't something we have time for right now.

For the time being we should aim for a minimal implementation that can handle ionic species concentrations as done in NEURON.

bcumming:
## Updates to modcc:

* add querying methods to modcc ion species type to simplify code generation
* add a `WriteBack` type that records metadata about a member state variable
  in a mechanism that contributes to the concentration of an ionic species.

cuda and c printers:
* add `write_back` implementation that adds weighted concentration
  contributions to global concentration fields.
* extend `uses_ion` interface to provide information about which concentration
  fields are modified by mechanisms.

module:
* Update `update_ion_symbols` lambda that is responsible for adding metadata
  about interactions between mechanism fields and external ion species fields.
  Specifically, create a writeback when the mechanism field is a state variable,
  or an indexed variable if it is a current/reversal potential, or an error
  otherwise.

## Updates to mechanisms

* add `test_ca.mod` mechanism that writes to the internal calcium concentration                          

## Updates to Arbor:

* backend callbacks that initialise concentration values.
* update `mechanism::uses_ion()` return value to additionally provide information
  about whether a mechanism modifies ion species concentration fields.
* add `mechanism::write_back()`
* update stimulus mechanism specialization to use this new interface.
* update `ion` type to calculate default concentration contributions in CVs 
  where the concentration is not fully described by mechanisms
* update `fvm_multicell` to calcluate the weights used by `ion` to make
  this calculation
* add unit test to check that the weights are calculated and applied correctly.

Fixes #373


modcc/writeback.hpp
```diff
@@ -0,0 +1,21 @@
...
+    std::string target_name;
+    // The ion channel being written to.
+    //      must not be ionKind::none
+    ionKind ion_kind;
+
+    WriteBack(std::string src, std::string tgt, ionKind k):
+        source_name(std::move(src)), target_name(std::move(tgt)), ion_kind(k)
+    {}
+};
+
```
halfflat: This seems like a class that could just live in `module.hpp` or even as a nested struct within `Module`.

modcc/cudaprinter.cpp
```diff
@@ -164,6 +167,36 @@ CUDAPrinter::CUDAPrinter(Module &m, bool o)
...
+        buffer().add_line("using value_type = arb::fvm_value_type;");
+
+        buffer().add_line("auto tid_ = threadIdx.x + blockDim.x*blockIdx.x;");
+        buffer().add_line("auto const n_ = params_.n_;");
+        buffer().add_line("if(tid_<n_) {");
+        buffer().increase_indentation();
+
+        auto ion_prefix = [](ionKind k) {
+            return k==ionKind::Ca? "ca": k==ionKind::Na? "na": "k";
+        };
```
halfflat: The mappings between ion kind and names occur in a number of places in the code now; can we bring them together in the one spot? (This will also ease the transition to general ions).

modcc/cudaprinter.cpp
```diff
@@ -164,6 +167,36 @@ CUDAPrinter::CUDAPrinter(Module &m, bool o)
...
+
+        for (auto& w: module_->write_backs()) {
+            auto& src = w.source_name;
+            auto& tgt = w.target_name;
+
+            auto idx = src + "_idx_";
+            buffer().add_line("auto "+idx+" = params_.ion_"+ion_prefix(w.ion_kind)+"_idx_[tid_];");
+            buffer().add_line("// 1/10 magic number due to unit normalisation");
+            buffer().add_line("params_."+tgt+"["+idx+"] = value_type(0.1)*params_.weights_[tid_]*params_."+src+"[tid_];");
```
halfflat: On one hand, the target of an ion write back is a (backend-specific) function of the corresponding variable, and on the other the `CPrinter` and `CUDAPrinter` handling of the `target_name` field from the `Module::write_backs()` call differs, and looks a bit magic without knowing the whole path from ION block parsing down to this bit of code.

I think it would simplify both the logic and the understanding of the code if the printers just used a list of ion-write variables. This information is already available from the `IonDep` data.

modcc/cudaprinter.cpp
```diff
@@ -451,67 +463,70 @@ CUDAPrinter::CUDAPrinter(Module &m, bool o)
...
+    // return true/false indicating if cell has dependency on k
+    auto const& ions = module_->neuron_block().ions;
+    auto find_ion = [&ions] (ionKind k) {
+        return std::find_if(
+            ions.begin(), ions.end(),
+            [k](IonDep const& d) {return d.kind()==k;}
+        );
+    };
+    auto has_ion = [&ions, find_ion] (ionKind k) {
+        return find_ion(k) != ions.end();
```
halfflat: Common code here with `CPrinter` could be moved to the `Module` interface. Would it simplify the interface and code generation if the `Module` could provide the sequence of `IonDep` ion descriptions?

src/backends/gpu/ions.hpp
```diff
@@ -14,6 +14,12 @@ void nernst(std::size_t n, int valency,
             const fvm_value_type* Xi,
             fvm_value_type* eX);
 
+// prototype for nernst equation cacluation
```
halfflat: Comment doesn't match (cut-and-paste error by the looks of it).

src/ion.hpp
```diff
@@ -65,14 +65,15 @@ public :
...
-        default_external_concentration(0)
+        default_int_concentration(0),
+        default_ext_concentration(0)
     {}
 
-    std::size_t memory() const {
-        return 4u*size() * sizeof(value_type)
-               +  size() * sizeof(iarray)
-               +  sizeof(ion);
+    void set_weights(const std::vector<value_type>& win, const std::vector<value_type>& wout) {
```
halfflat: A comment here and/or a change in the method name would help to indicate that these weights correspond to the contribution of the default values, which may not be the CV area when there are ion-concentration writing mechanisms present.

src/backends/gpu/ions.hpp
bcumming: fixed.

src/ion.hpp
bcumming: done: added a comment.

modcc/writeback.hpp
bcumming: It does no harm ehre. The `module.xpp` files are getting large enough as is in my opinion.

modcc/cudaprinter.cpp
bcumming: Good point. I moved it to `Module`

modcc/cudaprinter.cpp
bcumming: I have consoldiated this a bit, but I haven't done an exhaustive search.
This is going to get ripped up when generic ion species are introduced, so I am not inclined to really refactor fully now.

modcc/cudaprinter.cpp
bcumming: This is going to get ripped up when generic ion species are introduced, so I am not inclined to really refactor fully now.

halfflat: I'll probably hack on it in the mechanism PR of doom.

PR #400, opened 2017-11-24
Fix binary operator parentheses bug.

### LINKED ISSUE #399: Incorrect binary expression generation in modcc with operators of same precedence.
Binary operation expressions with a binary sub-expression with the same operator precedence are incorrectly rendered.

| Expression | Rendered |
|--|--|
| `a/(b/c)` | `a/b/c` |
| `(a*b)/c` | `(a*b)/c` (redundant parentheses) |
| `a-(b-c)` | `a-b-c` |
| `a<(b>0)` | `a<b> 0` |

Common code across the C-printer and CUDA-printer gives the same errors for both back-ends.

halfflat:
Fixes #399.

Update `test_printers.cpp`:
* Fix compilation and scope management issues, and re-include in CMakeLists.txt.
* Update to use `verbose_print` (run-time verbosity).
* Apply expected output tests to `CPrinter` and `CUDAPrinter` outputs.

Address binary operation rendering issue:
* Move C-style expression rendering common to `CPrinter` and `CUDAPrinter` to a specialized `Visitor`, `CExprRenderer`.
* Use operator associativity to determine need for parentheses around sub-expressions of a binary expression.

### DIFF
--- a/modcc/cexpr_render.cpp
+++ b/modcc/cexpr_render.cpp
@@ -1,0 +1,101 @@
+#include <ostream>
+#include <unordered_map>
+
+#include "cexpr_render.hpp"
+#include "error.hpp"
+
+void CExprRenderer::render_as_call(const char* sub, Expression* e) {
+    out_ << sub << '(';
+    e->accept(this);
+    out_ << ')';
+}
+
+void CExprRenderer::render_as_call(const char* sub, Expression* e1, Expression* e2) {
+    out_ << sub << '(';
+    e1->accept(this);
+    out_ << ", ";
+    e2->accept(this);
+    out_ << ')';
+}
+
+void CExprRenderer::visit(NumberExpression* e) {
+    out_ << " " << e->value();
+}
+
+void CExprRenderer::visit(UnaryExpression* e) {
+    // Place a space in front of minus sign to avoid invalid
+    // expressions of the form: (v[i]--67)
+    static std::unordered_map<tok, const char*> unaryop_tbl = {
+        {tok::minus, " -"},
+        {tok::exp,   "exp"},
+        {tok::cos,   "cos"},
+        {tok::sin,   "sin"},
+        {tok::log,   "log"}
+    };
+
+    if (!unaryop_tbl.count(e->op())) {
+        throw compiler_exception(
+            "CExprRender: unsupported unary operator "+token_string(e->op()), e->location());
+    }
+
+    const char* op_spelling = unaryop_tbl.at(e->op());
+    Expression* inner = e->expression();
+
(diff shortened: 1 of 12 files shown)


modcc/cexpr_render.cpp
```diff
@@ -0,0 +1,112 @@
+#include <ostream>
+#include <unordered_map>
+
+#include "cexpr_render.hpp"
+#include "error.hpp"
+
+void CExprRenderer::render_as_call(const char* sub, Expression* e) {
```
bcumming: Can we use `emit` in place of `render`? Not because `render` is bad, but because we already use `emit` elsewhere in modcc.

modcc/token.hpp
```diff
@@ -77,6 +77,15 @@ enum class tok {
     reserved, // placeholder for generating keyword lookup
 };
 
+namespace std {
+    // note: necessary before C++14 (refer: lwg dr#2148).
+    template <> struct hash<tok> {
```
bcumming: newline before `struct`

tests/modcc/test_printers.cpp
```diff
@@ -1,55 +1,93 @@
...
+TEST(scalar_printer, statement) {
+    std::vector<testcase> testcases = {
+        {"y=x+3",            "y=x+3"},
+        {"y=y^z",            "y=std::pow(y,z)"},
+        {"y=exp((x/2) + 3)", "y=exp(x/2+3)"},
+        {"z=a/b/c",          "z=a/b/c"},
+        {"z=a/(b/c)",        "z=a/(b/c)"},
+        {"z=(a*b)/c",        "z=a*b/c"},
+        {"z=a-(b+c)",        "z=a-(b+c)"},
+        {"z=(a>0)<(b>0)",    "z=a>0<(b>0)"}
```
bcumming: How about a test for `z=a- -2` to test the case of subtracting a negative value.

modcc/cudaprinter.cpp
```diff
@@ -708,43 +709,7 @@ void CUDAPrinter::visit(LocalVariable *e) {
...
-            buffer() << "log(";
-            e->expression()->accept(this);
-            buffer() << ")";
-            return;
-        default :
-            throw compiler_exception(
-                "CUDAPrinter unsupported unary operator " + yellow(token_string(e->op())),
-                e->location());
-    }
+    cexpr_render(e, buffer().text(), this);
```
bcumming: Nice reduction of duplicated code.

modcc/cexpr_render.cpp
halfflat: Should I use `emit` in the class name and wrapper call as well? Easy enough to change.

PR #404, opened 2017-11-28
Tidy `modcc` driver, remove optimize flag.

halfflat:
* Remove optimization option (use SIMD options for vectorization).
* Remove arbor utility library dependencies from modcc (pending separation of utility lib from arbor lib source).
* Split target (cpu, gpu) specification from vectorization architecture (avx2, avx512).
* Remove `Options` singleton; replace with structure local to `modcc.cpp`.
* Tidy `modcc` option parsing and main function; allow a single invocation of `modcc` to generate code for multiple backends.
* Rename generated sources to include backend target in filename.
* Always run a constant simplification pass on generated procedures.
* Remove file i/o code from `Module` and `modcc` main function; move functionality to new functions in `io` namespace. (Note: in on-going mechanism revamp, other i/o utility code will reside in the `io` namespace and subdirectory.)
* Remove classes `ConstantFolderVisitor` and `ExpressionClassifierVisitor` that are no longer used.
* Modify CMakeLists.txt files, `backends/*/fvm.cpp` to reflect the new filenames of generated sources.
* Small formatting changes in `modcc` source to reflect coding guidelines (incomplete).


modcc/backends/base.hpp
```diff
@@ -8,15 +8,18 @@
...
+enum class simdKind {
+    none, avx2, avx512
+};
+
 namespace modcc {
 
-using arb::util::enable_if_t;
+template <bool V, typename R = void>
+using enable_if_t = typename std::enable_if<V, R>::type;
+
```
vkarak: I suppose that's related to the separation of the utility lib from the arbor source, right?

halfflat: Exactly, yes. I plan to put it back once we have a separate utility source directory :)

halfflat:
Argh. No `std::hash` specialization for enum defect strikes again. Will amend.

modcc/io/bulkio.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+// Read or write the contents of a file in toto.
+
+#include <string>
+#include <iterator>
+#include <fstream>
+
+namespace io {
+
+template <typename Assignable>
```
vkarak: It looks like `Assignable` is a container. I see that your only requirement here is that `A` provides an `assign()` function, so perhaps you don't need another concept name, e.g., `SequenceContainer` etc. Would it make sense to have sth like `AssignableContainer`, 'cause initially I got a bit confused. I thought you need sth to support assignment and there are already a `CopyAssignable` and `MoveAssignable` concepts with a different meaning.

halfflat: Certainly happy to rename! Perhaps `HasAssign`?

vkarak: `HasAssign` sounds fine for me.

modcc/io/bulkio.hpp
```diff
@@ -0,0 +1,49 @@
...
+// Read or write the contents of a file in toto.
+
+#include <string>
+#include <iterator>
+#include <fstream>
+
+namespace io {
+
+template <typename HasAssign>
+void snarf(std::istream& in, HasAssign& A) {
```
bcumming: I appreciate that `snarf` and friends have history, but I would prefer to use something more descriptive for people who have been programming less than 20 years.

PR #411, opened 2017-12-05
410 rss cell bug

w-klijn:
The rss cell had a single time variable per cell that was set per cell group.
It was standard set to epoch at end of the epoch.

This causes problems with spike time steps bigger then epoch and when there are cells with different settings as observed by Avleen.

Each cell now has an unique time variable.
Change includes a test that went from red to green after the change.


w-klijn:
Intersting, this problem will be present with basically all time stepped neurons. 
I will also change this on the ipss_cell. I already use they approach in some corned cases. Thus extending this to the whole cell should be easy

bcumming:
You are right, it is a problem for many neurons that we time step. @halfflat and I decided that it would be best to have the "expected" behavior whenever it is reasonable to implement.

src/rss_cell_group.hpp
```diff
@@ -38,12 +38,17 @@ class rss_cell_group: public cell_group {
 
     void advance(epoch ep, time_type dt, const event_lane_subrange& events) override {
         for (auto& cell: cells_) {
-            auto t = std::max(cell.start_time, cell.t);
+            auto t_start = std::max(cell.start_time, cell.t);
             auto t_end = std::min(cell.stop_time, ep.tfinal);
 
+            std::size_t step = 0;
```
bcumming: To have full reproducable sequences, use `t = cell.start_time+cell.step*cell.period`, and increment `cell.step` (and set it to 0 when resetting the cell).

w-klijn: Done

PR #414, opened 2017-12-07
Implement cell-local event generators

### LINKED ISSUE #401: Generic event generation on cells. [4]
Implement generic event generation directly on cell postsynaptic event input queues.

Steps:
- [x] recipe support for describing the event generators on a given gid
- [x] generic event-generator type that users can customize to generate whatever sane and not-so-sane event streams
- [x] merge events from communicator with those from cell-local generators

Optimization of the k-way merge of event streams is will be a seperate task, to be implemented when and if merging is a bottleneck.

bcumming:
Implement generic `event_generator`, and some common generators that should be useful to users out of the box:
* poisson sequence                                                                                       
* uniform regular sequence
* from a user-prescribed sequence
* from a `std::vector` of post synaptic events.

Extend `recipe` API to provide an `recipe::event_generators(gid)` that returns a list of `event_generator`s attached to a cell with `gid`.

Update the `model` class and `merge_events` method to support event generators.

Fixes #401


src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+        if (!std::is_sorted(events_.begin(), events_.end())) {
+            util::sort(events_);
+        }
+    }
+
+    postsynaptic_spike_event next() override {
+        return it_==events_.end()? terminal_pse(): *it_;
+    }
+
+    void pop() override {
```
w-klijn: I would expect a pop() to return something. And compared with next() it might be a little confusing

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+    std::vector<postsynaptic_spike_event> events_;
+    std::vector<postsynaptic_spike_event>::const_iterator it_;
+};
+
+// Generator that for events in a generic sequence
+template <typename Seq>
+struct seq_generator: public event_generator {
+    using pse = postsynaptic_spike_event;
+    seq_generator(Seq& events):
+        events_(events),
```
w-klijn: In the vector_backed_generator you use a specific std::move()
Why not use this at this location?

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+private:
+    std::vector<postsynaptic_spike_event> events_;
+    std::vector<postsynaptic_spike_event>::const_iterator it_;
+};
+
+// Generator that for events in a generic sequence
+template <typename Seq>
+struct seq_generator: public event_generator {
```
w-klijn: The implementation appears the same as the vector_backed_generator (except) the sorting.

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+};
+
+// Generates a stream of events
+//  * with the first event at time t_start
+//  * subsequent events at t_start+n*dt, ∀ n ∈ [0, ∞)
+//  * with a set target and weight
+struct regular_generator: public event_generator {
+    using pse = postsynaptic_spike_event;
+
+    regular_generator(time_type tstart, time_type dt, cell_member_type target, float weight):
```
w-klijn: I would extend the interface with t_end (maybe as an default parameter time_type.max())

src/event_generator.hpp
bcumming: I agree: I sometimes find it confusing that `pop` doesn't return the value at the front of the queue, but that is how I have seen it implemented in most places, and it is also the interface of `std::queue`:

http://en.cppreference.com/w/cpp/container/queue

To get the next value from the generator, do you have suggestions? `std::queue` uses `front`, and I think that `peek` is often used too.

I am sure that @halflat has an opinion.

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+    postsynaptic_spike_event next() override {
+        return {target_, t_start_+(step_*dt_), weight_};
+    }
+
+    void pop() override {
+        ++step_;
+    }
+
+    void advance(time_type t0) override {
+        t0 = t0<t_start_? t_start_: t0;
```
w-klijn: I would put a short comment how this works.

src/event_generator.hpp
bcumming: Because the sequence might not own the underlying storage/state that generates the events.

The `vector_backed_generator` keeps a copy of the vector, so we pass it to the constructor by value, then move it into place. This is a common strategy for "sink parameters".

src/event_generator.hpp
bcumming: There is some repetition.

The main difference is that the `Seq` type is much more general. It could, for example, be an infinite sequence generated lazily. So we can't sort it.

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+
+    time_type t_start_;
+    std::size_t step_;
+    time_type dt_;
+    cell_member_type target_;
+    float weight_;
+};
+
+// Generates a stream of events
+//  * At times described by a Poisson point process with rate 1/dt
```
w-klijn: This is ambiguous. Is this per second per milli second?
Typically I would expect it defined as (spikes/second)

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+template <typename RandomNumberEngine>
+struct poisson_generator: public event_generator {
+    using pse = postsynaptic_spike_event;
+
+    poisson_generator(time_type tstart,
```
w-klijn: Again I would add and end time

src/event_generator.hpp
bcumming: Ha, you are right! I should just use `std::max(to, t_start)`

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+        return {target_, next_, weight_};
+    }
+
+    void pop() override {
+        next_ += exp_(rng_);
+    }
+
+    void advance(time_type t0) override {
+        while (next_<t0) {
+            pop();
```
w-klijn: Could this mean that if multiple pops occur untill t0 that they can not be emitted with next?

src/event_generator.hpp
bcumming: You are right, I should mention dt is has units `ms`.

Isn't desribing a Poisson process using either expected rate or frequency six of one, half a dozen of the other? I just used the convention that @halfflat used for the `schedule` stuff... because he cares much more about the semantics of these things than I do!

src/merge_events.cpp
```diff
@@ -0,0 +1,205 @@
...
+#include <util/span.hpp>
+#include <util/unique_any.hpp>
+#include <profiling/profiler.hpp>
+
+namespace arb {
+
+namespace impl {
+
+// The tournament tree data structure is used to merge k sorted lists of events.
+// See online for high-level information about tournament trees.
```
w-klijn: Would this be a good link to add (google I feel lucky)
http://www.geeksforgeeks.org/tournament-tree-and-binary-heap/

src/event_generator.hpp
bcumming: They are effectively being discarded. This method removes events from the queue until `next().time>=t0`, i.e. until we have the first event to be delivered on or after `t0`.

src/merge_events.cpp
bcumming: I thought about adding that exact link! But decided against it because links can go stale over time.
When googling "tournament tree" the first few hits are good references, so I figure folks in the future will just google it for themselves.
Keep in mind that this is a low level part of the library, designed just for the task in this file.

src/merge_events.cpp
```diff
@@ -0,0 +1,205 @@
...
+    return event(0).time >= t;
+}
+
+postsynaptic_spike_event tourney_tree::head() const {
+    return event(0);
+}
+
+// Remove the smallest (most recent) event from the tree, then update the
+// tree so that head() returns the next event.
+void tourney_tree::pop() {
```
w-klijn: Pop without a return value

src/merge_events.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+bool tourney_tree::empty(time_type t) const {
+    return event(0).time >= t;
+}
+
+postsynaptic_spike_event tourney_tree::head() const {
+    return event(0);
+}
+
+// Remove the smallest (most recent) event from the tree, then update the
```
w-klijn: This explanation talks about the tree, but the implementation pops from the input lane

bcumming: It removes an event from the appropriate input lane, places it at a leaf node of the tree, then re-heapifys the tree.
This is all predicated on someone understanding what a tournament tree. I am not inclined to fully document the algorithm, because this is used in just one spot, and not exported for use elsewhere (well, it is exported in the `impl` namespace so that it can be unit tested).

src/merge_events.hpp
```diff
@@ -0,0 +1,83 @@
...
+//
+//   [----------------------] lc
+//          [---------------] events
+//          [------) generators
+//
+// The output list, stored in lf, will contain all the following:
+//  * all events in events list
+//  * events in lc with time >= t₀
+//  * events from each generator with time < t₁
+// All events in lc that are to be delivered before t₀ are discared, along with
```
w-klijn: This comment appears to contradict itself. events on generators are discarded. But the next event on the generator is one that is discarded.

src/merge_events.cpp
```diff
@@ -0,0 +1,205 @@
...
+        auto lc_it = lower_bound(lc.begin(), lc.end(), t0, event_time_less());
+        auto lc_range = util::make_range(lc_it, lc.end());
+        generators[1] = make_event_generator<seq_generator<decltype(lc_range)>>(lc_range);
+
+        // Perform k-way merge of all events in events, lc and the generators
+        // that are due to be delivered in the interval [t₀, t₁)
+        impl::tourney_tree tree(generators);
+        while (!tree.empty(t1)) {
+            lf.push_back(tree.head());
+            tree.pop();
```
w-klijn: pop_head() ?

src/merge_events.cpp
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        // Find first event in lc with delivery time >= t1
+        lc_it = lower_bound(lc.begin(), lc.end(), t1, event_time_less());
+        // Find first event in events with delivery time >= t1
+        auto ev_it = lower_bound(events.begin(), events.end(), t1, event_time_less());
+        const auto m = lf.size();
```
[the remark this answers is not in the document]
bcumming: It is the number of events already in `lf`.
I kept the name short because:
* it is used just on the very next line
* I use a seperate variable because the code on the next line was becoming to verbose.

src/merge_events.hpp
bcumming: No events from generators are discarded in practice: `merge_events` takes all events from the generators with delivery time < `t1`. It stops reading events from the generators when it finds the first event in a generator that is to be delivered on or after `t1`. That event will be taken on the next call to `merge_events`, in the following epoch.
So you are right, this comment is a bit confusing.

src/event_generator.hpp
w-klijn: "Removes an element from the front of the queue. Effectively calls c.pop_front()"

src/event_generator.hpp
w-klijn: For the lesser programming gods, add this as class documentation?

src/event_generator.hpp
w-klijn: Since this will end up being user facing code, I would make it in spikes per second.

src/merge_events.cpp
w-klijn: https://web.archive.org/web/20170611232149/http://www.geeksforgeeks.org/tournament-tree-and-binary-heap/

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+    void advance(time_type t) override {
+        it_ = std::lower_bound(events_.begin(), events_.end(), t, event_time_less());
+    }
+
+private:
+    std::vector<postsynaptic_spike_event> events_;
+    std::vector<postsynaptic_spike_event>::const_iterator it_;
+};
+
+// Generator that for events in a generic sequence
```
kabicm: without "that"?

src/merge_events.cpp
```diff
@@ -0,0 +1,205 @@
...
+    // Allocate space for the tree nodes
+    heap_.resize(nodes_);
+    // Set the leaf nodes
+    for (auto i=0u; i<leaves_; ++i) {
+        heap_[leaf(i)] = i<n_lanes_?
+            key_val(i, input[i]->next()):
+            key_val(i, terminal_pse()); // null leaf node
+    }
+    // Walk the tree to initialize the non-leaf nodes
+    setup(0);
```
kabicm: Since this is a DFS (preorder) traversal of the tree, I would rather call it dfs(0) or walk(0), but that's a matter of taste and setup(0) is also perfectly fine. It's hard to please everyone :D

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+    {
+        reset();
+    }
+
+    postsynaptic_spike_event next() override {
+        return {target_, next_, weight_};
+    }
+
+    void pop() override {
+        next_ += exp_(rng_);
```
kabicm: As we have discussed, this might not work with the counter based random number generators as Random123. 

However, I think the easiest way to make this work with Random123 (or some other counter-based generator) is to make a wrapper around Random123 which will be compatible with the standard library c++ distribution so that we can invoke it as exp_(random123_generator).

src/event_generator.hpp
bcumming: I am going with `pop`, because it is what `std::queue` and `std::priority_queue` use.
`pop_front()` is not part of the interface for the STL queue types: it is an interface that must be provided by the optional user-defined container used by the queue.

src/event_generator.hpp
bcumming: I have added some documentation.

src/event_generator.hpp
bcumming: Done.

src/event_generator.hpp
```diff
@@ -0,0 +1,209 @@
...
+};
+
+// Generates a stream of events
+//  * At times described by a Poisson point process with rate 1/dt
+template <typename RandomNumberEngine>
+struct poisson_generator: public event_generator {
+    using pse = postsynaptic_spike_event;
+
+    poisson_generator(time_type tstart,
+                      time_type dt,
```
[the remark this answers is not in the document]
bcumming: I changed it to `lambda`, because both you and Wouter suggested that the rate was more intuitive.

src/model.cpp
```diff
@@ -14,15 +15,30 @@
...
     cell_local_size_type lidx = 0;
-    for (auto i: util::make_span(0, decomp.groups.size())) {
-        for (auto gid: decomp.groups[i].gids) {
+    const auto& grps = decomp.groups;
+    for (auto i: util::make_span(0, grps.size())) {
+        for (auto gid: grps[i].gids) {
+            // Store mapping of gid to local cell index.
             gid_to_local_[gid] = lidx++;
+
+            // Set up the event generators for cell gid.
```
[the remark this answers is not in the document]
bcumming: Which other classes are you saying?

Combining post-synaptic events from the communicator and the event generators is performed in `model`. It used to be handled inside the `cell_group` implementations, but it was refactored to:
* expose more parallelism: this way we can parallelize the merging operation with one thread per cell during the communication phase, which is a massive optimization for the GPU back end and LIF cells, which use cell groups with many cells.
* reduce the complexity of cell_group implementations, which no longer have to provide an `enqueue_events` interface: instead they are passed the events to be delivered in the next epoch already sorted and ready for deliver.

That is why we store the set of `event_generators` for each local cell inside the `model`.

tests/unit/test_event_generators.cpp
```diff
@@ -0,0 +1,222 @@
...
+    }
+    gen.reset();
+
+    // Test advance
+    gen.advance(10.1);
+    EXPECT_EQ(gen.next().time, time_type(10.5));
+    gen.advance(12);
+    EXPECT_EQ(gen.next().time, time_type(12));
+
+    // Test for rounding problems with large time values.
```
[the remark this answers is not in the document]
bcumming: Done.

src/event_generator.hpp
bcumming: I made a it `spikes_per_ms`, because making it per second made a mess of the units.
This will change to `spikes_per_second` when we fix all the units.

src/event_generator.hpp
bcumming: Added.

PR #415, opened 2017-12-11
Python miniapp

apeyser:
Run the miniapp from python as pyminiapp.miniapp(arguments....)


apeyser:
Primarily, I'd like people's opinions on directory structure & cmake infrastructure. The code itself is trivial.

apeyser:
Now it's been converted from simply applying the command line to creating an object for setting command line-type parameters.

python/miniapp/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+set(PYMINIAPP_SOURCES
```
w-klijn: I would expect the same pattern to develop as the c++ side: multiple miniapps. So the directory structure could mimic this?

apeyser: Yes, this is what I've done on the feature/python branch which add pyminiapp + pyarbor

bcumming:
Closing this because it has rotted, and work on using pybind11 is now under way.

PR #421, opened 2017-12-19
Add granule cell mechanisms

### LINKED ISSUE #417: Add kdrmt, nax and kamt ion channels
These ion channels are required for olfactory bulb modelling.

This task involves
* adding the channels
* modifying them to be Arbor compatible
* updating modcc if required

bcumming:
Add some built in math functions to modcc:
* `min`, `max` prefix binary operators
* `abs` unary operator
* `exprelr` unary operator, defined as `exprelr(x)=x/(exp(x)-1)` with `exprelr(1)=1`. This operation is frequently used, e.g. in HH, and handled in an ad-hoc manner by mechanism authors, e.g. using `vtrap` .

Each new math function requires C++, AVX2, AVX512 and CUDA implementations.
These were the first AVX512 functions that are not covered by the SVML library, so the implementation of intrinsincs in `src/backends/intrin.hpp` was reworked into two separate implementation files.
I experiment with some less-verbose intrinsices for the AVX512 versions.

The `min` and `max` were implemented as binary operators. These are the first prefix binary operators added to NMODL, so some small changes were made to the parser to handle them.
                                                                                                         
Added three new mechanisms `nax.mod`, `kdrmt.mod` and `kamt.mod`.

Fixes #417


src/math.hpp
```diff
@@ -80,6 +80,28 @@ int signum(T x) {
...
+template <typename T>
+T min(const T& lhs, const T& rhs) {
+    return rhs<lhs? rhs: lhs;
+}
+
+// Return maximum of the two values
+// Note: returns second argument in case of equality
+template <typename T>
+T max(const T& lhs, const T& rhs) {
+    return lhs<rhs? lhs: rhs;
```
halfflat: This also returns minimum.

halfflat: Is there a good reason not to wrap `std::min` or `std::fmin` (for stronger guarantees) instead?

tests/unit/test_math.cpp
```diff
@@ -285,3 +285,16 @@ TEST(quaternion, rotate) {
...
+    constexpr double dmax = std::numeric_limits<double>::max();
+    constexpr double deps = std::numeric_limits<double>::epsilon();
+    double inputs[] = {-1.,  -0.,  0.,  1., -dmax,  -dmin,  dmin,  dmax, -deps, deps, 10*deps, 100*deps, 1000*deps};
+
+    for (auto x: inputs) {
+        if (std::fabs(x)<deps) EXPECT_EQ(1.0, exprelr(x));
+        else                   EXPECT_EQ(x/std::expm1(x), exprelr(x));
+    }
+}
+
```
halfflat: Should add tests for new `min` and `max`. (Also for new CUDA kernels: time for a `test_math.cu`?).

src/backends/gpu/intrinsics.hpp
```diff
@@ -39,3 +39,24 @@ inline float cuda_atomic_sub(float* address, float val) {
...
+__device__
+inline T min(T lhs, T rhs) {
+    return rhs<lhs? rhs: lhs;
+}
+
+// Return maximum of the two values
+template <typename T>
+__device__
+inline T max(T lhs, T rhs) {
+    return lhs<rhs? lhs: rhs;
```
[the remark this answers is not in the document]
bcumming: No it doesn't.
(sound of someone furiously typing `git commit -am "fix dumb mistake"; git push` in the background)

src/backends/multicore/intrin_avx512.hpp
```diff
@@ -0,0 +1,80 @@
...
+    const vecd_avx512 vecd_avx512_two  = set(2.0);
+    const vecd_avx512 vecd_avx512_nan  = set(std::numeric_limits<double>::quiet_NaN());
+    const vecd_avx512 vecd_avx512_inf  = set(std::numeric_limits<double>::infinity());
+    const vecd_avx512 vecd_avx512_ninf = set(-std::numeric_limits<double>::infinity());
+}
+
+//
+// Operations on vector registers.
+//
+// shorter, less verbose wrappers around intrinsics
```
vkarak: I like the wrappers. I think we should template them on the backend (AVX2, AVX512 etc.) and expose them to the upper levels.

src/backends/multicore/intrin_avx512.hpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+inline vecd_avx512 min(vecd_avx512 x, vecd_avx512 y) {
+    // substitute values in x with values from y where x>y
+    return blend(greater(x, y), x, y);
+}
+
+inline vecd_avx512 exprelr(vecd_avx512 x) {
+    const auto ones = set(1);
+    return blend(less(ones, add(x, ones)), div(x, expm1(x)), ones);
```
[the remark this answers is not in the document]
halfflat: We should<sup>*</sup> implement compile-time S-expressions with preprocessor macros.

<sup>*</sup> _should not_

tests/unit/test_intrin.cpp
```diff
@@ -46,6 +52,76 @@ TEST(intrin, exp256) {
     }
 }
 
+TEST(intrin, abs256) {
+    constexpr size_t simd_len = 4;
+
+    __m256d vvalues[] = {
+        _mm256_set_pd(-42.,  -0.,  0.,  42.),
+        _mm256_set_pd(-dmin,  -dmax,  -dmax, dqnan),
```
vkarak: I think we should also test for `-Inf` and `Inf`.

tests/unit/test_intrin.cpp
```diff
@@ -46,6 +52,76 @@ TEST(intrin, exp256) {
...
+    constexpr size_t simd_len = 4;
+
+    // TODO: the third set of test values is commented out because it currently
+    // fails, because our implementation uses x/(exp(x)-1), when we should use
+    // x/expm(1) to handle rounding errors when x is close to 0.
+    // This test can be added once we have an implementation of expm1.
+    __m256d vvalues[] = {
+        _mm256_set_pd(-1.,  -0.,  0.,  1.),
+        _mm256_set_pd(-dmax,  -dmin,  dmin,  dmax),
+        //_mm256_set_pd(-deps, deps, 10*deps,  100*deps),
```
vkarak: Perhaps, we should also add a test for NaNs here.

tests/unit/test_intrin.cpp
bcumming: I am in two minds about this: sure for completeness, but on the other hand we never intend to have NaN or Inf in our models.

tests/unit/test_math.cpp
bcumming: Added.

modcc/expression.hpp
```diff
@@ -41,6 +41,8 @@ class ExpUnaryExpression;
 class LogUnaryExpression;
 class CosUnaryExpression;
 class SinUnaryExpression;
+class AbsUnaryExpression;
+class ExprelrUnaryExpression;
```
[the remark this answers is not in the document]
bcumming: They need not be. Fixed.

modcc/parser.cpp
```diff
@@ -1230,6 +1236,34 @@ expression_ptr Parser::parse_primary() {
...
+                auto op = token_;
+                // handle infix binary operators, e.g. min(l,r) and max(l,r)
+                get_token(); // consume operator keyword token
+                if (token_.type!=tok::lparen) {
+                    error("expected opening parenthesis '('");
+                    return nullptr;
+                }
+                get_token(); // consume (
+                auto lhs = parse_expression();
+                if (!lhs) return nullptr;
```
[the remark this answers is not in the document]
bcumming: fixed.

src/math.hpp
bcumming: I am not too concerned about stronger guarentees. So we could use `std::min`... I have just used a ternary (with a test to make sure I got it right).

CMakeLists.txt
```diff
@@ -205,11 +205,11 @@ set(ARB_VECTORIZE_TARGET "none" CACHE STRING "CPU target for vectorization {KNL,
 set_property(CACHE ARB_VECTORIZE_TARGET PROPERTY STRINGS none KNL AVX2 AVX512)
 
 if(ARB_VECTORIZE_TARGET STREQUAL "KNL")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXXOPT_KNL}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CXXOPT_KNL} -DVEC_ARC_KNL")
```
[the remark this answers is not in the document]
bcumming: I updated it to `SIMD_ARCH`.

src/backends/multicore/intrin_avx512.hpp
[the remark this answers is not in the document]
bcumming: Just wrapping the vector intrinsics encourages that style. It works for me.

src/backends/multicore/intrin_avx512.hpp
bcumming: The wrappers over simple operations like `add` don't need to be templated I think.
You can just use function overloading to do the trick (as opposed to having to provide specializations for different vector registers).

tests/unit/test_intrin.cu
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    __global__
+    void test_max(double* x, double* y, double* result) {
+        const auto i = threadIdx.x;
+        result[i] = max(x[i], y[i]);
+    }
+
+}
+
```
halfflat: Just missing the test for `exprelr`.

PR #427, opened 2017-12-21
Fix single child branch bug, and remove `cell_tree`

### LINKED ISSUE #426: Error building cells with branches with only one child
There is a bug inside the cell tree graph code that is triggered when attempting to build a cell that has a branch point with only one child. e.g.
```C++
// cell has structure
//      soma-dendrite-dendrite                                                                                                                                                                                      
cell c;
c.add_soma(1, {0,0,1});
c.add_cable(0, ...);
c.add_cable(1, ...);
c.model(); // this throws an exception
```

bcumming:
Add flag to the `tree` constructor to signal if the `parent_index` used for construction describes compartment or segment connections.

The `cell_tree` wrapper around `tree` was not being used in anger, so I removed it at the same time... which simplifies the code somewhat.

Fixes #426.


src/tree.hpp
```diff
@@ -34,42 +34,55 @@ class tree {
...
+    /// The parent_index can be in one of two forms:
+    ///   1) describes connectivity of compartments, from which the tree
+    ///      needs to be derived;
+    ///   2) gives the parent of each segment in the tree.
+    /// There is no way to determine the case from inspection of parent_index alone,
+    /// so the derive_segments parameter determines the case:
+    ///     true:   case 1  (default)
+    ///     false:  case 2
+    //template <typename I>
+    tree(std::vector<int_type> parent_index, bool derive_segments=true) {
```
halfflat: This just needs a little more documentation about what the `tree` data structure is representing. An additional issue is that the `derive_segments` parameter not only indicates to `tree` what it should do in terms of processing, but also changes the semantics of the first argument, which is a bit confusing.

I'd prefer an interface which either:
1. Always took in a segment parent index, with it being the caller's responsibility to construct that from a compartment-parent index first using e.g. `make_parent_index`, or:
2. Two constructors differentiated by a tag class, to make the change in semantics explicit, e.g.
```
explicit tree(std::vector<int_type> segment_parent_index);
tree(tree_compartment_parents_t, std::vector<int_type> compartment_parent_index);
```

halfflat: Bit of web trawling indicates that the process by which a tree is transformed into a homeomorphic pre-image is called 'reduction'. The 'reduced' tree is the smallest tree that maps into the tree homeomorphically (i.e. preserving labels and child relationships), and is what you get essentially by replacing everything that looks like `a`—`b`—`c` with `a`—`c` recursively.

So an appropriate tag type for the second constructor in the comment above might be `tree_reduce_t`.

PR #428, opened 2017-12-22
397: Inhomogeneous poisson spike source

w-klijn:
This is a input spike source that generates Poisson spike trains with a time varying rate described as a list of time rate pairs. The rates can be interpolated if needed. Each configurable time_step a dice is rolled and a spike might be generated.

Internal time keeping is done with an unsigned long with times calculated from that, this alleviates the float rounding issues discussed in slack.
All times are round up to the first multiple of the sample time step.
Interpolation is done at t + .5dt times. 

A mini_app based on simple_recipes.hpp is used. It generates configurable time-varying curves with the default settings matching a 'validation' curve in the accompanying python script. That plots the resulting binned instantaneous spike rate.

A documentation page is added including a start of some additional cell types.

I did a small performance test: 93% of the run-time of the cell is spend in random number generation. Improving performance necessitates improved rng (block generation is used in NEST). 

Upstream changes have been merged in.


example/ipss_app/ipss_util.hpp
```diff
@@ -0,0 +1,126 @@
...
+        // create the cell group
+        arb::ipss_cell_group sut(gids,
+            ipss_recipe(n_cells,
+            { begin, end, sample_delta, rates_per_time, interpolate }));
+
+        // Split whatever time we have into 10 equal epochs
+        arb::time_type time_step = (end - begin) / 10.0;
+
+        // run the cell
+        for (std::size_t idx = 0; idx < 10; ++idx) {
```
apeyser: Could also use `for (const auto idx: make_span(0, n_cells))` here (and in other for loops).

example/ipss_app/ipss_util.hpp
```diff
@@ -0,0 +1,126 @@
...
+        // return a copy, the cells will be destructed after the end
+        // of this function
+        return std::vector<arb::spike>(sut.spikes());
+    }
+
+    // Simple spike writer output spikes according to the gpf standard
+    // gid, time
+    void write_spikes_to_path(const std::vector<arb::spike> spikes, std::string path) {
+        std::ofstream outfile(path);
+        if (outfile) {
```
apeyser: I'd just do
```
if (! outfile) {
   throw ...
}

for (auto spikes: spikes) ...
```

The logic is clearer to me that way -- check for errors and throw, and then just write the correct code.

example/ipss_app/parse_and_plot.py
```diff
@@ -0,0 +1,208 @@
...
+    silently on errors
+    """
+    spikes_per_cell = collections.defaultdict(list)
+    with open(path , "r") as f:
+        for line in f:
+            tokens = line.split(",")
+            try:
+                gid = int(tokens[0].strip())
+                time = float(tokens[1].strip())
+            except:
```
apeyser: Is there only one possible exception here? Couldn't the exception tell us something other than failed?

example/ipss_app/parse_and_plot.py
```diff
@@ -0,0 +1,208 @@
...
+    time_rate = []
+
+    with open(path , "r") as f:
+        for line in f:
+            tokens = line.split(",")
+            try:
+                time = float(tokens[0].strip())
+                rate = float(tokens[1].strip())
+            except:
+                print "Failed parsing:", tokens[0], ", ", tokens[1]
```
apeyser: Same -- maybe the exception tells us something?

example/ipss_app/parse_and_plot.py
```diff
@@ -0,0 +1,208 @@
...
+    """
+    Perform a simple binning on the collected spikes before duration
+    Assuming the times are in ms resolution
+    bin_size (ms)
+    duration (ms)
+    """
+    n_bins = int(math.floor(duration / bin_size))
+    bins=[0]*n_bins
+
+    for gid, spikes in spikes_per_cell.items():
```
apeyser:
```
for spike in (s for _, spikes in spikes_per_cell.items() for s in spikes):
  ...
```

Maybe not... but maybe just replace `gid` with `_` since it's not used.

example/ipss_app/parse_and_plot.py
```diff
@@ -0,0 +1,208 @@
...
+
+def plot_histogram_and_target_curve(bins, times=None, rates=None, plot_target=True):
+    """
+    Plot the histogram and target curve in one figure
+    """
+    plt.hist([idx for idx in range(len(bins))], len(bins), weights = bins)
+    if plot_target:
+        plt.plot(times, rates, linewidth=7,alpha=0.7)
+
+    plt.title("Instantanious spike rate for a population of ipss cells ")
```
apeyser: `Instantaneous` spike rate (spelling)

example/ipss_app/parse_and_plot.py
```diff
@@ -0,0 +1,208 @@
...
+    plt.xlabel("Time (ms.)")
+    plt.ylabel("Instantanious spike rate (spikes / second)")
```
apeyser: Same

example/ipss_app/parse_and_plot.py
```diff
@@ -0,0 +1,208 @@
...
+    # plot the binned spikes and the optional curve
+    plot_histogram_and_target_curve(bins, times, rates, plot_target)
+
+def evaluate_string_true(s="True"):
+    """
+    Evaluates if s is a string with True value 
+    True -> 'true', '1', 't', 'yes', 'y'
+    Capitalization is ignored
+    """
+    return s.lower() in ['true', '1', 't', 'yes', 'y']
```
apeyser: Don't know whether a tuple would be implemented better than a list here, since the tuple is fixed length.

example/ipss_app/parse_and_plot.py
```diff
@@ -0,0 +1,208 @@
...
+
+if __name__ == "__main__":
```
apeyser: And this is why python should have had switch/case statements --- but it doesn't!

src/cell_group_factory.cpp
```diff
@@ -31,6 +32,9 @@ cell_group_ptr cell_group_factory(const recipe& rec, const group_description& gr
     case cell_kind::data_spike_source:
         return make_cell_group<dss_cell_group>(group.gids, rec);
 
+    case cell_kind::inhomogeneous_poisson_spike_source:
```
apeyser: We should put an issue up about making this generic so that cell groups can be loaded dynamically rather than requiring someone to find this case to add new groups.

src/ipss_cell.hpp
```diff
@@ -0,0 +1,201 @@
...
+        // spikes per sample_delta_  and the time to step_idx from t = 0.0
+        for (auto entry : desc.rates_per_time) {
+            auto spikes_per_sample_delta = (entry.second / 1000.0) * sample_delta_;
+            auto time_step = time_to_step_idx(entry.first, sample_delta_);
+            rates_per_step_.push_back({ time_step, spikes_per_sample_delta });
+        }
+
+        // Add and extra rate entry at max long, copy of the last original entry
+        // This allows current and next iterator.
+        rates_per_step_.push_back({ -1, rates_per_step_.cend()->second });
```
apeyser: `rates_per_step.back().second`

In fact, it looks to me like this should be an error: `cend()` should reference one past the end, so this is an undefined call past the end of the vector

src/ipss_cell.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+        // Sanity checks
+        if (start_step_ > stop_step_) {
+            throw std::logic_error("The start time is after the end time in the Inhomogeneous Poisson Spike Source.");
+        }
+
+        if (start_step_ < rates_per_step_.cbegin()->first) {
```
apeyser: `rates_per_step.front().first`

src/ipss_cell.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+        // set the pointer to the first and second entry in the rates vector
+        current_rate_it_ = rates_per_step_.cbegin();
+        next_rates_it_ = current_rate_it_;
+        next_rates_it_++;
+        update_rates();
+
+        // Step through the rate_change vector if begin_step does not fall in the
+        // range of the current pointers
+        while (start_step_ >= next_rates_it_->first) {
```
apeyser: Can't we fall of the end here by incrementing `next_rates_it` or are we guaranteed to find `start_step_`?

src/ipss_cell.hpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    /// Advance the Poisson generator until end of epoch
+    /// If a rate change is occurred during this ep the new rate will be used
+    /// including possible interpolation between steps.
+    void advance(epoch ep, std::vector<spike>& spikes)
+    {
+        // Convert the epoch end to step
+        unsigned long epoch_end_step = time_to_step_idx(ep.tfinal, sample_delta_);
+
```
apeyser: I'd remove this space since the next line is a test against the initialization.

src/ipss_cell.hpp
```diff
@@ -0,0 +1,201 @@
...
+        // If the start time of the neuron is after the end of the epoch
+        if (start_step_ > epoch_end_step) {
+            return;
+        }
+
+        auto end_step = std::min(stop_step_, epoch_end_step);
+
```
apeyser: Same here.

src/ipss_cell.hpp
```diff
@@ -0,0 +1,201 @@
...
+        // epoch is after end of cell
+        if (current_step_ >= end_step) {
+            return;
+        }
+
+        // We are in an active epoch, start stepping!
+        do {
+            // Should we change the rates in the current step?
+
+            if (current_step_ == next_rates_it_->first) {
```
apeyser: Again, are we guaranteed to never fall off the end of this?

src/ipss_cell.hpp
```diff
@@ -0,0 +1,201 @@
...
+            // rate = base rate + nr steps since start rate * delta per step
+            double spike_rate = base_rate_ + (current_step_ - current_rate_it_->first) * rate_delta_step_ ;
+
+            // This dice roll takes 93% of the runtime of this cell.
+            auto dice_roll = distribution_(generator_);
+            if (dice_roll < spike_rate) {
+                spikes.push_back({ { gid_, 0 }, current_step_ * sample_delta_ });
+            }
+
+        } while (++current_step_ < end_step);
```
apeyser: And again, are we guaranteed to not fall off? That we will find end_step in time?

src/ipss_cell.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+    }
+
+    // Small helper function for 'rounding' up to whole multiples of dt
+    // public for testing
+    static  unsigned long time_to_step_idx(time_type time, time_type dt) {
+        unsigned long whole_multiples = static_cast<unsigned long>(time / dt);
```
apeyser: This could be auto

apeyser: Could call `return static_cast<unsigned long>(std::ceil(time / dt))` for the whole function, I think.

src/ipss_cell.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+        return whole_multiples;
+    }
+
+private:
+    /// We need to do some book keeping when we have a rate_change step
+    ///   - Calculate the base rate and the delta per step
+    void update_rates() {
+        // First calculate the rate per time step (if we interpolate)
+        auto steps_this_rate = next_rates_it_->first - current_rate_it_->first;
```
apeyser: I'm not sure that the compiler is smart enough to recognize that if `! interpolate_`, this all collapses down to:
```
rate_delta_step_ = 0;
base_rate_ = current_rate_it_->second
```

So maybe an `if else` is needed here instead.

src/ipss_cell_description.hpp
```diff
@@ -0,0 +1,35 @@
...
+    bool interpolate;
+
+    // rates_per_time: A vector of spike_rates each starting at the supplied time
+    // The first time-rate pair should have a time before the start_time
+    ipss_cell_description(time_type start_time, time_type stop_time, time_type sample_delta,
+        std::vector<std::pair<time_type, double>> rates_per_time, bool interpolate = true) :
+        start_time(start_time),
+        stop_time(stop_time),
+        sample_delta(sample_delta),
+        rates_per_time(std::move(rates_per_time)),
```
apeyser: This took me a second to think -- but I think it's right, that we've constructed a new temporary vector copy and we can take ownership here.

tests/unit/test_ipss_cell_group.cpp
```diff
@@ -0,0 +1,313 @@
...
+            next_rates_ptr++;
+            // new baserate and dt
+            rate_per_dt = (next_rates_ptr->second - rates_ptr->second) /
+                (next_rates_ptr->first - rates_ptr->first);
+            base_rate = interpolate ? rates_ptr->second + 0.5 * rate_per_dt : rates_ptr->second;
+
+            // Normalize to ms!!
+            rate_per_dt = rate_per_dt / 1000.0 * sample_delta;
+            base_rate = base_rate / 1000.0 * sample_delta;
+
```
apeyser: Why all these extra newlines?

tests/unit/test_ipss_cell_group.cpp
```diff
@@ -0,0 +1,313 @@
...
+
+        }
+        double spike_rate = interpolate ? base_rate + (step - rates_ptr->first) * rate_per_dt : base_rate;
+
+        auto dice_roll = distribution(gen);
+
```
apeyser: And more extra new lines

tests/unit/test_ipss_cell_group.cpp
```diff
@@ -0,0 +1,313 @@
...
+        if (dice_roll < spike_rate) {
+
```
apeyser: More...

tests/unit/test_ipss_cell_group.cpp
```diff
@@ -0,0 +1,313 @@
...
+    time_type rate = 10.0;  // Hz
+    time_type sample_delta = 0.10; // 0.1 ms
+    bool interpolate = false;
+
+    // The rate changes we want
+    std::vector<std::pair<time_type, double>> rates_per_time;
+    rates_per_time.push_back({ 0.0, 0.0 });
+    rates_per_time.push_back({ 10.0, rate });
+    rates_per_time.push_back({ 20.0, rate *11 });
+
```
apeyser: You just love white space, doncha?

bcumming:
The cell groups that generate spike sources, i.e. `rss` and `dss` cell types, have been refactored in #496.

They have been replaced with a single `spike_source_cell_group` type. The time points at which spikes are generated are provided by a time sequence, for which we currently have regular, poisson and a vector backed implementation.

The benefit of this approach is that to make a new spike source cell type, you have to supply only the time sequence, which makes testing simpler (you only have to test the time sequence, and not the rest of the cell group infrastructure).

This approach als makes it possible for the user to define a time sequence, like this inhomogeneous poisson sequence, outside of the core library, and build a cell group with it.

PR #430, opened 2018-01-04
Make cell type copyable by default

bcumming:
This PR addresses two fixes:
1. Implement copy constructor for `cell` type
2. Add some sane defaults for `recipe` calls.

### Copying cells

Until now copying of cells required a constructor that took a special clone tag, to force users to explicitly opt-in to copying cells, which may be expensive to copy. The problem with this approach is that interfaces that require copyable types, like `util::any`, don't work with cells. 

This PR removes this restriction, with the understanding that
1. In the future we will be refactoring the cell types to be more amenable to copying cheaply
2. Copying cells has not been a performance problem in the wild: let's cross that bridge when we get to it.

### Recipe defaults

`arb::recipe` is an abstract base class, which required that all of the recipe interface be implemented in classes that inherit from `arb::recipe`. This update adds default implementations for some of the `recipe` interface, to make writing recipes simpler.

The methods with default implementations are:
* `recipe::num_sources(gid)` returns 0                                                                                                                                                                              
* `recipe::num_targets(gid)` returns 0
* `recipe::num_probes(gid)` returns 0
* `recipe::event_generators(gid)` returns an empty list (no generators)
* `recipe::connections_on(gid)` returns empty list (no connections)
* `recipe::get_probe(probe_id)`: throws runtime exception

### DIFF
--- a/example/miniapp/miniapp_recipes.cpp
+++ b/example/miniapp/miniapp_recipes.cpp
@@ -86,7 +86,7 @@ public:
         return ncell_ + 1;  // We automatically add a fake cell to each recipe!
     }
 
-    util::any get_cell_description(cell_gid_type i) const override {
+    util::unique_any get_cell_description(cell_gid_type i) const override {
         // The last 'cell' is a spike source cell. Either a regular spiking
         // or a spikes from file.
         if (i == ncell_) {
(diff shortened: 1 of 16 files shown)


halfflat:
I'm not convinced we should give up our move semantics so readily - discuss off line?

PR #438, opened 2018-01-10
Pybind11 python bindings.

bcumming:
Not for merging in current state.

A proof of concept Python wrapper that uses pybind11. Not feature complete, but it is not far off being useful. It has enough functionality to define a simple recipe and run it as a model.

This PR is fishing for comments. If folk are pleased with the general direction of the work, it could be mended, ammended and extended to a proper PR.

A good spot to start is the draft documentation in `doc/python.rst`.
Then have a look at the simple example in `python/example/rss_recipe.py`.


src/recipe.hpp
```diff
@@ -57,17 +57,17 @@ class recipe {
     virtual cell_size_type num_cells() const = 0;
 
     // Cell description type will be specific to cell kind of cell with given gid.
-    virtual util::unique_any get_cell_description(cell_gid_type gid) const = 0;
+    virtual util::any get_cell_description(cell_gid_type gid) const = 0;
```
halfflat: I still think we should stick to `unique_any` in this interface.

python/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+# Set up pybind11, which is used to generate python bindings.
+# We can just add the pybind11 directory, instead of find_package().
+set(PYBIND11_CPP_STANDARD -std=c++11)
+add_subdirectory(pybind11)
+
+pybind11_add_module(pyarb arb.cpp)
+
+target_link_libraries(pyarb LINK_PUBLIC ${ARB_LIBRARIES})
+target_link_libraries(pyarb LINK_PUBLIC ${EXTERNAL_LIBRARIES})
+
```
halfflat: (This relates also to the use of the global setting of `CMAKE_POSITION_INDEPENDENT_CODE` in the top level CMakeLists.txt.)

We should decide to either have PIC for all libraries, independent of python support, or build the static libraries without it and the dynamic ones with. Getting CMake to do the latter is tedious but possible.

Extra care, ultimately, will have to be taken to use a shared TBB library for `pyarb` if we allow a static TBB library for `libarbor.a`.

python/arb.cpp
```diff
@@ -0,0 +1,235 @@
...
+        };
+    }
+};
+
+std::unique_ptr<spike_recorder> make_spike_recorder(arb::model& m) {
+    auto r = arb::util::make_unique<spike_recorder>();
+    m.set_global_spike_callback(r->callback());
+    return r;
+}
+
```
halfflat: The C++ symbols should probably have static linkage or be in their own (or an anonymous, or the `arb`) namespace.

python/arb.cpp
```diff
@@ -0,0 +1,235 @@
...
+#include "print.hpp"
+#include "recipe.hpp"
+
+namespace pb = pybind11;
+
+template <typename T>
+arb::util::any wrap_any(T value) {
+    return arb::util::any(std::move(value));
+}
```
halfflat: I don't think this is being used…

python/arb.cpp
```diff
@@ -0,0 +1,235 @@
...
+            "Total number of cells in the global model (sum over all domains)")
+        .def_readonly("groups", &arb::domain_decomposition::groups,
+            "Descriptions of the cell groups on the local domain")
+        .def("gid_domain",
+            [](const arb::domain_decomposition& d, arb::cell_gid_type gid) {
+                return d.gid_domain(gid);
+            }, "The domain of cell with global identifier gid.", "gid"_a)
+        .def("__str__",  &domain_decomposition_string)
+        .def("__repr__", &domain_decomposition_string);
+
```
halfflat: Will this ultimately be constructible in python user code? Cf. by-hand construction of decompositions possible with C++ API.

python/arb.cpp
```diff
@@ -0,0 +1,235 @@
...
+        .def("__str__",  &cell_member_string)
+        .def("__repr__", &cell_member_string);
+
+    // spike recording
+    pb::class_<spike_recorder> spike_recorder(m, "spike_recorder");
+    spike_recorder
+        .def(pb::init<>())
+        .def_readonly("spikes", &spike_recorder::spike_store);
+    m.def("make_spike_recorder", &make_spike_recorder);
+
```
halfflat: I think there is the risk of a spike recorder being GCed and the corresponding implicit reference to the spike vector going away while the model is running.

For python purposes, I think it's worth considering having a stack or list of spike sink callbacks, with the corresponding objects deregistering themselves in the event of their destruction (or have reference counting with e.g. `shared_ptr`.) The infrastructure for this could be python-library specific, or we could add it as a feature in the Arbor library that sits on top of the model `set_global_spike_callback` interface. Callbacks could even be python functions.

There is also the question of global vs local spikes for this sort of interface.

python/arb.cpp
```diff
@@ -0,0 +1,235 @@
...
+                     .def("checkpoint", &arb::util::meter_manager::checkpoint);
+
+    // wrap meter_report type such that print(meter_report) works
+    pb::class_<arb::util::meter_report> meter_report(m, "meter_report");
+    meter_report.def("__str__", &meter_report_string)
+                .def("__repr__",&meter_report_string);
+
+    m.def("make_meter_report", &arb::util::make_meter_report,
+          "Generate a meter_report from a set of meters.");
+}
```
halfflat: The `measurement` and `meter_report` classes probably do not need to be exposed; a shim for the `meter_manager` that reports in JSON (using Python's JSON support) would be simpler.

python/example/rss_recipe.py
```diff
@@ -0,0 +1,51 @@
...
+recipe = rss_recipe()
+
+# get_node_info returns a description of the hardware resources available
+# to this process.
+node = arb.get_node_info()
+print(node)
+
+# Perform load balancing. The output will be a domain_decomposition, which
+# describes which cells are to be instantiated on the local domain.
+decomp = arb.partition_load_balance(recipe, node)
```
halfflat: An interface for the partitioner that doesn't require the user to explicitly provide the node information would be more straightforward to use; this might be more feasible when we have easier ways to construct decompositions explicitly.

python/print.hpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include <string>
+#include <sstream>
+
+#include <domain_decomposition.hpp>
+#include <hardware/node_info.hpp>
+#include <profiling/meter_manager.hpp>
+#include <recipe.hpp>
+#include <rss_cell.hpp>
+
```
halfflat: If these are all going to be included in the main source anyhow, I think it would be clearer to have the definition of the to-string functions next to the point of their use in the adaptors.

python/CMakeLists.txt
bcumming: I had to use PIC for the static libraries, otherwise it isn't possible to link the `libarbor.a` into the python module.

Why would we want to use PIC everywhere all the time, instead of just when building Python?

python/arb.cpp
bcumming: Right you are.
It is left over cruft from an earlier version.

python/arb.cpp
bcumming: Sure. I just put them here as a POC of how such a convenience function could be written to help Python users.

python/arb.cpp
bcumming: Well, if the user somehow conspires to have their spike recorder get GC'ed during model execution, they won't be able to use the output.

python/arb.cpp
bcumming: You could make this constructable in Python user code by either:

1. use `pybind11:init<>()` to give it a constructor that can be called from python (and keep the current non-editable once-constructed interface)
2. or make the fields writeable by using `def_readwrite()` instead of `def_readonly`.

It is probably more "pythonic" (I feel a bit silly saying that word) to use option 2, right?

python/example/rss_recipe.py
bcumming: I agree. It is on my todo list to write a version that just takes the recipe, and implicitly determines the node configuration.

python/CMakeLists.txt
halfflat: (Talked about this offline, but commenting for the record.) The issue is one of consistency: building with or without Python shouldn't change the build artefacts, i.e. libarbor.a

python/arb.cpp
halfflat: True, but we should strive to avoid memory corruption or segfaults regardless :)

.ycm_extra_conf.py
```diff
@@ -56,6 +56,10 @@
     'tests/ubench/google-benchmark/include',
     '-I',
     '/cm/shared/apps/cuda/8.0.44/include',
+    '-I',
```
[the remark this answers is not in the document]
bcumming: We can remove it when we have more developers.

It has some hard coded paths, but they are ignored if not found by YouCompleteMe.

python/print.hpp
bcumming: I have them in a seperate file because there is enough code in the `arb.cpp` file, without cluttering it with the print functions. I figured that a good compromise is to have all "stringify" code in one place.

src/merge_events.cpp
```diff
@@ -17,6 +17,21 @@ namespace arb {
 
 namespace impl {
 
+constexpr inline
```
apeyser: Why is this in this PR?

src/recipe.hpp
```diff
@@ -41,11 +41,13 @@ using cell_connection_endpoint = cell_member_type;
 // two in the current code. These two types could well be merged.
 
 struct cell_connection {
-    cell_connection_endpoint source;
-    cell_connection_endpoint dest;
+    cell_connection_endpoint source = {0,0};
```
apeyser: Why these defaults?

python/recipe.hpp
```diff
@@ -0,0 +1,172 @@
...
+    virtual cell_size_type num_targets(cell_gid_type) const {
+        return 0;
+    };
+    virtual std::vector<pybind11::object> event_generators(cell_gid_type gid) const {
+        auto guard = pybind11::gil_scoped_acquire();
+        return {};
+    };
+};
+
+class recipe_trampoline: public recipe {
```
apeyser: So this is the shim to call py from C, right?

apeyser: I'm having trouble understanding why there needs to be two shims -- this one and the one labeled shim which then calls this one.

apeyser: No -- I think I see, this is the shim for python to C, and the one labeled "shim" goes the other way, but they both end up calling python.

bcumming:
I am going to close this now, and create a new PR in a while that introduces the Pybind11 bindings on top of the infrastructure that was added in #458.

bcumming:
Thanks all for comments, I will take them into account when making the new PR.

PR #440, opened 2018-01-15
Event Generator example

bcumming:
Add an example miniapp for event generators.


example/generators/event_gen.cpp
```diff
@@ -0,0 +1,179 @@
...
+    // Create cell with just a single compartment for the soma:
+    //    soma diameter: 18.8 µm
+    //    mechanisms: pas [default params]
+    //    bulk resistivitiy: 100 Ω·cm [default]
+    //    capacitance: 0.01 F/m² [default]
+    //    synapses: 1 * expsyn
+    arb::util::unique_any get_cell_description(cell_gid_type gid) const override {
+        arb::cell c;
+
+        c.add_soma(18.8/2.0);
```
w-klijn: Magic value 2.0

example/generators/event_gen.cpp
```diff
@@ -0,0 +1,179 @@
...
+        // example with one cell and two generators, the seeds are hard-coded.
+
+        // Add excitatory generator
+        gens.push_back(
+            arb::make_event_generator<pgen>(
+                cell_member_type{0,0}, // Target synapse (gid, local_id).
+                w_e,                   // Weight of events to deliver
+                RNG(29562872),         // Random number generator to use
+                t0,                    // Events start being delivered from this time
+                hz_to_freq(500)));     // 50 Hz average firing rate
```
w-klijn: Comment does not fit with the used value

example/generators/readme.md
```diff
@@ -0,0 +1,228 @@
+# Event Generator Example
+
+A miniapp that demonstrates how to describe event generators in a `arb::recipe`.
+
+This miniapp uses a simple model of a single cell, with one compartment
+corresponding to the soma.
+The soma has one synapse, to which two event generators, one inhibitory and one excitatory, are attached.
+
+## The Recipe
```
w-klijn: I would add a list of important concepts introduced in this example

example/generators/readme.md
```diff
@@ -0,0 +1,228 @@
...
+        EXPECTS(gid==0); // There is only one cell in the model
+        return 1;
+    }
+```
+
+### Event Generators
+
+For our demo, we want to attach two Poisson event generators to the synapse on our cell.
+
+1. An excitatory synapse with spiking frequency λ\_e and weight w\_e.
```
w-klijn: What are the units for the frequency and weight?

example/generators/readme.md
```diff
@@ -0,0 +1,228 @@
...
+            arb::make_event_generator<pgen>(
+                cell_member_type{0,0}, w_i, RNG(86543891), t0, lambda_i));
+
+        return gens;
+    }
+```
+
+The `recipe::event_generators(gid)` method returns a vector of `event_generator_ptr` that are attached to the cell with `gid`.
+
+The `event_generator_ptr` is an alias for a `unique_ptr` wrapped around an `event_generator`.
```
w-klijn: I would put this in a separate notes section. It distracts from the story

example/generators/readme.md
```diff
@@ -0,0 +1,228 @@
...
+
+In the implementation, an empty vector is created, and the generators are created and `push_back`ed into the vector one after the other.
+
+The helper function `make_event_generator` is used to simplify the process of creating an event generator and wrapping it in a `unique_ptr`.
+It has the same semantics as [http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique](`std::make_unique`):
+
+* It takes as a template parameter the specialized type of `event_generator`, in this case `pgen`.
+* Then it takes as arguments the arguments to pass to the constructor of `pgen`.
+
+Of the arguments passed to the Poisson event generators, the random number generator state require further explanation.
```
w-klijn: The random number generation warning should be stressed a little more.
For the future (python) we need to make sure the neuroscientist does not have to worry about this.

example/generators/readme.md
```diff
@@ -0,0 +1,228 @@
...
+Each Poisson generator has its own private random number generator state.
+The initial random number state is provided on construction.
+For a real world model, the state should have a seed that is some reproducable hash of `gid` and the generator id, to ensure reproducable random streams.
+For this simple example, we use hard coded seeds to initialize the random number state.
+
+### Sampling Voltages
```
w-klijn: is a nice explanation. It should have its own section in the online doc. Or a link to this because it is an important part of getting info out.

src/model.cpp
```diff
@@ -51,35 +60,42 @@ model::model(const recipe& rec, const domain_decomposition& decomp):
...
-    event_lanes_[0].resize(communicator_.num_local_cells());
-    event_lanes_[1].resize(communicator_.num_local_cells());
+    event_lanes_[0].resize(num_local_cells);
+    event_lanes_[1].resize(num_local_cells);
 }
 
 void model::reset() {
     t_ = 0.;
 
+    // Reset cell group state.
```
w-klijn: No irony implied: Thank you for these comments. Reading the code will take x times as long. So these small sentences speed up understanding the code a lot

src/model.cpp
```diff
@@ -135,28 +149,23 @@ time_type model::run(time_type tfinal, time_type dt) {
...
-                merge_events(
-                    epoch_.tfinal,
-                    epoch_.tfinal+std::min(t_+t_interval, tfinal),
-                    event_lanes(epid)[i],
-                    events[i],
-                    event_generators_[i],
-                    event_lanes(epid+1)[i]);
-            });
+        const auto t0 = epoch_.tfinal;
+        const auto t1 = std::min(tfinal, t0+t_interval);
```
w-klijn: Having these tfinal looking this much alike is an error waiting to happen.
What are the local variables and how they are changed in the different steps is also a little unclear.

src/model.cpp
```diff
@@ -248,28 +279,22 @@ util::optional<cell_size_type> model::local_cell_index(cell_gid_type gid) {
...
-    std::set<cell_size_type> modified_lanes;
+    // Push all events that are to be delivered to local cells into the
+    // pending event list for the event's target cell.
     for (auto& e: events) {
         if (e.time<t_) {
-            throw std::runtime_error("model::inject_events(): attempt to inject an event at time " + std::to_string(e.time) + ", when model state is at time " + std::to_string(t_));
+            throw std::runtime_error(
+                "model::inject_events(): attempt to inject an event at time "
+                + std::to_string(e.time)
+                + ", when model state is at time "
```
w-klijn: I would add that it should not be before/after: The error displays the state, not what is wrong with it.

example/generators/readme.md
bcumming: The creator of a model does have to worry about this, regardless of whether they are in C++ or Python land. We can add some functionality/tools to make this easier, but it is something that users just have to deal with.

example/generators/event_gen.cpp
bcumming: The comment above states that the soma has diameter 18um. The argument is a radius.

example/generators/event_gen.cpp
bcumming: Right you are. I somehow lost changes where I set the frequencies more clearly, like in the docs.

src/model.cpp
bcumming: Pleasure!
The list of tasks in the `reset` method was growing, so it was time to add some comments to make the steps clearer.

src/model.cpp
bcumming: Personally I don't find it too distracting: one is an argument passed to the run method, the other is a member of `epoch`: so long as each variable is clearly defined, the user can figure out what is going on.

The cause of confusion is that local variables, like `tfinal`, are being accessed inside a lambda, which adds an extra layer/scope between variable definition and variable use. We need to wrap the tasks in lambdas to pass them to the threading back end for execution/schedulting. Althernative approaches, like declaring the lamdas directly inline when passing them to the threading back end, or making them stand alone functions complicate code in other ways... I can't see how to make this much clearer.

src/model.cpp
bcumming: I have updated the message to indicate that the problem is that the injected event is before the current model time, which is not reasonable/permitted.

w-klijn:
Github had a delay in displaying my conflict merge...

PR #443, opened 2018-01-24
Parallelize connection generation in communicator constructor

### LINKED ISSUE #442: Parallelize connection generation in communicator
Construction of the `communicator` class is not fully parallelized with threads on each domain.

The loop that queries the recipe for the incoming connections on each local cell is not parallelized, when it could be. As a result, building networks where calls to the `recipe::connections_on(gid)` method are expensive (i.e. when non-trivial calculation is required to generate the connections on a cell) is a serial bottleneck.

Building models could be accelerated by generating the connections in parallel, then performing the less expensive analysis of the connections serially.

bcumming:
Parallelize the loop that collects all of the connections into local cells in the constructor of `communicator`.
This speeds up model building for recipes that have nontrivial computation to perform on calls to `recip::connections_on(gid)`.

Fixes #442.

### DIFF
--- a/src/communication/communicator.hpp
+++ b/src/communication/communicator.hpp
@@ -45,6 +45,7 @@ public:
         using util::make_span;
         num_domains_ = comms_.size();
         num_local_groups_ = dom_dec.groups.size();
+        num_local_cells_ = dom_dec.num_local_cells;
 
         // For caching information about each cell
         struct gid_info {
@@ -52,6 +53,7 @@ public:
             cell_gid_type gid;              // global identifier of cell
             cell_size_type index_on_domain; // index of cell in this domain
             connection_list conns;          // list of connections terminating at this cell
+            gid_info() = default;           // so we can in a std::vector
             gid_info(cell_gid_type g, cell_size_type di, connection_list c):
                 gid(g), index_on_domain(di), conns(std::move(c)) {}
         };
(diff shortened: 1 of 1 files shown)


src/communication/communicator.hpp
```diff
@@ -64,30 +66,37 @@ class communicator {
...
+        threading::parallel_for::apply(0, gids.size(),
+            [&](cell_size_type i) {
+                auto gid = gids[i];
+                gid_infos[i] = gid_info(gid, i, rec.connections_on(gid));
+            });
 
-        cell_local_size_type n_cons = 0;
-        cell_local_size_type n_gid = 0;
+        cell_local_size_type n_cons =
+            util::sum_by(gid_infos, [](const gid_info& g){ return g.conns.size(); }, 0);
```
halfflat: Integer typing here is a bit obscure. Is the `0` added to `sum_by` to force an `int` return value, that then gets promoted to the the unsigned `cell_local_size_type`?

I think it would be clearer to leave out the '0', and let the conversion from `size_t` to `cell_local_size_type` be implicit.

bcumming: Good point, will change!

w-klijn:
I did a clean build only adding a time eater to a recipe and indeed the speedup is as expected.

PR #444, opened 2018-01-26
Value semantics for event_generators

bcumming:
Use type erasure tricks to remove abstract base class for `event_generator`.

This simplifies all the code that uses `event_generator`s (not radically, but it is simpler).

### DIFF
--- a/example/brunel/brunel_miniapp.cpp
+++ b/example/brunel/brunel_miniapp.cpp
@@ -120,8 +120,8 @@ public:
         return cell;
     }
 
-    std::vector<event_generator_ptr> event_generators(cell_gid_type gid) const override {
-        std::vector<arb::event_generator_ptr> gens;
+    std::vector<event_generator> event_generators(cell_gid_type gid) const override {
+        std::vector<arb::event_generator> gens;
 
         std::mt19937_64 G;
         G.seed(gid + seed_);
@@ -131,12 +131,7 @@ public:
         time_type t0 = 0;
         cell_member_type target{gid, 0};
 
-        gens.push_back(arb::make_event_generator<pgen>(
-                        target,
-                        weight_ext_,
-                        G,
-                        t0,
-                        lambda_));
+        gens.emplace_back(pgen(target, weight_ext_, G, t0, lambda_));
         return gens;
     }
 
--- a/example/generators/event_gen.cpp
+++ b/example/generators/event_gen.cpp
@@ -72,7 +72,7 @@ public:
     }
 
     // Return two generators attached to the one cell.
-    std::vector<arb::event_generator_ptr> event_generators(cell_gid_type gid) const override {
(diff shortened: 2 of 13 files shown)


example/generators/event_gen.cpp
```diff
@@ -88,21 +88,19 @@ class generator_recipe: public arb::recipe {
         double w_i = -0.005;
 
         // Make two event generators.
-        std::vector<arb::event_generator_ptr> gens;
+        std::vector<arb::event_generator> gens;
 
         // Add excitatory generator
         gens.push_back(
```
kabicm: I would rather use "emplace_back" here.

example/generators/readme.md
```diff
@@ -104,47 +104,33 @@ The implementation of this with hard-coded frequencies and weights is:
...
-The `recipe::event_generators(gid)` method returns a vector of `event_generator_ptr` that are attached to the cell with `gid`.
-
-The `event_generator_ptr` is an alias for a `unique_ptr` wrapped around an `event_generator`.
-
-```C++
-using event_generator_ptr = std::unique_ptr<event_generator>;
-```
+The `recipe::event_generators(gid)` method returns a vector of `event_generator`s that are attached to the cell with `gid`.
 
 In the implementation, an empty vector is created, and the generators are created and `push_back`ed into the vector one after the other.
```
kabicm: `emplace_back` instead of `push_back` maybe?

src/event_generator.hpp
```diff
@@ -11,38 +11,122 @@
...
+    return postsynaptic_spike_event{cell_member_type{0,0}, max_time, 0};
+}
+
 // An event_generator generates a sequence of events to be delivered to a cell.
 // The sequence of events is always in ascending order, i.e. each event will be
 // greater than the event that proceded it, where events are ordered by:
 //  - delivery time;
 //  - then target id for events with the same delivery time;
 //  - then weight for events with the same delivery time and target.
-struct event_generator {
```
kabicm: I would maybe add a comment explaining that here we use a type erasure method to avoid inheritance.

example/brunel/brunel_miniapp.cpp
```diff
@@ -131,12 +131,7 @@ class brunel_recipe: public recipe {
         time_type t0 = 0;
         cell_member_type target{gid, 0};
 
-        gens.push_back(arb::make_event_generator<pgen>(
-                        target,
-                        weight_ext_,
-                        G,
-                        t0,
-                        lambda_));
+        gens.emplace_back(pgen(target, weight_ext_, G, t0, lambda_));
```
kabicm: This is not so important, but out of curiosity: would constructing target (a pair) in-place with {gid, 0} avoid copying it and make it a tiny more efficient?

bcumming: `cell_member_type` is a POD type, so I don't expect it would make any difference.
http://en.cppreference.com/w/cpp/concept/PODType

PR #445, opened 2018-01-29
put arb_cell_description in arb namespace

bcumming:


kabicm:
It's definitelly a good point that lif_cell description should be added into the arbor namespace! Thanks Ben!

PR #447, opened 2018-02-27
Multithreading friendly profiler

bcumming:
Replace the profiler with a simpler design that works for nested multi-threaded regions.

The old profiler recorded the opening and closing of profiling regions inside a tree structure, which didn't work when regions spawned more asynchronous tasks, e.g. with a call to `threading::parallel_for`. The new approach stores records in a flat data structure, which is post-processed into a tree structure for output and analysis.

All parts of the application that are called in `model::run()` are included in profile regions, and it has been tested with MPI and CUDA back ends.

A simple data structure is used to present the results from the profiler, for which a printer to stdout is implemented. In the future, a method for combining results from accross multiple MPI ranks could be implemented.

Documentation on how to use the profiler is in the Sphynx documentation.


src/profiling/profiler.hpp
```diff
@@ -1,255 +1,56 @@
...
+profile profiler_summary();
+std::size_t profiler_region_id(const char* name);
+void profiler_print(const profile& prof, float threshold=1);
 
-/// iterate and stop them
-void profilers_stop();
+// enter a profiling region
+#define PE(name) \
+    static std::size_t name ## _profile_region_tag__ = arb::util::profiler_region_id(#name); \
+    arb::util::profiler_enter(name ## _profile_region_tag__);
```
halfflat: An alternative to using the region name as the key would be to use the `__COUNTER__` preprocessor macro to generate unique names. It's not standard, but it's quite portable across compilers. This would also obviate the need for the `PR` macro, and we could use a separator like `/` in region names, if that appears more natural.

src/model.cpp
```diff
@@ -134,36 +131,30 @@ time_type model::run(time_type tfinal, time_type dt) {
...
         const auto t0 = epoch_.tfinal;
         const auto t1 = std::min(tfinal, t0+t_interval);
         setup_events(t0, t1, epoch_.id);
-        PL(2);
-
-        PL(2);
     };
 
+    util::profiler_start(); // start the profilers just before time stepping begins
+
```
halfflat: Mentioned elsewhere, but tying the profiler active lifetime to `model::run()` means we can't use the profiler to track e.g. cell construction costs. As argued above, I think we should separate these concerns.

src/profiling/profiler.cpp
```diff
@@ -1,442 +1,401 @@
...
+        std::size_t last = str.find('_');
+        while (last != std::string::npos) {
+            cont.push_back(str.substr(first, last - first));
+            first = last + 1;
+            last = str.find('_', first);
         }
-        is_running_nvprof = false;
+        cont.push_back(str.substr(first, last - first));
+        return cont;
     }
```
halfflat: An alternative would be to make a `std::istringstream` on `str` and use `std::getline()` with the delimiter — no more efficient, but it might read more simply.

src/profiling/profiler.cpp
```diff
@@ -1,442 +1,401 @@
...
+    profile_node(std::string n):
+        name(std::move(n)), time(0), count(npos) {}
+};
 
-void profiler_node::print(std::ostream& stream, double threshold) {
-    // convert threshold from proportion to time
-    threshold *= value;
-    print_sub(stream, 0, threshold, value);
+namespace data {
+    arb::util::profiler profiler;
```
halfflat: I don't think we use this convention for global objects elsewhere. I've a personal preference for a `g_` prefix on the global variable in question, as it's very much in the programmer's face :) We can take this one to bikeshed.

src/profiling/profiler.hpp
```diff
@@ -1,255 +1,56 @@
...
-/// reset profilers
-void profilers_restart();
+// re-enter profiling region
+#define PR(name) \
+    arb::util::profiler_enter(name ## _profile_region_tag__);
 
-/// print the collated profiler to std::cout
-void profiler_output(double threshold, bool profile_only_zero);
+// leave a profling region
+#define PL arb::util::profiler_leave
```
halfflat: The macros should expand to nothing if the profiler is disabled at compile time, or else we'll have spurious calls remain to empty functions that are unlikely to be optimized out.

doc/profiler.rst
```diff
@@ -0,0 +1,255 @@
...
+Compilation
+-----------
+
+There are some non-trivial overheads associated with using the profiler, so it is not enabled by default.
+The profiler can be enabled at compile time, by setting the CMake flag ``ARB_WITH_PROFILING``.
+For example to compile a debug build with profiling turned on:
+
+.. code-block:: bash
+
+    cmake .. ARB_WITH_PROFILING=ON
```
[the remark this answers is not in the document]
bcumming: Yes, yes it will.

src/model.cpp
```diff
@@ -99,7 +97,7 @@ void model::reset() {
     current_spikes().clear();
     previous_spikes().clear();
 
-    util::profilers_restart();
+    util::profiler_restart();
```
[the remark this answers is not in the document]
bcumming: Yes, I agree. This is just a bit of sloppy design on my part.

I had in mind adding profiler callbacks into the setup code in the future. It is now easy to generate "profiler reports", so reports for different phases (setup, advance), and different simulations could be created and saved separately.

src/profiling/profiler.cpp
```diff
@@ -1,442 +1,401 @@
...
+struct profile_accumulator {
+    std::size_t count=0;
+    double time=0.;
+};
+
+// Records the accumulated time spent in profiler regions on one thread.
+// There is one recorder for each thread.
+class recorder {
+    // used to mark that the recorder is not currently timing a region.
+    static constexpr std::size_t npos = std::numeric_limits<std::size_t>::max();
```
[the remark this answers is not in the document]
bcumming: -1 is a very obtuse way of saying the largest possible unsigned 64 bit integer.

src/profiling/profiler.cpp
```diff
@@ -1,442 +1,401 @@
...
-            snprintf(buffer, sizeof(buffer), "%-25s%10.3f%10.1f",
-                            label.c_str(), float(other), percentage);
-            stream << buffer << std::endl;
-        }
+void recorder::enter(std::size_t index) {
+    if (index_!=npos) {
+        throw std::runtime_error("recorder::enter without matching recorder::leave");
+    }
+    index_ = index;
+    start_time_ = timer_type::tic();
```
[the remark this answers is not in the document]
bcumming: Yes, good spot.

doc/profiler.rst
```diff
@@ -0,0 +1,255 @@
...
+
+.. container:: example-code
+
+    .. code-block:: cpp
+
+            // get a profile summary
+            util::profile report = util::profiler_summary();
+
+            // print a summary of the profiler to stdout
+            util::profiler_profiler_print(report);
```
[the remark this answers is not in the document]
bcumming: Printing to `stdout` was a stop gap. `operator<<` makes sense: will do.

I liked the idea of having a profiler summary report available programatically too.

src/profiling/profiler.hpp
```diff
@@ -1,255 +1,56 @@
...
+// enter a profiling region
+#define PE(name) \
+    static std::size_t name ## _profile_region_tag__ = arb::util::profiler_region_id(#name); \
+    arb::util::profiler_enter(name ## _profile_region_tag__);
 
-/// reset profilers
-void profilers_restart();
+// re-enter profiling region
+#define PR(name) \
+    arb::util::profiler_enter(name ## _profile_region_tag__);
```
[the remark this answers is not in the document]
bcumming: Yes, it was added later. I will update the docs.

BTW, it stands for "profiler reenter", so that we can have two regions in the same scope that contribute to the same region (e.g. all regions in a function that work with "events" contribute to an "events" region).

src/profiling/profiler.hpp
bcumming: I am not very comfortable with using `__COUNTER__`, because it is non-standard... but it is tempting.

src/profiling/profiler.cpp
bcumming: To the bikeshed!

src/profiling/profiler.hpp
halfflat: We do also use `#pragma once` :) Not sure though of their respective spans of support across compilers. With the limitation of sticking to one `PE` per line, using `__LINE__` is a possible fallback.

halfflat: Hmm. Everything (including MSVC) on Godbolt supports it, as does xlc v. 13 at least.

src/profiling/profiler.cpp
bcumming: This works, and isn't too complicated!

src/profiling/profiler.cpp
bcumming: Added a static member `static profiler::get_global_profiler()` that returns a reference to a static `profiler`, similarly to the cthread thread pool implementation.l

src/profiling/profiler.cpp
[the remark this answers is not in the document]
bcumming: We could argue over this all day! For the record, I think that if you want the largest unsigned integer, it is best to ask for it explicitly.
For what it is worth, the standard also feels the need to explain itself: `This constant is defined with a value of -1, which because size_t is an unsigned integral type, it is the largest possible representable value for this type.`

src/profiling/profiler.hpp
bcumming: Done.

src/profiling/profiler.hpp
bcumming: I have implemented it. Seems to work pretty well... we can roll it back if it causes problems in the future.

src/profiling/profiler.cpp
```diff
@@ -1,442 +1,401 @@
...
-            children.push_back(n);
-        }
+void recorder::leave() {
+    if (index_==npos) {
+        throw std::runtime_error("recorder::leave without matching recorder::enter");
     }
-
-    value += other.value;
+    accumulators_[index_].count++;
+    accumulators_[index_].time += timer_type::toc(start_time_);
```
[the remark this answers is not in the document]
bcumming: Done. Good point.

src/profiling/profiler.hpp
```diff
@@ -1,255 +1,65 @@
...
-#endif
 
-/// get a reference to the thread private profiler
-/// will lazily create and start the profiler it it has not already been done so
-profiler& get_profiler();
+    // enter a profiling region
+    #define REGION_TAG_NAME(x) x##_profile_region_tag__
+    #define PE(name) \
+        { \
+            static std::size_t REGION_TAG_NAME(name) = arb::util::profiler_region_id(#name); \
```
halfflat: With the new block-based wrapper (which is very cool IMO!) there is no longer a need to use `REGION_TAG_NAME` to derive a unique identifier. Allows also change to e.g. '/' as default separator.

bcumming: Done.

bcumming:
Let's keep the tree data structure "internal" for now. I would wait until we try to extend the profiler output functions, and use the requirements for those extensions to decided its final interface.

PR #448, opened 2018-03-02
refactor git submodule support in cmake

bcumming:
In some places our CMake scripts were attempting to check out git submodules when required, if they have not already been checked out. The code that does this was cut and pasted, and was getting unwieldy.

To minimise the responsibilities of CMake, this PR
* removes calls to git
* introduces a function `check_git_submodule` that can be used to test if a git submodule is installed, and print a helpful message that informs the user how to check it out if needed.
* introduces a function `add_error_target` that makes a target that prints a message then quits with an error. This can be used to generate a proxy target when a problem is detected during CMake setup.  This means that an error is only generated when building a target with a missing dependency, instead of an error during CMake setup.
* refactors the CMake setup for the `docs` and `ubenches` targets to use these new features.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,6 +302,6 @@ if (rtdtheme_avail)
 else()
     add_error_target( docs
         "Generating Sphinx documentation"
-        "The git submodule for the read the docs them needs to be checked out.")
+        "The git submodule for the read the docs is not available")
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -302,6 +302,6 @@ if (rtdtheme_avail)
 else()
     add_error_target( docs
         "Generating Sphinx documentation"
-        "The git submodule for read the docs is not available.")
+        "The git submodule for read the docs is not available")
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,6 +114,12 @@ if(UNWIND_FOUND)
 endif()
 
 #----------------------------------------------------------
+# Git, used to automatically check out submodules if not
+# already checked out.
+#----------------------------------------------------------
+find_package(Git)
+
+#----------------------------------------------------------
 # CUDA support
 #----------------------------------------------------------
 option(ARB_WITH_CUDA "use CUDA for GPU offload" OFF)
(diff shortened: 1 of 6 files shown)


apeyser:
As I said in the pybind reviews, why are we doing this in CMake? And if so, I should at least be able to turn it off.

bcumming:
@apeyser, that is a good question. I found myself asking the same thing when I was making this update.

It was added originally with the micro benchmarks `tests/ubench`, to check out google benchmark if users tried to build the micro benchmarks, which is the only component that requires google benchmark.

The options for such dependencies are:

1. check out submodules automatically if not already checked out. (current approach)
2. make users update the sub-modules manually (or do a recursive checkout).
3. copy library dependencies into the repository (like we do with google test and TCLAP).

The problem with 1 is that it makes the cmake a bit more complicated, but it means that users have an easier time building the code.

The problem with 2 is that users have to muck about with git, which will annoy them. To make the build process user friendly, we would still need to check whether the git submodule has been checked out, and give users a meaningful message, which isn't much less work than also attempting to get the dependency.

The problem with 3: the size of the repo gets much bigger, and updating to new versions of some libraries will be very heavy. For small dependencies like google test and TCLAP, it isn't much bother, but for larger libraries that we are likely to update fairly frequently, not so much.

If I could make one complaint about our current approach, it would be that the first time we run cmake, the submodules for the documencation style and google benchmark are checked out automatically, regardless of whether they are needed. This could be fixed to only check out those dependencies if the user requests the `doc` or `ubench` target respectively. 

@halfflat and @apeyser, feedback?

bcumming:
I have updated the PR so that CMake no longer tries to call git.

If a submodule is not available:
* a helpful warning that tells how to update the submodules is printed when configuring CMake.
* a proxy target is created for targets the require the missing submodule, which will print a useful error message when the user tries to build them.

CMakeLists.txt
```diff
@@ -285,9 +285,23 @@ if(NOT use_external_modcc)
...
 add_subdirectory(tests)
 add_subdirectory(example)
 add_subdirectory(lmorpho)
-add_subdirectory(doc)
+if (rtdtheme_avail)
+    add_subdirectory(doc)
+else()
+    add_error_target( docs
+        "Generating Sphinx documentation"
+        "The git submodule for the read the docs them needs to be checked out.")
```
halfflat: Error message here is a bit garbled.

bcumming: Fix/Ungarbled.

PR #449, opened 2018-03-02
Python setup

bcumming:
Set up the layout of the python wrappers. None of the python interface is implemented here, so that we can focus on the structure and workflow before getting into the nitty gritty of the implementation.

All of the python implementation and examples are in the `python` path:

* `python/pybind11` the pybind11 library is installed as a git submodule in this path.
* `python/example` python example scripts, that implement examples from the documentation go in here.

The CMake build script only includes the `python` path if the user explicitly requests the python interface by setting `ARB_WITH_PYTHON=on`, so that CMake will only look for Python if explicitly requested.

The C++ implementation of the Python wrappers are in the `python` path, and are compiled into `pyarb.so` in `lib`. The module interface is implemented in `python/pyarb.cpp`, which implements a simple "hello world" function as a placeholder for proper content in this PR.

Documentation for the Python wrappers is in `doc/python.rst`, which is put in the `Users` part of the documentation.


vkarak:
Sorry, for the the `jenkins-cscs` spam messages. I'm playing a bit with the CI. Didn't expect to show up sth here...

CMakeLists.txt
```diff
@@ -113,6 +113,12 @@ if(UNWIND_FOUND)
     list(APPEND EXTERNAL_LIBRARIES ${UNWIND_LIBRARIES})
 endif()
 
+#----------------------------------------------------------
```
apeyser: I'm not sure we should do submodule checkouts in CMake.

cmake/GitSubmodule.cmake
```diff
@@ -0,0 +1,36 @@
+
+# Call to ensure that the git submodule in location `path` is loaded.
+# If the submodule is not loaded, the function will attempt to check it
+# out using git.
+# Sets the following variable `name_AVAIL` to `ON` if the submodule is
+# available on completion, or `OFF` otherwise.
+
+function(git_submodule name path)
```
apeyser: At least if we do this, there should be a flag to turn it off.

tests/ubench/CMakeLists.txt
```diff
@@ -26,27 +26,8 @@ set(gbench_cmake_args
...
-        endif()
-    else()
-        set(git_failed "git not found")
-    endif()
-
-    if(git_failed)
-        message(WARNING "Unable to update the google-benchmark submodule: ${git_failed}")
-    endif()
-endif()
+include(GitSubmodule)
```
apeyser: Same with this... I can imagine situations where I don't want to update via git when I run CMake.

CMakeLists.txt
bcumming: I wrote a lengthy response in the other PR.

What do you propose as an alternative here? Should we attempt to detect the `.git` path, and if it is not present, print a user-friendly error message that advises the user on how to get the dependency.

I agree on principle that putting tools like git inside CMake is not really nice. On the other hand, errors caused by not having submodules available should annoy users as little as possible.

cmake/GitSubmodule.cmake
bcumming: Or at least make it only run when the user requests a target that depends on the submodule, which is currently how it works for python, but not for docs or ubenches.

tests/ubench/CMakeLists.txt
bcumming: This attempts to find git, but doesn't cause any errors if it doesn't find it.
How about we then make it only attempt to use git if the required dependency is not found *and* a target that needs a dependency is built, and then print a sane error message if it fails?

CMakeLists.txt
bcumming: The CMake PR has been updated to remove git from CMake.

apeyser:
The git refactor is now merged --- I think we only need to place this on top of that and make sure there's no conflicts in order to merge.

doc/CMakeLists.txt
```diff
@@ -1,41 +1,22 @@
...
-
-endif()
-
 # a static path is required to avoid warning messages from sphinx-build
 file(MAKE_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/static")
 
 # configure target for the sphinx-generated html docs
 find_package(Sphinx)
 
+set(error_msg "Error:")
```
apeyser: Is this merged in right? Why do we need this check, if we have the error target with the same check?

bcumming:
I am closing this because GitHub (or I) has gotten confused.
A new PR will be opened.

PR #451, opened 2018-03-14
Improve TBB vs. CMake

### LINKED ISSUE #332: Issues with CMake vs TBB vs HPC systems.
PR #331 uses a new mechanism for building and linking against TBB. While it solves the build problem for now, there are some deficiencies in the approach which are hard to address, thanks to the combination of CMake generally and TBB inflexibility regarding linking in particular:
  1. The build now requires that the TBB source package be downloaded at configure time, but unrestricted access to the Internet is not always possible on HPC systems.
  2. The forced dynamic linking adds a requirement that the build filesystem and directory be accessible from the execution environment, which again is not guaranteed in systems with separate front end machines and compute nodes.
  3. There is no way currently to instruct the build to use a specific version of TBB that may already be installed or customized for the environment.
  3. It adds a bunch of TBB boilerplate CMake code to the project, which is broadly undesirable.

In principle these can be addressed by spending enough time dicking about with CMake scripts and TBB's `make`-based build process.

bcumming:
This replaces the CMake templates provided by TBB with a much more sane alternative!

The TBB CMake templates had a very strange workflow, that involved downloading the TBB source and compiling it, which made it impossible to configure the TBB build, and caused problems on systems without connection to the internet.

We replace this with a fork of the TBB repository maintained by Github user @wjakob:
https://github.com/wjakob/tbb
This fork provides a sane `CMakeLists.txt` that can be configured from our CMake setup.
It is added as a git submodule, so it can be downloaded with the rest of the repository, hence not requiring connection to the internet during CMake configuration.

It could be extended to use a user-provided build of TBB to use instead of building it.

fixes #332.


.travis.yml
```diff
@@ -17,8 +17,7 @@ env:
     # test single node/rank with different threading back ends
     - BUILD_NAME=serial   WITH_THREAD=serial   WITH_DISTRIBUTED=serial
     - BUILD_NAME=cthread  WITH_THREAD=cthread  WITH_DISTRIBUTED=serial
-    # The tbb backend sometimes fails to build because CMake. Turn it off until our CMake TBB support is stable.
-    #- BUILD_NAME=tbb      WITH_THREAD=tbb      WITH_DISTRIBUTED=serial
+    - BUILD_NAME=tbb      WITH_THREAD=tbb      WITH_DISTRIBUTED=serial
     # test mpi
     - BUILD_NAME=mpi      WITH_THREAD=cthread  WITH_DISTRIBUTED=mpi
```
apeyser: Should we test mpi also with tbb?

bcumming: Positive:
* more test coverage

Neutral:
* we already test the combination of threads + mpi (with threads provided by cthread)

Negative:
* tests take longer to run.

Your call.

apeyser: After looking inside of TBB, I can well imagine weird interactions and breakages --- it's not quite as professional as I'd imagined. I'd go for longer time here.

bcumming: I have added it.

PR #456, opened 2018-03-20
Fix depricated call warnings in CUDA 9

### LINKED ISSUE #379: Fix CUDA 9 warp intrinsics warnings
CUDA 9 introduced more sophisticated warp-level synchronization. Intrinsics like `__ballot` and `__shfl` were depricated to be replaced with more specific instructions like `__shfl_sync`.

We are getting warnings with CUDA 9, e.g.
```
/users/bcumming/repos/arbor/src/backends/gpu/kernels/reduce_by_key.hpp(92):
  warning: function "__ballot"
/cm/shared/apps/cuda-9.0/include/sm_20_intrinsics.h(407):
  here was declared deprecated ("__ballot() is deprecated in favor of __ballot_sync()
  and may be removed in a future release (Use -Wno-deprecated-declarations to suppress
  this warning).")
```

bcumming:
CUDA 9 introduced new, fine-grained, thread synchronization primitives.
In doing so, it introduced new forms of the warp intrinsics like `__shfl_up`, depricating the old symbols in the process.

It will be a while before we can use 9 as the default minimum, so we have to support compilers that expect the new and old behavior.

There are two options: wrap the intrinsics in question, or pass nvcc a flag to not issue warnings about depricated symbols. I go for the approach of wrapping, because I would rather keep the compiler warning turned on.

Fixes #379.

### DIFF
--- a/src/backends/gpu/kernels/reduce_by_key.hpp
+++ b/src/backends/gpu/kernels/reduce_by_key.hpp
@@ -8,6 +8,43 @@ namespace gpu {
 
 namespace impl{
 
+constexpr unsigned mask_all = 0xFFFFFFFF;
+
+// Wrappers around the CUDA warp intrinsics used in this file.
+// CUDA 9 replaced the warp intrinsics with _sync variants, and
+// depricated the old symbols.
+// These wrappers can be removed CUDA 9 becomnse the minimum
+// version used by Arbor.
+
+template <typename T>
+static __device__ __inline__
+T arb_shfl(T x, unsigned lane) {
+#if __CUDACC_VER_MAJOR__ < 9
+    return __shfl(x, lane);
+#else
+    return __shfl_sync(mask_all, x, lane);
+#endif
+}
+
+template <typename T>
+static __device__ __inline__
+unsigned arb_shfl_up(T x, unsigned delta) {
+#if __CUDACC_VER_MAJOR__ < 9
+    return __shfl_up(x, delta);
+#else
+    return __shfl_up_sync(mask_all, x, delta);
+#endif
+}
+
+static __device__ __inline__
+unsigned arb_ballot(int pred) {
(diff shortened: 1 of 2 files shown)


src/backends/gpu/kernels/reduce_by_key.hpp
```diff
@@ -8,6 +8,43 @@ namespace gpu {
 
 namespace impl{
 
+constexpr unsigned mask_all = 0xFFFFFFFF;
```
apeyser: Style question: would it be more readable to just have two regions in this file, a __CUDACC_VER_MAJOR < 9 and not with 3 definitions rather than this line-wise preprocessor testing?

src/backends/gpu/kernels/reduce_by_key.hpp
```diff
@@ -22,6 +59,9 @@ unsigned rounddown_power_of_2(std::uint32_t i) {
 // with two 32 bit shuffles.
 // get_from_lane is a wrapper around __shfl() for both single and double
 // precision.
+// TODO: CUDA 9 provides a double precision version of __shfl that
```
apeyser: Why not know with an #if test? Since anyhow we have a bunch of preprocessor tests for this?

bcumming: Because it is another `#if` test, and I would rather just replace this wholesale when the time comes to use CUDA9.

src/backends/gpu/kernels/reduce_by_key.hpp
bcumming: Not too big of a deal, either way.
Two regions would make for 5 additional lines of code (remove 6, add 11), because the function prototypes would have to be replicated, which makes more room for cut and paste error.
This approach also makes it very clear that the only thing that varies is the intrinsic.

PR #458, opened 2018-03-21
Python setup

bcumming:
Set up the layout of the python wrappers. None of the python interface is implemented here, so that we can focus on the structure and workflow before getting into the nitty gritty of the implementation.

All of the python implementation and examples are in the `python` path:

* `python/pybind11` the pybind11 library is installed as a git submodule in this path.
* `python/example` python example scripts, that implement examples from the documentation go in here.

The CMake build script only includes the `python` path if the user explicitly requests the python interface by setting `ARB_WITH_PYTHON=on`, so that CMake will only look for Python if explicitly requested.

The C++ implementation of the Python wrappers are in the `python` path, and are compiled into `pyarb.so` in `lib`. The module interface is implemented in `python/pyarb.cpp`, which implements a simple "hello world" function as a placeholder for proper content in this PR. 

Documentation for the Python wrappers is in `doc/python.rst`, which is put in the `Users` part of the documentation.


bcumming:
This PR is a fresh attempt at PR #449, which was having issues being merged with the `origin/python` branch used for development of the python backend.

PR #459, opened 2018-03-22
Installation Guide

bcumming:
* Added an installation guide to the Read The Docs
* Removed the outdated build/install information from README.md
* Link from README to Read The Docs
* Updated the splash page for Read The Docs


doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
+    Clang       4.0          Clang 3.8 and later probably work.
+    Apple Clang 9
+    Intel       17.0.1       Needs GCC 5 or later for standard library.
+    =========== ============ ============================================
+
+.. Note::
+    The ``CC`` and ``CXX`` environment variables are used to specify the compiler executable
+    to the CMake build scripts. If these are not set, CMake will automatically choose a compiler.
+    On the system that the test below was performed, if the ``CC`` and ``CXX`` variables are
+    not set, CMake used ``/usr/bin/c++``, which was GCC version 4.8.5.
```
halfflat: The comment is a bit ungrammatical, and without specifying the system, it's oddly vague. Can you say instead that on many systems the default C++ compiler may be too old to compile Arbor, for example, GCC 4.8.5?

doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
 
-To make these docs you also need:
+.. Note::
+    Is is commonly assumed that to get the best performance one should use a vendor-specific
+    compiler (e.g. the Intel, Cray or IBM compilers). These compilers are often better at
+    auto-vectorizing loops, however for everything else GCC and Clang nearly always generate
+    more efficient code.
 
-  * Sphinx
+    The main computational loops in Arbor are based on
```
halfflat: I would say 'generated from' rather than 'based on' here, and say below that this code is vectorized explicitly.

doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
 
-.. _install_desktop:
+.. _building:
 
-Basic Installation
-======================================
+Building Arbor
+==============
 
 Before building an optimzed version for your target system, it is a good idea to build a debug version:
```
halfflat: It's not clear why this is a good idea.

I'd propose stating that:
* Arbor can be built in debug or release mode.
* Debug mode is useful for troubleshooting (link to section).
* Release mode should be used for installation, benchmarking etc.

doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
+              -DARB_VECTORIZE=ON         \
+
+
+.. _threading:
+
+Multi Threading
+---------------
+
+Arbor provides three threading back ends, one of which is selected at compile time.
+by setting the ``ARB_THREADING_MODEL`` CMake option:
```
halfflat: Grammar bit wonky. How about:

>  Arbor provides three possible multithreading implementations. The implementation is selected at configuration time by setting the ``ARB_THREADING_MODEL`` CMake option:

doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
+-----------
+
+Arbor supports NVIDIA GPUs using CUDA. The CUDA back end is enabled by setting the CMake ``ARB_WITH_CUDA`` option.
+
+.. code-block:: bash
+
+    cmake .. -DARB_WITH_CUDA=ON
+
+.. Note::
+    Abor requires CUDA version >= 8, and targets P100 GPUs.
```
halfflat: We should make it clear that earlier GPUs are not yet supported by Arbor.

doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
+    # configure with mpi, tbb threading and compiled with optimizations
+    cmake .. -DARB_DISTRIBUTED_MODEL=mpi \      # Use MPI
+             -DCMAKE_BUILD_TYPE=release  \      # Optimizations on
+             -DARB_THREADING_MODEL=tbb   \      # TBB threading library
+
+    # run unit tests for global communication on 2 MPI ranks
+    mpirun -n 2 ./tests/global_communication.exe
+
+The first step to building with MPI support is to set the ``CC`` and ``CXX`` environment variables to refer to the mpi compiler wrappers.
+
```
halfflat: Out of the box, we don't need to use wrappers (e.g. bog standard Linux build). I think it would be useful to clarify that on some HPC systems one might need to use the wrappers in order to link against the right MPI libraries or whatever.

doc/install.rst
bcumming: done.

doc/install.rst
bcumming: Things are more likely to break and tests are more likely to fail when compiling with optimization on (see our experiences with Power 8 and Arm). I wanted to have a "most likely to work" set of steps that use the least number of flags presented up front.

I agree that this sentence is a misleading, because obviously if you want to use/install Arbor a debug build is a terrible idea.

I have reworded things, and used release mode for that example workflow.

doc/install.rst
bcumming: 95% of the time people use compiler wrappers, particularly if compiling on a cluster.
I have amended things to point out that this example uses compiler wrappers, but that it is perfectly possible to compile without them.

apeyser: Right -- the mpi support in cmake is horrible because it tries to deconstruct the black boxes of the wrappers into cmake data structures by probing them, and then reconstruct cmake command lines that do the same thing as the mpiwrappers would -- obviously impossible in general.

So maybe point out that it is highly suggested to **not** use cmake mpi deduction, but just use the mpi wrapper as the compiler (not said explicitly).

doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
+    # set compiler wrappers
+    $ export CC=`which cc`
+    $ export CXX=`which CC`
+
+Note that the C and C++ compiler wrappers are called ``cc`` and ``CC``
+respectively on Cray systems.
+
+CMake detects that it is being run in the Cray programming environment, which makes
+our lives a little bit more difficult (CMake sometimes tries a bit too hard to help).
```
apeyser: Dry... :smile:

doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
+    mkdir build_modcc
+    cd build_modcc
+
+    # configure and make modcc
+    cmake ..
+    make -j modcc
+
+    # set PATH to find modcc
+    cd ..
+    export PATH=`pwd`/build_modcc/modcc:$PATH
```
apeyser: I use `-Duse_external_modcc=ON -DMODCC_BIN=$MODCC_BIN` instead. Maybe that option should go here as well?

doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
+    from which Arbor generates vectorized code, which can be compiled very efficiently by GCC
+    and Clang.
+    This allows us focus on GCC and Clang for faster compilation times, fewer
+    compiler bugs to work around, and support for recent C++ standards.
+
+.. Note::
+    The IBM xlc compiler versions 13.1.4 and 13.1.6 have been tested for compiling on
+    IBM power 8. Arbor contains some patches to work around xlc compiler bugs,
+    however we do not recommend using xlc because GCC produces faster code,
+    with lower comilation times.
```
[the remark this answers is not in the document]
apeyser: actually, lower comilation times -> smaller compilation times

doc/install.rst
bcumming: Good thing I wrote this instead of @halfflat.

doc/install.rst
```diff
@@ -1,58 +1,655 @@
...
-======================================
+Getting the Code
+================
+
+The easiest way to acquire the latest version of Arbor is to check the code out from
+the `Github repository <https://github.com/eth-cscs/arbor>`_:
+
+.. code-block:: bash
+
+    git clone https://github.com/eth-cscs/arbor.git --recursive
```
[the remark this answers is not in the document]
bcumming: Which genius changed the flag? There are no other arguments to `git clone` that use the word `recursive`.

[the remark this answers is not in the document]
bcumming: I will use `--recurse-submodules` of course, above was just me ranting.

doc/install.rst
bcumming: `use_external_modcc` is intended for this purpose (hence it is not an `option`).
I would rather leave it undocumented. Sam and I have some ideas for how to make cross compilation work "nearly always", just have to get around to implementing it when we have a bit more time.
God I hate CMake.

PR #464, opened 2018-03-29
Pointer/padded vector mechanism state.

halfflat:
As prelude to the SIMD printer work, use padded vectors for data storage on non-GPU mechanisms, and non-owning pointers to shared data in corresponding generated code.

* Merge master branch.
* Change padded_allocator propagation semantics to always propagate alignment/padding on assignment.
* Collect printer sources in modcc under `printer/`.
* Disable non-cpu printers in modcc until implementation.
* Use pointers instead of views in generated mechanisms.
* Move more of the FVM integration step functionality into `shared_state` methods, allowing more flexibility in back-end representations and simplifying the integration calling code.
* Virtualize the `fvm_lowered_cell` interface, and move `mc_cell_group` implementations into own file.
* Split and move multicore back-end functionality into separate sources for independent compilation.
* Split ion species description data (now `ion_info`) from mechanism shared ion state.
* New arb::math:: function `next_pow2` for unsigned integral types.
* New `index_into` implementation that supports bidirectional access (moved to util::).
* Add virtual destructors for virtual base classes.


src/backends/multicore/mechanism.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    auto dest_n = util::size(dest);
+    auto source_n = util::size(source);
+
+    auto n = source_n<dest_n? source_n: dest_n;
+    auto tail = std::copy_n(begin(source), n, begin(dest));
+    std::fill(tail, end(dest), fill);
+}
+
+void mechanism::instantiate(fvm_size_type id, backend::shared_state& shared, const layout& w) {
```
bcumming: `w` isn't very descriptive.
`positions` or `pos_data`?

src/backends/multicore/mechanism.cpp
```diff
@@ -0,0 +1,153 @@
...
+    //
+    // For entries in the padded tail of weight_, set weight to zero.
+    // For indices in the padded tail of node_index_, set index to last valid CV index.
+    // For indices in the padded tail of ion index maps, set index to last valid ion index.
+
+    node_index_ = iarray(width_padded, pad);
+    copy_extend(w.cv, node_index_, w.cv.back());
+
+    weight_     = array(width_padded, pad);
+    copy_extend(w.weight, weight_, 0);
```
bcumming: quick comment here to state why zero specifically is chosen for the weight.

src/backends/multicore/mechanism.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    // Allocate and initialize state and parameter vectors.
+
+    auto fields = field_table();
+    std::size_t n_field = fields.size();
+
+    data_ = array(n_field*width_padded, NAN, pad);
+    for (std::size_t i = 0; i<n_field; ++i) {
+        fvm_value_type*& field_ptr = *std::get<1>(fields[i]);
+
```
bcumming: A comment here about  exactly which pointer is being set here, i.e. in the generated mechanism code, is neccesary.

src/fvm_lowered_cell_impl.hpp
```diff
@@ -0,0 +1,426 @@
...
+
+    // Discrerize mechanism data.
+
+    fvm_mechanism_data mech_data = fvm_build_mechanism_data(*catalogue, cells, D);
+
+    // Create shared cell state.
+    // (SIMD padding requires us to check each mechanism for alignment/padding constraints.)
+
+    unsigned data_alignment = 1;
+    for (auto& m: mech_data.mechanisms) {
```
bcumming: do we have an `util::max` to use with `util::transform_view`?

src/ion.hpp
```diff
@@ -37,116 +31,13 @@ inline std::string to_string(ionKind k) {
...
-
-private:
-    iarray node_index_;
-    array iX_;          // (nA) current
-    array eX_;          // (mV) reversal potential
-    array Xi_;          // (mM) internal concentration
-    array Xo_;          // (mM) external concentration
-    array weight_Xi_;   // (1) concentration weight internal
-    array weight_Xo_;   // (1) concentration weight external
+// Ion (species) description
```
bcumming: Move the comment that explains the legacy NEURON naming scheme for ion species variables to the fvm shared state whither it was refactored.

src/backends/multicore/stimulus.cpp
```diff
@@ -23,19 +21,13 @@ class stimulus: public arb::multicore::mechanism {
...
-        auto vec_i_range = util::make_range(vec_i_, vec_t_+n);
-        auto vec_ci_range = util::make_range(vec_ci_, vec_t_+n);
-
-        auto vec_t = util::indirect_view(util::indirect_view(vec_t_range, vec_ci_range), node_index_);
-        auto vec_i = util::indirect_view(vec_i_range, node_index_);
-
         for (size_type i=0; i<n; ++i) {
-            auto t = vec_t[i];
+            auto cv = node_index_[i];
+            auto t = vec_t_[vec_ci_[cv]];
```
bcumming: Good catch.

halfflat: It was all okay, I think, but the underscore vs no-underscore indirect view wrangling was confusing and unnecessary — it was put in as a quick way to get things working with the pointer representations originally.

PR #471, opened 2018-04-09
Domain decomposition and simulation C++ API docs

bcumming:
Add two new documentation pages for the C++ API
  1. Domain decomposition page that covers `domain_decomposition`, `node_info` and `partition_load_balance`.
  2. Simulation page that describes `arb::simulation` API interface.

Also
  * fixed some small typos elsewhere in the docs.
  * use `std::move` when adding spike callbacks to `arb::simulation` (useful if callbacks are stateful).

### DIFF
--- a/doc/cpp_domdec.rst
+++ b/doc/cpp_domdec.rst
@@ -1,0 +1,180 @@
+Domain Decomposition
+====================
+
+Definitions
+-----------
+
+Domain decomposition
+    A description of the distribution of the model over the available
+    computational resources. In practice, the description partitions the
+    cells in the model as follows:
+
+        * group the cells into *cell groups* of the same kind of cell;
+        * assign each cell group to either a CPU core or GPU on a specific MPI rank.
+
+    The number of cells in each cell group depends on different factors,
+    including the type of the cell, and whether the cell group will run on a CPU
+    core or the GPU.
+
+    See :cpp:class:`arb::domain_decimposition`.
+
+Load balancer
+    A distributed algorithm that determines the domain decomposition using the
+    model recipe and and description of the available computational resources as
+    inputs.
+
+    See :cpp:func:`arb::partition_load_balance`.
+
+Hardware
+--------
+
+.. cpp:namespace:: arb::hw
+
+.. cpp:class:: node_info
+
+    Information about the computational resources available to a simulation, typically a
+    subset of the resources available on a physical hardware node.
+    When used for distributed simulations, where a model will be distributed over more than
+    one node, a :cpp:class:`hw::node_info` represents the resources available to the local
+    MPI rank.
+
+    .. container:: example-code
+
+        .. code-block:: cpp
+
+            // Make node that uses one thread for each available hardware thread,
+            // and one GPU if any GPUs are available.
+            hw::node_info node;
+            node.num_cpu_cores = threading::num_threads();
+            node.num_gpus = hw::num_gpus()>0? 1: 0;
+
+    .. cpp:function:: node_info() = default
+
+        Default hardware description with 1 CPU core and no GPUs.
+
+    .. cpp:function:: node_info(unsigned cores, unsigned gpus)
+
+        Constructor that sets the number of :cpp:var:`cores` and :cpp:var:`gpus` available.
+
+    .. cpp:member:: unsigned num_cpu_cores = 1
+
+        The number of CPU cores available.
+
+        By default it is assumed that there is one core available.
+
+    .. cpp:member:: unsigned num_gpus = 0
+
+        The number of GPUs available.
+
+        By default it is assumed that there are no GPUs.
+
+Load Balancers
+--------------
+
+Load balancing generates a :cpp:class:`domain_decomposition` given a :cpp:class:`recipe`
+and a description of the hardware on which the model will run. Currently Arbor provides
+one load balancer, :cpp:func:`partition_load_balance`, and more will be added over time.
+
+If the model is distributed with MPI, the partitioning of cells is distributed,
+and MPI communication is used to generate the domain decomposition.
+The returned :cpp:class:`domain_decomposition` describes the cell groups
+on the local MPI rank.
+
+.. Note::
+    Often power users know their model and the hardware resources that
+    they are running on very well, and know best how to partition their model
+    workload to get optimal performance.
+    The :cpp:class:`domain_decomposition` type is kept very simple, and is
+    independent of any load balancing algorithm, so power users can supply their
+    own domain decomposition without using one of the built-in load balancers.
+
+.. cpp:namespace:: arb
+
+.. cpp:function:: domain_decomposition partition_load_balance(const recipe& rec, hw::node_info nd)
+
+    Makes an :cpp:class:`domain_decomposition` that distributes the cells
+    in the model described by :cpp:var:`rec` over the hardware resources described
+    by `hw::node_info`.
+
+Decomposition
+-------------
+
+Documentation for the data structures used to describe domain decompositions.
+
+.. cpp:namespace:: arb
+
+.. cpp:enum-class:: backend_kind
+
+    Used to indicate which hardware backend to use for running a :cpp:class:`cell_group`.
+
+    .. cpp:enumerator:: multicore
+
+        Use multicore backend for all computation.
+
+    .. cpp:enumerator:: gpu
+
+        Use gpu back end when supported.
+
+        .. Note::
+            Can only be set if the :cpp:class:`cell_group` type supports the GPU backend.
+
+.. cpp:class:: domain_decomposition
+
+    Meta data that describes a domain decomposition.
+    Solely responsible for describing the distribution of cells across cell groups and domains.
+    A load balancing algorithm generates the domain decomposition,
+    for example see :cpp:func:`partition_load_balance`.
+
+    .. cpp:function:: bool is_local_gid(cell_gid_type gid) const
+
+        Tests whether a gid is on the local domain.
+
+    .. cpp:member:: std::function<int(cell_gid_type)> gid_domain
+
+        Return the domain id of cell with gid.
+        Supplied by the load balancing algorithm that generates the domain
+        decomposition.
+
+    .. cpp:member:: int num_domains
+
+        Number of domains that the model is distributed over.
+
+    .. cpp:member:: int domain_id
+
+        The index of the local domain.
+        Equal to 0 for non-distributed models, and corresponds to the MPI rank for distributed runs.
+
+    .. cpp:member:: cell_size_type num_local_cells
+
+        Total number of cells in the local domain.
+
+    .. cpp:member:: cell_size_type num_global_cells
+
+        Total number of cells in the global model
+        (sum of :cpp:member:`num_local_cells` over all domains).
+
+    .. cpp:member:: std::vector<group_description> groups
+
+        Descriptions of the cell groups on the local domain.
+
+.. cpp:class:: group_decomposition
+
+    The indexes of a set of cells of the same kind that are group together in a
+    cell group in a :cpp:class:`arb::simulation`.
+
+    .. cpp:function:: group_description(cell_kind k, std::vector<cell_gid_type> g, backend_kind b)
+
+        Constructor.
+
+    .. cpp:member:: const cell_kind kind
+
+        The kind of cell in the group.
+
+    .. cpp:member:: const std::vector<cell_gid_type> gids
+
+        The gids of the cells in the cell group, sorted in ascending order.
+
+    .. cpp:member:: const backend_kind backend
+
+        The back end on which the cell group is to run.
+
--- a/doc/cpp_domdec.rst
+++ b/doc/cpp_domdec.rst
@@ -16,11 +16,11 @@ Domain decomposition
     including the type of the cell, and whether the cell group will run on a CPU
     core or the GPU.
 
-    See :cpp:class:`arb::domain_decimposition`.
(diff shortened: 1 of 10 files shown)


doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+    cells in the model as follows:
+
+        * group the cells into *cell groups* of the same kind of cell;
+        * assign each cell group to either a CPU core or GPU on a specific MPI rank.
+
+    The number of cells in each cell group depends on different factors,
+    including the type of the cell, and whether the cell group will run on a CPU
+    core or the GPU.
+
+    See :cpp:class:`arb::domain_decimposition`.
```
halfflat: ytpo

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+
+Load balancer
+    A distributed algorithm that determines the domain decomposition using the
+    model recipe and and description of the available computational resources as
```
halfflat: and and

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+describes the cell groups on the local MPI rank.
+
+.. Note::
+    Power users know their model and the hardware resources that
+    they are running on very well, and know best how to partition their model
+    workload to get optimal performance.
+    The :cpp:class:`domain_decomposition` type is kept very simple, and is
+    independent of any load balancing algorithm, so power users can supply their
+    own domain decomposition without using one of the built-in load balancers.
+
```
halfflat: I wouldn't use the term 'power user' — rather, I think it would be better to state that in cases where the provided load balancers are inadequate, one can provide a `domain_decomposition` object constructed by hand to better suit the model.

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+.. cpp:namespace:: arb
+
+.. cpp:function:: domain_decomposition partition_load_balance(const recipe& rec, hw::node_info nd)
+
+    Construct a :cpp:class:`domain_decomposition` that distributes the cells
+    in the model described by :cpp:var:`rec` over the hardware resources described
+    by `hw::node_info`.
+
```
halfflat: A description of the algorithm it employs to do so would be useful: it is the information a prospective user would use to determine if they are likely to be satisfied with using this balancer, or if they should write their own balancer or decomposition object.

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+    .. cpp:enumerator:: multicore
+
+        Use multicore backend.
+
+    .. cpp:enumerator:: gpu
+
+        Use gpu back end.
+
+        .. Note::
+            Can only be set if the :cpp:class:`cell_group` type supports the GPU backend.
```
halfflat: I'd say 'only meaningful' rather than 'only be set'; it can be set for cell groups without a GPU implementation, but it won't mean anything. (Perhaps we should throw an error in that circumstance.)

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+
+.. cpp:class:: domain_decomposition
+
+    Describes a domain decomposition. Solely responsible for describing the
```
halfflat: I'd join the sentence fragments with 'and is', just for better readability.

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+    distribution of cells across cell groups and domains.
+    Holds cell group descriptions (:cpp:member:`groups`) for cells assigned to
```
halfflat: I'd turn these sentence fragments into complete sentences, e.g. "It holds…" and "The domain decomposition object also has meta-data…".

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+    the local domain, and a helper function (:cpp:member:`gid_domain`) used to
+    look up which domain a cell has been assigned to.
+    Also has meta-data about the number of cells in the global model, and the
+    number of domains over which the model is destributed.
```
halfflat: It's not clear if the object _specifies_ the number of cells and domains, or whether the values are derived elsewhere. As it's actually the latter, we should be clear in stating that these values must correspond to the actual number of cells in the recipe and the number of domains.

Ideally we wouldn't have the redundant specification of information, and we wouldn't need a copy in the domain decomposition object — possible future refactoring of this?

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+
+    A load balancing algorithm generates the domain decomposition, for example see
+    :cpp:func:`partition_load_balance`.
+
+    .. cpp:function:: bool is_local_gid(cell_gid_type gid) const
+
+        Tests whether the cell with :cpp:var:`gid` is on the local domain.
+
```
halfflat: As this is implemented in terms of `gid_domain`, we should state this is a convenience function that performs the `gid_domain(gid)==domain_id` test: it can't be overridden.

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+    .. cpp:member:: std::function<int(cell_gid_type)> gid_domain
+
+        A callback function that returns the domain id that a cell assigned to
+        (using global identifier :cpp:var:`gid`).
+        Set by the load balancing algorithm that generates the domain decomposition.
```
halfflat: While technically a callback, I think it is better couched in terms of supplying a way of querying the (implicit) mapping from cell gids to domains: we don't specify when or how it would be called, and it should be a pure function in that it shouldn't have side-effects and the result should depend solely on the gid parameter.

I don't think there is any need to state that this is supplied by the load balancing algorithm: if the object comes from that, then of course it is, but we also want to make it clear that users are not tied to using that balancer.

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+        for distributed runs.
+
+    .. cpp:member:: cell_size_type num_local_cells
+
+        Total number of cells in the local domain.
+
+    .. cpp:member:: cell_size_type num_global_cells
+
+        Total number of cells in the global model
+        (sum of :cpp:member:`num_local_cells` over all domains).
```
halfflat: That the sum of the `num_local_cells` must equal `num_global_cells` should be described as an (unenforced) constraint or consistency requirement (there is nothing in principle stopping a user from supplying a domain decomposition object for which this is not true.) We should also state that this sum must (for consistency) equal the number of cells as reported by the recipe.

doc/cpp_domdec.rst
```diff
@@ -0,0 +1,186 @@
...
+
+        The kind of cell in the group.
+
+    .. cpp:member:: const std::vector<cell_gid_type> gids
+
+        The gids of the cells in the cell group, **sorted in ascending order**.
+
+    .. cpp:member:: const backend_kind backend
+
+        The back end on which the cell group is to run.
```
halfflat: When discussing consistency, it should be made explicit that the group descriptions are non-overlapping and their gids exhaust the set of local gids.

doc/cpp_simulation.rst
```diff
@@ -0,0 +1,139 @@
...
+        * **Experimental inputs** that can change between model runs, such
+          as external spike trains.
+
+    Simulations provide an interface for executing and interacting with the model:
+
+        * **Advance model state** from one time to another and reset model
+          state to its original state before simulation was started.
+        * **I/O** interface for sampling simulation variables (e.g. compartment voltage
+          and current) and spike output.
+
```
halfflat: 'Simulation variable' is a new concept here which is only vaguely defined. How about an 'interface for monitoring simulation state during execution'?

doc/cpp_simulation.rst
```diff
@@ -0,0 +1,139 @@
...
+    **Types:**
+
+    .. cpp:type:: communicator_type = communication::communicator<communication::global_policy>
+
+        Type used for distributed communication of spikes and global synchronization.
+
+    .. cpp:type:: spike_export_function = std::function<void(const std::vector<spike>&)>
+
+        User-supplied callack function used as a sink for spikes generated
```
halfflat: callback

doc/cpp_simulation.rst
```diff
@@ -0,0 +1,139 @@
...
+    .. cpp:function:: simulation(const recipe& rec, const domain_decomposition& decomp)
+
+    **Experimental inputs:**
+
+    .. cpp:function:: void inject_events(const pse_vector& events)
+
+        Add events directly to targets.
+        Must be called before calling :cpp:func:`simulation::run`, and must contain events that
+        are to be delivered at or after the current simulation time.
+
```
halfflat: I don't think `pse_vector` is described anywhere.

doc/cpp_simulation.rst
```diff
@@ -0,0 +1,139 @@
...
+    **Updating Model State:**
+
+    .. cpp:function:: void reset()
+
+        Reset the state of the simulation to its original state before
+        :cpp:func:`simulation::run` was called.
+
```
halfflat: 'Initial state' is probably sufficient.

doc/cpp_simulation.rst
```diff
@@ -0,0 +1,139 @@
...
+    .. cpp:function:: time_type run(time_type tfinal, time_type dt)
+
+        Run the simulation from current simulation time to :cpp:var:`tfinal`,
+        with maximum time step size :cpp:var:`dt`.
+
+    .. cpp:function:: void set_binning_policy(binning_kind policy, time_type bin_interval)
+
+        Set event binning policy on all our groups.
+
```
halfflat: What a binning policy is is not described. Add a link to a placeholder or stub?

doc/cpp_simulation.rst
```diff
@@ -0,0 +1,139 @@
...
+    .. cpp:function:: void remove_all_samplers()
+
+        Remove all samplers from probes.
+        (see the :ref:`sampling_api` documentation.)
+
+    .. cpp:function:: std::size_t num_spikes() const
+
+        The total number of spikes that occurred since either construction or
+        last resetting the model.
+
```
halfflat: Is this spikes generated, or delivered? I.e. how are spikes 'in flight' counted with respect to this number? We should also clarify that this is a global (across all domains) count.

doc/cpp_simulation.rst
```diff
@@ -0,0 +1,139 @@
...
+    .. cpp:function:: void set_global_spike_callback(spike_export_function export_callback)
+
+        Register a callback that will perform an export of the global spike vector.
+        Will be called on the MPI rank/domain with id 0.
+
+    .. cpp:function:: void set_local_spike_callback(spike_export_function export_callback)
+
+        Register a callback that will perform an export of the rank local spike vector.
+        Will be called on each MPI rank/domain with a copy of the local spikes.
```
halfflat: What the local and global spike vectors are should be described (at least in brief), or have a link to a stub that promises to describe the semantics of distributed spike generation and exchange.

doc/cpp_domdec.rst
bcumming: Good point.

doc/cpp_domdec.rst
bcumming: point.

doc/cpp_domdec.rst
bcumming: Ah, yes, right you are. I thought that you were passively aggresively suggesting that more exposition was required ;)

doc/cpp_domdec.rst
bcumming: I added a brief description and note about potential load balancing problems.

doc/cpp_domdec.rst
bcumming: Fixed.

doc/cpp_domdec.rst
bcumming: Done.

doc/cpp_domdec.rst
bcumming: I am actually quite comfortable with this description: it is just stating what the domain_decomposition represents. Where the values came from ins't really the concern of this description.

bcumming: I have added an extra "note" that lists requirements like `num_global_cells` must be the sum of `num_local_cells`.

doc/cpp_domdec.rst
bcumming: Not actually used in anger anywhere, and a bit redundant, so I have removed it.

doc/cpp_domdec.rst
bcumming: I added a note about these consistency requirements.

doc/cpp_domdec.rst
bcumming: See above.

doc/cpp_simulation.rst
bcumming: I am inclined to leave it for now. The link will automatically be added when the docs for event binning are added.

doc/cpp_simulation.rst
bcumming: I thought about whether to say "global" model. I think that the "global" is redundant: "the model" really means "the model".

I will update to clarify that it is generated spikes.

doc/cpp_simulation.rst
bcumming: I would leave this undefined for now, because to define that I need to define events, which requires defining generic events, time types, ... and so on.

The next documentation to write is about spikes, events and communication. When `pse_vector` is defined somewhere, the link will automatically be added to this part of the documentation.

PR #483, opened 2018-04-27
Complete CUDA back end for the new mechanism infrastructure.

bcumming:
- Add CudaPrinter
    * Use the CPrinter as a base and provide overrides for calls and variables.
    * Reuse the other printer visitors and `cexpr_emit`.
    * Small refactorings to expose some of the CPrinter code to CUDAPrinter
    * Remove the `namespace_components` to simplify code.

- Add std::ostream support for printing gpu `arb::gpu::shared_state`
  (similar to existing printer for multicore shared state).

- Fix bugs in the refactoring of the GPU back end.

- Added `std::ostream` support for `device_view` types.
    * For debugging convenience.


modcc/printer/printerutil.cpp
```diff
@@ -100,3 +100,41 @@ NetReceiveExpression* find_net_receive(const Module& m) {
...
+        data_var=ion_pfx+".reversal_potential";
+        break;
+    case sourceKind::ion_iconc:
+        data_var=ion_pfx+".internal_concentration";
+        break;
+    case sourceKind::ion_econc:
+        data_var=ion_pfx+".external_concentration";
+        break;
+    default:
+        throw compiler_exception("unrecognized indexed data source", sym->location());
```
w-klijn: Would it be an idea to also add  (sym->data_source()) to the exception?

modcc/printer/printerutil.hpp
```diff
@@ -94,9 +92,21 @@ struct module_variables_t {
...
+std::vector<ProcedureExpression*> module_normal_procedures(const Module& m);
+
 // Extract key procedures from module.
 
 APIMethod* find_api_method(const Module& m, const char* which);
 
 NetReceiveExpression* find_net_receive(const Module& m);
 
+struct indexed_variable_info {
+    std::string data_var;
```
w-klijn: Why the var postfix? would data and index suffice ?

src/backends/gpu/mechanism.cpp
```diff
@@ -51,6 +51,8 @@ void mechanism::instantiate(fvm_size_type id, backend::shared_state& shared, con
 
     mechanism_ppack_base* pp = ppack_ptr(); // From derived class instance.
 
+    pp->width_ = width_;
+
```
w-klijn: The 2/3 letter variable name scheme appears to breaking down in this file:
Trying to understand what is going on here I found:
ion_
n_ion
oion
Without really being clear if these are class members or local function members.
I would suggest helping the reader of the code by using consistent informative names -or- source code comments.

modcc/printer/cudaprinter.cpp
```diff
@@ -53,20 +69,21 @@ std::string emit_cuda_cpp_source(const Module& module_, const std::string& ns) {
...
+    out << "\n" << namespace_declaration_open(ns_components) << "\n";
 
     emit_common_defs(out, module_);
 
     out <<
-        "void " << class_name << "_nrn_init_(int, " << ppack_name << "&);\n"
-        "void " << class_name << "_nrn_state_(int, " << ppack_name << "&);\n"
-        "void " << class_name << "_nrn_current_(int, " << ppack_name << "&);\n"
-        "void " << class_name << "_write_ions_(int, " << ppack_name << "&);\n";
+        "void " << class_name << "_nrn_init_(" << ppack_name << "&);\n"
```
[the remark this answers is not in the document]
bcumming: Historical reasons: the mechanism interface in NEURON named these funtions `nrn_state`, `nrn_init` etc, and modcc was originally designed to generate code that might be used by NEURON.
No reason not to change now, but would require a fair few changes.

modcc/printer/printerutil.hpp
bcumming: An "indexed variable" is really two variables when printing: the array (the data variable) and the index (the index variable).
This struct doesn't talk about the data and index, it describes the data variable and index variable that are going to be used in the printed output/
Does that make sense?

modcc/printer/cudaprinter.cpp
```diff
@@ -252,4 +362,106 @@ void emit_common_defs(std::ostream& out, const Module& module_) {
     out << popindent << "};\n\n";
 }
 
+void emit_api_body_cu(std::ostream& out, APIMethod* e, bool is_point_proc) {
+    auto body = e->body();
+    auto indexed_vars = indexed_locals(e->scope());
+
+    if (!body->statements().empty()) {
+        out << "int tid_ = threadIdx.x + blockDim.x*blockIdx.x;\n";
+        out << "if (tid_<n_) {\n" << indent;
```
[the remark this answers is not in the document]
bcumming: We use such comments on closing `}' for namespaces, e.g.`}  // namespace arb`.
The trailining `}` for namespaces can be quite confusing, particularly if you don't indent namespaces.

Here the braces are usually quite close together, and are pretty easy to match up (look for matching `popindent` for each `indent`.)

src/backends/gpu/mechanism.cpp
bcumming: I had a look for the problem, and found the following
  * there is no variable named `ion_` (or `ion`)
  * there is no variable named `n_ion`, There is one named `n_ion_` (a class member because it has the trailing `_`)
  * `oion` is an `util::optional` wrapper, and is used in the 5 lines below where it is returned from a function.

We have standardised practice of using a trailing underscore, e.g. `n_ion_`, to indicate class members, which is being followed in this class.

I renamed `n_ion_` and `n_field` to `num_ions_` and `num_fields` respectively, to make their meaning a bit clearer.

halfflat: I can't argue against greater clarity, but I will assert `n_foo` is a perfectly reasonable way of naming a variable that holds the number of foos.

modcc/printer/cudaprinter.cpp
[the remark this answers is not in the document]
halfflat: As we don't typically indent within the namespace scopes, it becomes more important to clarify where that lonely random closing brace is coming from. One can otherwise though follow it through the indentation, which IMO means it is unnecessary in most other contexts.

modcc/printer/cudaprinter.cpp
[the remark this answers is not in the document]
halfflat: I think we should change them, but in a separate PR.

modcc/printer/cudaprinter.cpp
```diff
@@ -252,4 +362,106 @@ void emit_common_defs(std::ostream& out, const Module& module_) {
...
+                    return true;
+                }
+            }
+            return false;
+        };
+
+        uses_ion(ionKind::K)  && out << "auto kid_  = params_.ion_k_index_[tid_];\n";
+        uses_ion(ionKind::Ca) && out << "auto caid_ = params_.ion_ca_index_[tid_];\n";
+        uses_ion(ionKind::Na) && out << "auto naid_ = params_.ion_na_index_[tid_];\n";
+
```
halfflat: Rather than hard-coding the index names and ion kinds, which will need to be replaced when we start supporting arbitrary ion species, I'd rather use code analogous to that in `cprinter.cpp:emit_simd_api_body()`, which explicitly gathers the indexed variables and their 'index names' as provided by `decode_indexed_variable`. This also covers the `gid_` and `cid_` case (`cid_` is I believe currently unused, as we're still pull `dt` from the cv-indexed variable), and means we don't have to worry about `__attribute__((unused))` etc.

This also obviates the need for the `uses_ion` lambda.

modcc/printer/cudaprinter.cpp
```diff
@@ -252,4 +362,106 @@ void emit_common_defs(std::ostream& out, const Module& module_) {
...
+            case ionKind::none: return "gid_";
+            case ionKind::Ca:   return "caid_";
+            case ionKind::Na:   return "naid_";
+            case ionKind::K:    return "kid_";
+            default :
+                throw compiler_exception("CudaPrinter unknown ion type", s->location());
+        }
+    }
+    return "";
+}
```
halfflat: Related to comment above about using the hard-coded ionKinds, `decode_indexed_variable` provides a variable name for the index as well as the data. This would provide names that are common with the cpu printer, and also will make arbitrary ions easier to implement.

src/backends/gpu/shared_state.hpp
```diff
@@ -121,7 +125,7 @@ struct shared_state {
     void reset(fvm_value_type initial_voltage, fvm_value_type temperature_K);
 };
 
-// For debugging only:
+// For debugging only.
 std::ostream& operator<<(std::ostream& o, const shared_state& s);
```
halfflat: With the `friend` declaration above, this declaration for `operator<<` is redundant and can be removed.

modcc/modcc.cpp
```diff
@@ -254,17 +254,17 @@ int main(int argc, char **argv) {
         // If no output prefix given, use the module name.
         std::string prefix = opt.outprefix.empty()? m.module_name(): opt.outprefix;
 
-        io::write_all(build_info_header(m, "arb"), prefix+".hpp");
+        io::write_all(build_info_header(m), prefix+".hpp");
```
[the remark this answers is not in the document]
bcumming: Reverted.

modcc/printer/cudaprinter.cpp
bcumming: All the code that generates the kernels has been significantly improved while at the same time accomodating the new mechanism infrastructure.
This can be improved in a seperate PR.

modcc/printer/cudaprinter.cpp
bcumming: See comment above.

src/backends/gpu/shared_state.hpp
bcumming: Now I look at it, the `operator<<` doesn't need to be a friend because `shared_state` is a `struct`. So instead I have removed the `friend` declaration from inside the struct.

PR #484, opened 2018-05-02
Mechanism Refactor: multicore and simd.

bcumming:
# Overview

This PR represents a large body of work that refactor the mechanism infrastructure.

The goals of the refactoring:
* Reduce the amount of code generated by modcc for mechanism implementations.
* Reduce the coupling between the mechanism abstraction and the specifics of the FVM discretization, making it possible to provide other discretization schemes, such as for single-compartment cells.
* Split the mechanism metadata/parameter schemata from generated mechanism code.
* Allow mechanisms to be stored in a catalogue as prototypes, independent of any back-end.
* Allow users to separately compile user-supplied mechanisms and incorporate them at run time.

In this patch, the CUDA implementation is incomplete.

## Mechanisms

The `mechanism` class is now a partially-abstract base class with a reduced
public interface:

```C++
virtual mechanism_fingerprint fingerprint() const;
virtual string internal_name() const;
virtual mechanismKind kind() const;
virtual size_t memory() const;
virtual mechanism_ptr clone() const;

virtual void set_global(string param, fvm_value_type value);
virtual void set_parameter(string key, const vector<fvm_value_type>& values);

virtual void nrn_init();
virtual void nrn_state();
virtual void nrn_current();
virtual void deliver_events();
virtual void write_ions();

fvm_size_type mechanism_id() const;
```

For each back-end `B` there is a subclass `concrete_mechanism<B>` which adds the method:
```C++
void instantiate(fvm_size_type id, typename B:shared_state& state, const mechanism::layout& layout);
```
The shared state is a back-end specific class that comprises the data for a cell-group that is commonly read or modified by mechanism instances, including the membrane voltages, ion concentrations, etc.

The `layout` structure is just a pair of vectors, giving the mechanism CV indices and weight contributions.

The common initialization/instantiation code then is the responsibility of a back-end supplied subclass, which also acts as a super-class for the final generated mechanism code. For the multicore/CPU back-end, the hierarchy then is:

`arb::mechanism`
↓
`arb::concrete_mechanism<arb::multicore::backend>`
↓
`arb::multicore::mechanism`
↓
`arb::mechanism_cpu_hh` (generated class)

For the multicore back-end, `arb::multicore::mechanism` manages the instantiation for both scalar and explicit SIMD versions of mechanisms and provides protected members that point into the shared state for the membrane voltage vector, time vector, etc.

The instantiation process requires some introspection into the final generated mechanism code, both to set parameter values in the mechanism and to compute and initialize indices into shared ion state. These are implemented by virtual protected methods that should be overridden as required in the generated mechanism code which provide maps from parameter names or ion kinds to the corresponding field:
```C++
    using global_table_entry = std::pair<const char*, value_type*>;
    using mechanism_global_table = std::vector<global_table_entry>;

    using field_table_entry = std::pair<const char*, value_type**>;
    using mechanism_field_table = std::vector<field_table_entry>;

    using field_default_entry = std::pair<const char*, value_type>;
    using mechanism_field_default_table = std::vector<field_default_entry>;

    using ion_state_entry = std::pair<ionKind, ion_state_view*>;
    using mechanism_ion_state_table = std::vector<ion_state_entry>;

    using ion_index_entry = std::pair<ionKind, iarray*>;
    using mechanism_ion_index_table = std::vector<ion_index_entry>;

    virtual mechanism_field_table field_table();
    virtual mechanism_global_table global_table();
    virtual mechanism_ion_state_table ion_state_table();
    virtual mechanism_ion_index_table ion_index_table();
```
In addition, the generated multicore mechanism must provide implementations for
the generic top-level mechanism methods `clone()`, `fingerprint()`, `internal_name()`, `kind()` and the integration methods `nrn_init()`, `nrn_state()`, `nrn_current()` and `write_ions()`.

The GPU back-end internal interface is similar.

### Mechanism fingerprints

The `mechanism_fingerprint` class and the mechanism `fingerprint()` method is
intended to provide a way to ensure that mechanism metadata and implementations
both derive from the same underlying description. At this point no hash is
performed on the NMODL source to verify fingerprint fidelity — the current
implementation is a placeholder.

### Mechanism catalogue

A mechanism catalogue provides a map from mechanism names both to mechanism metadata and to concrete implementations. A system of mechanism 'derivation' replaces the 'specialized mechanism' system; a derived mechanism shares the same mechanism schema (and by default, the same set of implementations) but overrides one or more of its global parameters.

Two catalogues are provided by the library:
* `builtin_mechanisms()` — mechanisms implemented within the library itself, independent of `modcc`. Currently this contains just the builtin `stimulus` mechanism.
* `global_default_catalogue()` — mechanisms derived from the included NMODL descriptions.

There are two steps to providing a mechanism for a catalogue:
1. Add the mechanism metadata to the catalogue object with  `mechanism_catalogue::add(const string& name, mechanism_info info)`.
2. Register implementations for the mechanism with `mechanism_catalogue::register_implementation(const string& name, unique_ptr<concrete_mechanism<B>> prototype)`.

The catalogue provides `operator[]` to retrieve mechanism metadata, and `instance<B>(name)` to return an instance, if registered, of the mechanism `name` for back-end `B`.

The global default catalogue is generated at build time using a python script `mechanism/generate_default_catalogue`. Generated mechanism sources are stored in the build directory.

## FVM refactor

The functionality provided by `fvm_multicell` has been split between the discretization, provided by `fvm_discretize` and `fvm_build_mechanism_data`(declared in `fvm_layout.hpp`), and the initialization and integration code, now provided by `fvm_lowered_cell_impl<backend>`, which in turn implements the abstract interface `fvm_lowered_cell` (see below).

The cell group class `mc_cell_group` is no longer parameterized on `fvm_lowered_cell<backend>`, but instead invokes methods on the provided `fvm_lowered_cell` pointer, simplifying the build somewhat.

### FVM discretization

Discretization is performed in two phases, independent of the back-end. `fvm_discretize` takes a vector of `cell` objects and returns an `fvm_discretization` object, which comprises all the required information for
segment location to CV mapping, the set up of the voltage solver matrix, and the computation of mechanism weights and parameter values.

`fvm_build_mechanism_data` takes a mechanism catalogue, the same vector of `cell` objects, and the discretization, and then computes the mechanism area contributions, ion default concentration area contributions, mechanism parameter values and synapse–target index mappings.

### FVM lowered cell

The lowered cell now provides a smaller virtualized interface that is presented to the cell group:
```C++
struct fvm_lowered_cell {
    virtual void reset() = 0;
    virtual void initialize(
        const std::vector<cell_gid_type>& gids,
        const recipe& rec,
        std::vector<target_handle>& target_handles,
        probe_association_map<probe_handle>& probe_map) = 0;
    virtual fvm_integration_result integrate(
        fvm_value_type tfinal,
        fvm_value_type max_dt,
        std::vector<deliverable_event> staged_events,
        std::vector<sample_event> staged_samples,
        bool check_physical = false) = 0;
    virtual fvm_value_type time() const = 0;
    virtual ~fvm_lowered_cell() {}
};
```
The `integrate` method constitutes the whole integration loop for one epoch,
and the returned structure provides the spikes and sample data.

The backend-specific implementations `fvm_lowered_cell_impl<backend>` hold the mechanism shared state as provided by the back-end class, stores for sampled values and events, threshold detectors,
and the voltage solver matrix

A signed `fvm_index_type` (alias for `int`) has been added for the representation of indices (such as CV index, ion instance index etc.) to replace `fvm_size_type` in fvm layouts and mechanisms. This was chosen as an alternative to making `unsigned` versions of all our SIMD implementation classes.

### Ions

The public `ion.hpp` header now contains only backend- and mechanism-independent structures for describing ion species: `ion_info` describes ion charge, deault concentrations; the `ionKind` enum is as before. All other ion-related state is relegated to the backend-specific shared-state data.

### GPU backend

Sources for the GPU backend have been grouped by functionality, so that `shared_state.cu` contains the CUDA kernels and wrappers that are used by `shared_state.cpp` and so on.

Most of the original GPU backend code has been refactored to admit the new mechanism infrastructure.

## Cell class

The required user-visible class for cell specification have been moved to `cell.hpp` and `segment.hpp`, specifically: `i_clamp`, `segment_location` and `mechanism_desc` (which replaces `mechanism_spec`: too many things were being called a spec or specification!)

When mechanism descriptions are added to a segment, there is no check done against mechanism metadata, but this could be added (given a catalogue) as an additional feature.

The `cable1d_neuron` global data now comprises:
* The mechanism catalogue to use.
* Default ion concentrations and charges.
* Global temperature (only used for Nernst at the moment).
* Global initial membrane potential.

## Modcc

The command line options of `modcc` have been changed to reflect the new code generation required for the separated mechanism infrastructure and explicit SIMD generation:
* `-s` requests explicitly vectorized code for the cpu target.
* `-S simd-abi-spec` overrides the SIMD ABI class used for the vectorized implementation; the default is `native`.

For a mechanism `foo`, `modcc` will always generate a `foo.hpp` containing the mechanism metadata (parameter names, fingerprint, etc.), and then backend-specific sources as requested:
* `foo_cpu.cpp` is generated for the cpu target.
* `foo_gpu.cu` and `foo_gpu.cpp` are generated for the gpu target.

### Printers

* Collect printer sources in modcc under `printer/`.
* Functionality expected to be common across the printers has been moved to `printerutil.hpp` and `printerutil.cpp`.
* String to file I/O implemented in routines `read_all` and `write_all` in `io/bulkio.hpp`.
* Indent-friendly source code generation is implemented using a `std::streambuf` filter `io::prefixbuf` defined in `io/prefixbuf.hpp`, together with manipulators and a corresponding `std::ostream`-derived wrapper.
* Printers are accessed via top-level functions, e.g. `emit_cpp_source`, `emit_cuda_cu_source` etc. which use visitor classes internally.

The cpu target printer generates both scalar and explicitly vectorized implementations.
The gpu target (CUDA) printer is incomplete, but generates non-functional stubs for the GPU kernels.

### Indexed variables

The distinction between regular `IndexedVariable` expressions and `CellIndexedVariable` expressions has been removed. Indexed variables now explicitly hold an enum `sourceKind` describing what external quantity is being referenced, and it is the responsibility of the printer to provide the appropriate indexing implementation (though at this point, nothing was using the cell-indexed `dt`, only the CV-indexed one.)

Valid source kinds: voltage, current, dt, ion current, ion reversal potential, ion internal concentration, ion external concentration.

State variables can 'shadow' an ion concentration — these are assigned in the generated `write_ions` method.

## SIMD class changes

* Incorporate SIMD module printing into modcc cprinter.
* Add `simd_cast` operation between SIMD value types of the same width, and with  `std::array`. (Note: this was tested and used in an early development version  of the code, but not in this version. It was still a lacuna in the original SIMD wrappers, so it has been left in.)
* Restructure SIMD gather/scatter API to use a `simd::indirect` expression,  which encapsulates a pointer and SIMD offset.
* Add `simd::index_constraint` scoped enum to describe knowledge of contention  in indirect indices, so that we can branch on this to the appropriate  implementation.
* Add SIMD concrete implementation routines `reduce_add` for horizontal reduction and `element0` for access to first lane scalar value.
* Add SIMD value method `sum()` that exposes implementation `reduce_add`.
* Add SIMD concrete implementation routine `compound_indexed_add` that provides  the implementation for `indirect(p, simd_indices) += simd_value` construction.
* Fix SIMD `implbase` bug where some static methods were using the `implbase`  fall-back functions instead of the derived class specialized implementations.
* Move SIMD mathematical functions into friend routines of `simd_impl` in order  to resolve implicit conversions from scalars in mixed SIMD-scalar operations.
* Use a templated `tag` class to dispatch on SIMD concrete implementation types,  to avoid problems with incomplete types in method signatures.
* Remove old SIMD intrinsics.

## CMake

* Downcase some variables in `CMakeLists.txt` files to  distinguish them  visually from CMake keywords and variables.
* Split arbor modcc vectorization option (now `ARB_VECTORIZE`) and  target-architecture optimization (now `ARB_ARCH`).
* For `arbor` and `arbormech` targets, and in particular not the `modcc` target,  use `ARB_ARCH` to generate corresponding target-appropriate binaries,  including, for example, appropriate SIMD support.
* Extend `CompilerOptions.cmake` to map as best as able between the various target architecture names (we use the gcc names) and the correct option to pass to the compiler based on the compiler and platform.
* Add work-around for misidentification by CMake of XL C as Clang.
* As a temporary work-around, include `arbormech` library twice on link line to resolve circular arbor–arbormech dependencies.

## Unit Tests

* Extend repertoire of generic sequence equality/near equality testing support  in `common.hpp`.
* Add warning suppression for icc for the malloc instrumentation code.
* SIMD unit tests for indirect expressions, compound indirect add, reduction.
* Make some exact tests into floating point 'near' tests when comparing computed areas and lengths in swc and fvm layout tests, to account for compiler (e.g. icc) performing semantically inequivalent floating point operation reordering or fusion at `-O3`.
* Split out some of the CUDA tests into separate .cpp/.cu files for  separate-compilation purposes.

## Other

* The `padded_allocator` has been modified to propagate alignment/padding on move and copy (these semantics make their use much easier and safer in the multicore mechanism instantiation code).
* Map/table searching utilities in `util/maputil.hpp`.
* Fixes for correct sequence type categorization and `begin/end` ADL.
* Fixes for type guards for range methods that take universal references.
* Removal of some redundant code in range utilities through the use of universal references.
* Add new range view `reverse_view` for ranges delineated by bidirectional iterators.
* Add single argument form of `make_span` to count up from zero, and associated helper `count_along` that gives a span that indexes a supplied container.
* Moved `prefixbuf` to `modcc` source.
* Make sequence positive and negative tests in algorithms generic.
* Add `private`-subverting helper code/macro to `tests/unit/common.hpp` to reduce the number of public testing-only interfaces in the library code.
* Add virtual destructors for virtual base classes.
* Add new arb::math:: functions: `next_pow2` for unsigned integral types, `round_up` to round a number away from zero to next largest magnitude multiple.
* New `index_into` implementation that supports bidirectional access (moved to `util::` namespace).
* Fix problem in `test_ca.mod` with uninitialized ion current.
* Rework dangerous `memory::array(Iter, Iter)` constructor to be less dangerous (and do the expected thing).
* Allow ranges to be constructed from other ranges if the iterators are compatible.


halfflat:
In the interests of getting the merge in, we'll patch up any SIMD doc issues afterwards.

PR #485, opened 2018-05-07
Runtime distributed context

### LINKED ISSUE #472: Virtualize global communication
Add runtime support for choosing the global communication model (MPI or serial).
The choice of communication model is currently hard coded at compile time.

* `arb::simulation` takes a communication context that provides an interface for performing all ocmmunication tasks.
* Allow creation of communication contexts from a user-supplied MPI communicator.

bcumming:
Move from choosing the distributed communication model from a compile time
choice (the old `arb::communication::communication_policy` type) to a run time
decision.

Previously, communication was handled by static methods on the compile time
`communication_policy`.  Motivations behind the new design include:
  * Increasing flexibility of the library: uses can use a non-distributed
    communication when the library has been compiled with MPI support.
  * Getting Arbor to work better with other libraries by allowing the use of a
    user supplied MPI communicator.

The `arb::global_context` type defines the interface for global communication
operations:
  * defines the methods that must be provided by an implementation;
  * uses type erasure tricks to simplify the implementations. Implementations
    don't have to inherit from `arb::global_context`, and can use templates
    to implement methods that are overloaded on types;
  * Two implementations provided:
    1. `arb::mpi_context`
    2. `arb::serial_context`
    3. TODO: implement dryrun context.

Interfaces to functions and types that use distributed communication now accept
a pointer to their communication context.

Other changes:
  * Remove the `arb::communication` namespace (move to `arb` namespace)
  * Changes to the low level MPI library wrappers include:
    * Add `mpi_error` exception type, derived from `std::runtime_error`.
    * Wrap all MPI API calls to throw `mpi_error` when `MPI_SUCCESS` is not
      returned.
    * Add `arb::mpi::scoped_guard` for RAII handling of `MPI_Init` and
      `MPI_Finalize`.                                                                                                                                                                                               
  * Rewrite all examples and tests to use the new interface.
  * Update documentation for load balancer and simulation, whose interfaces are
    changed.

Fixes #472


halfflat:
Just on naming, I think rather than `global_context`, the class should be called something like `distributed_context`, or anticipating the extension to encompass other facilities, `execution_context`.

halfflat:
This doesn't appear to modify the top-level `CMakeLists.txt`; we should remove the `ARB_DISTRIBUTED_MODEL` option, perhaps replacing it with an `ARB_WITH_MPI` flag to conditionally compile the MPI code.

bcumming:
There are pros and cons of each approach.

I prefer the type erasure approach because it puts all the complexity in one spot (the interface implementation), and may reduce the complexity of calling code.

## OO Approach

### pros

* simpler interface class implementation
* overall design probably easier for new developers to understand (type erasure might be a new idea)

### cons

* implementations become more complicated and more maintenance burden, because each virtual function has to be implemented on each implementation.
* user code may become more complicated by using pointer to base class.                                                       

## Type-erasure Approach

### pros

* simple implementations
* templates on implementations mean that adding support for new types only requires extending interface, not implementations.
* default construction with serial context is simple.
 
### cons

* interface implementation is more complicated.

halfflat:
I don't think that in the OO approach that implementations are more complicated; the gain from the type-erasure approach (with regards to implementation complexity) is that one can use templated implementations directly. But if one is already using a macro-foreach, the difference is only a few lines.
E.g. in an OO implemenation of the MPI context, the reductions look like:
```
struct mpi_context: global_context {
    // ... other methods ...

    #define ARB_MPI_CONTEXT_GENERIC_DEFNS_(X)\
    virtual X min(X x) const override { return mpi::reduce(x, MPI_MIN, comm_); }\
    virtual X max(X x) const override { return mpi::reduce(x, MPI_MAX, comm_); }\
    virtual X sum(X x) const override { return mpi::reduce(x, MPI_SUM, comm_); }\
    virtual std::vector<X> gather(X x, int root) const override { return mpi::gather(x, root, comm_); }

    PP_FOREACH(ARB_MPI_CONTEXT_GENERIC_DEFNS_, ARB_GLOBAL_CONTEXT_SCALARS_)
};
```

Value-semantics for the contexts is convenient, but less so than for our other type-erased classes, as we pass it to methods by pointer regardless.

halfflat:
OO vs type-erasure aside, using a `FOREACH`-like macro can simplify the implementation. Cut-and-paste from proof-of-concept code:
```
// Implementation macros for PP_FOREACH:

#define PP_FOREACH_1_(M, A, ...)  M(A)
#define PP_FOREACH_2_(M, A, ...)  M(A) PP_FOREACH_1_(M, __VA_ARGS__)
#define PP_FOREACH_3_(M, A, ...)  M(A) PP_FOREACH_2_(M, __VA_ARGS__)
#define PP_FOREACH_4_(M, A, ...)  M(A) PP_FOREACH_3_(M, __VA_ARGS__)
#define PP_FOREACH_5_(M, A, ...)  M(A) PP_FOREACH_4_(M, __VA_ARGS__)
#define PP_FOREACH_6_(M, A, ...)  M(A) PP_FOREACH_5_(M, __VA_ARGS__)
#define PP_FOREACH_7_(M, A, ...)  M(A) PP_FOREACH_6_(M, __VA_ARGS__)
#define PP_FOREACH_8_(M, A, ...)  M(A) PP_FOREACH_7_(M, __VA_ARGS__)
#define PP_FOREACH_9_(M, A, ...)  M(A) PP_FOREACH_8_(M, __VA_ARGS__)
#define PP_FOREACH_10_(M, A, ...) M(A) PP_FOREACH_9_(M, __VA_ARGS__)
#define PP_GET_11TH_ARGUMENT_(a1, a2, a3, a4, a5, a6, a7, a8, a9, a10, a11, ...) a11

// Apply macro in first argument to each of the remaining arguments (up to 10).

#define PP_FOREACH(M, ...)\
PP_GET_11TH_ARGUMENT_(__VA_ARGS__, PP_FOREACH_10_, PP_FOREACH_9_, PP_FOREACH_8_, PP_FOREACH_7_, PP_FOREACH_6_, PP_FOREACH_5_, PP_FOREACH_4_, PP_FOREACH_3_, PP_FOREACH_2_, PP_FOREACH_1_)(M, __VA_ARGS__)
```

src/communication/serial_context.hpp
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include <vector>
+
+#include <communication/gathered_vector.hpp>
+#include <spike.hpp>
+
+namespace arb {
+
+struct serial_context {
```
noraabiakar: If the serial_context is for single machine cases and not doing any communication, why not call it single_machine_context?

halfflat: Can we envisage ever having to distinguish between NUMA domains (as relates to the execution context) in a non-MPI run? If so, we might need to be more careful with the naming. I agree though that `serial_context` is probably not the best name.

halfflat: How about `local_context`? 'Local' in these sense of being an antonym of distributed.

src/communication/distributed_context.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+    void barrier() const {
+        impl_->barrier();
+    }
+
+    std::string name() const {
+        return impl_->name();
+    }
+
+    PP_FOREACH(PUBLIC_COLLECTIVES, COLLECTIVE_TYPES);
```
halfflat: Similarly, we should probably rename this macro to `ARB_PP_FOREACH` in order to reduce the chance of macro name collisions.

tests/unit/test_serial_context.cpp
```diff
@@ -0,0 +1,77 @@
...
+    EXPECT_EQ(42u,  ctx.min(42u));
+}
+
+TEST(serial_context, gather)
+{
+    arb::serial_context ctx;
+
+    EXPECT_EQ(std::vector<int>{42}, ctx.gather(42, 0));
+    EXPECT_EQ(std::vector<double>{42}, ctx.gather(42., 0));
+}
```
halfflat: As `std::string` is a special case for gather, we should include that explicitly in the unit test.

tests/global_communication/test_mpi.cpp
```diff
@@ -117,25 +110,24 @@ TEST(mpi, gather_string) {
...
         }
     }
 }
 
 TEST(mpi, gather) {
-    using policy = mpi_global_policy;
-
-    int id = policy::id();
+    mpi_context ctx(MPI_COMM_WORLD);
+    int id = ctx.id();
```
halfflat: This sort of mixes the `mpi` functionality tests with `mpi_context`. Is it possible to split the tests so that `test_mpi` just checks the MPI wrappers, and a different set of tests e.g. `test_mpi_context` tests the `distributed_context` implementation?

src/communication/serial_context.hpp
bcumming: I think we should encourage at least one MPI rank per NUMA domain, because getting NUMA right is hard, and we don't have a pressing reason to not use MPI between NUMA regions.

I have gone with `local_context` and `mpi_context`.

These contexts are quite difficult to name... Sam and I spent quite a while discussing whether "context" is the right word. The next step is to implement similar abstractions for threading, GPUs, and so on. These will all be packed into an "execution context", at which point we can enfore more sensible naming.

tests/global_communication/test_mpi.cpp
bcumming: I removed all `g_context` references from this file. The global context was being used soley to get the rank and number of mpi ranks, so I replaced those calls with `mpi::rank` and `mpi::size`.

There should be seperate unit tests for the `distributed_context` (there were none before, and there are still none now). However it isn't a super high priority, because `mpi_context` is a very thin wrapper around these calls, so these are the key ones to test).

tests/unit/test_serial_context.cpp
bcumming: Done.

src/communication/distributed_context.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+    std::vector<std::string> gather(std::string value, int root) const {
+        return impl_->gather(value, root);
+    }
+
+    private:
```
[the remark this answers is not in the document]
bcumming: Fixed. Vim always trys to indent those.

src/communication/distributed_context.hpp
bcumming: Done.

src/profiling/meter_manager.cpp
```diff
@@ -1,5 +1,5 @@
+#include <communication/distributed_context.hpp>
```
apeyser: Here may be en example where the templating should stay at the current level, but the communicator itself should have a handle that erases the templating.

src/communication/mpi.cpp
```diff
@@ -5,56 +5,73 @@
...
 }
 
-void finalize() {
-    MPI_Finalize();
+// MPI exception class.
+
+mpi_error::mpi_error(const char* msg, int code):
+    error_code_(code)
+{
+    thread_local char buffer[MPI_MAX_ERROR_STRING];
```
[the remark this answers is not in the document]
bcumming: To reduce stack size.

src/communication/mpi.hpp
```diff
@@ -18,229 +18,247 @@
...
+    mpi_error(const char* msg, int code);
+    const char* what() const throw() override;
+    int error_code() const;
+
+private:
+    std::string message_;
+    int error_code_;
+};
+
+struct scoped_guard {
```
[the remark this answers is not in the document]
bcumming: It is in the `mpi` namespace, so it will appear as `arb::mpi::scoped_guard` or `mpi::scoped_guard` wherever it is used.

src/communication/communicator.hpp
```diff
@@ -35,16 +35,17 @@ namespace communication {
...
 
-template <typename CommunicationPolicy>
 class communicator {
 public:
-    using communication_policy_type = CommunicationPolicy;
-
     communicator() {}
 
-    explicit communicator(const recipe& rec, const domain_decomposition& dom_dec) {
+    explicit communicator(const recipe& rec,
```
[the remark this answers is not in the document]
bcumming: Dry run didn't work for models that have more than one cell kind (e.g. multicompartment cells and a spike generator cell).
To work properly it takes a non-trivial amount of refactoring, so I "removed" it.

The dry run will involve the following:
  * implement dry run context, similar to `mpi_context` and `local_context`.
  * either add some smarts to the domain decomposition algorithms + recipe, or add restrictions to recipes to enforce valid dry run models.

src/communication/distributed_context.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+#if defined(ARB_HAVE_MPI)
+#   include "mpi_context.hpp"
+#endif
+#include "local_context.hpp"
+
+
+namespace arb {
+
+#define ARB_PUBLIC_COLLECTIVES_(T) \
```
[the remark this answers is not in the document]
bcumming: This way it is very simple to
  * use a context: because it's interface is type erased.
  * define a new context implementation: see the `mpi_context` and `local_context` implementations, which are very simple and clean, with no need to inherit from or know about the `distributed_context` type.
  * I like this specific type erasure because the user doesn't need to mess around with pointers to base class.

src/profiling/meter_manager.cpp
bcumming: I don't quite follow. Which templates?
This refactoring removes the template from the `communicator` and passes the communicator a handle.
The handle has to perform virtual dispatch at some point: i.e. it has to hold onto pointers to the implementation of the various API methods.
You can do this with an abstract base class (i.e. typical OO C++), or with the type erasure method we use, or with a roll your own vtable.
The `distributed_context` type is the wrapper/handle type. It holds onto the implementation (`mpi_context` or `local_context`), and redirects user calls to their implementation in the "held" implementation.

src/communication/mpi.hpp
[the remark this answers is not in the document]
bcumming: It isn't really intended for users to use it: it is for use in our miniapps, tests and benchmarks.
An application writer will already have their own methods for intitializing and finalizing MPI.

src/communication/communicator.hpp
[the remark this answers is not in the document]
halfflat: I have 'ideas' for robust dry run semantics, but I'll pop them in an issue.

PR #492, opened 2018-05-22
profile multicore mechanism state and current calls individually

bcumming:
The built in profiler generates timings for state and current for individual multicore mechanisms.

Modcc generates and `PE(advance_integrate_{state,current}_X)` profiler calls (along with corresponding `PL()` for calls to multicore mechanism `nrn_state` and `nrn_current` API calls.

No timings are made for the gpu back end, which is not properly supported by the current  profiling tools.

### DIFF
--- a/modcc/printer/cprinter.cpp
+++ b/modcc/printer/cprinter.cpp
@@ -14,6 +14,13 @@ using io::indent;
 using io::popindent;
 using io::quote;
 
+constexpr bool with_profiling() {
+#ifdef ARB_HAVE_PROFILING
+    return true;
+#else
+    return false;
+#endif
+}
 
 void emit_procedure_proto(std::ostream&, ProcedureExpression*, const std::string& qualified = "");
 void emit_simd_procedure_proto(std::ostream&, ProcedureExpression*, const std::string& qualified = "");
@@ -79,8 +86,10 @@ std::string emit_cpp_source(const Module& module_, const std::string& ns, simd_s
         "#include <cstddef>\n"
         "#include <memory>\n"
         "#include <" << arb_header_prefix() << "backends/multicore/mechanism.hpp>\n"
-        "#include <" << arb_header_prefix() << "profiling/profiler.hpp>\n"
         "#include <" << arb_header_prefix() << "math.hpp>\n";
+
+    with_profiling() &&
+        out << "#include <" << arb_header_prefix() << "profiling/profiler.hpp>\n";
 
     if (with_simd) {
         out << "#include <" << arb_header_prefix() << "simd/simd.hpp>\n";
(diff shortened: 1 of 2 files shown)


modcc/printer/cprinter.cpp
```diff
@@ -81,6 +88,9 @@ std::string emit_cpp_source(const Module& module_, const std::string& ns, simd_s
         "#include <" << arb_header_prefix() << "backends/multicore/mechanism.hpp>\n"
         "#include <" << arb_header_prefix() << "math.hpp>\n";
 
+    with_profiling() &&
```
apeyser: Really? Abusing `&&` to form an if? That's very Perl-ish...
Just bite the bullet and add `if (with_profiling())`...

modcc/printer/cprinter.cpp
```diff
@@ -280,11 +290,15 @@ std::string emit_cpp_source(const Module& module_, const std::string& ns, simd_s
     out << popindent << "}\n\n";
 
     out << "void " << class_name << "::nrn_state() {\n" << indent;
+    with_profiling() && out << "PE(advance_integrate_state_" << name << ");\n";
```
apeyser: And...

src/fvm_lowered_cell_impl.hpp
```diff
@@ -203,11 +203,9 @@ fvm_integration_result fvm_lowered_cell_impl<Backend>::integrate(
 
         // Integrate mechanism state.
 
-        PE(advance_integrate_state);
```
apeyser: We don't want to measure this any more?

apeyser: Ah, I see, this is supposed to come out of modcc now

modcc/printer/cprinter.cpp
halfflat: It makes things line up better. Ultimately this is my fault for using the same construction elsewhere in the modcc printer source. (But I still think it looks prettier.)

src/fvm_lowered_cell_impl.hpp
bcumming: Yes. We only mark "leaf" nodes in the profiling tree, and the leaf nodes are now inside the calls to `nrn_current` and `nrn_state`

modcc/printer/cprinter.cpp
bcumming: I wouldn't want to see this pattern used anywhere else, but it is fairly useful for these stream heavy operations in the modcc printers.

apeyser: Okay then -- maybe we put a comment that Sam is very naughty and lesser beings are not to follow his bad example.

bcumming: The pattern is used throughout the printers: so I don't think that there is a single sensible location to put the comment that will be seen by all unfortunate souls that wander into the printers.

It is a bit of an odd idiom, but used consistently throughout the printers, so I think readers might just have to "pick it up".

PR #494, opened 2018-05-28
Simd partition by constraint

noraabiakar:
Changes have been made to the simd implementation of mechansim functions: 

- The node_index array (array of indices that specifies for each mechanism the CVs where it is present), is now partitioned into 4 arrays according to the constraint on each simd_vector in node_index:
    1. contiguous array: contains the indices of all simd_vectors in node_index where the elements in simd_vector are contiguous
    2. constant array: contains the indices of all simd_vectors in node_index where the elements in simd_vector are identical
    3. independent array: contains the indices of all simd_vectors in node_index where the elements in simd_vector are independent (no repetitions) but not contiguous 
    4. none array: contains the indices of all simd_vectors in node_index where the none of the above constraints apply

    When mechanism functions are executed, they loop over each of the 4 arrays separately. This allows for optimizations in every category. 

- The modcc compiler was modified to generate code for the previous changes, including the optimizations per constraint:
    1. contiguous array: we use vector load/store and vector arithmetic. 
    2. constant array: we load only one element and broadcast it into a simd_vector; we use vector arithmetic; we reduce the result; we store one element.   
    3. indepndent array: we use vector scatter/gather and vector arithmetic. 
    4. none array: we cannot operate on the simd_vector in parallel, we loop over the elements to read, perform arithmetic and write back 

- Added a mechanism benchmark for pas, hh and expsyn

- Moved/modified some functions in simd.hpp to ensure that the correct implementation of a function is being called.

### DIFF
--- a/src/backends/multicore/mechanism.cpp
+++ b/src/backends/multicore/mechanism.cpp
@@ -19,6 +19,7 @@
 #include <backends/multicore/mechanism.hpp>
 #include <backends/multicore/multicore_common.hpp>
 #include <backends/multicore/fvm.hpp>
+#include <backends/multicore/partition_by_constraint.hpp>
 
 namespace arb {
 namespace multicore {
@@ -127,6 +128,7 @@ void mechanism::instantiate(fvm_size_type id, backend::shared_state& shared, con
     node_index_ = iarray(width_padded_, pad);
     copy_extend(pos_data.cv, node_index_, pos_data.cv.back());
     copy_extend(pos_data.weight, make_range(data_.data(), data_.data()+width_padded_), 0);
+    gen_constraint(node_index_, constraint_index_);
 
     for (auto i: ion_index_table()) {
         util::optional<ion_state&> oion = value_by_key(shared.ion_data, i.first);
--- a/src/backends/multicore/mechanism.cpp
+++ b/src/backends/multicore/mechanism.cpp
@@ -128,7 +128,7 @@ void mechanism::instantiate(fvm_size_type id, backend::shared_state& shared, con
     node_index_ = iarray(width_padded_, pad);
     copy_extend(pos_data.cv, node_index_, pos_data.cv.back());
     copy_extend(pos_data.weight, make_range(data_.data(), data_.data()+width_padded_), 0);
-    gen_constraint(node_index_, constraint_indices_, width_);
+    generate_index_constraint_partitions(node_index_, index_constraints_, width_);
 
     for (auto i: ion_index_table()) {
         util::optional<ion_state&> oion = value_by_key(shared.ion_data, i.first);
@@ -143,7 +143,7 @@ void mechanism::instantiate(fvm_size_type id, backend::shared_state& shared, con
         ion_index = iarray(width_padded_, pad);
         copy_extend(indices, ion_index, util::back(indices));
 
-        if(!compatible_constraint_indices(node_index_, ion_index))
+        if(!compatible_index_constraints(node_index_, ion_index))
             throw std::runtime_error("ion_index and node_index not compatible");
     }
 
--- a/src/backends/multicore/mechanism.cpp
+++ b/src/backends/multicore/mechanism.cpp
@@ -128,7 +128,7 @@ void mechanism::instantiate(fvm_size_type id, backend::shared_state& shared, con
     node_index_ = iarray(width_padded_, pad);
     copy_extend(pos_data.cv, node_index_, pos_data.cv.back());
     copy_extend(pos_data.weight, make_range(data_.data(), data_.data()+width_padded_), 0);
-    gen_constraint(node_index_, constraint_index_);
+    gen_constraint(node_index_, constraint_indices_, width_);
 
     for (auto i: ion_index_table()) {
         util::optional<ion_state&> oion = value_by_key(shared.ion_data, i.first);
@@ -142,6 +142,9 @@ void mechanism::instantiate(fvm_size_type id, backend::shared_state& shared, con
         auto& ion_index = *i.second;
         ion_index = iarray(width_padded_, pad);
         copy_extend(indices, ion_index, util::back(indices));
+
+        if(!compatible_constraint_indices(node_index_, ion_index))
+            throw std::runtime_error("ion_index and node_index not compatible");
     }
 
 }
--- a/src/backends/multicore/mechanism.cpp
+++ b/src/backends/multicore/mechanism.cpp
@@ -27,6 +27,9 @@ namespace multicore {
 using util::make_range;
 using util::value_by_key;
 
+constexpr unsigned simd_width = S::simd_abi::native_width<fvm_value_type>::value;
+
+
 // Copy elements from source sequence into destination sequence,
 // and fill the remaining elements of the destination sequence
 // with the given fill value.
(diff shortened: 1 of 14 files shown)


tests/unit/test_partition_by_constraint.cpp
```diff
@@ -0,0 +1,187 @@
...
+    iarray input_index(input_size_);
+    iarray expected_indices_contiguous, expected_indices_constant,
+            expected_indices_independent, expected_indices_serial;
+    multicore::constraint_partitions output_constraint;
+
+
+    const int c = 5;
+//shuffle here
+    for (unsigned i = 0; i < input_size_; i++) {
+        input_index[i] = (i < input_size_ / 4 ? i / ((simd_width_ + 1)/ 2) :
```
bcumming: This ternary expression can be made clearer by removing some parenthesis `()` and realignment, e.g.:
```
        input_index[i] = i<input_size_/4   ? i/((simd_width_ + 1)/ 2): 
                         i<input_size_/2   ? i*2:
                         i<input_size_*3/4 ? c:
                         i;
```

tests/unit/test_partition_by_constraint.cpp
```diff
@@ -0,0 +1,187 @@
...
+            expected_indices_independent.push_back(i);
+        else if (i < input_size_* 3/ 4 && i % simd_width_ == 0)
+            expected_indices_constant.push_back(i);
+        else if (i % simd_width_ == 0)
+            expected_indices_contiguous.push_back(i);
+    }
+
+    multicore::generate_index_constraint_partitions(input_index, output_constraint, input_size_);
+
+    if (simd_width_ != 1) {
```
bcumming: Like I noted above, you can probably simplify the body of this if statement to
```C++                                                      
EXPECT_EQ(expected_indices_contiguous, output_constraint.contiguous);
EXPECT_EQ(expected_indices_constant, output_constraint.constant);
EXPECT_EQ(expected_indices_independent, output_constraint.independent);
EXPECT_EQ(expected_indices_serial, output_constraint.none); 
```                                                         
Which could be make easier to read by using shorter variable names:                                      
```C++                                                      
EXPECT_EQ(expected_contiguous,  output.contiguous);
EXPECT_EQ(expected_constant,    output.constant);
EXPECT_EQ(expected_independent, output.independent);
EXPECT_EQ(expected_serial,      output.none);
```

src/backends/multicore/mechanism.cpp
```diff
@@ -27,6 +27,9 @@ namespace multicore {
 using util::make_range;
 using util::value_by_key;
 
+static constexpr unsigned simd_width_ = S::simd_abi::native_width<fvm_value_type>::value;
```
bcumming: We use the trailing underscore `name_` for member variables of classes.
The following should be sufficient:
```
constexpr unsigned simd_width = S::simd_abi::native_width<fvm_value_type>::value;
```
Note that you don't need `static` to restrict the visibility of a `constexpr` variable in this context.

src/backends/multicore/partition_by_constraint.hpp
```diff
@@ -18,86 +18,97 @@ struct constraint_partitions {
...
-               (node_constraint == index_constraint::independent &&
-                ion_constraint == index_constraint::contiguous))) {
-                return false;
-            }
+bool compatible_index_constraints(T& node_index, T& ion_index, unsigned simd_width){
+    for (unsigned i = 0; i < node_index.size(); i+= simd_width) {
+        auto ptr_n = &node_index[i];
+        auto ptr_i = &ion_index[i];
+
+        auto nc = idx_constraint(ptr_n, simd_width);
```
bcumming: I think that it is probably clearer to be more direct:
```C++
auto nc = idx_constraint(&node_index[i], simd_width);
auto ic = idx_constraint(&ion_index[i], simd_width);
```

src/fvm_lowered_cell_impl.hpp
```diff
@@ -56,6 +56,10 @@ class fvm_lowered_cell_impl: public fvm_lowered_cell {
 
     value_type time() const override { return tmin_; }
 
+    std::vector<mechanism_ptr>& mechanisms() {
```
bcumming: Does this have to be a non-const reference, i.e. can we make the return type `const std::vector<mechanism_ptr>&`?

A con-const reference would allow a users to modify the contents of the class. This isn't necessarily a massive problem, because this is part of our internal API, but it still good to avoid.

src/simd/simd.hpp
```diff
@@ -220,37 +212,37 @@ namespace simd_detail {
...
 
-                    scalar_type a[width];
-                    Impl::copy_to(s, a);
+                        scalar_type a[width];
+                        Impl::copy_to(s, a);
 
-                    for (unsigned i = 0; i<width; ++i) {
-                        p[o[i]] += a[i];
+                        for (unsigned i = 0; i<width; ++i) {
+                         p[o[i]] += a[i];
```
bcumming: indentation here is a bit off.

src/simd/simd.hpp
```diff
@@ -220,37 +212,37 @@ namespace simd_detail {
...
+                        Impl::copy_to(v, p);
+                    }
+                    break;
                 case index_constraint::constant:
-                {
-                    p += ImplIndex::element0(index);
-                    *p += Impl::reduce_add(s);
-                }
-                break;
+                    {
```
bcumming: these braces are not needed on this `case`

tests/ubench/mech_vec.cpp
```diff
@@ -12,6 +12,18 @@ using namespace arb;
 using backend = arb::multicore::backend;
 using fvm_cell = arb::fvm_lowered_cell_impl<backend>;
 
+int find_mechanism(const std::string &mech_name, fvm_cell &cell) {
```
bcumming: The error handling code is the same for each benchmark (i.e. print error message and `exit(1)`), so I would put it in this function, and also return the `mechanism_ptr` directly.
```C++
mechanism_ptr& find_mechanism(const std::string& name, fvm_cell& cell) {
    auto &mechs = cell.mechanims();
    auto it = std::find_if(mechs.begin(),
                           mechs.end(),
                           [&](mechanism_ptr& m){return m->internal_name()==name;});
    if (it==mechs.end()) {
        std::cerr << "couldn't find mechanism with name " << name << "\n";
        exit(1);
    }   
    return *it;
}

// which can be called directly
auto& m = find_mechanism("hh", cell);
```
I haven't tested this, but you get the idea!

src/fvm_lowered_cell_impl.hpp
bcumming: I see that it does have to be for the benchmark. Maybe we should put a comment above this line that points out that these are exposed for testing purposes only.

tests/ubench/mech_vec.cpp
noraabiakar: That makes a lot of sense.

src/simd/simd.hpp
```diff
@@ -158,14 +159,102 @@ namespace simd_detail {
 
         template <typename IndexImpl, typename = typename std::enable_if<width==simd_traits<IndexImpl>::width>::type>
         void copy_from(indirect_expression<IndexImpl, scalar_type> pi) {
-            value_ = Impl::gather(tag<IndexImpl>{}, pi.p, pi.index);
+            switch (pi.constraint) {
```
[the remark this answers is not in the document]
halfflat: And just on this style issue, following a principle of indenting one level per scope (with a special exception for namespaces) and having labels de-indented by one level, `switch` indenting becomes:
```
switch (foo) {
case bar:
    do_thing();
    break;
case xyzzy:
case quux:
    do_other_thing();
}
```
and so on.

PR #496, opened 2018-05-31
generalize time sequences

bcumming:
# Changes to libarbor                                                                                                                                                                                               

Time sequences were added in `src/time_sequence.hpp`:
- added new `time_seq` type that implements a type-erasure interface for the
  concept of a time sequence generator.
- added poisson, regular and vector-backed implementations of the time sequence
  concept.

Event generators:
- The poisson, regular and vector-backed implementations of the event generator
  concept were refactored to use the.

Cell groups:
- Removed the `dss_cell_group` and `rss_cell_group` and associated types.
- Added a generic spike source cell  that generates a sequence of spikes
  at time points specified by a `time_seq`. Using this approach, an
  additional `cell_group` specialization is not required for each type of
  sequence, and user-defined sequences can be used with minimal overhead.

# Unit tests

- Added unit tests for `time_seq`.
- Simplified `event_generator` unit tests, because much of the testing
  of the sequences was moved to the `time_seq` tests.
- Added unit tests for `spike_source_cell_group`.

# Changes to miniapp

- simplified the miniapp by removing the command line options for using an input spike chain from file.
- updated the miniapp recipe to use `spike_source` cell group instead of `dss_cell_group`.


src/time_sequence.hpp
```diff
@@ -0,0 +1,251 @@
...
+    vector_time_seq(std::vector<time_type> seq):
+        seq_(std::move(seq))
+    {
+        // Ensure that the time values are sorted.
+        if(!std::is_sorted(seq_.begin(), seq_.end())) {
+            util::sort(seq_);
+        }
+        reset();
+    }
+
```
halfflat: I would like to see this either be an `EXPECTS` test, or to make the semantics explicit: that the vector time sequence returns the next time in the vector that is not strictly less than the previous time. The latter has the benefit of being robust to user input without requiring an initial traversal over the whole vector.

src/spike_source_cell_group.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <time_sequence.hpp>
+
+namespace arb {
+
+spike_source_cell_group::spike_source_cell_group(std::vector<cell_gid_type> gids, const recipe& rec):
+    gids_(std::move(gids))
+{
+    time_sequences_.reserve(gids_.size());
+    for (auto gid: gids_) {
+        time_sequences_.push_back(util::any_cast<time_seq>(rec.get_cell_description(gid)));
```
halfflat: Because a `time_seq` is more general in application than a spike source 'cell', and because they are essentially an implementation detail for the description of a spike source, I think they should have a (trivial) wrapper, e.g. something like
```
struct spike_source {
    explicit spike_source(time_seq times): times_(std::move(times)) {}
    time_seq times_;
};
```
It keeps the recipe usage consistent — each cell group kind has its own description class — and helps the code document itself.

src/event_generator.hpp
```diff
@@ -16,6 +17,11 @@ postsynaptic_spike_event terminal_pse() {
     return postsynaptic_spike_event{cell_member_type{0,0}, max_time, 0};
 }
 
+inline
+bool is_terminal_pse(const postsynaptic_spike_event& e) {
+    return e.time==max_time;
```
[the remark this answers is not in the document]
bcumming: As tempting as it is to use `end_of_all_time` (and the test `is_end_of_all_times()`), I am going to change it to `terminal_time`.

src/spike_source_cell_group.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+void spike_source_cell_group::advance(epoch ep, time_type dt, const event_lane_subrange& event_lanes) {
+    PE(advance_sscell);
+
+    for (auto i: util::make_span(0, gids_.size())) {
+        auto& tseq = time_sequences_[i];
+        const auto gid = gids_[i];
+
+        while (tseq.next()<ep.tfinal) {
```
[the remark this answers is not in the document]
bcumming: Done. I changed it to `front`.

src/spike_source_cell_group.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+cell_kind spike_source_cell_group::get_cell_kind() const {
+    return cell_kind::spike_source;
+}
+
+void spike_source_cell_group::advance(epoch ep, time_type dt, const event_lane_subrange& event_lanes) {
+    PE(advance_sscell);
+
+    for (auto i: util::make_span(0, gids_.size())) {
```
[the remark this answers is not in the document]
bcumming: Thanks, I did not know this.

src/time_sequence.hpp
bcumming: The later would be more robust to user input, however it would not give what I feel would be the expected behavior for the user.

This is a user facing interface, and I feel that sending in unsorted values is the sort of error we should reasonably expect from users. In which case I am not comfortable modifying our semantics to avoid testing user inputs. I think we just have to do the test.

src/spike_source_cell_group.cpp
bcumming: Done. There is now a `spike_source_cell` wrapper kind.

src/time_sequence.hpp
halfflat: I'd still prefer then to throw than sort, and have the name be e.g. `checked_vector_time_seq` to make it clear. Or else move validation closer to the user, e.g. in `vector_backed_generator` and `spike_source_cell`.

src/spike_source_cell_group.hpp
```diff
@@ -6,6 +6,10 @@
 
 namespace arb {
 
+struct spike_source_cell {
+    time_seq seq;
+};
+
 class spike_source_cell_group: public cell_group {
```
halfflat: I can postpone this to the installable-target work, but it would be good to have the public-facing class (`spike_source_cell`) in a separate header from the library implementation class (`spike_source_cell_group`).

halfflat:
So, I'd like still to rename (and possibly promote outside the class) `time_seq::dummy_seq`, and still arguing about `vector_time_seq`. Other issue about splitting `spike_source_cell` out from `spike_source_cell_group.hpp` I can do later. 

`tests/unit/test_rss_cell_group.cpp` and `tests/unit/test_dss_cell_group.cpp` should be removed, too.

PR #500, opened 2018-06-04
Benchmark cell type

### LINKED ISSUE #493: Add benchmark cell type
Add a cell type to the model that can be used for benchmarking.

A benchmark cell is parameterized on
* spiking pattern
* time taken to integrate a single cell over a fixed time interval

When updated, a benchmark cell would generate spikes, then busy wait until the update time has been reached.

This cell type can be used to test the scalability of the simulation framework.

### LINKED ISSUE #501: Benchmark cell miniapp
Add a miniapp that uses the benchmark cell type to build networks for simulation infrastructure testing.

bcumming:
Add a new cell type, and corresponding `cell_group` implementation, for benchmarking the simulator library architecture.

Add an `benchmark_cell_group`, where each cell in the group
  * generates a spike train prescribed by a `time_seq`
  * takes a prescribed time interval per cell to perform the  `cell_group::advance` method.
With this cell type, one can easily build arbitrary networks with prescribed spiking and cell update overheads.

A miniapp that uses this cell type to build a benchmark model is implemented in `example/bench`.

Fixes #493 
Fixes #501


src/cell_group_factory.cpp
```diff
@@ -23,6 +24,9 @@ cell_group_ptr cell_group_factory(const recipe& rec, const group_description& gr
     case cell_kind::lif_neuron:
         return make_cell_group<lif_cell_group>(group.gids, rec);
 
+    case cell_kind::benchmark:
```
apeyser: I wish I could think of a better way to centralize all these switch mappings. New developers will spend a lot of time finding where each mapping happens.

src/common_types.hpp
```diff
@@ -68,9 +68,10 @@ using sample_size_type = std::int32_t;
 // group equal kinds in the same cell group.
 
 enum class cell_kind {
-    cable1d_neuron,      // Our own special mc neuron
-    lif_neuron,          // Leaky-integrate and fire neuron
-    spike_source,        // Cell that generates spikes at a user-supplied sequence of time points
+    cable1d_neuron,   // Our own special mc neuron.
```
apeyser: Some kind of enum mapping object or conversion operator for the enum maybe.

src/benchmark_cell.hpp
```diff
@@ -0,0 +1,21 @@
...
+namespace arb {
+
+// Cell description returned by recipe::cell_description(gid) for cells with
+// recipe::cell_kind(gid) returning cell_kind::benchmark
+
+struct benchmark_cell {
+    // Describes the time points at which spikes are to be generated.
+    time_seq time_sequence;
+
+    // Time taken in μs to advance the cell one ms of simulation time.
```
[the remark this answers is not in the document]
bcumming: Yeah, the units are a hot mess.

I could change it to a "relative to real time" value.
  * value of 1 corresponds ot real time
  * value of 10 corresponds to 10 tmes faster than real time

src/cell_group_factory.cpp
bcumming: We want to make it easy for users to define their own cell/cell_group types.

Some sort of central catalogue of cell types, and associated methods, properties, etc will be required to make that happen.

src/benchmark_cell.hpp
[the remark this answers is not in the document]
bcumming: Changed to `realtime_ratio` parameter: 1 -> realtime, 0.1 -> 10x faster than realtime.

example/bench/bench.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include <util/ioutil.hpp>
+
+#include "recipe.hpp"
+
+using namespace arb;
+
+int main(int argc, char** argv) {
+    try {
+        distributed_context context;
+        #ifdef ARB_HAVE_MPI
```
apeyser: Do we have a standard for #ifdef alignment? I tend to like them flush left, since they are pre-processing, rather than part of the code logic itself.

example/bench/parameters.cpp
```diff
@@ -0,0 +1,111 @@
...
+    auto it = j.find(name);
+    if (it==j.end()) {
+        return arb::util::nullopt;
+    }
+    T value = std::move(*it);
+    j.erase(name);
+    return std::move(value);
+}
+
+bench_params read_options(int argc, char** argv) {
```
apeyser: Add comment about command line options, ie: 
`bench.exe <config-filename>?`
with config-filename referencing where we define the json variables

example/bench/parameters.cpp
```diff
@@ -0,0 +1,111 @@
...
+    std::ifstream f(fname);
+
+    if (!f.good()) {
+        throw std::runtime_error("Unable to open input parameter file: "+fname);
+    }
+
+    nlohmann::json json;
+    json << f;
+
+    if (auto o  = extract<std::string>("name", json)) {
```
apeyser: So, none of these parameters can be something equivalent to false? No empty strings, no 0 num-cells, no 0 duration, ...? Do we have that documented somewhere?

example/bench/parameters.cpp
```diff
@@ -0,0 +1,111 @@
...
+        params.network.fan_in = *o;
+    }
+    if (auto o  = extract<double>("realtime-ratio", json)) {
+        params.cell.realtime_ratio = *o;
+    }
+    if (auto o  = extract<double>("spike-frequency", json)) {
+        params.cell.spike_freq_hz = *o;
+    }
+
+    for (auto it=json.begin(); it!=json.end(); ++it) {
```
apeyser: Couldn't this just be a range expression
```
for (auto it: json) {}
```

apeyser: Maybe the problem is what a `json::iterator` dereferences to?

example/bench/parameters.hpp
```diff
@@ -0,0 +1,43 @@
...
+#include <ostream>
+#include <string>
+
+#include <common_types.hpp>
+
+using arb::time_type;
+
+struct bench_params {
+    struct cell_params {
+        double spike_freq_hz = 20;   // Frequency in hz that cell will generate (poisson) spikes.
```
apeyser: Isn't 10 Hz a bit more realistic?

example/bench/recipe.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <time_sequence.hpp>
+
+#include "recipe.hpp"
+
+cell_size_type bench_recipe::num_cells() const {
+    return params_.num_cells;
+}
+
+arb::util::unique_any bench_recipe::get_cell_description(cell_gid_type gid) const {
+    using RNG = std::mt19937_64;
```
apeyser: Why all capital?

example/bench/recipe.cpp
```diff
@@ -0,0 +1,54 @@
...
+    arb::benchmark_cell cell;
+    cell.realtime_ratio = params_.cell.realtime_ratio;
+    cell.time_sequence = arb::poisson_time_seq<RNG>(RNG(gid), 0, 1e-3*params_.cell.spike_freq_hz);
```
apeyser: Comment on what's happening here? Are we seeding a sequence based on the gid?

example/bench/recipe.cpp
```diff
@@ -0,0 +1,54 @@
...
+    return std::move(cell);
```
apeyser: Is this needed? Isn't local variable and constructor elision part of the standard? (Which is why C++ is such a nightmare --- such a basic thing requires "checking the standard")

example/bench/recipe.cpp
```diff
@@ -0,0 +1,54 @@
...
+    return arb::cell_kind::benchmark;
+}
+
+std::vector<arb::cell_connection> bench_recipe::connections_on(cell_gid_type gid) const {
+    const auto n = params_.network.fan_in;
+    std::vector<arb::cell_connection> cons;
+    cons.reserve(n);
+    using RNG = std::mt19937_64;
+    RNG rng(gid);
+    // issue: invalid gid for source in a connection
```
apeyser: Don't understand this comment

example/bench/recipe.cpp
```diff
@@ -0,0 +1,54 @@
...
+    std::uniform_int_distribution<cell_gid_type> dist(0, params_.num_cells-2);
+    for (unsigned i=0; i<n; ++i) {
+        cell_gid_type src = dist(rng);
+        if (src>=gid) ++src;
```
apeyser: I don't understand what is happening here. Why are we increment src when it's `>= gid`? If it was just `gid`, I'd assume we don't want ourselves to be the source, and so we make it "the next cell", but this ... I don't get. And what if we go over the max gid? What if the src is now outside of the range?

example/bench/recipe.cpp
```diff
@@ -0,0 +1,54 @@
...
+        cons.push_back(con);
+    }
+
+    return cons;
+}
+
+cell_size_type bench_recipe::num_targets(cell_gid_type gid) const {
+    // Only one target, to which all incoming connections connect.
+    // It might be the case that we can parameterize this, in which case the connections
+    // generated in connections_on should end on random cel-local targets.
```
apeyser: cel-local -> cell-local

src/benchmark_cell_group.cpp
```diff
@@ -0,0 +1,88 @@
...
+        // Start timer.
+        auto start = high_resolution_clock::now();
+
+        while (tseq.front()<ep.tfinal) {
+            spikes_.push_back({{gid, 0u}, tseq.front()});
+            tseq.pop();
+        }
+
+        // Wait until the expected time to advance has elapsed.
+        while (duration_type(high_resolution_clock::now()-start).count() < duration_us);
```
apeyser: Okay --- we're not just waiting, we want to eat up cpu time until the clock has elapsed. So avoiding yields. (Maybe add to comment so no one ever attempts to 'optimize' this)

example/bench/parameters.cpp
bcumming: Sure.

example/bench/parameters.cpp
bcumming: This code will do a reasonable job of testing that the correct type was used for each parameter, but you are right, it doesn't then check that the parameters are in valid ranges.

This is intended to be simple example code, so I was trying to tread the line between readability and a reasonable level of robustness (hence the efforts to warn users about unused parameters, which would normally indicate a typo).

example/bench/parameters.cpp
bcumming: You are right: `json::iterator` dereferences to the `value`, not the `key`, unlike STL containers which give a `std::pair<key,value>`.

bcumming: And I want to print out all the unused keys, which would be a dead giveaway that there is a typo in the input.

example/bench/parameters.hpp
bcumming: I have no idea... I will change it!

example/bench/recipe.cpp
bcumming: Good question. I have used this pattern in a few places, and it was originally cut and paste from code where `RNG` was a template parameter.
I replaced it with `rng_type`.

example/bench/recipe.cpp
bcumming: Yes, we are passing it a random generator seeded on `gid`. This way, regardless of how the cell is distributed over MPI ranks/threads/time zones, it will generate a reproducable sequence of spikes.
I have added a comment that explains this in more detail.

example/bench/recipe.cpp
bcumming: It is needed, because we are not returning a value of type `arb::benchmark_cell`. Instead we are returning an `arb::unique_any`. When saying `return X;`, we are saying to create a `unique_any` constructed with `X`, which by default will use the copy constructor. By manually turning it into an rvalue reference, we get the move constructor.

example/bench/recipe.cpp
bcumming: Sorry, this was a remnant from debugging.
It has been replaced with the following comment:
```C++
// Generate n incoming connections on this cell with random source, where
// the source can't equal gid (i.e. no self-connections).                                                                                                                                                           
// We use a random distribution that will uniformly draw values from the 
// union of the two ranges: [0, gid-1] AND [gid+1, num_cells-1].
// To do this, we draw random values in the range [0, num_cells-2], then
// add 1 to random values ≥ gid.
```
Which should help with your comment below about why `src` is incremented when greater than or equal to `gid`.

example/bench/recipe.cpp
bcumming: I have added detailed comments that explain what is goind on here.
We draw values in the range [0, num_cells-2], so by adding one we get uniform values in the range [0, num_cells-1], hence we never get a source with gid greater than or equal to num_cells.

example/bench/recipe.cpp
bcumming: Fixed.

src/benchmark_cell_group.cpp
bcumming: I have added the following comment:
```
// Wait until the expected time to advance has elapsed. Use a busy-wait                                                                                                                                             
// so that the resources of this thread are tied up until the interval
// has elapsed, to emulate a "real" cell.
```
Though such an optimization would completely miss the point of this cell type!

example/bench/recipe.cpp
apeyser: Probably should comment when you have to coerce the compiler like that.

src/benchmark_cell_group.cpp
apeyser: Exactly!

History shows people go through and miss the point and "fix things" -- see the great Debian ssh debacle.

PR #506, opened 2018-06-19
Feature/lib install target part i

halfflat:
## CMake and build refactoring

*   Use CUDA as first-class language (leading to CMake 3.9 minimum version requirement).

*   Use 'modern CMake' interface libraries for compiler options, include file and library dependency tracking. Interface library targets:
    * `arbor-deps`: compiler options and library requirements for the `libarbor.a` static library, as governed by configure-time options and environment.
    * `arbor-private-headers`: include path for non-installed headers, as required by unit tests and arbor itself.
    * `arbor-aux`: helper classes and utilities used across tests and examples.
    * `ext-json`, `ext-tclap`, `ext-tbb`, `ext-benchmark`, `ext-sphinx_rtd_theme`: externally maintained software that we include (directly or via submodule) in the `ext/` subdirectory.
 
*   Single static library `libarbor.a` includes all built-in modules and CUDA objects.

*   Simply configuration options:
    *  `ARB_WITH_TRACE`, `ARB_AUTORUN_MODCC_ON_CHANGES` `ARB_SYSTEM_TYPE` removed.
    * External `modcc` is provided by `ARB_MODCC` configuration option; if provided `modcc` is still buildable, but is not included in the default target.
    * `ARB_PRIVATE_TBBLIB`, defaulting to `OFF`, instructs the build to make TBB from the included submodule.

*   Extend `ErrorTarget` functionality to provide a dummy target or an error target based on a condition.

*   Generate header version defines and library version variables based on git status and project version, via new script `include/git-source-id`.

*   All generated binaries now placed in `bin/` subdirectory at build.

*  Install targets installs: public headers (incomplete); static library; `modcc` tool; `lmorpho` executable; `html` documentation (examples, tests and validation data are currently not installed).

*   Executable targets have had the `.exe` suffix removed; unit tests are labelled `unit` (arbor unit tests), `unit-modcc` (modcc unit tests), `unit-local` (distributed tests with local context), `unit-mpi` (distributed tests with MPI context).

*   More graceful handling of configure-time detection of `nrniv`, Julia and required Julia modules for validation data generation.

*  Add `cmake/FindJulia.cmake`, `cmake/FindTBB.cmake`  package finders, and adjust `cmake/FindUnwind.cmake` to use link library-style properties.

*  Adjust travis script to test `unit-local` and `unit-mpi` if appropriate.

*  Simply documentation `conf.py`.

## Source relocation and reorganization

* All external project sources and files moved to `ext/`.
* Source code refactoring to decouple library-using code from the configure-time definitions that govern arbor behaviour: removes conditional code in public headers that depends upon `ARB_WITH_X`-type definitions at compile time. Affected code is is in the public interfaces for MPI, the threading implementation, and the profiler.
* Remove `util/debug.hpp`; split out functionality for pretty-printing from assertion handling.
* Make FVM cell non-physical voltage check a run-time cell-group parameter.
* Move spike double buffer implementation to `simulation.cpp`.
* Make timer utility wrap POSIX `clock_gettime` independent of threading configuration.
* Make `mpi_error` derive from `system_error` and follow C++11 `system_error` semantics.
* `EXPECTS` macro replaced by `arb_assert` macro.
* JSON dependency removed from `libarbor.a` and header files: moved to auxiliary library.
* Publicly visible macros garner an `ARB_` prefix as required.
* Move SWC test file to `test/unit` directory.
* Work-in-progress splitting of public from private includes: as a convention not entirely adhered to as yet, private headers within arbor source are included with `""`, public headers with `<>`.

## Modcc interface changes

* Expose via `--namespace` option the functionality that sets the namespace in generated code.
* Use `--profile` option to add profiler hooks to generated code; uses public function interface directly rather than `PE/PL` macros in order to avoid public `PE` and `PL` defines.


CMakeLists.txt
```diff
@@ -1,313 +1,248 @@
...
+    set(ARB_WITH_CUDA TRUE)
 
-    # unfortunate workaround for C++ detection in system mpi.h
-    add_definitions(-DMPICH_SKIP_MPICXX=1 -DOMPI_SKIP_MPICXX=1)
-    set_property(DIRECTORY APPEND_STRING PROPERTY COMPILE_OPTIONS "${MPI_C_COMPILE_FLAGS}")
-endif()
+    target_compile_options(arbor-deps INTERFACE
+        $<$<COMPILE_LANGUAGE:CUDA>:--diag_suppress=integer_sign_change
+        $<$<COMPILE_LANGUAGE:CUDA>:--diag_suppress=unsigned_compare_with_zero)
+    target_compile_definitions(arbor-deps ARB_HAVE_GPU)
```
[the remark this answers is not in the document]
bcumming: I think I figured that one out. The question then becomes "if `arbor-deps` is an interface, why do I have to say `INTERFACE` every time I interact with it?".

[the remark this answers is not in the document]
halfflat: That's a very good question, Ben! Does anyone else have a question they would like to ask the CMake developers?

mechanisms/CMakeLists.txt
```diff
@@ -1,29 +1,33 @@
...
-# Generate mechanism implementations for host/cpu environment
+# Generate mechanism implementations.
 
 set(mech_dir "${CMAKE_CURRENT_BINARY_DIR}/generated")
 file(MAKE_DIRECTORY "${mech_dir}")
 
+set(modcc_flags)
 if(ARB_VECTORIZE)
-    set(modcc_simd "-s")
+    list(APPEND modcc_flags "-s")
```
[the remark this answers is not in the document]
halfflat: Yup, good point, will amend.

CMakeLists.txt
```diff
@@ -1,313 +1,248 @@
...
-    add_definitions(-DARB_HAVE_MPI)
+    enable_language(CUDA)
+    set(ARB_WITH_CUDA TRUE)
 
-    # unfortunate workaround for C++ detection in system mpi.h
-    add_definitions(-DMPICH_SKIP_MPICXX=1 -DOMPI_SKIP_MPICXX=1)
-    set_property(DIRECTORY APPEND_STRING PROPERTY COMPILE_OPTIONS "${MPI_C_COMPILE_FLAGS}")
-endif()
+    target_compile_options(arbor-deps INTERFACE
+        $<$<COMPILE_LANGUAGE:CUDA>:--diag_suppress=integer_sign_change
```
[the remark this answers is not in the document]
halfflat: Yeah, I think this is a combination of sloppy testing and merge-editing. Will get revise all the CUDA building again.

CMakeLists.txt
```diff
@@ -1,313 +1,248 @@
...
-if(UNWIND_FOUND)
-    add_definitions(-DWITH_UNWIND)
-    include_directories(${UNWIND_INCLUDE_DIR})
-    list(APPEND EXTERNAL_LIBRARIES ${UNWIND_LIBRARIES})
+
+set(ARB_THREADING_MODEL "cthread" CACHE STRING "set the threading model, one of cthread/tbb/serial")
+set_property(CACHE ARB_THREADING_MODEL PROPERTY STRINGS cthread tbb serial )
+
+if(ARB_THREADING_MODEL STREQUAL "tbb")
+set(ARB_WITH_TBB TRUE)
```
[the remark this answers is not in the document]
halfflat: Could explain the TBB travis failure, hmm!

include/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/arbor)
+if(ARB_WITH_ASSERTIONS)
+configure_file(arbor/assert_macro.hpp.enabled arbor/assert_macro.hpp COPYONLY)
+else()
+configure_file(arbor/assert_macro.hpp.disabled arbor/assert_macro.hpp COPYONLY)
+endif()
+
+add_library(arbor-public-headers INTERFACE)
+target_include_directories(arbor-public-headers INTERFACE
```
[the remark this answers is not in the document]
halfflat: Yeah, I'll try and come up with a comment that properly balances concision, informativeness, and profanity.

modcc/CMakeLists.txt
```diff
@@ -12,25 +16,24 @@ set(MODCC_SOURCES
...
+set_target_properties(libmodcc PROPERTIES OUTPUT_NAME modcc)
 
-add_executable(modcc modcc.cpp)
+add_executable(modcc ${modcc_sources})
+target_link_libraries(modcc libmodcc ext-tclap)
 
-target_link_libraries(modcc LINK_PUBLIC compiler)
+set_target_properties(modcc libmodcc PROPERTIES EXCLUDE_FROM_ALL ${ARB_WITH_EXTERNAL_MODCC})
+
+install(TARGETS modcc RUNTIME DESTINATION ${CMAKE_INSTALL_LIBEXECDIR})
```
bcumming: This is installed in `CMAKE_INSTALL_LIBEXECDIR`, which on my system is `<prefix>/libexe`.
`lmorpho` is installed in `CMAKE_INSTALL_BINDIR`, i.e. `<prefix>/bin`.
I would rather see modcc get installed in `bin`, what do you think?

halfflat: Put it in `libexec` on the grounds that it was a tool used to build the library, rather than a user executable, but on reflection this was wrong thinking.

If `modcc` were only ever used by `libarbor` — say to build mechanisms from NMODL at run time — then it would go here. But it is in fact a user-tool, and should go in `bin` as you suggest.

include/arbor/communication/mpi_error.hpp
```diff
@@ -0,0 +1,110 @@
+#pragma once
```
bcumming: This header is public, however it is only used in non-public code. And only in non-public cpp files at that.

halfflat: `mpi_error` should be able to be consumed though in user code, i.e. one should be able to write a `catch (mpi_error& e)` clause in an application.

halfflat:
CMake error is something they addressed in a later version of CMake. I'll attempt a workaround tomorrow (or we can bump the minimum CMake version).

test/unit/test_algorithms.cpp
```diff
@@ -9,10 +9,19 @@
 #include <arbor/util/compat.hpp>
 
 #include "algorithms.hpp"
-#include "threading/threading.hpp"
 #include "util/index_into.hpp"
 #include "util/meta.hpp"
 
+// (Pending abstraction of threading interface)
```
bcumming: How about indenting some of these includes? The pattern of "always align preprocessor to the left gutter" makes for ugly code in my opinion.

```
#include <arbor/version.hpp>
#if defined(ARB_TBB_ENABLED)
    #include "threading/tbb.hpp"
#elif defined(ARB_CTHREAD_ENABLED)
    #include "threading/cthread.hpp"
#else
    #include "threading/serial.hpp"
#endif
```

test/unit/test_lif_cell_group.cpp
```diff
@@ -158,7 +159,7 @@ TEST(lif_cell_group, spikes) {
     path_recipe recipe(2, 1000, 0.1);
 
     hw::node_info nd;
-    nd.num_cpu_cores = threading::num_threads();
+    nd.num_cpu_cores = arb::thread_count();
```
bcumming: To minimise the surface area of this PR, lI would avoid renamings like`num_threads` to `thread_count`.
Also, `num_threads` is a better variable name.

bcumming:
I would prefer that you finish this PR as is: i.e. get it passing tests and installing "something", even if the installed target is not very useful.
Then further pull requests to get the installed target into shape.
There are a lot of changes like wholesale renaming that are going on, and i can't really review them because they are lost in the noise of 490 change files.

test/unit/test_mc_cell_group_gpu.cpp
```diff
@@ -1,10 +1,11 @@
 #include "../gtest.h"
 
-#include <backends.hpp>
-#include <common_types.hpp>
-#include <epoch.hpp>
-#include <fvm_lowered_cell.hpp>
-#include <mc_cell_group.hpp>
+#include <arbor/common_types.hpp>
+
+#include "backends.hpp"
```
bcumming: Why are we using `"backends.hpp"` instead of `<backends.hpp>`. These files are not in the current path, so this is a bit misleading.

CMakeLists.txt
```diff
@@ -159,19 +166,12 @@ endif()
...
-    enable_language(CUDA)
     set(ARB_WITH_CUDA TRUE)
 
-    target_compile_options(arbor-deps INTERFACE
-        $<$<COMPILE_LANGUAGE:CUDA>:--diag_suppress=integer_sign_change
-        $<$<COMPILE_LANGUAGE:CUDA>:--diag_suppress=unsigned_compare_with_zero)
-    target_compile_definitions(arbor-deps ARB_HAVE_GPU)
+    add_compile_options(
+        "$<$<COMPILE_LANGUAGE:CUDA>:-Xcudafe=--diag_suppress=integer_sign_change>"
+        "$<$<COMPILE_LANGUAGE:CUDA>:-Xcudafe=--diag_suppress=unsigned_compare_with_zero>")
```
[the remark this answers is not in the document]
halfflat: Nearly correct: they get resolved when they make the Makefiles.

ext/CMakeLists.txt
```diff
@@ -12,24 +12,38 @@ target_include_directories(ext-tclap INTERFACE tclap/include)
 # Alias system TBB or build locally and export that, according
 # to ARB_PRIVATE_TBBLIB setting.
 
-if(ARB_REQUIRE_TBB)
+find_package(TBB)
+if(ARB_PRIVATE_TBBLIB OR NOT TBB_FOUND)
```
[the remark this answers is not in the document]
halfflat: I''m not really concerned about the exact semantics here: there is more than one 'right way'. If you'd prefer that users must explicitly set the ARB_PRIVATE_TBBLIB option to use the included TBB, then that's also cool.

ext/CMakeLists.txt
```diff
@@ -12,24 +12,38 @@ target_include_directories(ext-tclap INTERFACE tclap/include)
...
+    # a subdirectory and 2) having been excluded-from-all, the behaviour
+    # might have been undefined anyway. Seriously.
+
+    install(FILES "$<TARGET_FILE:tbb_static>" "$<TARGET_FILE:tbbmalloc_static>" DESTINATION ${CMAKE_INSTALL_LIBDIR} OPTIONAL)
+else()
+    add_library(ext-tbb ALIAS TBB::tbb)
+endif()
+
+if(ARB_WITH_TBB)
     if(ARB_PRIVATE_TBBLIB)
```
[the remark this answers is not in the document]
halfflat: It already is an explicit option that defaults to OFF :)

modcc/printer/cudaprinter.cpp
```diff
@@ -218,7 +218,7 @@ std::string emit_cuda_cu_source(const Module& module_, const printer_options& op
         "#include <" << arb_private_header_prefix() << "backends/event.hpp>\n"
         "#include <" << arb_private_header_prefix() << "backends/multi_event_stream_state.hpp>\n"
         "#include <" << arb_private_header_prefix() << "backends/gpu/cuda_common.hpp>\n"
-        "#include <" << arb_private_header_prefix() << "backends/gpu/math.hpp>\n"
+        "#include <" << arb_private_header_prefix() << "backends/gpu/math_cu.hpp>\n"
```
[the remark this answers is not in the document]
halfflat: It resolved some ambiguities in header paths.

test/unit/test_algorithms.cpp
halfflat: I generally prefer not to indent directives unless it's really necessary for clarity, as the scoping rules for them are completely different from the scoping rules of the source code around them.

test/unit/test_lif_cell_group.cpp
halfflat: You know, I hate those `num_`names. But yeah, it doesn't matter what it is called. It's not replacing the underlying call to the thread implementation, but wrapping it.

test/unit/test_mc_cell_group_gpu.cpp
halfflat: The convention I'm trying to adhere to is that public headers get the angle brackets, and private ones get the quotes, just to help keep things straight. We could e.g. rename the arbor source directory back to `src`, or maybe `lib`, and then qualify the private headers with that directory prefix. In terms of confusing  header paths though, it was already a huge mess, and I think this is making it less so.

halfflat:
The refactorings are unfortunately necessary if we want to keep configuration-based code out of header files that are included by other code which does not (and should not) set the same configuration-based defines.

modcc/printer/cudaprinter.cpp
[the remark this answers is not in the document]
halfflat: Also it's clearer: having multiple headers called "math.hpp" is confusing.

ext/CMakeLists.txt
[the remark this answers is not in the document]
bcumming: I thought that the approach you were taking was to explicitly set `ARB_PRIVATE_TBBLIB`

ext/CMakeLists.txt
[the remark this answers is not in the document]
bcumming: OK. That didn't show up when I did a search of the changeset.

ext/CMakeLists.txt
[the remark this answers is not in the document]
halfflat: Well, the logic there kinda captures what I thought would be useful: if TBB is requested and we can't find it, or the user requests the in-source one, then we build the in-source one. I'm totally happy with the alternative, of it being an error if TBB isn't found and the explicit request is not made. Do you have a preference?

test/unit/test_algorithms.cpp
bcumming: It isn't about scoping, it is about making it easier to read and understand.
The additional clarity overides any potential scoping confusion for a proficient C++ developer.
There is an argument to be made, particularly at the top of a file like this, that there is a sort of  "preprocessor scoping".

test/unit/test_lif_cell_group.cpp
bcumming: I know you don't like them, but that is the name we have, and wrapping it with a different name is renaming it ;)

test/unit/test_algorithms.cpp
halfflat: Consistency is also desirable, and having one convention for when it is just directives, and another when it is a mix of directives and code, is also unappealing. In short sequences like this I can't see it being a big gain in readability to indent them. It matters little in this instance, but I don't want to concede that indenting these guys is necessarily desirable.

bcumming:
How about finishing this PR up with an imperfect setup. Then in subsequent PRs I will be able to follow the steps taken to keep configuration based code out of header files.
It doesn't have to be perfect at first.

halfflat:
As far as I can tell from this point, indenting and name wrangling aside, the only remaining issue is working around CMake brokenness in 3.9.

test/unit/test_mc_cell_group_gpu.cpp
bcumming: We have discussed this in the past, and moved to using `<>` in the unit tests.

halfflat: If we see `<foo/bar.hpp>` it's not at all clear that that is referring to a private `foo/bar.hpp` or a for a `bar.hpp` belonging to a system-installed package called `foo`. I am not saying that that is a terrible thing, only that it doesn't give the clarity that you desire.

halfflat: I'd be quite happy with a `<src/arbor/foo.hpp>` or `"src/arbor/foo.hpp"` though.

halfflat: Also note that when the public/private split is complete, the number of private headers used by the tests will be very much reduced.

test/unit/test_algorithms.cpp
bcumming: We discussed this with @vkarak 2 years ago when working on our coding standards.
After ~much bloodletting~ a discussion the decision was to indent preprocessor directives, though we never wrote it down.

Taking the original `src/threading/threading.hpp` file as a reference:
```
#if defined(ARB_HAVE_TBB)
    #include "tbb.hpp"
#elif defined(ARB_HAVE_CTHREAD)
    #include "cthread.hpp"
#else
    #define ARB_HAVE_SERIAL
    #include "serial.hpp"
#endif
```

halfflat: Pop it in the coding standards.

test/unit/test_mc_cell_group_gpu.cpp
halfflat: And (hopefully) lastly, this distinction between <> and "" for public/private is quite common.

ext/CMakeLists.txt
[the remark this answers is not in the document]
bcumming: What you have is fine then! I thought that your preference was for the in-source TBB to be strictly opt-in.
I actually prefer this approach to making it strictly opt-in, because I like using the in-source version, and don't feel like explicitly asking for it!

modcc/printer/cudaprinter.cpp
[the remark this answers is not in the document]
bcumming: It just fells like redundant renaming. I would also look for `X/math.hpp` instead of `X/math_x` where `x` is restating `X`.
Not a big deal though!

test/unit/test_algorithms.cpp
bcumming: We should copy those standards from the old `cell_algorithms` wiki over the the new wiki, where we might update them from time to time.

test/unit/test_mc_cell_group_gpu.cpp
bcumming: How about a new macro `include_private_header(foo.hpp)`?

Joking aside, `<src/arbor/foo.hpp>` is a bit long winded, and we probably shouldn't change the directory structure of the source code just to accomodate the internal use of headers in our unit tests.

Thinking about it, the reason this makes me uncomfortable is that the quoting is being use as shorthand for "any header not part of the public API", when it is usually used as shorthand for "headers relative to this path".

bcumming: We could have grouping rules for the unit tests, e.g.

```
#include <iostream>
#include <numeric>

#include <arbor/common_types.hpp>
#include <arbor/simulation.hpp>

#include <foo.hpp>
#include <internal.hpp>
...
```

There will still be a lot of inclusion of private headers after the refactoring, because we will have tests for all the private functionality, right?

bcumming: OK, I have been doing some reading, and it would appear that using quotes for project files is a common practice.
e.g. https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes

Let's argue about this in front of a black board.

PR #508, opened 2018-06-25
Move cell description types to public includes.

halfflat:
Further work to public install target.

* Move SIMD classes, cell description classes, simple sampler to public include.
* Rename `cell` to `mc_cell`, `segment` to `mc_segment`, and remove `_description` from cell description class names and includes.
* Move `compartment_model` out of `mc_cell` interface and use only in `fvm_layout.cpp`.
* (Provisionally) remove area/volume methods on `mc_cell` and `mc_segment`.


include/arbor/lif_cell.hpp
```diff
@@ -3,7 +3,7 @@
 namespace arb {
 
 // Model parameteres of leaky integrate and fire neuron model.
-struct lif_cell_description {
+struct lif_cell {
```
bcumming: OK. I see, each cell type/description should have an `X_cell` description type in the public api.

include/arbor/mc_segment.hpp
```diff
@@ -54,8 +54,13 @@ struct mechanism_desc {
     }
 
     double get(const std::string& key) const {
-        auto optv = util::value_by_key(param_, key);
-        return optv? *optv: throw std::out_of_range("no field "+key+" set");
+        auto i = param_.find(key);
+        if (i==param_.end()) {
```
bcumming: I have a memo from a C++ developer in the future, it reads simply "thank you"!

bcumming: though why the `else`?

include/arbor/simd/avx.hpp
```diff
@@ -8,8 +8,8 @@
 #include <cstdint>
 #include <immintrin.h>
 
-#include <simd/approx.hpp>
-#include <simd/implbase.hpp>
+#include <arbor/simd/approx.hpp>
+#include <arbor/simd/implbase.hpp>
```
bcumming: Why is SIMD moving to the public API? Of all the things to hide from users, that would intuitively be at the top of my list!

include/arbor/time_sequence.hpp
```diff
@@ -33,9 +32,10 @@ class time_seq {
 
     template <
         typename Impl,
-        typename = typename util::enable_if_t<
+        typename = typename std::enable_if<
```
bcumming: gargh. Bring on C++14.

include/arbor/time_sequence.hpp
```diff
@@ -123,8 +123,8 @@ struct vector_time_seq {
         seq_(std::move(seq))
     {
         // Ensure that the time values are sorted.
-        if(!std::is_sorted(seq_.begin(), seq_.end())) {
-            util::sort(seq_);
+        if (!std::is_sorted(seq_.begin(), seq_.end())) {
+            std::sort(seq_.begin(), seq_.end());
```
bcumming: Is there a tipping point where the convenience of using the nice util functionality in the public API code outweighs the downsides of sharing it?

arbor/backends/multicore/shared_state.hpp
```diff
@@ -12,12 +12,10 @@
 #include <arbor/fvm_types.hpp>
 #include <arbor/ion.hpp>
 #include <arbor/util/enumhash.hpp>
+#include <arbor/simd/simd.hpp>
```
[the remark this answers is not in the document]
halfflat: Will fix.

arbor/cell_group_factory.hpp
```diff
@@ -2,11 +2,12 @@
...
-#include <cell_group.hpp>
-#include <domain_decomposition.hpp>
-#include <recipe.hpp>
-#include <util/unique_any.hpp>
+#include <arbor/util/unique_any.hpp>
+
+#include "backends.hpp"
+#include "cell_group.hpp"
+#include "domain_decomposition.hpp"
+#include "recipe.hpp"
```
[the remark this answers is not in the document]
halfflat: Yes, definitely!

arbor/fvm_compartment.hpp
```diff
@@ -6,84 +6,15 @@
 #include <arbor/common_types.hpp>
```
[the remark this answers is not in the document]
halfflat: For this file, it is because it is a very specifically FVM-decomposition of the compartments into components suitable for the FVM discretization as performed now in `fvm_layout`.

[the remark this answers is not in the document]
halfflat: `segment` is prefixed only because it is quite a generic name being used in a fairly specific context. When it comes to cell type naming, we should definitely have consensus between the cell kind names and the class names; here I aligned with the name of the cell group type.

I think a good choice of names is going to require either another sort of mechanism-aware cell (quite likely, with one- or explicit-compartment models) or another sort of discretization of 3-d embedded morphologies (e.g. according to a finite difference scheme, but this is very unlikely to be implemented). Then we'll be forced to straighten all this out.

arbor/fvm_layout.cpp
```diff
@@ -132,7 +151,7 @@ fvm_discretization fvm_discretize(const std::vector<cell>& cells) {
         seg_comp_bounds.clear();
         auto seg_comp_part = make_partition(
             seg_comp_bounds,
-            transform_view(c.segments(), [](const segment_ptr& s) { return s->num_compartments(); }),
+            transform_view(c.segments(), [](const mc_segment_ptr& s) { return s->num_compartments(); }),
```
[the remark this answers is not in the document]
halfflat: If we rename segment, we need to rename this guy. But it's definitely contingent on the former.

include/arbor/mc_segment.hpp
halfflat: No semantic reason; just thought it aided clarity. Would happily remove.

include/arbor/time_sequence.hpp
halfflat: Oh yep. Next on the code quality TODO list.

include/arbor/time_sequence.hpp
halfflat: Possibly, but I think we're fairly removed from that point.
(As you know, I don't think this guy should be calling sort anyhow!)

include/arbor/simd/avx.hpp
halfflat: This has to be exposed so that `modcc` users can compile their modules.

halfflat: (More than just simd has to be exposed for this, of course.)

arbor/fvm_layout.cpp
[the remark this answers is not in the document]
bcumming: Then why not `soma_mc_segment`?
Just trolling.

PR #518, opened 2018-07-04
Feature/lib install target part 3

halfflat:
This time we're moving `recipe.hpp` and `simulation.hpp`, plus the requirements they bring.

Code changes:
* Pimplize `simulation`.
* Consolidate arbor exceptions: all non-cell kind specific exceptions that might be expected to reach user code now have consistent messages and fit in an exception hierarchy based at `arb::arbor_exception`. Internal errors throw an `arb::arbor_internal_error` exception.
* Renamed `postsynaptic_spike_event` to `spike_event`. (Note: `pse_vector` name is unchanged.)
* Repurposed `pprintf` and moved it into `strprintf.h` — further consolidation is a TODO.
* Made a generic `util::to_string` to avoid redundancy of `operator<<` overloads and other `to_string` definitions. Defaults to ADL `to_string`, `std::to_string`, and finally tries using `operator<<`.


include/arbor/simulation.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <arbor/recipe.hpp>
+#include <arbor/sampling.hpp>
+#include <arbor/schedule.hpp>
+#include <arbor/util/handle_set.hpp>
+
+namespace arb {
+
+using spike_export_function = std::function<void(const std::vector<spike>&)>;
+
+struct simulation_state;
```
bcumming: A comment here along the lines of "forward declaration of pimpl implementation of simulation."

bcumming: Also a comment that mentions that we this is using PIMPL, and why.

include/arbor/util/uninitialized.hpp
```diff
@@ -43,15 +43,15 @@ class uninitialized {
     using const_rvalue_reference= const X&&;
 
     pointer ptr() {
-        // COMPAT: xlC 13.1.4 workaround:
+        // COMPAT: xlC 13.1.4 workaround (still broken in 14.1.0):
```
bcumming: Are we just going to remove this and refuse to compile with xlc?
xlc is worse than useless: generates incorrect code that is 10x slower than gcc.

include/arbor/simulation.hpp
halfflat: I'll add something brief in next tranche.

include/arbor/util/uninitialized.hpp
halfflat: We will! I wanted to make the xlC deprecation its own PR. (It could be the next one, even.)

PR #522, opened 2018-07-06
Migrate source/build to c++14                                                                                                                   

### LINKED ISSUE #358: Migrate to C++14 [4]
Moving to C++14 for the (non-kernel) sources will allow us to remove a number of work-arounds in the code (e.g. the `DEDUCED_RETURN_TYPE` macro and the lack of aggregate initializations for PODs with default value) and simplify a bunch of the type-wrangling currently performed.

We can _probably_ assume that a sufficiently capable C++14 compiler exists on our target platforms.

Tasks:
1. Try it! Use C++14 as the language for compilation, and see what breaks, on various platforms.
2. Simplify code using the existing we're-not-using-C++14-yet work-arounds.

halfflat:
* Update `CMakeLists.txt` for C++14 option.
* Update to gcc 6 minimum.
* Update travis CI from gcc-5 to gcc-6
* Use `std::..._t` style type traits, replacing `util::` aliases.
* Use `std::cbegin`, `std::cend`, and `std::make_unique`, replacing `util::` versions.
* Remove `DEDUCED_RETURN_TYPE` macros.
* Remove redundant return type specifications.
* Use correct ADL for `begin` and `end` in (almost all) the range utilities.
* Remove redundant `mechinfo` ctor (aggregate initialization suffices).
* Use lambda capture initializers where appropriate.
* Use generic `std::equal_to`.
* Use variable templates for `math::infinity` and `math::pi`.
* Remove `enum_hash` workaround.
* Use `""s` string literals where we were using our own `""_s` construction.
* Use generic lambda for recursive lambda instead of `std::function` wrapper.
* Use generic lambda for generic arithmetic tests.

Fixes #358.


arbor/util/rangeutil.hpp
```diff
@@ -261,12 +263,15 @@ template <
...
+    using std::end;
+
     if (util::empty(seq)) {
         return {Value{}, Value{}};
     }
 
-    auto i = std::begin(seq);
-    auto e = std::end(seq);
+    auto i = begin(seq);
+    auto e = end(seq);
```
bcumming: It is overkill to have `using std::begin` when it is used once (ditto `std::end`).

arbor/util/rangeutil.hpp
```diff
@@ -297,11 +302,14 @@ bool is_sorted(const Seq& seq) {
...
     typename Seq,
     typename Proj,
-    typename Compare = std::less<typename std::result_of<Proj (typename sequence_traits<const Seq&>::value_type)>::type>
+    typename Compare = std::less<std::result_of_t<Proj (typename sequence_traits<const Seq&>::value_type)>>
 >
 bool is_sorted_by(const Seq& seq, const Proj& proj, Compare cmp = Compare{}) {
-    auto i = std::begin(seq);
-    auto e = std::end(seq);
+    using std::begin;
+    using std::end;
```
bcumming: Same here? Is this some sort of ADL shenanigans?

halfflat: Yep, it's exactly ADL shenanigans. `begin` is a customization point, and the 'right way' to get the beginning iterator is through this sort of construction.

PR #531, opened 2018-07-10
Feature/lib install target part 4

halfflat:
All example code and validation tests no longer require access to private include directories. This provides the minimal requirement for an installable target

Note that it is still not possible to separately build mechanisms from NMODL with just the public includes, and there is not yet any package configuration file creation for use with CMake or pkg-config.

* Replace `hw::node_info` with `domain_info`, describing local resources for the purposes of domain decomposition.
* Group processor counting and gpu counting implementation under `node_info.cpp`.
* Remove `domain_decomposition` dependency from `cell_group_factory.hpp` so we can use the latter to test for backend support for a cell kind.
* Add `arb::cell_kind_implementation()` which performs the mapping from cell kind and backend kind to a `cell_group_ptr`-producing function (this will then become the site for custom cell group kind mapping support in future work).
* Move headers for aux library into `aux/include/aux`, so that there is less ambiguity in header location within test and example code.
* Simplify a little the affinity scanning code.
* Add `sysconf()` and `std::thread::hardware_concurrency()` fallbacks for processor count determination.
* Move `util::mask_stream`, `util::path`, `util::scope_exit` into aux header.
* Remove `util::nop`.
* Remove `io::exporter` and `io::exporter_spike_file`; replace with simpler interface in `aux::spike_emitter`.
* Move `load_balance.hpp`, `math.hpp`, `scwio.hpp` into public includes.
* Move path, glob and scope_exit utilities to aux lib.
* Move cell group size-based partitioning functionality into partition_load_balance, through the use of new `partition_hint` structure.
* Remove brunel example specific partitioner.
* Move brunel example option code out of `::arb` namespace.
* Move swc routines out of `arb::io` and into `arb::`.
* Specialize public swc-routines to take a vector of swc_record objects instead of an arbitrary sequence; place implementations into swcio.cpp.
* Remove dependency on `util/strprintf.hpp` from lmorpho.
* Remove `arb::math::min`, `arb::math::max`.
* Remove range utility dependence from validation tests: add piece-wise linear interpolator in test/validation/interpolate.hpp; remove `times` and `values` adaptors for simple trace data; add test/validation/util.hpp with generic size function and a `to_string` for `backend_kind`.
* Add `aux::open_or_throw` helper function to replicate no-overwrite spike output behaviour in example applications.
* Add `aux::strsub` simple string substitution function for use in examples, so as to avoid a `util::strprintf` or `util::pprintf` dependency.
* Remove arbor private header dependencies from examples and validation tests.


arbor/hardware/node_info.cpp
```diff
@@ -1,19 +1,51 @@
...
-// If unable to determine the number of available cores, assumes that there is one
-// core available.
-node_info get_node_info() {
-    auto res = num_cores();
-    unsigned ncpu = res? *res: 1u;
-    return {ncpu, num_gpus()};
+unsigned node_gpus() {
+#ifdef ARB_HAVE_GPU
+    int n;
+    if (!cudaGetDeviceCount(&n)) {
```
bcumming: Compare explicitly to `CUDA_SUCCESS`

arbor/hardware/node_info.cpp
```diff
@@ -1,19 +1,51 @@
...
+        return static_cast<unsigned>(n);
```
bcumming: You use `static_cast<unsigned>` here, and `(unsigned)` below.
I am comfortable with using the second option.

arbor/hardware/node_info.cpp
```diff
@@ -1,19 +1,51 @@
...
+    return 0;
+}
+
+unsigned node_processors() {
+    // Attempt to get count first from affinity information if available.
+    unsigned n = get_affinity().size();
+
+    // If no luck, try sysconf.
+#ifdef _SC_NPROCESSORS_ONLN
+    if (!n) {
```
bcumming: As much as I like the idiom, `n==0` is a clearer statement of intent.

arbor/partition_load_balance.cpp
```diff
@@ -1,16 +1,20 @@
...
 #include "util/span.hpp"
 
 namespace arb {
 
-domain_decomposition partition_load_balance(const recipe& rec,
-                                            hw::node_info nd,
-                                            const distributed_context* ctx)
+domain_decomposition partition_load_balance(
+    const recipe& rec,
+    domain_info nd,
```
bcumming: I don't fell that `domain_info` is a good name, because I think of a domain as a chunk/part of the model that is running on the local resources.
A more appropriate name might make it clear that this refers to hardware resources.

arbor/partition_load_balance.cpp
```diff
@@ -69,16 +77,30 @@ domain_decomposition partition_load_balance(const recipe& rec,
...
-            groups.push_back({k, std::move(kind_lists[k]), backend_kind::gpu});
+        partition_hint hint;
+        if (auto opt_hint = util::value_by_key(hint_map, k)) {
+            hint = opt_hint.value();
+        }
+
+        backend_kind backend = backend_kind::multicore;
+        std::size_t group_size = hint.cpu_group_size;
+
+        if (hint.prefer_gpu && nd.num_gpus && has_gpu_backend(k)) {
```
bcumming: I know you won't like it, but `nd.num_gpus>0` is clearer.

arbor/simulation.cpp
```diff
@@ -162,7 +162,11 @@ simulation_state::simulation_state(
     // Generate the cell groups in parallel, with one task per cell group.
     cell_groups_.resize(decomp.groups.size());
     foreach_group_index(
```
bcumming: I didn't notice these `foreach_group` functions being added in the last PR.
Should they take the input function as a "forwarding reference", and `std::forward` it onto the tasking system, for completeness sake?

example/bench/bench.cpp
```diff
@@ -11,31 +11,31 @@
...
-using namespace arb;
+namespace profile = arb::profile;
 
 int main(int argc, char** argv) {
     try {
-        distributed_context context;
-        #ifdef ARB_HAVE_MPI
-        mpi::scoped_guard guard(&argc, &argv);
+        arb::distributed_context context;
+#ifdef ARB_HAVE_MPI
```
bcumming: We argued about this very recently. And I am certain that we decided to indent preprocessors.

example/brunel/brunel_miniapp.cpp
```diff
@@ -240,37 +235,31 @@ int main(int argc, char** argv) {
...
-                sim.set_local_spike_callback(
-                    [&](const std::vector<spike>& spikes) {
-                        file_exporter->output(spikes);
-                    }
-                );
+            auto rank = context.id();
+            aux::path p = options.output_path;
+            p /= aux::strsub("%_%.%", options.file_name, rank, options.file_extension);
+
+            if (options.single_file_per_rank) {
```
bcumming: How about a single file for the whole simulation? For a demo, I don't see any point making this an option?
I suppose it was this way in the past to test or justify the existence of this option in the spike_exporter.

example/brunel/brunel_miniapp.cpp
```diff
@@ -304,12 +293,12 @@ int main(int argc, char** argv) {
     return 0;
 }
 
-void banner(hw::node_info nd, const distributed_context* ctx) {
+void banner(domain_info nd, const distributed_context* ctx) {
```
bcumming: Yeah, I reiterate that `domain_info` is not a name I like. `node_info` was more descriptive, but not great either.
`domain_resources`, `local_resources`, `hw_resources`, `local_hw`... ?

halfflat: So, we're splitting a problem into subdomains, and assign each domain to _thing_, which is an element of our set of distributed processes. Quick gander on the web seems to indicate that everyone just calls these 'processes', and so we need a name for the information used to describe the configuration and resources that a process should use to run its part of the problem.

I don't want to use `hw` as the hardware operates at a different, lower level of abstraction. 'Local' describes where these resources come from, but not what they are.

How about one of:
* `process_resource` (or shorter variations, e.g. `proc_res`, `process_res`, `proc_resource`)
* `process_config` (or shorter variations).

example/bench/bench.cpp
halfflat: But these aren't inside a preprocessor scope.

example/brunel/brunel_miniapp.cpp
halfflat: The original code (which was copied from the original miniapp) supported either having each rank write their own spikes, or instead having rank 0 write all of them (see the `set_global_spike_callback` a few lines below). We can certainly nuke this, but here I'm just preserving the existing behaviour.

arbor/simulation.cpp
halfflat: If our tasking system can pull in rvalues, then definitely.

example/bench/bench.cpp
bcumming: But But, you also indented those (see the code that determines the processor affinity.
I find the "just left indent regardless of the scope in which it resides" makes the code harder to read. Just like not indenting `case` statements ;)

arbor/partition_load_balance.cpp
halfflat: Will add, no worries.

arbor/partition_load_balance.cpp
halfflat: Sure — suggestions below.

example/brunel/brunel_miniapp.cpp
bcumming: `proc_resources`?

halfflat: Conclusion from pers. comm: `proc_allocation`

arbor/hardware/node_info.cpp
halfflat: I think is such a common C idiom that there is no need to change.

arbor/hardware/node_info.cpp
halfflat: Cool beans, will change.

arbor/hardware/node_info.cpp
halfflat: `CUDA_SUCCESS` (and success return values in pretty much every C or C++ library) is defined to be zero, and will never change. Similarly `MPI_SUCCESS`. I can't see how making the explicit comparison is helpful, and in fact makes it look like these values might possibly not be zero, which is misleading.

The convention that a zero return from a function (that's not returning a pointer) indicates success is really well established, and there is any amount of code along the lines of `if (some_library_call()) { deal_with_errror(); }`

bcumming: I don't think this is a good argument. You are assuming the value of `CUDA_SUCCESS`, which is not explicitly defined in the documentation anywhere. Why not state what is intended in the code?

Of course, it won't change, but it is a big undocumented leap from `cudaGetDeiceCount==CUDA_SUCCESS` to `!cudaGetDeiceCount`.

arbor/hardware/node_info.cpp
bcumming: I have been avoiding it everywhere, because it isn't explicit (though I don't enjoy the extra typing either). At such a low level it is not a bad thing.

PR #536, opened 2018-07-19
Return view from schedule, replace time_seq.

### LINKED ISSUE #504: Merge `schedule` and `time_seq`
The two classes/wrappers provide nearly identical functionality in different circumstances. The only significant difference comprises:
*  `time_seq` presents a queue interface to time values, avoiding the heap allocations of
*  `schedule`, which returns vectors of time values in an interval, to avoid per-time point virtual function call overheads.

If we replaced the `vector`-returning interface of `schedule` with a const pointer/iterator pair to mutable data, the two interfaces could be merged and we would have a low-overhead bulk method and a queue interface in the same classes.

halfflat:
Reduce redundant functionality across event_generator, time_seq
and schedule by providing a low-heap overhead interface to schedule
and using that for time sequences in event_generator and specialized
cell groups.

* Have schedule return pair of pointers as view to generated times.
* Fix missing DEBUG/TRACE functionality.
* Use rate instead of mean_dt for Poisson schedule.
* Move merge_events() functionality to simulator.cpp.
* Migrate event_generator to event span interface.
* Migrate tourney_tree to event span interface.
* Only invoke tournet_tree merge if generators have events in the epoch.
* Use schedule for times in event_generator implementations.
* Replace seq_generator with explicit_generator that keeps a copy of events.
* Replace vector_backed_generator and poisson_generator with schedule_generator.
* Replace time_seq uses with schedule.
* Add default empty schedule.
* Move rounding error test for regular time sequence into schedule test.

Fixes #504


bcumming:
I am concerned with choosing to return events as a pointer range.

I feel that it is a more complicated design that an iterator oriented solution, however I don't know if it is justified.

The reason for doing this is optimization to avoid per-event virtual call overheads, however:
1. we don't have any evidence that these are actually an overhead that we have to avoid. Measurements of the poisson generator showed that all the overheads were in the random number generation.
2. I doubt that avoiding virtual calls compensates for the overheads of memory allocation and cache pressure associated with storing all pending events in buffers.

We would like users to be able to prescribe time point sequences, and this is much easier for an iterator solution.
We could write adaptors for user-supplied iterator based methods, but I am afraid that this is additional complexity that isn't justified.

halfflat:
Well, I honestly don't think the range approach is so complicated as to need an adaptor; if it is, then it's easy to make one, that's all.

This interface was chosen because it: does avoid virtual call overheads in event generation; and was the smallest adaption from the existing schedule code interface. Rather than have two different time point generation interfaces, it is certainly better to have one unless there are deleterious performance considerations.

bcumming:
I agree that we should have a single time sequence generation interface. That isn't the issue.

I think that the most complicated one has been selected: a pointer range return type requires that each generator maintain a `std::vector` (or similar) buffer internally. It minimised changes to event schedule code, at the expense of increasing changes to all code that used time sequences.

halfflat:
I chose the schedule interface because it was around longer and had more tests associated with it. I think it also has some inherent advantages (batched generation minimizing virtual calls). Do you really want me to redo everything with the queue approach?

bcumming:
I would like to have had this discussion before starting on it.

The event delivery and sampling code used to be pretty much identical, until I spent a lot of time and testing refactoring the event delivery for performance. What we had was quite well optimised, and I had in mind to make a similar refactoring of the sampling code, which would have been a good time to discuss the pros and cons of the two approaches.

halfflat:
I had the ticket up, with a proposed solution, said I was going to do it on slack. Look, I'll nuke the PR and rehash everything in terms of queues.

But I do resent having existing code duplicated in functionality in the first place: if the schedule code could not be made efficient (I believe, based on some testing, that the current view approach is just as efficient), then it should have been changed, and then used elsewhere. This would have avoided a confusion between two APIs, different unit tests, etc. etc.

halfflat:
Reopened (as per pers. comm.), updated branch along code-comment lines, updating documentation for the event generator example and the scheduling API document.

halfflat:
Reopening as per pers. comm.

halfflat:
Rebased on master, updated relevant docs, address code PR comments.

PR #547, opened 2018-07-25
Remove `memory::array` from `arb::tree`

bcumming:
This small refactor simplifies the interface and implementation of the `tree` type.
* use `std::vector` instead of `memory::array` for internal storage in `arb::tree`
* return a `util::range` intstead of a view for `tree::children(int)`
* remove unused functionality for changing the root of a tree.

### DIFF
--- a/arbor/tree.hpp
+++ b/arbor/tree.hpp
@@ -99,9 +99,11 @@ public:
         return children_;
     }
 
-    /// return the list of all children of branch b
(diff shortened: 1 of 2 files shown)


bcumming:
The tree rebalancing code was implemented way back in the day when I was experimenting and thought it might be useful.
To this day, nobody has needed to use it! Deleting that functionality didn't break any tests or miniapps, so out it went.

PR #548, opened 2018-07-25
fix profiler

noraabiakar:
Add initialize method to the profiler to set up the needed threading parameters given a simulation's task system.

### DIFF
--- a/arbor/profile/profiler.cpp
+++ b/arbor/profile/profiler.cpp
@@ -3,10 +3,12 @@
 #include <ostream>
 
 #include <arbor/profile/profiler.hpp>
+#include <arbor/arbexcept.hpp>
 
 #include "threading/threading.hpp"
 #include "util/span.hpp"
 #include "util/rangeutil.hpp"
+#include "util/strprintf.hpp"
 
 namespace arb {
 namespace profile {
@@ -83,6 +85,8 @@ public:
 class profiler {
     std::vector<recorder> recorders_;
 
+    std::unordered_map<std::thread::id, std::size_t> thread_ids_;
+
     // Hash table that maps region names to a unique index.
     // The regions are assigned consecutive indexes in the order that they are
     // added to the profiler with calls to `region_index()`, with the first
@@ -96,9 +100,14 @@ class profiler {
     // Used to protect name_index_, which is shared between all threads.
     std::mutex mutex_;
 
+    // Flag to indicate whether the profiler has been initialized with the task_system
+    bool init_ = false;
+
 public:
     profiler();
+    profiler(unsigned nthreads);
 
+    void initialize(task_system_handle& ts);
     void enter(region_id_type index);
     void enter(const char* name);
     void leave();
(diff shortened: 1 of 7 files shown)


arbor/profile/profiler.cpp
```diff
@@ -96,9 +100,14 @@ class profiler {
     // Used to protect name_index_, which is shared between all threads.
     std::mutex mutex_;
 
+    // Flag to indicate whether the profiler has been initialized with the task_system
+    bool init_ = false;
+
 public:
     profiler();
+    profiler(unsigned nthreads);
```
bcumming: I think the profiler should simply default construct to uninitialized state. A constructor that takes a number of threads while leaving it uninitialized is a bit confusing.

arbor/profile/profiler.cpp
```diff
@@ -169,17 +178,33 @@ profiler::profiler() {
...
+
+void profiler::initialize(task_system_handle& ts) {
+    recorders_.resize(ts.get()->get_num_threads());
+    thread_ids_ = ts.get()->get_thread_ids();
+    init_ = true;
+}
+
 void profiler::enter(region_id_type index) {
-    recorders_[threading::thread_id()].enter(index);
+    if(!init_)
```
bcumming: Style point: curly braces `{}` for `if` statements.

arbor/profile/profiler_macro.hpp
```diff
@@ -4,6 +4,9 @@
 
 #ifdef ARB_HAVE_PROFILING
 
+    // initialize profiler
+    #define PI(ts) arb::profile::profiler_initialize(ts)
```
bcumming: The convenience macros `PE` and `PL` are defined to make them easier to type and less invasive in code.
I don't think a macro is needed for initializing the profiler, because it will only be done once, and it is nice to be clear/explicit about it.

arbor/simulation.cpp
```diff
@@ -220,6 +220,7 @@ time_type simulation_state::run(time_type tfinal, time_type dt) {
     // If spike exchange and cell update are serialized, this is the
     // minimum delay of the network, however we use half this period
     // to overlap communication and computation.
+    PI(task_system_);
```
bcumming: This should be done in the calling code.

The profiler is turned on/off, started/stopped etc by the calling code. In our case this will be in the miniapps.

arbor/profile/profiler.cpp
```diff
@@ -169,17 +178,33 @@ profiler::profiler() {
...
+        throw arbor_exception(util::pprintf(
+                "Profiler not initialized with correct task_system parameters"));
+    recorders_[thread_ids_.at(std::this_thread::get_id())].enter(index);
 }
 
 void profiler::enter(const char* name) {
+    if(!init_)
```
bcumming: How about just returning if uninitialized? That way if the application has been compiled with profiler support on, but the user does not initialize the profiler, nothing is recorded.

PR #555, opened 2018-07-26
Manage distributed_context using shared pointers

noraabiakar:
Execution_context now holds a shared pointer distributed_context and a shared pointer to a task_system.


arbor/communication/communicator.hpp
```diff
@@ -46,10 +46,10 @@ class communicator {
                           const domain_decomposition& dom_dec,
                           const execution_context* ctx)
     {
-        distributed_ = &ctx->distributed;
+        distributed_ = ctx->distributed;
         thread_pool_ = ctx->thread_pool;
 
-        num_domains_ = distributed_->size();
+        num_domains_ = distributed_.get()->size();
```
bcumming: `std::shared_ptr` overloads `operator->`, so you can replace `distributed_.get()->size()` with `distributed_->size()`.

example/miniapp/miniapp.cpp
```diff
@@ -49,17 +49,17 @@ int main(int argc, char** argv) {
     try {
 #ifdef ARB_MPI_ENABLED
         with_mpi guard(argc, argv, false);
-        context.distributed = mpi_context(MPI_COMM_WORLD);
+        context.distributed = distributed_context_handle(new distributed_context(mpi_context(MPI_COMM_WORLD)));
```
apeyser: Should this be `make_shared` ?

apeyser: Yes -- I think so. `distributed_context_handle` is just a `typedef` to `std::shared_ptr<distributed_context>`, so we should either directly call `std::make_shared<distributed_context>` or create a call that maps onto that.

include/arbor/execution_context.hpp
```diff
@@ -14,16 +14,19 @@ namespace threading {
...
 
 struct execution_context {
-    // TODO: use a shared_ptr for distributed_context
-    distributed_context distributed;
+    distributed_context_handle distributed;
     task_system_handle thread_pool;
 
-    execution_context(): thread_pool(arb::make_thread_pool(arb::num_threads())) {};
-    execution_context(proc_allocation nd): thread_pool(arb::make_thread_pool(nd.num_threads)) {};
+    execution_context(): distributed(new distributed_context),
```
apeyser: Same thing here -- `make_shared` would be the proper replacement for `new`

example/bench/bench.cpp
```diff
@@ -35,21 +35,21 @@ int main(int argc, char** argv) {
     try {
         arb::execution_context context;
 #ifdef ARB_MPI_ENABLED
-        aux::with_mpi guard(&argc, &argv);
-        context.distributed = mpi_context(MPI_COMM_WORLD);
+        with_mpi guard(argc, argv, false);
+        context.distributed = std::make_shared<arb::distributed_context>(arb::mpi_context(MPI_COMM_WORLD));
```
halfflat: Now that we're using a `shared_ptr` for the distributed context, we can have context-makers like `arb::mpi_context` return a `shared_ptr` directly.

This points towards moving away from value-like semantics for the type-erasing `distributed_context` class, but that's something I think I will have to arm wrestle @bcumming over, and can be the topic of a later issue.

arbor/local_alloc.cpp
```diff
@@ -1,13 +1,14 @@
...
-#include <arbor/threadinfo.hpp>
+#include <arbor/execution_context.hpp>
+#include <threading/threading.hpp>
 
 #include "hardware/node_info.hpp"
 
 namespace arb {
 
-proc_allocation local_allocation() {
+proc_allocation local_allocation(execution_context* ctx) {
```
halfflat: Unless we have a good reason to use pass-by-pointer, which tends to obscure the lifetime requirements when used in a function interface, I'd recommend defaulting to pass-by-value or pass-by-const-reference. Presuming that we keep `execution_context` as essentially a bag of shared pointers, I'd prefer to use pass-by-value when it's going to be copied and used in some stateful object, and pass-by-const-reference otherwise where it makes sense.

PR #563, opened 2018-08-03
Bugfix/build osx macports

### LINKED ISSUE #562: compile error on Mac with gcc7.3.0 in arbor/arbor/profile/meter_manager.cpp
@halfflat 

# Issue:
code does not compile on Mac with gcc 7.3.0 or clang 9.1.0

## Build: 
built with gcc 7.3.0/clang 9.1.0  using cmake (version 3.11.4) in release mode
```
/opt/local/bin/gcc-mp-7 --version                                                                                                                                                                                                               
gcc-mp-7 (MacPorts gcc7 7.3.0_1) 7.3.0

cmake -DCMAKE_C_COMPILER=/opt/local/bin/gcc-mp-7 -DCMAKE_CXX_COMPILER=/opt/local/bin/g++-mp-7 -DCMAKE_BUILD_TYPE:STRING=Release ..
-- The C compiler identification is GNU 7.3.0
-- The CXX compiler identification is GNU 7.3.0
-- Checking whether C compiler has -isysroot
-- Checking whether C compiler has -isysroot - yes
-- Checking whether C compiler supports OSX deployment target flag
-- Checking whether C compiler supports OSX deployment target flag - yes
-- Check for working C compiler: /opt/local/bin/gcc-mp-7
-- Check for working C compiler: /opt/local/bin/gcc-mp-7 -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Checking whether CXX compiler has -isysroot
-- Checking whether CXX compiler has -isysroot - yes
-- Checking whether CXX compiler supports OSX deployment target flag
-- Checking whether CXX compiler supports OSX deployment target flag - yes
-- Check for working CXX compiler: /opt/local/bin/g++-mp-7
-- Check for working CXX compiler: /opt/local/bin/g++-mp-7 -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - found
-- Found Threads: TRUE  
-- Could NOT find Unwind (missing: Unwind_LIBRARIES) 
-- Found Sphinx: /opt/local/bin/sphinx-build  
-- Configuring done
-- Generating done
-- Build files have been written to:
```

unwind is not found although installed *libunwind @5.0.1 (devel)* and *libunwind-headers @5.0.1 (devel)* with macports.

## Error: 
happens in arbor/arbor/profile/meter_manager.cpp:27:30 (min) and again in 27:54 (max)

```
Scanning dependencies of target arbor
[ 32%] Building CXX object arbor/CMakeFiles/arbor.dir/profile/meter_manager.cpp.o
arbor/arbor/profile/meter_manager.cpp: In constructor 'arb::profile::measurement::measurement(std::__cxx11::string, std::__cxx11::string, const std::vector<double>&, const distributed_context_handle&)':
arbor/arbor/profile/meter_manager.cpp:27:30: error: call of overloaded 'min(const long unsigned int&)' is ambiguous
     if (ctx->min(num_readings)!=ctx->max(num_readings)) {
                              ^
In file included from arbor/include/arbor/execution_context.hpp:6:0,
                 from arbor/include/arbor/profile/meter_manager.hpp:7,
                 from arbor/arbor/profile/meter_manager.cpp:3:
arbor/include/arbor/distributed_context.hpp:13:7: note: candidate: float arb::distributed_context::min(float) const
     T min(T value) const { return impl_->min(value); }\
       ^

...

arbor/arbor/profile/meter_manager.cpp:27:54: error: call of overloaded 'max(const long unsigned int&)' is ambiguous
     if (ctx->min(num_readings)!=ctx->max(num_readings)) {
                                                      ^
In file included from arbor/include/arbor/execution_context.hpp:6:0,
                 from arbor/include/arbor/profile/meter_manager.hpp:7,
                 from arbor/arbor/profile/meter_manager.cpp:3:
arbor/include/arbor/distributed_context.hpp:14:7: note: candidate: float arb::distributed_context::max(float) const
     T max(T value) const { return impl_->max(value); }\
       ^

...
```

## Suggestion of source of error: 
Looks like a *macro issue* for max() and min() and an inability to decide between *int and float*.

halfflat:
Two MacPorts/gcc7 issues:
1. `std::uint64_t` is `unsigned long long` on OS X, breaking an assumption about `size_t` in the `distributed_context` interface.
2. Problems with missing errno defines in the standard library headers.

With MacPorts gcc7, the installed `c++config.h` defines `_GLIBCXX_HAVE_EOWNERDEAD` and `_GLIBCXX_HAVE_ENOTRECOVERABLE`, but the corresponding errno defines are not provided by
`sys/errno.h` unless `__DARWIN_C_SOURCE`, which takes its value from `_POSIX_C_SOURCE` if defined, is greater than or equal to 200809L. Technically a MacPorts configuration bug? but easily worked around.

* Use basic integral types for communication collectives interfaces.
* Define `_POSIX_C_SOURCE` to be 200809L for `glob.cpp`.

Fixes #562.

### DIFF
--- a/arbor/communication/mpi.hpp
+++ b/arbor/communication/mpi.hpp
@@ -48,15 +48,19 @@ struct mpi_traits<T> {  \
     constexpr static bool is_mpi_native_type() { return true; } \
 };
 
-MAKE_TRAITS(double, MPI_DOUBLE)
-MAKE_TRAITS(float,  MPI_FLOAT)
-MAKE_TRAITS(int,    MPI_INT)
-MAKE_TRAITS(long int, MPI_LONG)
-MAKE_TRAITS(char,   MPI_CHAR)
-MAKE_TRAITS(unsigned int, MPI_UNSIGNED)
-MAKE_TRAITS(size_t, MPI_UNSIGNED_LONG)
-static_assert(sizeof(size_t)==sizeof(unsigned long),
-              "size_t and unsigned long are not equivalent");
+MAKE_TRAITS(float,              MPI_FLOAT)
+MAKE_TRAITS(double,             MPI_DOUBLE)
+MAKE_TRAITS(char,               MPI_CHAR)
+MAKE_TRAITS(int,                MPI_INT)
+MAKE_TRAITS(unsigned,           MPI_UNSIGNED)
+MAKE_TRAITS(long,               MPI_LONG)
+MAKE_TRAITS(unsigned long,      MPI_UNSIGNED_LONG)
+MAKE_TRAITS(long long,          MPI_LONG_LONG)
+MAKE_TRAITS(unsigned long long, MPI_UNSIGNED_LONG_LONG)
+
+static_assert(std::is_same<std::size_t, unsigned long>::value ||
+              std::is_same<std::size_t, unsigned long long>::value,
+              "size_t is not the same as unsigned long or unsigned long long");
 
 // Gather individual values of type T from each rank into a std::vector on
 // the root rank.
--- a/aux/glob.cpp
+++ b/aux/glob.cpp
@@ -1,6 +1,6 @@
 // POSIX headers
 extern "C" {
-#define _POSIX_C_SOURCE 2
+#define _POSIX_C_SOURCE 200809L
 #include <glob.h>
 }
 
--- a/include/arbor/distributed_context.hpp
+++ b/include/arbor/distributed_context.hpp
@@ -27,7 +27,7 @@ namespace arb {
     T sum(T value) const override { return wrapped.sum(value); }\
     std::vector<T> gather(T value, int root) const override { return wrapped.gather(value, root); }
 
-#define ARB_COLLECTIVE_TYPES_ float, double, int, std::uint32_t, std::uint64_t
+#define ARB_COLLECTIVE_TYPES_ float, double, int, unsigned, long, unsigned long, long long, unsigned long long
 
 // Defines the concept/interface for a distributed communication context.
 //


arbor/communication/mpi.hpp
```diff
@@ -48,15 +48,19 @@ struct mpi_traits<T> {  \
...
-MAKE_TRAITS(long int, MPI_LONG)
-MAKE_TRAITS(char,   MPI_CHAR)
-MAKE_TRAITS(unsigned int, MPI_UNSIGNED)
-MAKE_TRAITS(size_t, MPI_UNSIGNED_LONG)
-static_assert(sizeof(size_t)==sizeof(unsigned long),
-              "size_t and unsigned long are not equivalent");
+MAKE_TRAITS(float,              MPI_FLOAT)
+MAKE_TRAITS(double,             MPI_DOUBLE)
+MAKE_TRAITS(char,               MPI_CHAR)
+MAKE_TRAITS(int,                MPI_INT)
```
apeyser: Should we be explicit about int being signed int for documentary purposes?

apeyser: So now we have traits for types that should never be used? I'm wondering if there's a way to 'break' the types by default, and then fix the size_t type.

halfflat: I think it is safe to assume that people with the fingers in the code at this point won't need the signedness of int to be made explicit — indeed, I think adding in signed just here would make the less confident coder wonder if there is something they are missing elsewhere, and thus increase confusion.

We already had traits for types that don't have an interface in `distributed_context`, but I think that's okay. Having the traits for the basic numeric scalar types is fine IMO. I wouldn't object if `short` and `unsigned short` were added too, for example.

halfflat: Regarding the problem of `size_t`: the maximally portable approach involves a huge amount of type traits guff, and the other portable alternative is to do configure-time checks against the various int types. I'm really, really reluctant to put that work in :)

You can all curse me when we port it to the first system where `size_t`, `uintmax_t`, and `uint128_t` are all the same while `unsigned long long` is 64 bit.

bcumming: I vote for keeping it simple @halfflat.

How about a quick `static_assert` that borks when the highly unlikely scenario that you just described occurs, with a comment attached that explains the situation?

halfflat: @bcumming The appropriate `static_assert` is in the PR :)

PR #565, opened 2018-08-08
Global temperature for NMODL mechanisms.

### LINKED ISSUE #386: Setting temperature in model

halfflat:
Global temperature for mechanisms.

* Make 'celsius' magic in modcc: now an indexed variable.
* Add a new temperature data source for indexed variables.
* Add support to printers for indexed variables that reference a scalar.
* Check that indexed variables aren't used in PROCEDURE blocks (this is a problem not just for 'celsius').
* Modify built-in mod files to pass celsius as a parameter to rates() procedures.
* Add global temperature to shared_state classes, and initialize through backend mechanism superclasses.
* Add some infrastructure for unit-test only mechanisms.
* Set modcc flags globally in top level CMakeLists.txt.
* Add test mechanism/module for checking celsius setting.
* Add unit test for multicore and gpu mechanism celsius setting.
* Make common mechanism private field data access helper for unit tests.
* Use helper in temperature, synapses tests.
* Fix warning in `distributed_context.hpp` about errant semicolon.
* Use `ARB_CXXOPT_ARCH` not `CXXOPT_ARCH` in CMakeLists.txt files.

Fixes #386.

### DIFF
--- a/test/unit/CMakeLists.txt
+++ b/test/unit/CMakeLists.txt
@@ -1,10 +1,40 @@
+# Build unit test mechanisms.
+
+set(test_mechanisms
+    celsius_test
+)
+
+include(${PROJECT_SOURCE_DIR}/mechanisms/BuildModules.cmake)
+
+set(external_modcc)
+if(ARB_WITH_EXTERNAL_MODCC)
+    set(external_modcc MODCC ${modcc})
+endif()
+set(test_mech_dir ${CMAKE_CURRENT_BINARY_DIR}/mechanisms)
+
+build_modules(
+    ${test_mechanisms}
+    SOURCE_DIR mod
+    DEST_DIR "${test_mech_dir}"
+    ${external_modcc}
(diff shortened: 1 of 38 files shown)


test/unit/CMakeLists.txt
```diff
@@ -1,10 +1,40 @@
+# Build unit test mechanisms.
+
+set(test_mechanisms
+    celsius_test
```
bcumming: Let's make a ticket for moving the other test-only mechanisms (e.g. `test_kin1` etc) into here.

bcumming: Also a more detailed comment above that makes it clear that these are mechanisms used only in unit tests, hence why we don't make them elsewhere.

halfflat: Updated comment in-place.

PR #566, opened 2018-08-16
Create gpu_context and manage it as part of execution_context

noraabiakar:
* Create gpu_context as part of execution context
* gpu_context contains information about GPU availability, managed_memory availability; atomic double availability 
* We still need compile time information about the GPU in cuda_atomic.hpp for atomicAdd(double*, double*). This is because the function is only defined when the program is compiled  for sm_60 or more. 
* Choose between ON and OFF for ARB_GPU in CMake. If ON compile for K20, K80, and P100


arbor/backends/gpu/stack.hpp
```diff
@@ -56,7 +58,9 @@ class stack {
         return *this;
     }
 
-    explicit stack(unsigned capacity): storage_(create_storage(capacity)) {}
+    explicit stack(unsigned capacity, unsigned has_concurrent_managed_access):
```
halfflat: That `unsigned has_concurrent_managed_access` should probably be a `bool`.

include/arbor/gpu_context.hpp
```diff
@@ -0,0 +1,18 @@
...
+    bool has_gpu_;
+    size_t attributes_;
+    gpu_context();
+
+#ifdef ARB_GPU_ENABLED
+    bool has_concurrent_managed_access();
+    bool has_atomic_double();
+#endif
+};
+}
```
halfflat: Rather than have the GPU-dependent code in the public header, we should arrange things such that to users of the code, the context is opaque (e.g. just a forward declaration `struct gpu_context`) — it might not even need its own header, instead being forward-declared in `execution_context.hpp`.

Then the actual contents of the `gpu_context` structure would be present in a private library header for consumption by the library code. The `#ifdef` guard would then check `ARB_HAVE_GPU` as before.

arbor/gpu_context.cpp
```diff
@@ -0,0 +1,46 @@
...
+enum gpu_flags {
+    has_concurrent_managed_access = 0,
+    has_atomic_double = 1
+};
+
+size_t get_attributes() {
+    size_t attributes = 0;
+    cudaDeviceProp prop;
+    cudaGetDeviceProperties(&prop, 0);
+    if(prop.concurrentManagedAccess)
```
bcumming: Formatting:
* space after `if`
* curly braces.

arbor/backends/gpu/stack.hpp
noraabiakar: Thanks for pointing it out!

CMakeLists.txt
```diff
@@ -169,29 +169,24 @@ endif()
...
+    set(cuda_code_1 sm_37)
+    set(cuda_arch_1 compute_37)
+    set(cuda_code_2 sm_60)
+    set(cuda_arch_2 compute_60)
 
-    target_compile_options(arbor-private-deps INTERFACE
-        $<$<COMPILE_LANGUAGE:CUDA>:-arch=${cuda_arch}>)
+    add_compile_options($<$<COMPILE_LANGUAGE:CUDA>:-gencode=arch=${cuda_arch_0},code=${cuda_code_0}>)
+    add_compile_options($<$<COMPILE_LANGUAGE:CUDA>:-gencode=arch=${cuda_arch_1},code=${cuda_code_1}>)
+    add_compile_options($<$<COMPILE_LANGUAGE:CUDA>:-gencode=arch=${cuda_arch_2},code=${cuda_code_2}>)
```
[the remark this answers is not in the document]
noraabiakar: We don't. I'll simplify

arbor/backends/gpu/managed_ptr.hpp
```diff
@@ -42,8 +43,8 @@ class managed_ptr {
     using pointer = element_type*;
     using reference = element_type&;
 
-    managed_ptr():
-        concurrent_managed_access(device_concurrent_managed_access())
+    managed_ptr(bool has_concurrent_managed_access):
```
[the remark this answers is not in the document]
noraabiakar: managed_ptr has a local synchronize method used in its constructor and destructor. I think these make sense to remain as they are. Were you referring to the host_access method?

CMakeLists.txt
```diff
@@ -169,30 +165,18 @@ endif()
...
-        set(arb_cuda_arch 370)
-    elseif(ARB_GPU_MODEL MATCHES "P100")
-        set(cuda_arch sm_60)
-        set(arb_cuda_arch 600)
-    else()
-        message(FATAL_ERROR "-- GPU architecture '${ARB_GPU_MODEL}' not supported. Use one of {none, K20, K80, P100}")
-    endif()
+    add_compile_options($<$<COMPILE_LANGUAGE:CUDA>:-gencode=arch=compute_35,code=sm_35>)
+    add_compile_options($<$<COMPILE_LANGUAGE:CUDA>:-gencode=arch=compute_37,code=sm_37>)
+    add_compile_options($<$<COMPILE_LANGUAGE:CUDA>:-gencode=arch=compute_60,code=sm_60>)
```
halfflat: This will apply the options to every nvcc invocation, where we actually just need to compile for all these targets on the library: the unit tests should be fine using whatever the defaults are. I'll edit this one in-place to change these to `target_compile_options`.

arbor/gpu_context.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+    gpu_context(): has_gpu_(false), attributes_(0) {}
+};
+
+#else
+
+enum gpu_flags {
+    has_concurrent_managed_access = 0,
+    has_atomic_double = 1
+};
```
halfflat: Rather than have the bit positions in the flags, I think it's simpler (and more traditional) to store the mask values in the flags, i.e. (1, 2, 4, 8 etc.).

PR #572, opened 2018-08-21
Work-around for gcc version < 8.2 versus std::fma

### LINKED ISSUE #568: FMA arithmetic tests failing for vectorization on daint
Some SIMD unit tests are failing when Arbor is compiled with vectorization support on Daint

To build:

```
module load daint-mc
module swap PrgEnv-cray PrgEnv-gnu
module swap gcc/7.3.0
module load CMake/3.11.4

export CRAYPE_LINK_TYPE=dynamic
export CC=`which gcc`; export CXX=`which g++`;


cmake .. -DARB_VECTORIZE=ON -DARB_ARCH=broadwell
```

The failing tests are `S/simd_value/*.arithmetic`, specifically those that compare `std::fma` with our vectorized `fma` operation.

halfflat:
Use a `compat::fma` wrapper for `std::fma` to avoid a bug in the tree optimizer in GCC version < 8.2.

See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87046
Fixes #568.

### DIFF
--- a/include/arbor/util/compat.hpp
+++ b/include/arbor/util/compat.hpp
@@ -37,4 +37,14 @@ inline void compiler_barrier_if_icc_leq(unsigned ver) {
 #endif
 }
 
+// Work-around for bad vectorization of fma in gcc version < 8.2
+
+template <typename T>
+#if defined(__GNUC__) && (100*__GNUC__ + __GNUC_MINOR__ < 802)
+__attribute((optimize("no-tree-vectorize")))
+#endif
+inline auto fma(T a, T b, T c) {
+    return std::fma(a, b, c);
+}
+
 } // namespace compat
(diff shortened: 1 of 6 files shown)


include/arbor/util/compat.hpp
```diff
@@ -37,4 +37,14 @@ inline void compiler_barrier_if_icc_leq(unsigned ver) {
 #endif
 }
 
+// Work-around for bad vectorization of fma in gcc version < 8.2
+
+template <typename T>
+#if defined(__GNUC__) && (100*__GNUC__ + __GNUC_MINOR__ < 802)
```
bcumming: Have you tested for gcc 8.1? I suppose gcc 8.x isn't available on daint, so I assume that you verified 8.2 by looking at the assembly generated for your reduced reproducer on your desktop?
It would be nice to draw a line at gcc 8.

halfflat: The 8.2 test was based on the comment on the gcc bug report, but correct behaviour in the arbor unit test has not been verified with 8.2 — I didn't want to have to install gcc on greina :)

PR #588, opened 2018-09-05
Clarify vectorization-enabled build errors.

### LINKED ISSUE #587: Clang fails to compile the vectorized backends
I have tried using Clang 6.0.1. This is the build script I use:

```
mkdir build
cd build
cmake .. -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DARB_WITH_ASSERTIONS=ON -DARB_VECTORIZE=ON #-DARB_ARCH=native
make -j
```

I have tried also enabling the ARB_ARCH variable, but again fails. The errors I am getting are of several kinds. Here are some of them:

```
mechanisms/generated/kdrmt_cpu.cpp:118:20: error: no matching constructor for initialization of 'arb::simd_value' (aka 'arb::simd::simd_detail::simd_impl<void>')

arbor/simd/simd.hpp:106:50: error: array has incomplete element type 'const arb::simd::simd_detail::simd_impl<void>::scalar_type' (aka 'const void')

etc.
```
Same behaviour with AppleClang 9.1.0.
Non-vectorized version compiles fine.

halfflat:
Fixes #587.

* Eliminate Clang warnings from GCC-tree-optimization bug work-around.
* Error with static-assert if simd type is used with a missing simd abi.
* Clarify install documentation regarding use of ARB_VECTORIZE with ARB_ARCH.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -287,16 +287,17 @@ Vectorization
 -------------
 
 Explicit vectorization of computational kernels can be enabled in Arbor by setting the
-``ARB_VECTORIZE`` CMake flag:
+``ARB_VECTORIZE`` CMake flag. This option is typically used in conjunction with the
+``ARB_ARCH`` option to specify the target architecture: without SIMD support in Arbor
+for the architecture, enabling ``ARB_VECTORIZE`` will lead to a compilation error.
 
 .. code-block:: bash
 
-    cmake -DARB_VECTORIZE=ON
+    cmake -DARB_VECTORIZE=ON -DARB_ARCH=native
 
 With this flag set, the library will use architecture-specific vectorization intrinsics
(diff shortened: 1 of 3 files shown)


doc/install.rst
```diff
@@ -287,16 +287,17 @@ Vectorization
 -------------
 
 Explicit vectorization of computational kernels can be enabled in Arbor by setting the
-``ARB_VECTORIZE`` CMake flag:
+``ARB_VECTORIZE`` CMake flag. This option is typically used in conjunction with the
+``ARB_ARCH`` option to specify the target architecture: without SIMD support in Arbor
+for the architecture, enabling ``ARB_VECTORIZE`` will lead to a compilation error.
```
bcumming: :+1:

PR #590, opened 2018-09-07
wrapped meter_manager in Python, with support for printing meter_report

akuesters:
Also added a simple test profiler.py in Python/test for printing


python/context.cpp
```diff
@@ -15,10 +16,10 @@
 namespace arb {
 namespace py {
 
-struct context_shim {
+/*struct context_shim {
```
bcumming: Remove these commented out lines completely.

python/profiler.cpp
```diff
@@ -0,0 +1,56 @@
...
+    // metering
+    //
+    pybind11::class_<arb::profile::meter_manager> meter_manager(m, "meter_manager");
+
+    meter_manager
+        .def(pybind11::init<>())
+
+        // Need to use shimming of context due to pybind11 complaining about incomplete type of execution_cntext  
+        .def("start", [](arb::profile::meter_manager& manager, const arb::py::context_shim& ctx){\
+            manager.start(ctx.context);\
```
bcumming: You don't need the `\` on the end of split lines in C++. It is only every used when splitting strings over multiple lines.

python/profiler.cpp
```diff
@@ -0,0 +1,56 @@
...
+    //
+    pybind11::class_<arb::profile::meter_report> meter_report(m, "meter_report");
+
+    meter_report
+        .def("__str__", [](const arb::profile::meter_report& report){\
+            std::stringstream s;
+            s << "meter report:\n";
+            s << report;
+            return s.str();
+            //return "<pyarb.meter_report>";
```
bcumming: Remove this commented out line, and the one below for `__repr__`.

python/profiler.cpp
```diff
@@ -0,0 +1,56 @@
...
+        })
+        .def("__repr__", [](const arb::profile::meter_report& report){\
+            std::stringstream s;
+            s << "meter report:\n";
+            s << report;
+            return s.str();
+            //return "<pyarb.meter_report>";
+        });
+
+    m.def("make_meter_report", [](const arb::profile::meter_manager& manager, const arb::py::context_shim& ctx){\
```
bcumming: I would indent this as follows:

```C++
    m.def("make_meter_report",
        [](const arb::profile::meter_manager& manager, const arb::py::context_shim& ctx){
            return arb::profile::make_meter_report(manager, ctx.context);
        },
        "manager"_a,
        "context"_a,
        "Generate a meter_report from a set of meters.");
```

The `"manager"_a` and `"context"_a` are "string literals" that are short hand for `pybind11::arg("manager")`.
You can enable them by putting this on the first line of the `register_profilers` function:

```C++
```

PR #592, opened 2018-09-07
repair compiler warnings with AppleClang

bcumming:
Turns out that CMake thinks Clang and AppleClang are different things.

This PR makes both CMake and AppleClang disable the same compiler warnings.


bcumming:
Turns out that CMake thinks Clang and AppleClang are different things.

PR #595, opened 2018-09-10
Threading exceptions

noraabiakar:
Propagate exceptions in the threading backend to the wait() function in the task_group.


arbor/threading/threading.hpp
```diff
@@ -97,10 +97,27 @@ class task_system {
 
 class task_group {
 private:
+    struct exception_struct {
```
bcumming: The `_struct` is redundant, and could be replaced with something that describes what the struct is for, like `exception_state_`?
The members could also be simplified to
* `exception_ptr_` --> `exception_`
* `exception_mutext_` --> `mutex_`
* `bail` --> `error_`

arbor/threading/threading.hpp
```diff
@@ -97,10 +97,27 @@ class task_system {
...
+        std::atomic<bool> bail{false};
+        std::exception_ptr exception_ptr_;
+        std::mutex exception_mutex_;
+
+        bool have_exception() {
+            return bail.load(std::memory_order_relaxed);
+        }
+
+        void set_exception(std::exception_ptr ex) {
```
bcumming: Rename to `set`? This type has only one job: to manage exception state, so `exception_state::set` is suitably descriptive.

arbor/threading/threading.hpp
```diff
@@ -114,30 +131,41 @@ class task_group {
     class wrap {
         F f;
         std::atomic<std::size_t>& counter;
+        exception_struct& exception_status;
```
bcumming: I just noticed that the private members of this class don't have `_` on the end of the variable names, e.g. `counter` should be `counter_`, `f` should be `f_`, etc.

arbor/threading/threading.hpp
```diff
@@ -179,7 +208,18 @@ struct parallel_for {
         for (int i = left; i < right; ++i) {
           g.run([=] {f(i);});
         }
-        g.wait();
+        auto ex = g.wait();
+
+        if(ex) {
+            try
+            {
```
bcumming: one line `try {`.

arbor/threading/threading.hpp
```diff
@@ -179,7 +208,18 @@ struct parallel_for {
...
+                std::rethrow_exception(ex);
+            }
+            catch (const std::exception &e)
+            {
+                throw std::runtime_error(e.what());
```
bcumming: Instead of `throw::runtime_error(e.what())` should it be `std::rethrow_exception(e)`?

arbor/threading/threading.hpp
```diff
@@ -146,21 +174,22 @@ class task_group {
...
     template<typename F>
     void run(F&& f) {
         ++in_flight_;
-        task_system_->async(make_wrapped_function(std::forward<F>(f), in_flight_));
+        task_system_->async(make_wrapped_function(std::forward<F>(f), in_flight_, exception_status_));
     }
 
     // wait till all tasks in this group are done
-    void wait() {
+    std::exception_ptr wait() {
```
bcumming: I think it makes more sense to check for an exception here, and rethrow it.
Otherwise we will have exceptions being silently ignored.
@halfflat, what are your feelings about this?

Our motivating example, where an exception thrown during model building from a recipe would still not work as expected, without adding tests to the return value of every call to `wait` inside `simulation`.

bcumming: An alternative, would be to make the destructor of `exception_struct` rethrow the exception if it has not explicitly been acquired or rethrown.
That way, if we ignore the return value of `task_group::wait()`, any exception will be rethrown when it is destructed.

test/unit/test_threading_exceptions.cpp
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    std::vector<cell_connection> connections_on(cell_gid_type) const override {
+        return std::vector<cell_connection>();
+    }
+
+    cell_size_type n_, boom_;
+};
+
+void run_model_ctor(const recipe& rec) {
```
bcumming: This should be tested more thoroughly at a lower level.

Tests should cover:
* what is the expected behavior when more than one thread throws an exception?
* what state is a `task_group` in after an exception has been thrown?
* is the correct exception thrown on failure of a `task_group`?

arbor/threading/threading.hpp
```diff
@@ -97,10 +97,27 @@ class task_system {
 
 class task_group {
 private:
+    struct exception_struct {
+        std::atomic<bool> bail{false};
+        std::exception_ptr exception_ptr_;
+        std::mutex exception_mutex_;
+
+        bool have_exception() {
```
bcumming: This could be rewritten to convert to `bool`, i.e.
```C++
operator bool() const {
    return bail.load(std::memory_order_relaxed);
}
```

arbor/threading/threading.hpp
```diff
@@ -97,10 +97,27 @@ class task_system {
...
+            return bail.load(std::memory_order_relaxed);
+        }
+
+        void set_exception(std::exception_ptr ex) {
+            bail.store(true, std::memory_order_relaxed);
+            lock ex_lock{exception_mutex_};
+            exception_ptr_ = std::move(ex);
+        }
+    };
+
```
bcumming: I would propose the following:

* add a `get` method that will return the `exception_ptr` and reset `bail`  to false.
* make the destructor explicitly `std::rethrow_exception` on the `exception_pointer` if `bail` is still set.

This would mean that if users ignore the returned exception state, and it has been set, the exception is thrown for them.

arbor/threading/threading.hpp
noraabiakar: Thanks for the name suggestions and style comments. I have modified all the things you pointed out :)

arbor/threading/threading.hpp
noraabiakar: rethrow_exception takes an exception_ptr, so that won't work I think. Should I throw something other than a runtime_error?

arbor/threading/threading.hpp
noraabiakar: Is this in relation to your earlier comment: "An alternative, would be to make the destructor of exception_struct rethrow the exception if it has not explicitly been acquired or rethrown.
That way, if we ignore the return value of task_group::wait(), any exception will be rethrown when it is destructed." ? 
If yes, do you mean we should call the get() method inside wait() ? Because that wouldn't solve the problem. If the user ignores the returned exception state, the destructor still won't throw the exception. 
Or are you saying that we would expose the get() method to the user, and wait() would just return void? Or are you saying something completely different?

arbor/threading/threading.hpp
bcumming: Sorry, I didn't read that code very well when reviewing it. I see that you are already trying `std::rethrow_exception`.
Then, I wonder if just `throw;` will rethrow `e`?

bcumming:
@noraabiakar, did the propagation of exceptions from inner task groups to outer task groups "just work" like we thought it should?

noraabiakar:
@bcumming, yes it did! Turns out, my unit test was expecting a task group with zero tasks to throw an exception which of course won't happen. I fixed my loops so we don't have that problem.

PR #601, opened 2018-09-24
Integrating Mac OS X and clang compiler into Travis CI

akuesters:
changes: 
- .travis.yml:
  - added matrix for different osx's, since enumeration style only works for `env` and `compiler`

- scripts/travis/build.sh:
  - changed getting compiler version from ``${CXX} -dumpversion`` to ``${CXX} --version | grep -m1 ""`` 
  - added `--oversubscribe` flag to `mpiexec` on Mac to allow more processes on a node than processing elements
  - added `--mca btl tcp,self` flag for Open MPI to use the "tcp" and "self" BTLs for transporting MPI messages on Mac

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,64 +1,26 @@
 language: cpp
 sudo: false
+os: linux
+dist: trusty
+compiler: gcc
 
-matrix:
-  include:
-#################### OS X #####################
-    - os: osx 
-      osx_image: xcode10
-      env:
-        - MATRIX_EVAL="brew install gcc@6 && brew link --force --overwrite gcc@6 && CC=gcc-6 && CXX=g++-6"
-        - BUILD_NAME=cthread WITH_DISTRIBUTED=serial
-      compiler: gcc
-
-    - os: osx 
-      osx_image: xcode10
-      env:
-        - MATRIX_EVAL="brew install gcc@6 && brew link --force --overwrite gcc@6 && CC=gcc-6 && CXX=g++-6"
-        - BUILD_NAME=mpi WITH_DISTRIBUTED=mpi
-      compiler: gcc
-
-#################### LINUX #####################
-    - os: linux
-      dist: trusty
-      addons:
-        apt:
-          sources:
+addons:
+    apt:
+        sources:
             - ubuntu-toolchain-r-test
-          packages:
+        packages:
             - g++-6
             - openmpi-bin
             - libopenmpi-dev
-      env:
-        - MATRIX_EVAL="CC=gcc-6 && CXX=g++-6"
-        - BUILD_NAME=serial WITH_DISTRIBUTED=serial
-      compiler: gcc-6
 
-    - os: linux
-      dist: trusty
-      addons:
-        apt:
-          sources:
-            - ubuntu-toolchain-r-test
-          packages:
-            - g++-6
-            - openmpi-bin
-            - libopenmpi-dev
-      env:
-        - MATRIX_EVAL="CC=gcc-6 && CXX=g++-6"
-        - BUILD_NAME=mpi WITH_DISTRIBUTED=mpi
-      compiler: gcc-6
+env:
+    # test single node/rank with threading backend
+    - BUILD_NAME=cthread  WITH_DISTRIBUTED=serial
+    # test mpi with threading backend
+    - BUILD_NAME=mpi      WITH_DISTRIBUTED=mpi
 
-before_script:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then export HOMEBREW_NO_AUTO_UPDATE=1; brew cask uninstall --force oclint; fi
-  - if [[ ( "$TRAVIS_OS_NAME" == "osx" ) && ( "$WITH_DISTRIBUTED" == "mpi" ) ]]; then brew install open-mpi; fi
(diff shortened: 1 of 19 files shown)


.travis.yml
```diff
@@ -1,27 +1,127 @@
...
+      env:
+        - MATRIX_EVAL="brew install gcc@6 && brew link --force --overwrite gcc@6 && CC=gcc-6 && CXX=g++-6"
+        - BUILD_NAME=cthread WITH_DISTRIBUTED=serial
+      compiler: gcc-6
+
+## test gcc6 - mpi with threading backend ##
+    - os: osx
+      osx_image: xcode9.2
+      env:
+        - MATRIX_EVAL="brew install gcc@6 && brew link --force --overwrite gcc@6 && CC=gcc-6 && CXX=g++-6"
```
bcumming: Do these tests for gcc 6 take a long time to run by virtue of having to install gcc via homebrew?

.travis.yml
```diff
@@ -1,27 +1,127 @@
+######## Testing minimal compiler requirements ########
+# GCC          6.1.0        here: 6.4.0
+# Clang        4.0          here: 3.8,3.9 not compiling
```
bcumming: Does this mean that we are testing clang 4, and that 3.8 & 3.9 don't work?
And does the comment above say that we require GCC 6.1? and here we are testing GCC 6.4?

.travis.yml
```diff
@@ -1,27 +1,127 @@
...
+      # xcode9/9.1 broken, 9.2 still uses Apple LLVM version 9.0.0 (clang-900.0.39.2)
+      osx_image: xcode9.2
+      env:
+        - MATRIX_EVAL="CC=clang && CXX=clang++"
+        - BUILD_NAME=cthread WITH_DISTRIBUTED=serial
+      compiler: clang
+
+
+## test clang9 - mpi with threading backend ##
+    - os: osx
```
bcumming: Is there a way to give this test a name that clearly describes what it is testing.
If you look at the dashboard for the tests here:
https://travis-ci.org/eth-cscs/arbor/builds/432535277
It would be nice if it was clearly labeled with something like `os x, apple clang, mpi`

scripts/travis/build.sh
```diff
@@ -27,7 +27,11 @@ if [[ "${WITH_DISTRIBUTED}" = "mpi" ]]; then
     export OMPI_CXX=${CXX}
     CC="mpicc"
     CXX="mpicxx"
+    if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
```
bcumming: The else branch hasn't been indented.
An alternative way to do this is to "build up" the `launch` variable like this:
```bash
launch="mpiexec -n 4"
if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
    launch="$launch --oversubscribe --mca btl tcp,self"
fi
```

bcumming: Because the flags are not standard mpirun flags, it would be useful to have a small comment that explains why the `--oversubscribe` and `--mca` are used on os x.

.travis.yml
akuesters: In total, it takes roughly 8 min. The installing of gcc6 takes only 15s (according to Travis)

bcumming: Great!  Whenever I have installed gcc on my mac using Homebrew it has compiled gcc, which took a loooong time (admittedly the last time I did this was years ago).

.travis.yml
akuesters: Will need to research this option

.travis.yml
bcumming: OK. So, I would propose removing any mention of 3.8 and 3.9 from here, and updating the installation guide to not mention Clang 3.x.

akuesters: Will do. Where do I update the install guide?

bcumming: You can update the documentation in this PR.
The docs are generated from files in the `doc/` path, in this case `doc/install.rst`:
https://github.com/eth-cscs/arbor/blob/master/doc/install.rst
Just edit this in your text editor.

.travis.yml
bcumming: I don't know if it is possible, but it would be nice!

.travis.yml
akuesters: Do you want me to update gcc 6.1 to 6.4 as well as the note beneath where we still use 5.2.0?

bcumming: No. I have tested it with gcc 6.1 on Daint.
We have to support 6.1 at least.

bcumming: And yes, please move 5.2.0 to 6.1 wherever you see it!

akuesters: Ok, what about the gcc5.2.0 note? update to 6.1.0?

akuesters: ok

akuesters: Does the Intel compiler note need to be updated as well from "Needs GCC 5 or later for standard library" to "Needs GCC 6 ..."?

Further, for Travis I needed gcc6 for clang on linux as well for std lib. Shall I put that in the note for clang?

akuesters:
Having problems with osx clang `libc++abi.dylib: terminating with uncaught exception of type error` for (yet) unknown reasons. Need to debug

PR #620, opened 2018-10-11
fix weights in ring benchmark

bcumming:
Fix potential numeric instabilities in the ring benchmark caused by passing arguments to an event generator in the wrong order.


halfflat:
Fix potential numeric instabilities in the ring benchmark caused by passing arguments to an event generator in the wrong order.

PR #623, opened 2018-10-12
Bump version post 0.1 for development.

halfflat:
Bump version so we do not skew code from tagged v0.1.


halfflat:
Well, thank you, CMake. Thank you so much.
https://gitlab.kitware.com/cmake/cmake/issues/16716

bcumming:
The ironic applause was for Cmake.

@halfflat gets two thumbs up and our undying gratitude for doing CMake so that the rest of us don't have to.

PR #638, opened 2018-11-12
Fine-grained matrix GPU matrix solver.

bcumming:
A new Hines matrix solver implementation for the GPU that can solve a single tree in parallel with multiple threads.

Replaces the interleaved solver, which used a single thread to solve each matrix.
Branches with the same common root in the tree can be solved independently on each of the forward and backward solution passes.

* A matrix storage type, `arb::gpu::matrix_state_fine` that stores the branches of multiple trees for efficient backward and forward substitution.
* Extend the `arb::tree` data structure to support operations for choosing a new root node and determining a root node which minimises the maximum distance between the root and any of the trees leaves.
* Implementation of code for rebalancing a set of matrix trees, aka. a "forest" of trees.
* CUDA kernels for efficiently performing matrix assembly and matrix solution steps
* CMake option `ARB_WITH_GPU_FINE_MATRIX` for toggling the new solver (default `on`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ set(ARB_VALIDATION_DATA_DIR "${PROJECT_SOURCE_DIR}/validation/data" CACHE PATH
 #----------------------------------------------------------
 
 option(ARB_WITH_GPU "build with GPU support" OFF)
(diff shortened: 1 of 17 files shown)


bcumming:
Travis doesn't like this PR for some reason. Closing and reopening now.

PR #649, opened 2018-11-20
Workaround for CMake 3.12 bug passing -thread to nvcc

### LINKED ISSUE #645: GPU Build Fails with GCC 7.3.0 CUDA 9.2 CMake 3.12.3
As discussed via email, I'm reporting a problem I have with building the GPU version of Arbor with the following dependencies on JURECA

* GCC/7.3.0
* CUDA/9.2.88
* CMake/3.12.3
* MVAPICH2/2.3-GDR
* Some notes: GCC 7.3.0 is currently the oldest supported GCC on JURECA. Because of the -GDR version, MVAPICH2 is currently the preferred MPI implementation to be used together with GPUs; it (used to/) is supposed to have the best performance for CUDA-aware MPI calls.

I configure with

```
cmake ../src -DARB_WITH_GPU=ON -DARB_ARCH=haswell -DARB_VECTORIZE=ON -DARB_WITH_MPI=ON -DCMAKE_C_COMPILER=`which mpicc` -DCMAKE_CXX_COMPILER=`which mpicxx`
```

The issue reads the following

    [100%] Linking CUDA device code CMakeFiles/ring.dir/cmake_device_link.o
    nvcc fatal   : Unknown option 'pthread'

It seems to be related to `-pthread` being passed as a direct flag to `nvcc`, even though it's suppose to be passed to the underlying host compiler – see https://gitlab.kitware.com/cmake/cmake/issues/18008.

Using `make VERBOSE=1` to extract the failing line and then manually entering the corresponding directory and calling the same line,

* plus `-Xcompiler` in front of the `-pthread` option, or
* without `-pthread` altogether

seems to solve the issue.

```
make ring bench VERBOSE=1

OUTPUTOUTPUTOUTPUTOUTPUT…

cd /homeb/zam/aherten/NVAL/ICEI-Benchmarks/Arbor.jureca/src/arbor/build/example/ring && /usr/local/software/jureca/Stages/2018b/software/CMake/3.12.3-GCCcore-7.3.0/bin/cmake -E cmake_link_script CMakeFiles/ring.dir/dlink.txt --verbose=1
/usr/local/software/jureca/Stages/2018b/software/CUDA/9.2.88/bin/nvcc  -O3 -DNDEBUG -Xcompiler=-fPIC -Wno-deprecated-gpu-targets -shared -dlink CMakeFiles/ring.dir/ring.cpp.o -o CMakeFiles/ring.dir/cmake_device_link.o ../../lib/libarbor.a ../../lib/libarboraux.a ../../lib/libarbor.a -pthread -lcudadevrt -lcudart_static -lpthread -ldl
nvcc fatal   : Unknown option 'pthread'

cd /homeb/zam/aherten/NVAL/ICEI-Benchmarks/Arbor.jureca/src/arbor/build/example/ring

/usr/local/software/jureca/Stages/2018b/software/CUDA/9.2.88/bin/nvcc  -O3 -DNDEBUG -Xcompiler=-fPIC -Wno-deprecated-gpu-targets -shared -dlink CMakeFiles/ring.dir/ring.cpp.o -o CMakeFiles/ring.dir/cmake_device_link.o ../../lib/libarbor.a ../../lib/libarboraux.a ../../lib/libarbor.a -Xcompiler -pthread -lcudadevrt -lcudart_static -lpthread -ldl

OR

/usr/local/software/jureca/Stages/2018b/software/CUDA/9.2.88/bin/nvcc -O3 -DNDEBUG -Xcompiler=-fPIC -Wno-deprecated-gpu-targets -shared -dlink CMakeFiles/ring.dir/ring.cpp.o -o CMakeFiles/ring.dir/cmake_device_link.o ../../lib/libarbor.a ../../lib/libarboraux.a ../../lib/libarbor.a -lcudadevrt -lcudart_static -lpthread -ldl
```

I can do some more tests, on my to do list is:

* Use another MPI instead of MVAPICH GDR (@apeyser's idea, indirectly)
* Use another CMake
  - 3.13.0-rc3
  - 3.12.0

It could be Monday before I'll look at it again, though.

halfflat:
CMake wants to run a device link pass with nvcc despite
there being no CUDA seperable compilation enabled anywhere,
and _then_ passes on -pthread to that unnecessary nvcc
invocation when we use the Threads dependency. The latter,
at least, is fixed in CMake 3.13.

We used the prefer -pthread option for compatibility with
our earlier build configuration; turning it off will
hopefully have no consequence.

We also enable device linking on the arbor library. Which
is not needed, but if they are going to insist on doing it,
it should be on the library rather than the executable.

CMake then goes and does it on the executable anyway. Great.

Fixes #645.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,7 +334,7 @@ write_basic_package_version_file(
 #
 # To avoid CMake users of the installed arbor library conditionally requiring
 # that they add CUDA to their project language, explicitly munge the import
-# language and library dependencies on the installed target if ARP_WITH_GPU
+# language and library dependencies on the installed target if ARB_WITH_GPU
 # is set, via the variables arbor_override_import_lang and arbor_add_import_libs.
 # arbor_build_config records our build type in a way compatible with the
 # generated export cmake files.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -66,7 +66,8 @@ if (NOT CMAKE_BUILD_TYPE)
     set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "debug" "release")
 endif()
 
-# When we find threads, prefer -pthread option.
+# When we find threads, prefer not to use the -pthread option
+# in order to avoid a CMake 3.12 issue.
 set(THREADS_PREFER_PTHREAD_FLAG OFF)
 
 # Add CUDA as a language if GPU support requested.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,7 +67,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 
 # When we find threads, prefer -pthread option.
-set(THREADS_PREFER_PTHREAD_FLAG ON)
+set(THREADS_PREFER_PTHREAD_FLAG OFF)
 
 # Add CUDA as a language if GPU support requested.
 # (This has to be set early so as to enable CUDA tests in generator
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -330,6 +330,27 @@ write_basic_package_version_file(
 # Template file will use contents of arbor_export_dependencies to include the
 # required `find_dependency` statements, and arbor_supported_components will
 # be used to check feature support.
+#
+# To avoid CMake users of the installed arbor library conditionally requiring
+# that they add CUDA to their project language, explicitly munge the import
+# language and library dependencies on the installed target if ARP_WITH_GPU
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 
 # When we find threads, prefer -pthread option.
```
apeyser: Comment seems out of date

halfflat: Good catch, thanks!

PR #654, opened 2018-11-28
Refactor hardware detection to sup

### LINKED ISSUE #647: Move thread/rank/gpu device wrangling out of core arbor library
Based on discussion between @bcumming and @halfflat:

In terms of separating concerns and code hygiene, we're thinking it would be cleaner to have the environment inspection for data such as GPU ids, thread counts, core counts, etc. and their default assignments to allocated resources in a context be separate from the core library. E.g.

1. Application uses non-core routines to determine operating environment, including number of available or desired threads.
2. Application creates a context based on that determination.

The motivation for this is the problem of determining a sane default allocation of GPUs to ranks in the instance that multiple ranks on a node may see multiple shared GPUs — while we can pick a reasonable guess as to an appropriate assignment of resources, this is ultimately going to be very dependent upon the system configuration, batch environment, etc. etc. and we shouldn't hard code any such guess into Arbor itself.

Options:

1. We move this stuff into `sup`, and install it in parallel with `libarbor.a`. Downsides:
    * There are things in sup which are convenient for our examples and tests, but are otherwise fairly specific to our needs.
    * `sup` includes a JSON library dependency which we should avoid baking into an install.
    * Installing `sup` reduces our freedom in putting useful test/example helper code in there, and adds a documentation burden.

2. We move this into `sup`, and _don't_ install it. It then acts as example code for how an application might want to do this sort of resource query, but we're not providing any installed code to do it.

3. We make another library, e.g. `arbordrv` or `arborenv` (I dunno, we have options here) that does this resource wrangling and maybe some other related convenience code. This gets installed, and `sup` remains our personal playground.

My personal preference is option 3.

bcumming:
Refactoring that moves logic for determing available concurrency and finding GPUs from the core Arbor library to the sup library.

Before this patch, Arbor's would attempt to use the first available GPU by default.
This caused problems on systems with multiple GPUs and multiple ranks on the same node, where by default all ranks would attempt to use the same GPU.

Now thread pools and execution contexts are always default initialized to use one thread and no GPU, making no effort to determine if more resources are available.

The sup library now provides an interface for 
* querying a best guess at the number of available threads.
* finding a GPU, with and without MPI. Currently implements the old behavior of taking the first available GPU. The next PR will add logic for assigning GPUs in MPI jobs.

Fixes #647


example/bench/bench.cpp
```diff
@@ -33,16 +35,26 @@ int main(int argc, char** argv) {
...
     try {
+        arb::proc_allocation resources;
+        if (auto nt = sup::get_env_num_threads()) {
+            resources.num_threads = *nt;
+        }
+        else {
+            resources.num_threads = sup::thread_concurrency();
+        }
+
 #ifdef ARB_MPI_ENABLED
```
halfflat: This thread count code is repeated in each example, and will likely appear in pretty much every C++ code that uses the library. It's boilerplate code that could be wrapped up with e.g. `sup::get_default_threads()` or similar; people can still use the more fundamental functions if they wish.

example/bench/bench.cpp
```diff
@@ -33,16 +35,26 @@ int main(int argc, char** argv) {
...
         sup::with_mpi guard(argc, argv, false);
-        auto context = arb::make_context(arb::proc_allocation(), MPI_COMM_WORLD);
+        resources.gpu_id = sup::find_gpu(MPI_COMM_WORLD);
+        auto context = arb::make_context(resources, MPI_COMM_WORLD);
         {
-            int rank = 0;
+            int rank;
             MPI_Comm_rank(MPI_COMM_WORLD, &rank);
```
halfflat: Only a minor nit, but if `MPI_Comm_rank` were to fail, it wouldn't write to `rank`; having a default initialized value seems like good practice.

example/bench/bench.cpp
```diff
@@ -33,16 +35,26 @@ int main(int argc, char** argv) {
...
             is_root = rank==0;
         }
 #else
-        auto context = arb::make_context();
+        resources.gpu_id = sup::find_gpu();
```
halfflat: I think `find_gpu()` is not a good name; more on this below.

sup/concurrency.cpp
```diff
@@ -25,7 +30,7 @@ namespace threading {
 //
 // Throws std::runtime_error:
 //  ARB_NUM_THREADS or OMP_NUM_THREADS is set with invalid value.
-util::optional<size_t> get_env_num_threads() {
+arb::util::optional<size_t> get_env_num_threads() {
```
halfflat: While we're here, this would be a good opportunity to reduce the use of `arb::util::optional` in public interfaces: this is going to be of a concern when we're in a mixed C++17 environment and we move to `std::optional`. (We're currently only using `optional` in one other public method, in `arb::mc_segment`).

sup/concurrency.cpp
```diff
@@ -52,26 +57,39 @@ util::optional<size_t> get_env_num_threads() {
...
     if (!n) {
-        n = hw::node_processors();
+        n = std::thread::hardware_concurrency();
     }
 
-    return n? n: 1;
+    // If still zero, use one thread.
+    n = std::max(n, 1u);
+
+    return n;
```
halfflat: We can keep the comment, but using `max` here just adds code to no effect: there's nothing wrong with `n? n: 1`.

sup/gpu.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <arbor/version.hpp>
+
+#ifdef ARB_GPU_ENABLED
+#include <cuda_runtime.h>
+#endif
+
+#ifdef ARB_MPI_ENABLED
+#include <mpi.h>
+#endif
+
```
halfflat: This code doesn't depend upon Arbor having been built with MPI or CUDA or not, but rather on the current configuration environment.

When we split `sup` up into a local-for-us library and an installed-for-general-consumption component, the latter will have to have similar CMake guff that the arbor library itself has to make this all hang together; in the meantime, we can hand-write the CMake stuff for `sup` that handles the `ARB_WITH_MPI` and `ARB_WITH_CUDA` stuff as required.

sup/include/sup/gpu.hpp
```diff
@@ -0,0 +1,18 @@
...
+namespace sup {
+
+int find_gpu();
+
+#ifdef ARB_MPI_ENABLED
+int find_gpu(MPI_Comm comm);
+#endif
+
+} // namespace sup
+
```
halfflat: As above, `find_gpu(MPI_Comm)` does not require that `arbor` was built with MPI, but that MPI support is available at configure time. The MPI code can live in a different header+source that just includes `mpi.h`, but which is only conditionally included in the `sup` library based on the configure-time `ARB_WITH_MPI`.

The `find_gpu` names for these functions are too generic and don't really reflect what they do, which itself is very different in `find_gpu()` and `find_gpu(MPI_Comm)`.

The non-distributed version just checks to see if any GPU is visible, and so should have a corresponding name and return true or false. Alternately, if the interface doesn't change, it should have a name like `sup::default_gpu()` or similar.

The distributed one does this negotiation between ranks, and perhaps should be called something like `sup::find_private_gpu_id(MPI_Comm)` or something else along these lines.

example/bench/bench.cpp
bcumming: Not a minor nit at all: I should be testing the return value from MPI_Comm_rank... but wait a moment, I think I see a better solution.

example/bench/bench.cpp
bcumming: It is repeated in each example, but I don't have a problem with this because:
* I think this makes it much clearer what is going on: test the environment variable, then attempt to inspect the CPU set if the environment variable isn't set.
* the examples are meant to be stand alone, so a bit of repeated code isn't a crime.

I really feel that combining this logic into a single function obscures what is going on too much.
This level of granularity feels "about right" for what the `sup` library is trying to achieve: small well-defined tools/types/functions that are useful for all examples, tests, etc.

sup/concurrency.cpp
bcumming: I could change it to return 0 or -1 on failure to read an environment variable. How does that sound.

I have no problem with making the fix, but I don't see what the problem is with keeping it.
When we move to C++17, wouldn't we just change it to `std::optional`?
If a user is currently using C++17, then how does our current set up cause a problem (besides potential confusion about the difference between `std::optional` and `util::optional`)?

example/bench/bench.cpp
halfflat: I think though that almost every such example will have similar code, and a bunch of user code will too. By providing an easy-to-use default function, we make it easy for library users also to do the reasonable thing by default.

`sup` contains mainly things that we want to use in common across the examples and tests, which should not form part of the core interface. Wrapping boilerplate code is part of its very purpose — I really don't see this as a problem at all.

This thread counting code and the GPU negotiation code will be in whatever we will call our installed-with-Arbor helper library; any obscurity in its functionality can be well addressed simply with documentation, IMO.

sup/concurrency.cpp
bcumming: I personally prefer using `std::max` for this sort of logic. It is arguable whether a ternary or `max` is clearer... probably a matter of what you are used to.
The suggested ternary `n? n: 1` is short hand for `n>0? n: 1`, so while it is less "code", it isn't less logic for the reader to mentally parse.

That is my five cents, though I have no problem making the change.

example/bench/bench.cpp
bcumming: replaced with
```C++
root = arb::num_ranks(context) == 0;
```

sup/concurrency.cpp
apeyser: Isn't it possible to lift a name into a new namespace? Have an `arb::future`, where eventuall `arb::future::optional = std::optional`?

halfflat: The chief problem is one of interface: it's quite possible that we build Arbor with a C++17-capable compiler, but that users building against an installed version of it might want to use a different compiler with less robust C++ support (assuming all the ABIs are the same of course).

We should definitely move to `std::optional` when we can rely upon a C++17 build environment, but we shouldn't force a C++17 dependency through the interface, at least until compiler support is near universal.

sup/concurrency.cpp
apeyser: For me, both have equal cognitive load, but I'm less tempted to think about possible c++ weirdness with `n ? n : 1` than `max`. (And their not precisely equivalent but I have no idea what the compiler could do to either one).

sup/gpu.cpp
bcumming: I think I see what you are saying.

You are suggesting another set of `#defines`, e.g. `SUP_WITH_MPI`, `SUP_WITH_CUDA` from CMake?
We can't use the `ARB_WITH*` defines, because they shouldn't go past the public/private split.

We always want consistency between sup and arbor, i.e. if Arbor was compiled with MPI support, then sup must have matching MPI support.
The `libsup` installed with `libarbor` must be consistent, and the `libsup` and `libarbor` "used locally" for tests and the like must also always be consistent.

So `ARB_MPI_ENABLED` and `SUP_WITH_MPI` must always agree. Which makes me wonder if it isn't just simpler to use the one, even if it isn't a perfect split in theory.

To play the devil's advocate: two arguments for keeping it this way
* it is actually less confusing for users.
* one single variable, `ARB_XYZ_ENABLED`, has to be configured in CMake for every feature, instead of two (and the less CMake wrangling, the better).

sup/include/sup/gpu.hpp
bcumming: I would be for keeping the interfaces, and renaming them. `sup::default_gpu` and `sup::find_private_gpu` are quite good names for that purpose!

sup/gpu.cpp
halfflat: The `ARB_WITH_foo` defines are provided by the CMake infrastructure, and reflect the configuration environment. They are never exposed to user code, but there is no reason why they cannot also be used in the `sup` sources (obviously not the headers).

I can't see any need for the `sup` code to provide `arbor`-like feature macros.

The CMake guff will come from splitting  MPI and CUDA dependencies out from `arbor-private-deps` into an internal dependency target that will be a requirement for both `arbor` and `sup`.

halfflat: Addendum: the only reason why I've been advocating not using `ARB_WITH_foo` in our test sources is that it allows us to catch interface issues — it keeps us more honest when it come to corralling our configuration state.

example/bench/bench.cpp
bcumming: I must admit I am a bit uneasy suggesting that `ARB_NUM_THREADS` is a reasonable environment variable for users ;)
More seriously, user code might be similar, but would in all likelihood use different environment variables.
The examples might share code via `sup`, but at the same time they should be as simple to understand stand-alone. Because nobody reads the docs (they just create a ticket).

This feels a bit like 'half of one, six of the other'. I can change it, but I think that both of our positions are fairly reasonable.

example/bench/bench.cpp
bcumming: Yep, I agree with the proposed changes below.

sup/concurrency.cpp
bcumming: I think it is probably worth removing the `optional` from this interface, because it wouldn't be a massive loss (the error state is easy enough to encode in the return value with 0 or -1).
We might have a look at removing the `optional` on `mc_segment` so that this becomes a discussion we don't have to have.

If we were to keep `optional` in our interface (and similar types like `any`), I think Sam's idea of just using our `util::optional` until we move everything to C++17 is the way forward.
I want to avoid having multiple implementations of things in the library for different C++ standards: I would prefer to just implement everything in the minimum standard that we have selected for the library (currently C++14).

sup/gpu.cpp
bcumming: The point about minimising `ARB_WITH_foo` in the tests is a good one: we want to test the public interface whenever possible.

I am just trying to think about how to keep such `#ifdef MPI` out of the sup headers. When we have MPI support, there will be a header that exposes `find_private_gpu(MPI_Comm)`, which will obviously need to include `mpi.h`, like `sup/include/sup/with_mpi.hpp` currently does.

I would have to put the `find_private_gpu(MPI_Comm)` in a separate file to the `find_gpu()` (or whatever we call it, you get what I am saying).

Is this correct reasoning?

bcumming: The template option is a bit easier to implement, and is a pattern that we use elsewhere.
But compile errors are always preferable to linker errors.
I will think about it some more.

sup/concurrency.cpp
bcumming: The compiler should generate equally efficient code, it is just a matter of semantics I think.

https://godbolt.org/z/0ITuiT

halfflat: Man, I can never convince gcc to produce a conditional move when I want it to, damn it.

bcumming:
I have acted on all the suggested changes, except the logic that selects the number of threads to use in the examples:

```
        if (auto nt = sup::get_env_num_threads()) {
            resources.num_threads = nt; 
        }   
        else {
            resources.num_threads = sup::thread_concurrency();
        }   
```

Which I kept, because that is six of one, half a dozen of the other.

Issue #667, opened 2018-12-19
Python infrastructure merge


bcumming:
ZAdding the Python wrapper can be broken into two steps:
1. The Python infrastructure (directory structure, python binding interface, CMake integration, documentation format, etc)
2. Wrapping the individual parts of the Arbor public API in Python and documenting them.

To avoid a messy merge process, these steps will be performed separately, so that the discussion about the infrastructure can be separated from discussions about the python interface for each component.

## TODO list

- [x] Directory structure
- [x] Unit tests
- [ ] Documentation
- [x] Doc strings
- [x] Error handling (exceptions)
- [x] CMakeLists (compilation options, pybind11 CMake integration)

akuesters:
Created PR #668 to discuss after holiday (back on 21st January)

PR #668, opened 2018-12-21
Python PR #667 

akuesters:
concerns #667 

## Directory Structure
```
|- test\
    |- options.py
    |- unit\
        |- test_contexts.py
        |- runner.py
    |- unit-distributed\
        |- test_contexts_arbmpi.py
        |- test_contexts_mpi4py.py
        |- runner.py
```
In parent folder `test`: 
- `options.py`: set global options (define arg parser)

In subfolders `unit`/`unit_distributed`: 
- `test_xxxs.py`: define unittest class with test methods and own test suite (named: test module)
- `runner.py`: run all tests in this subfolder (which are defined as suite in test modules) 

### with `unittest` from SUBFOLDER: 
* to run all tests in subfolder:  
```             
python -m unittest [-v]
```
* to run module: 
```  
python -m unittest module [-v]
```  
, e.g. in `test/unit` use `python -m unittest test_contexts -v`
* to run class in module: 
```
python -m unittest module.class [-v]
```  
, eg. in `test/unit` use `python -m unittest test_contexts.Contexts -v`
* to run method in class in module: 
```  
python -m unittest module.class.method [-v]
```  
, eg. in `test/unit` use `python -m unittest test_contexts.Contexts.test_context -v`

### with `runner.py` and argument(s) `-v {0,1,2}` from SUBFOLDER: 
* to run all tests in subfolder:   
```  
python -m runner[-v2]
```   
or `python runner.py [-v2]`
* to run module: 
```  
python -m test_xxxs [-v2]
```   
or `python test_xxxs.py [-v2]`
* running classes or methods not possible this way

### from any other folder: 
* to run all tests:   
```
python path/to/runner.py [-v2]
```
* to run module: 
```  
python path/to/test_xxxs.py [-v2]
```   

## Adding new tests
1. In suitable folder `test/unit` (no MPI) or `test/unit_distributed` (MPI), create `test_xxxs.py` file
2. In  `test_xxxs.py` file, define 
  a) a unittest `class Xxxs(unittest.TestCase)` with test methods `test_yyy` 
  b) a suite function `suite()` consisting of all desired tests returning a unittest suite `unittest.makeSuite(Xxxs, ('test'))` (for all defined tests, tuple of selected tests possible); steering of which tests to include happens here!
  c) a run function `run()` with a unittest runner `unittest.TextTestRunner` running the `suite()` via `runner.run(suite())`
  d) a `if __name__ == "__main__":` calling `run()`
3. Add module to `runner.py` in subfolder by adding `test_xxxs`
  a) to import: in `try` add `import test_xxxs`, in `except` add `from test.subfolder import test_xxxs`
  b) to `test_modules` list

## Naming convention
- modules: `test_xxxs` (all lower case, ending with `s` since module can consist of multiple classes)
- class(es): `Xxxs` (first letter upper case, ending with `s` since class can consist of multiple test functions)
- functions: `test_yyy` (always starting with `test`since suite is build from all methods starting with `test`)


cmake/FindPythonModule.cmake
```diff
@@ -0,0 +1,30 @@
...
+    # A module's location is usually a directory, but for binary modules
+    # it's a .so file.
+    execute_process(COMMAND "${PYTHON_EXECUTABLE}" "-c" 
+      "import re, ${module}; print(re.compile('/__init__.py.*').sub('',${module}.__file__))"
+      RESULT_VARIABLE _${module}_status 
+      OUTPUT_VARIABLE _${module}_location
+      ERROR_QUIET OUTPUT_STRIP_TRAILING_WHITESPACE)
+    
+    if(NOT _${module}_status)
+      set(HAVE_${module_upper} ON CACHE INTERNAL "cmake sucks")
```
bcumming: There is some stray white space, and this correct but inappropriate text.
I have cleaned it up with a commit.

bcumming:
Some notes on the unit tests.

The `unit` and `distributed-unit` tests need separate runners.
If the tests are run from `python/test` with `mpirun -n 2 python -m unittest` the serial unit tests are run twice each.

The dependence on an environment variable makes control of the tests a bit confusing,
I would stick with just using command line arguments.
* the non-distributed tests can't give an error if you
  try running them inside their path without setting the
  `ARB_MPI_ENABLED` environment variable
* the test suite should be able to detect if Arbor supports
  MPI without checking an environment variable                                                   
* it should also be able to detect if mpi4py is supported.

bcumming:
I don't think that the unit tests should be run in MPI, unless they are directly require MPI (i.e. the distributed tests).

The output is confusing if run with `-n 2`, because each message is printed twice, etc.
How about having the separate runners, and not having a "global" runner?

I would propose that the user could use introspection of the Python module to test if it was compiled with MPI. e.g.

```
import arbor

if arbor.has_mpi():
    arbor.mpi_initialize()
```

The C++ library provides such a function:

https://github.com/arbor-sim/arbor/blob/master/include/arbor/context.hpp#L74

bcumming:
I made an incorrect suggestion. The `has_mpi(context)` function takes a context as an argument: so it requires a context with MPI enabled... which is too late!

Maybe you could write a new stand alone function and wrapper in `python/mpi.cpp` that returns `true`/`false` depending on whether Arbor was compiled with MPI?

akuesters:
Regarding `mpi4py` there needs to be a different function, eg has_mpi4py(), since otherwise mpi gets initialized/finalized too often. How would python see (from cmake?) which to choose from mpi4py, or arbor's own?

akuesters:
ah, ok. Based on C++'s ARB_MPI_ENABLED and ARB_WITH_MPI4PY which are available.

python/strings.hpp
```diff
@@ -0,0 +1,28 @@
+#pragma once
+
+/*
+ * Utilities for generating string representations of types.
+ */
+
+#include <string>
+
```
bcumming: remove the commented lines.

python/test/options.py
```diff
@@ -0,0 +1,57 @@
...
+    args = parser.parse_args()
+    return args
+
+"""
+bool_env
+    Goal:       get an environment variable coerced to a boolean value.
+    Arguments:  var_name  - name of the environment variable
+                default   - default to use if `var_name` is not specified in environment
+    Returns:    `var_name` or `default` coerced to a boolean using the following rules
+                False: "False", "false", "OFF", "off", "Off", ""; True: any other non-empty string
```
bcumming: I think we can do without this (see other comments), but I prefer to keep it simple
```
# turn on MPI
export ARB_WITH_MPI
# turn it off
unset ARB_WITH_MPI
```
to avoid testing all those strings ;)

python/test/options.py
```diff
@@ -0,0 +1,57 @@
+# -*- coding: utf-8 -*-
+#
+# options.py
+
+import argparse
+
+import os
+
+"""
+parse_arguments
```
bcumming: This documentation string isn't very clear to me.
The namespace argument is not used, so remove it.
You probably don't need any of the documentation, because the function itself is very clear.

python/test/runner.py
```diff
@@ -0,0 +1,40 @@
...
+from unit_distributed import test_contexts_mpi4py
+# add more if needed
+
+test_modules = [\
+    test_contexts,\
+    test_contexts_arbmpi,\
+    test_contexts_mpi4py\
+] # add more if needed
+
+"""
```
bcumming: How about a single string: "returns a suite of all the tests".
The Goal and name are redundant.

python/test/unit/runner.py
```diff
@@ -0,0 +1,44 @@
...
+except ModuleNotFoundError:
+    from test import options
+    from test.unit import test_contexts
+    # add more if needed
+
+test_modules = [\
+    test_contexts\
+] # add more if needed
+
+"""
```
bcumming: See comment above.

python/test/unit/test_contexts.py
```diff
@@ -0,0 +1,83 @@
...
+# to be able to run .py file from child directory
+import sys, os
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '../../')))
+
+try:
+    import options
+except ModuleNotFoundError:
+    from test import options 
+
+"""
```
bcumming: Again, I would just have the single string: "all tests for non-distributed arb.context"

doc/python.rst
```diff
@@ -0,0 +1,29 @@
+Python
+======
+
+Arbor provides a full set of python bindings that expose all functionality.
```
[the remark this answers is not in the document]
akuesters: Done. Next PR on documentation still work in progress

include/arbor/arbexcept.hpp
```diff
@@ -95,4 +95,11 @@ struct range_check_failure: arbor_exception {
     double value;
 };
 
+// Python wrapper errors
```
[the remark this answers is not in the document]
akuesters: In order to use `pprintf` from `arbor/util/strprintf.hpp` I would need to set the include path in cmake to `arbor/util`, otherwise the compiler won't find `util/meta.hpp` which gets included in `arbor/util/strprintf.hpp`. Or have duplicates or not use `pprintf` or leave python exception where it is. Any suggestions?

python/context.cpp
```diff
@@ -0,0 +1,80 @@
...
+        .def(pybind11::init(
+            [](int threads, pybind11::object gpu, pybind11::object mpi){
+                arb::proc_allocation alloc(threads, gpu.is_none()? -1: pybind11::cast<int>(gpu));
+                if (mpi.is_none()) {
+                    return context_shim(arb::make_context(alloc));
+                }
+                auto& c = pybind11::cast<mpi_comm_shim&>(mpi);
+                return context_shim(arb::make_context(alloc, c.comm));
+            }),
+            "threads"_a=1, "gpu"_a=pybind11::none(), "mpi"_a=pybind11::none())
```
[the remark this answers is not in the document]
akuesters: In pybind11 doc`s I couldn`t find a way to include the doc string directly to the argument. 
I could add a new`readwrite` or `readonly`? As in

```
.def_readwrite("threads", &arb::proc_allocation::num_threads,
            "The number of threads to use in the context's thread pool.")
```
Alternatively, I could add the string to the constructor as in 

```
.def(pybind11::init(
            [](int threads, pybind11::object gpu, pybind11::object mpi){
                arb::proc_allocation alloc(threads, gpu.is_none()? -1: pybind11::cast<int>(gpu));
                if (mpi.is_none()) {
                    return context_shim(arb::make_context(alloc));
                }
                auto& c = pybind11::cast<mpi_comm_shim&>(mpi);
                return context_shim(arb::make_context(alloc, c.comm));
            }),
            "threads"_a=1, "gpu"_a=pybind11::none(), "mpi"_a=pybind11::none(),
            "Arguments:\n \t threads: \t The number of threads available locally for execution.\n \t gpu: \t The index of the GPU to use: defaults to -1 for no GPU. \n \t threads: \t mpi: The MPI communicator: defaults to None for no MPI. ")
```

python/mpi.cpp
```diff
@@ -0,0 +1,99 @@
...
+    throw arb::mpi_error(MPI_ERR_OTHER, "The argument is not an mpi4py communicator");
+}
+
+#endif
+
+// Some helper functions for initializing and finalizing MPI.
+// Arbor requires at least MPI_THREAD_SERIALIZED, because the communication task
+// that uses MPI can be run on any thread, and there will never be more than one
+// concurrent communication task.
+
```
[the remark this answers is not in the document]
akuesters: From Sam on Slack (to have it in one place): 
Had a chat with @bcumming about this this morning.

My position (and I think I convinced Ben) is that ultimately, we should either demand that the python user code handle all the MPI environment and just provide an interface for getting a communicator for the context  (the 'pure' approach), or if we're going to provide an easy interface, that we handle everything — initialization, clean up, and so on.

Exposing `MPI_Initialize` and `MPI_Finalize` is convenient from the point of view of writing the wrappers, but IMO is just pushing unnecessary complexity onto the user — the user isn't going to be able to do anything else useful with the MPI context we provide unless we start wrapping whole chunks of the MPI interface, which I think we all agreed should be avoided.

Concretely, we should keep the MPI_Initialize wrapper and friends for getting the PR through, but then we should change the interface so that one can initialize the arbor context with e.g. `mpi=auto`; the python library would then run the initialization if required, and set up a MPI_Finalize on clean-up.

Whether MPI is running or not, and which rank the python user code sees, can be exposed via the existing functionality that sits on `arb::execution_context`.

python/test/options.py
akuesters: Done.

python/test/runner.py
akuesters: Done.

python/test/unit/runner.py
akuesters: Done.

python/test/unit/test_contexts.py
akuesters: Done.

python/context.cpp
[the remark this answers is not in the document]
bcumming: You are right: there is no way to have docstrings for individual arguments (I just checked the pybind11 source code to make sure).

So your suggestion of using a docstring for the constructor is a good one, however I would format the string as follows to make it a bit easier to read (and without the tab `\t` characters):
```
            "threads"_a=1, "gpu"_a=pybind11::none(), "mpi"_a=pybind11::none(),
            "Arguments:\n"
            "  threads: The number of threads available locally for execution.\n"
            "  gpu:     The index of the GPU to use, defaults to None for no GPU.\n"
            "  mpi:     The MPI communicator, defaults to None for no MPI.\n");
```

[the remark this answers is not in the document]
akuesters: Done. Also added docstrings to other con`tors in context and proc_allocation class.

include/arbor/arbexcept.hpp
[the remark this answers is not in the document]
akuesters: Done.

python/test/options.py
akuesters: Done. 
* Removed python argument reader, 
* added boolean function (+wrapper) in `mpi.cpp` to check if arbor was compiled with MPI/mpi4py,
* removed `test/runner.py` (so that test are now to be run in regarding foolders `unit`or `unit_distributed` via the respective runners as `python -m runner` instead of `python -m unittest`)

bcumming:
Looking good, we are getting there.

There are a few things that still need organizing in the unit tests.

The instructions that are in the commit message can go into a `readme.md` in the `python/tests` path.    

The `test_contexts_arbmpi` case crashes because MPI is not initialized/finalized correctly.

The logic for deciding if `test_contexts_arbmpi` has to initialize/finalize
MPI is simpler with logic like this pseudo-code:
        
```     
    setUpClass:
        self.local_mpi = false
        if not arb.mpi_is_initialized()
            arb.mpi_init()
            self.local_mpi = true
            
    tearDownClass:
        if self.local_mpi
            arb.mpi_finalize()
``` 

Then, if the tests are run through the runner `runner.py`, the runner should set up MPI,
and all tests won't try to initialize or shut down MPI. If tests are run stand alone,
without the wrapper, they will set up MPI.

Another problem is that each unit test outputs for each MPI rank, e.g.

```
 > mpirun -n 1 python -m unittest 
.......
----------------------------------------------------------------------
Ran 7 tests in 0.001s 

OK
 > mpirun -n 2 python -m unittest
..............
----------------------------------------------------------------------
Ran 7 tests in 0.001s 

OK

----------------------------------------------------------------------
Ran 7 tests in 0.001s

OK
``` 

There are three outputs here:
1. `Ran X tests in Ys`
2. `OK`
3. `......` with one dot for each test.
4. Error messages that describe a test failure (there are none above,
   because the unit tests pass)

It is best if only one MPI rank (say rank 0) makes the output for 1, 2 and 3.
Each MPI rank can output an error message when a test case fails on that rank.

For the C++ unit tests, we wrote a custom Google test runner that ran the test
on all ranks, then did a barrier-reduction step with rank 0 printing whether a
test passed or failed.

I don't know how easy (or if it is possible at all) it is to do this for the python
unit testing framework. It might have to be a task for a later date.

akuesters:
RE MPI rank output: 

For non-zero ranks all messages are redirected now to os.devnull. 
Since this is handled in the `TextTestRunner()` via the parameter `stream`, which is called before actually running the test, the handle `result.wasSuccessful()` cannot be used to differentiate between success and error messages on single ranks.

bcumming:
For now the MPI unit tests can be accepted, though we will have to write a custom runner if and when we start more elaborate MPI testing in python.

Issue #637, opened 2018-10-26
AVX512 performance


halfflat:
Before the gather/scatter vectorization optimizations, AVX512 was considerably faster than AVX2 on KNL. Now, AVX2 is faster. So it seems likely that our use of vector memory operations on KNL is suboptimal, and this may also explain the worse performance we see on Skylake-X when AVX512 is used instead of AVX2.

Investigate and fix if possible!

*Update:* might not actually be slower on KNL, but the speed up is much smaller than we'd expect.

halfflat:
How much work would it be to resurrect the dynamic dispatch on constraint solution, with the single pass through memory? It would be good to compare.

To be honest, I have another reason for re-considering that approach: it greatly simplifies adding different shared storage data and layouts to the multi-core back end in both modcc and the shared state classes.

noraabiakar:
It shouldn't be too much work. I need to modify the modcc printer and test. I'll implement it and compare.

noraabiakar:
After re-implementing the dynamic dispatch version, I found that performance is worse on significant platforms for significant tests. VTune profiling shows almost 2x retired instructions compared to the static version. This is likely due to branch misprediction. 
I also tried to alleviate some of the memory access issues in the static version, specifically the "none" index_constraints in the compound_index_add function. A simple change insures we only read/write every distinct index in a vector once from/to memory. This gives around 20% improvement for expsyn_current. The fix is in PR #673

halfflat:
Might look again at AVX512 optimized indirect compound assignment in the future, but this looks pretty well sorted now!

PR #673, opened 2019-01-18
Optimize vectorized compound_indexed_add

noraabiakar:
Optimize "none" index_constraint specialization of compound_indexed_add. 
Only read/write each distinct memory index once per vector. 
Related to issue #637

### DIFF
--- a/include/arbor/simd/simd.hpp
+++ b/include/arbor/simd/simd.hpp
@@ -220,8 +220,15 @@ namespace simd_detail {
                     scalar_type a[width];
                     Impl::copy_to(s, a);
 
+                    scalar_type temp = 0;
                     for (unsigned i = 0; i<width; ++i) {
-                            p[o[i]] += a[i];
+                        unsigned curr_index_ = o[i];
+                        unsigned next_index_ = i == width - 1 ? o[i]+1 : o[i+1];
+                        temp += a[i];
+                        if(curr_index_ != next_index_) {
+                            p[curr_index_] += temp;
+                            temp = 0;
+                        }
                     }
                 }
                 break;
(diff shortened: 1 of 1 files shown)


include/arbor/simd/simd.hpp
```diff
@@ -220,8 +220,15 @@ namespace simd_detail {
...
                     for (unsigned i = 0; i<width; ++i) {
-                            p[o[i]] += a[i];
+                        unsigned curr_index_ = o[i];
+                        unsigned next_index_ = i == width - 1 ? o[i]+1 : o[i+1];
+                        temp += a[i];
+                        if(curr_index_ != next_index_) {
+                            p[curr_index_] += temp;
+                            temp = 0;
+                        }
                     }
```
halfflat: Fix is sound! I think simplifying (maybe?) the code slightly will save us a comparison and make it a little easier to read, e.g.
```
    scalar_type temp = 0;
    for (unsigned i = 0; i<width-1; ++i) {
        temp += a[i];
        if (o[i] != o[i+1]) {
            p[o[i]] += temp;
            temp = 0;
        }
    }
    p[o[width-1]] = temp;
```

noraabiakar: That does look better, I modified it :)

halfflat:
Argh, I led you astray. There's a bug in my suggested code: it never adds `a[width-1]`. I'll submit a PR.

noraabiakar:
Oops, I should have caught that. Thanks!

halfflat:
It's doubly my fault, because there's a bug in the unit test that allowed this error to pass in the first place.

PR #679, opened 2019-01-25
Split out 'arborenv' as an installable library from the sup library.

### LINKED ISSUE #647: Move thread/rank/gpu device wrangling out of core arbor library

halfflat:
Make a new installed library `libarborenv.a` covering a subset of the `sup` library functionality, with corresponding installed CMake target `arbor::arborenv`.

* Move NVML or CUDA 10 API decision for GPU UUID discovery to top level CMake.
* Move affinity, concurrency, MPI init guard, and gpu detection and negotiation functionality out of `sup` and into new library `arborenv`.
* Move `include/arbor` in project tree to `arbor/include/arbor` (for consistency across `sup`, `arbor`, and `arborenv` subdirectories.)
* Wrangle more explicit library dependency adding CMake code into the installed `arbor-config.cmake`, to help mitigate [CMake issue #18614](https://gitlab.kitware.com/cmake/cmake/issues/18614).
* Have `arborenv` code throw `std::runtime_exception` instead of `arb::arbor_error`. (We are still using `arb::mpi_error` though for a failure in `with_mpi`.)
* Move `scope_exit` into the `arb::util` namespace.
* Merge `affinity.hpp` into `concurrency.hpp`.
* Rename `gpu.hpp` to `gpu_env.hpp` in `arborenv` includes.

Fixes #647.


bcumming:
There is a build error with Travis for all of the MPI builds:
```
[100%] Building CXX object test/unit-distributed/CMakeFiles/unit-mpi.dir/test.cpp.o
/home/travis/build/arbor-sim/arbor/test/unit-distributed/test.cpp:13:10: fatal error: 
      'sup/with_mpi.hpp' file not found
#include <sup/with_mpi.hpp>
         ^~~~~~~~~~~~~~~~~~
1 error generated.
```

arborenv/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+
+target_link_libraries(arborenv PUBLIC arbor arborenv-public-headers)
+target_link_libraries(arborenv PRIVATE arbor-config-defs arborenv-private-deps)
+
+if(ARB_WITH_GPU)
+    # So that cpp and hpp files can use cuda.h and cuda_runtime.h
+    target_include_directories(arborenv PRIVATE ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
+
+    # The arborenv lib needs to use the CUDA NVML library for CUDA 9
+    if (${CUDA_VERSION_MAJOR} LESS 10)
```
bcumming: This search is being done in two places: it was also added to `CMakeLists.txt`.

halfflat: Yup, that's definitely an edit I forgot to do.
Moved the NVML stuff up to top because we need to know there what library dependencies to munge in the generated `arbor-config.cmake` script; forgot to remove from `arborenv/CMakeLists.txt`.

PR #682, opened 2019-01-31
Feature/expunge spike emitter

### LINKED ISSUE #677: Expunge `spike_emitter`
The `spike_emitter` in the sup library is only used in the example code. Its role can just as easily be performed using a lambda, as is the case in the ring benchmark:

https://github.com/arbor-sim/arbor/blob/master/example/ring/ring.cpp#L220

This approach is easier to understand, so let's just deprecate the `spike_emitter`.

halfflat:
Remove `sup::spike_emitter` and replace usages in `miniapp` and `brunel-miniapp`.

* Replaces instances of `spike_emitter` in `miniapp` and `brunel-miniapp` with a callback that saves spikes to a vector and which is written to a file after the simulation ends.
* Remove unused fields from `cl_options` in `brunel-miniapp`, including spike output file components.
* Hard code spike output path to `spikes.gdf` in `brunel-miniapp` and write saved spikes after simulation end.
* Correctly set 'rank' in `miniapp` (used in spike output paths).
* Read spike output options from JSON config in `miniapp` unconditionally.
* Make line comments in `miniapp` and `brunel-miniapp` main functions a bit more consistent in
formatting.

Fixes #677.


halfflat:
Oops. Will remove, thanks!

Issue #683, opened 2019-02-04
Make a distinction between integration domains and cells


noraabiakar:
Make a distinction between integration domains and cells in the lowered/back-end implementation.

Mechanism state integration uses a per 'integration domain' time and dt — an integration domain would correspond to a supercell — while the matrix solver will still run per-cell. If time and dt vectors were indexed via an 'intdom_index' as opposed to 'cell_index', it would simplify the fvm solver state and avoid the need to run a second pass over the time vectors to get agreement over the supercells.

This would involve making a new index for the backends (mapping cv to integration domain), with a lot of subsequent renaming, and using the cell partition only for building the matrix structures. Events on cells within the same integration domain would also have to be explicitly merged; there would be one stream per integration domain in the multi_event_stream_state as opposed to one per cell.

PR #688, opened 2019-02-12
Integrate over integration-domains instead of cells in the lowered/back-end implementation

noraabiakar:
Addresses #683 

Changes: 
* Mechanism state integration uses a per 'integration-domain' time and dt. An integration domain corresponds to a supercell. The matrix solver still runs per-cell. 
* New index added to the backends: cv-to-intdom 
* Staged_events on the same integration domain are reordered and merged. There is one stream per integration domain in the multi_event_stream_state as opposed to one per cell.
* New map added to mc_cell_group: cell_to_intdom; used to merge staged events
* gap_junctions example modified: now has synapses


arbor/mc_cell_group.cpp
```diff
@@ -135,21 +85,54 @@ void mc_cell_group::advance(epoch ep, time_type dt, const event_lane_subrange& e
...
+    cell_size_type ev_begin = 0, ev_mid = 0, ev_end = 0;
     // skip event binning if empty lanes are passed
     if (event_lanes.size()) {
-        for (auto lid: util::count_along(gids_)) {
-            auto& lane = event_lanes[perm_gids_[lid]];
+
+        std::vector<cell_size_type> idx_sorted_by_intdom(cell_to_intdom_.size());
+        cell_size_type n = 0;
+        std::generate(idx_sorted_by_intdom.begin(), idx_sorted_by_intdom.end(), [&]{ return n++; });
+        std::sort(idx_sorted_by_intdom.begin(), idx_sorted_by_intdom.end(),
```
halfflat: Could use `util::sort_by` from `util/rangeutil.hpp` for this.

arbor/mc_cell_group.cpp
```diff
@@ -135,21 +85,54 @@ void mc_cell_group::advance(epoch ep, time_type dt, const event_lane_subrange& e
...
+
+    cell_size_type ev_begin = 0, ev_mid = 0, ev_end = 0;
     // skip event binning if empty lanes are passed
     if (event_lanes.size()) {
-        for (auto lid: util::count_along(gids_)) {
-            auto& lane = event_lanes[perm_gids_[lid]];
+
+        std::vector<cell_size_type> idx_sorted_by_intdom(cell_to_intdom_.size());
+        cell_size_type n = 0;
+        std::generate(idx_sorted_by_intdom.begin(), idx_sorted_by_intdom.end(), [&]{ return n++; });
```
halfflat: Probably better to use `std::iota`.

arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -67,6 +69,11 @@ class fvm_lowered_cell_impl: public fvm_lowered_cell {
         const recipe& rec,
         const fvm_discretization& D);
 
+    fvm_index_type fvm_intdom(
+        const recipe& rec,
+        const std::vector<cell_gid_type>& gids,
+        std::vector<fvm_index_type>& cell_to_intdom);
+
```
halfflat: This could do with a comment (and/or maybe a more descriptive name) indicating that it returns the number of integration domains — which might be better expressed as an `fvm_size_type` — and generates the `cell_to_intdom` map.

While essentially a private function, it is exposed for unit testing; comments here then allow us to cross check against a test implementation.

PS: Just noticed the comment on line 372; perhaps move that up here, slightly expanded?

arbor/fvm_layout.hpp
```diff
@@ -87,7 +88,9 @@ struct fvm_discretization {
     }
 };
 
-fvm_discretization fvm_discretize(const std::vector<mc_cell>& cells);
+fvm_discretization fvm_discretize(
+        const std::vector<mc_cell>& cells,
+        const std::vector<fvm_index_type>& cell_to_intdom);
```
halfflat: The FVM discretization currently doesn't do anything for gap junctions (and probably won't ever have to), so the generation of the CV to intdom map from the `cell_to_intdom` vector and the discretization's cell to CV map is better off in the `fvm_lowered_cell_impl` code. This way we can keep the discretization code focused on the cell to CV mapping.

arbor/mc_cell_group.cpp
```diff
@@ -135,21 +85,54 @@ void mc_cell_group::advance(epoch ep, time_type dt, const event_lane_subrange& e
...
+        std::sort(idx_sorted_by_intdom.begin(), idx_sorted_by_intdom.end(),
+            [&](cell_size_type a, cell_size_type b) {
+            return cell_to_intdom_[a] < cell_to_intdom_[b];
+        });
+
+        int prev_intdom = -1;
```
halfflat: There may well be as many intdoms as there are cells in the group, so for consistency/documentation it would be better to use either `cell_size_type` to match the lane counting, or `fvm_index_type` to match the `cell_to_intdom_` mapping.

arbor/mc_cell_group.cpp
```diff
@@ -135,21 +85,54 @@ void mc_cell_group::advance(epoch ep, time_type dt, const event_lane_subrange& e
 
     PE(advance_eventsetup);
     staged_events_.clear();
+
+    cell_size_type ev_begin = 0, ev_mid = 0, ev_end = 0;
```
halfflat: Being at the interface between the public 'cell_ABC' types and the back-end `fvm_XYZ` types means that there will be some compromises in type choices here. For indexes into the big list of staged events, it might be better to go with `fvm_index_type`.

A comment at the beginning of `advance()` clarifying what set of types is used for which information could be helpful, given the ambiguity here.

As an aside, the main reason why `fvm_index_type` is an `int` rather than `unsigned` is that otherwise we'd have to extend the vectorization library to support `unsigned` in an addition to `int`, and that looked like way too much work!

arbor/mc_cell_group.cpp
noraabiakar: Done

arbor/fvm_lowered_cell_impl.hpp
noraabiakar: I added a comment describing the function and changed the type

arbor/fvm_layout.hpp
noraabiakar: Moved the code to fvm_lowered_cell_impl; I can no longer unit test cv_to_intdom so I removed the checks from the unit tests.

arbor/mc_cell_group.cpp
noraabiakar: Changed to fvm_index_type

PR #708, opened 2019-02-28
Python feature: config file to present Arbor's support on mpi, gpu and version

### LINKED ISSUE #700: Python: config file to present Arbor's support on mpi, gpu and version
**Idea**: 
 have a single spot that users can go to with queries about what features have been enabled:
```
import arbor

arbor.config.mpi
arbor.config.mpi4py
arbor.config.gpu
arbor.config.version_major
arbor.config.version_minor

print(arbor.config)
```

```
mpi:        yes
mpi4py:     no
gpu:        yes
version:    v0.1
```

**Open Questions**: 
1. What sort of structure to use?
2. How to build it in the C++ arbor module implementation?
3. How to present it to users?

**First thoughts**:
1. Start with a dictionary `config = {"mpi": True, "mpi4py": False, "gpu": True, "version": "v0.1"};` since pybind11 has support for working with Python dictionaries directly. Later they can be converted to `namedtuples`.

akuesters:
Fixes #700

Usage in python as: 

```
>>>import arbor
>>>arbor.config()
{'mpi': True, 'mpi4py': True, 'gpu': False, 'version': '0.1.1-dev'}
>>>d = arbor.config()
>>>arbor.print_config(d)
Arbor's configuration:
     mpi    :       True
     mpi4py :       True
     gpu    :      False
     version:  0.1.1-dev
```


python/config.cpp
```diff
@@ -0,0 +1,13 @@
+#include <pybind11/pybind11.h>
+
+#include "config.hpp"
+
+namespace pyarb {
+
+void register_config(pybind11::module &m) {
+
+    m
```
bcumming: this indentation is a little odd. How about the following:
```
    m.def("config", &config, "Get Arbor's configuration.")
     .def("print_config",
          [](const pybind11::dict& d){return print_config(d);},
          "Print Arbor's configuration.");
```

python/config.hpp
```diff
@@ -0,0 +1,45 @@
...
+        dict[pybind11::str("version")] = pybind11::str(ARB_VERSION);
+    return dict;
+}
+
+void print_config(const pybind11::dict &d) {
+    std::stringstream s;
+    s << "Arbor's configuration:" << std::endl;
+
+    for (auto x: d) {
+        s << "     " << std::left << std::setw(7) << x.first << ": " << std::right << std::setw(10) << x.second << "\n";
```
bcumming: use indentation on long print lines:
```
    s << "     "
      << std::left << std::setw(7) << x.first << ": "
      << std::right << std::setw(10) << x.second << "\n";
```

python/config.cpp
akuesters: Done.

python/config.hpp
akuesters: Done.

PR #709, opened 2019-03-01
Python: finalises config feature

akuesters:
- Removed `config.hpp`
- Removed `mpi_compiled()` and `mpi4py_compiled()` 
- Replaced them with `config.mpi` and `config.mpi4py` in unit tests


python/config.cpp
```diff
@@ -1,9 +1,50 @@
...
+#include <iomanip>
+#include <ios>
 
-#include "config.hpp"
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
 
 namespace pyarb {
+    
+// Create and return a dictionary
```
bcumming: Say what the dictionary is, i.e. something like "returns a python dictionary that python users can use to look up which options the Arbor library was configured with at compile time."

python/config.cpp
```diff
@@ -1,9 +1,50 @@
...
+#else
+    dict[pybind11::str("gpu")]     = pybind11::bool_(false);
+#endif
+    dict[pybind11::str("version")] = pybind11::str(ARB_VERSION);
+    return dict;
+}
+
+void print_config(const pybind11::dict &d) {
+    std::stringstream s;
+    s << "Arbor's configuration:" << std::endl;
```
bcumming: just `<< Arbor's configuration:\n";
`std::endl` can have heavy performance cost, and should only be used when you really want to force that the output is flushed.

python/test/unit_distributed/test_contexts_arbmpi.py
```diff
@@ -15,10 +16,15 @@
 except ModuleNotFoundError:
     from test import options
 
+# check Arbor's configuration of mpi
```
bcumming: Remove the named tuple tests. We were interested in named tuples, if we could get our `arbor.config()` to return one, but we have settled on dictionaries.

python/test/unit_distributed/test_contexts_arbmpi.py
```diff
@@ -36,7 +42,7 @@ def setUpClass(self):
     def tearDownClass(self):
         #print("tearDown --- TestContextMPI class")
```
bcumming: Remove this commented out code.

python/test/unit_distributed/test_contexts_mpi4py.py
```diff
@@ -3,6 +3,7 @@
 # test_contexts_mpi4py.py
 
 import unittest
+from collections import namedtuple
```
bcumming: Again, remove all `namedtuple` mentions.

python/config.cpp
akuesters: Done

python/config.cpp
akuesters: Done

python/test/unit_distributed/test_contexts_arbmpi.py
akuesters: Done

python/test/unit_distributed/test_contexts_arbmpi.py
akuesters: Done

python/test/unit_distributed/test_contexts_mpi4py.py
akuesters: Done

PR #714, opened 2019-03-04
Final fixes to Python config PR

akuesters:
Missed to adjust `runner.py` to new config


python/test/unit_distributed/runner.py
```diff
@@ -4,7 +4,13 @@
 
 import unittest
 import arbor as arb
-if (arb.mpi4py_compiled() and arb.mpi_compiled()):
+
+# check Arbor's configuration of mpi
+dict = arb.config()
```
bcumming: Instead of calling this `dict`, how about `config`?

python/test/unit_distributed/runner.py
```diff
@@ -4,7 +4,13 @@
...
+config_mpi = dict["mpi"]
+config_mpi4py = dict["mpi4py"]
```
bcumming: And call these `has_mpi4py` or `with_mpi4py` or `mpi_enabled`.

python/test/unit_distributed/test_contexts_arbmpi.py
```diff
@@ -18,12 +18,13 @@
...
 # check Arbor's configuration of mpi
 dict = arb.config()
+config_mpi = dict["mpi"]
 mpi_check = dict["mpi"]
 
 """
 all tests for distributed arb.context using arbor mpi wrappers
 """
-@unittest.skipIf(mpi_check == False, "MPI not enabled!")
+@unittest.skipIf(config_mpi == False, "MPI not enabled!")
```
bcumming: We don't need the exclamation at the end of the string, i.e. just "MPI not enabled" is enough.

PR #720, opened 2019-03-14
Include Python build and unit tests in Travis CI

akuesters:
#fixes 715

Note: 
Python 3.6 is tested (serial and mpi), in particular: 
- 3.6.3 on linux with a virtualenv, 
- 3.6.5 as downgrade from 3.7 on osx as in https://stackoverflow.com/questions/52584907/how-to-downgrade-python-from-3-7-to-3-6?rq=1 (other solutions did not work)


scripts/travis/build.sh
```diff
@@ -54,24 +66,36 @@ cd $build_path
...
+if [[ "${WITH_DISTRIBUTED}" == "mpi" ]]; then
+    progress "C++ distributed unit tests (MPI)"
     make unit-mpi -j4        || error "building MPI distributed unit tests"
     ${launch} ./bin/unit-mpi || error "running MPI distributed unit tests"
 fi
 
+if [[ "${WITH_PYTHON}" == "true" ]]; then
+    progress "Building python module"
+    make pyarb -j4                                                           || error "building pyarb"
+    if [[ "${WITH_DISTRIBUTED}" == "serial" ]]; then
```
bcumming: Should we always run the serial unit tests for Python, just like we always run the serial C++ tests?

akuesters: Since I included the `WITH_PYTHON=true` in the `.travis.yml` for all entries in the matrix, I would leave it here. Basically, that way we already always run the serial unit test. I just introduced `WITH_PYTHON` in case we would like to deactivate python for mpi testing. Then we only have to set it to false in the mpi matrix entries.

bcumming: This looks like it won't run the serial test if `WITH_PYTHON==true` and `WITH_DISTRIBUTED=mpi`.

akuesters: Yes, you`re right. I'll change that.

akuesters: Done.

PR #723, opened 2019-03-25
Python feature addition for identifiers and event_generators/schedules

akuesters:
Added: 
- Identifiers (`cell_member`) + according tests
- `event_generator` and schedules (`regular_schedule`, `explicit_schedule`, `poisson_schedule`) + according tests

Open: 
- Methods of `schedule` class (e.g. `events(t0,t1)`, `reset`) if needed
- Connection of `events` (as `py::list` in `event_generator_shim`) to readonly argument `event_generator.sched` and/or `__str__` and `__repr__` strings
- Identifiers need to be added as needed (e.g. `cell_kind` for `recipe`)


python/event_generator.hpp
```diff
@@ -0,0 +1,115 @@
+#pragma once
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <arbor/common_types.hpp>
+#include <arbor/event_generator.hpp>
+#include <arbor/schedule.hpp>
+
```
bcumming: I don't think we need this header. All of the contents of this file can go into the `event_generators.cpp`.
The `X_schedule_shim` types don't need to be used anywhere else that I can see.

python/strings.cpp
```diff
@@ -32,4 +41,26 @@ std::string proc_allocation_string(const arb::proc_allocation& a) {
     return s.str();
 }
 
+std::string schedule_explicit_string(const explicit_schedule_shim& e) {
```
bcumming: Define these inside `event_generator.cpp` to avoid having to have `event_generator.hpp`

python/test/unit/test_event_generators.py
```diff
@@ -0,0 +1,138 @@
...
+        rs = arb.regular_schedule()
+        self.assertEqual(rs.tstart, np.finfo(np.float32).max)
+        self.assertEqual(rs.dt, 0)
+        self.assertEqual(rs.tstop, np.finfo(np.float32).max)
+        self.assertEqual(rs.tstart, rs.tstop)
+
+    def test_dt_contor_regular_schedule(self):
+        rs = arb.regular_schedule(0.025)
+        self.assertEqual(rs.tstart, 0)
+        self.assertEqual(float("{0:.3f}".format(rs.dt)), 0.025)
```
bcumming: Why do we cast like this to get the `dt` value?

python/test/unit/test_event_generators.py
```diff
@@ -0,0 +1,138 @@
...
+except ModuleNotFoundError:
+    from test import options
+
+"""
+all tests for event generators (regular, explicit, poisson)
+"""
+
+class RegularSchedule(unittest.TestCase):
+    def test_default_regular_schedule(self):
+        rs = arb.regular_schedule()
```
bcumming: I don't know about testing default construction of these, because the test has to be rewritten if the default values are changed. You can convince me otherwise, if you think it is a good idea.

python/event_generator.cpp
```diff
@@ -0,0 +1,137 @@
...
+            "Construct a Poisson schedule with arguments:\n"
+            "  tstart: Events in the Poisson schedule start being delivered from this time(in ms).\n"
+            "  freq:   The expected frequency (in Hz).\n"
+            "  seed:   The seed of the Mersenne Twister pseudo-random generator of 64-bit numbers with a state size of 19937 bits.")
+        .def_readwrite("tstart", &poisson_schedule_shim::tstart,
+            "Events in the Poisson schedule start being delivered from this time (in ms).")
+        .def_readwrite("freq", &poisson_schedule_shim::freq,
+            "The expected frequency (in Hz).")
+        .def_readwrite("seed", &poisson_schedule_shim::seed,
+            "The seed of the Mersenne Twister pseudo-random generator of 64-bit numbers with a state size of 19937 bits.")
```
bcumming: Simplify to "Set seed for the random number generator."

python/event_generator.cpp
```diff
@@ -0,0 +1,137 @@
...
+            "  dt:     0 ms.\n"
+            "  tstop:  tstart.")
+        .def(pybind11::init<arb::time_type>(),
+            "dt"_a,
+            "Construct a regular schedule starting at tstart=0 ms, ending at tstop=maximal numerical number (in ms) with argument:\n"
+            "  dt:      The distance between time points in the regular sequence (in ms).\n")
+        .def(pybind11::init<arb::time_type, arb::time_type, arb::time_type>(),
+            "tstart"_a, "tstop"_a, "dt"_a,
+            "Construct a regular schedule with arguments:\n"
+            "  tstart: Events in the regular sequence start being delivered from this time (in ms).\n"
```
bcumming: "delivery time of first event in sequence (ms)"

python/event_generator.cpp
```diff
@@ -0,0 +1,137 @@
...
+            "  tstop:  Events in the regular sequence stop being delivered after this time (in ms).\n"
```
bcumming: "no events delivered after this time (ms)"

python/event_generator.cpp
```diff
@@ -0,0 +1,137 @@
...
+            "  dt:     The distance between time points in the regular sequence (in ms).\n")
```
bcumming: "interval between time points (in ms)"

python/event_generator.cpp
```diff
@@ -0,0 +1,137 @@
...
+        .def_readwrite("tstart", &regular_schedule_shim::tstart,
+            "Events in the regular sequence start being delivered from this time (in ms).")
```
bcumming: I don't think we need to say "regular sequence" everywhere.
"delivery time of the first event in the sequence".

python/event_generator.hpp
akuesters: Done.

python/strings.cpp
akuesters: Done.

python/test/unit/test_event_generators.py
akuesters: Done by using `assertAlmostEqual` (checking if `round(a-b, 7) == 0`).

python/test/unit/test_event_generators.py
akuesters: Done. Set default values `tstart=None` and `tstop=None`.

python/event_generator.cpp
akuesters: Done (leaving out "Set" due to consistancy).

python/event_generator.cpp
akuesters: Done (added some "the" s).

python/event_generator.cpp
akuesters: Done.

python/event_generator.cpp
akuesters: Done.

python/event_generator.cpp
akuesters: Done.

python/event_generator.cpp
```diff
@@ -0,0 +1,262 @@
...
+    {}
+
+    regular_schedule_shim(pybind11::object t0, time_type deltat, pybind11::object t1):
+        tstart(t0),
+        dt(deltat),
+        tstop(t1)
+    {}
+
+    arb::schedule schedule() const {
+        return arb::regular_schedule(tstart.is_none()? arb::terminal_time: pybind11::cast<time_type>(tstart), dt, tstop.is_none()? arb::terminal_time: pybind11::cast<time_type>(tstop));
```
bcumming: These will give hard errors if `tstart` and `tstop` can't be coverted to `time_type`.

e.g.
```
s = arb.regulard_schedule()
s.tstart = 'cat'
```
When the string is cast to float, and exception will be thrown, which will give the user a confusing error message. I have just noticed that the same thing will happen in the context. 

We might wrap the `pybind11::cast` call in a try-catch block, that throws an exception with a meaningful error message, such as "tstart must be either None or a non-negative value".

akuesters: When initializing `tstart` and `tstop` with `pybind::none()` as `pybind11:object` the constructors of the `regular_schedule` won't throw exceptions, since every `object` is legitimate. Only when constructing the `event_generator` it throws an exception where it is too late. So should we consider `arb::time_type` and setting `None` as default argument (allowed in default setting of args) which is then set to `arb::terminal_time`? That way python should throw a "wrong argument type" error instead.

akuesters: If we keep `pybind1::object`'s the try-catch block only hits in `schedule()` which is called in the `make_event_generator()`

```
    arb::schedule schedule() const {
        time_type t0;
        time_type t1;
        try {
            t0 = tstart.is_none()? arb::terminal_time: pybind11::cast<time_type>(tstart);
        }
        catch (...){
            std::cerr << "casting error: argument tstart must be either None or a non-negative value." << "\n";
        }
        try {
            t1 = tstop.is_none()? arb::terminal_time: pybind11::cast<time_type>(tstop);
        }
        catch (...){
            std::cerr << "casting error: argument tstop must be either None or a non-negative value." << "\n";
        }
        return arb::regular_schedule(t0, dt, t1);
    }
```

akuesters: Is there a way to include a try-catch in an initialization list?

halfflat: There is, it turns out!
https://en.cppreference.com/w/cpp/language/function-try-block

bcumming: There is no way to put try-catch in an initialization list.
Instead of `std::cerr`, I would throw a `std::runtime_error` with our clear message. The error message could be reduced to "regular schedule tstart must be either None or non-negative number."

You are right that it is not good to catch the problem "late" when converting from Python shim to a C++ `schedule`, because the error thrown at a different location to where the user set the invalid value.

How about writing a "setter" for the shim, that will check the type of the value being set by the Python user? The pybind11 docs show how to control access to properties via setters and getters. You could implement the setters and getters as lambdas.

https://pybind11.readthedocs.io/en/master/classes.html#instance-and-static-fields

Here is some pseudo code that just sets without checking: you could add the checking to the setter lambda:
```
    .def_readwrite("tstart",
        [](regular_schedule_shim& s, pybind11::object v) {s.tstart=v;},
        [](regular_schedule_shim& s) {return s.tstart;},                                                                                    
        "The delivery time of the first event in the sequence (in ms).")
```

halfflat: I misunderstood — I thought you were referring to member initialization.

akuesters: With 

```
.def_property("tstart", &regular_schedule_shim::get_tstart, &regular_schedule_shim::set_tstart,
            "The delivery time of the first event in the sequence (in ms).")
```

using 

```
void set_tstart(const pybind11::object v) {
        try {
            if (!v.is_none()) {
                tstart = v;
                if (tstart.cast<time_type>() < 0. or !pybind11::isinstance<time_type>(tstart) or !pybind11::isinstance<std::string>(pybind11::str(tstart)) )
                    throw std::runtime_error("regular schedule tstart must be either None or a non-negative value.");
            }
            else {
                tstart = v;
                if (!v.is_none())
                    throw std::runtime_error("regular schedule tstart must be either None or a non-negative value.");
            }
        }
        catch (std::runtime_error& e) {
            std::cout << e.what() << "\n";
        }
    };

const pybind11::object &get_tstart() const { return tstart; }
```

I still get 

```
>>> import arbor
>>> rs = arbor.regular_schedule()
>>> rs
<regular_schedule: tstart None, dt 0 ms, tstop None>
>>> rs.tstart=-1
regular schedule tstart must be either None or a non-negative value.
>>> rs.tstart=True
regular schedule tstart must be either None or a non-negative value.
>>> rs.tstart="cat"
Unable to cast Python instance to C++ type (compile in debug mode for details)
>>> rs.tstart='dog'
Unable to cast Python instance to C++ type (compile in debug mode for details)
```

bcumming: I think it is because the `cast<time_type>` throws an exception with the message `"Unable to cast Python instance to C++ type (compile in debug mode for details)"`, which is caught and then just forwarded.

akuesters:
With the help of pybind11's automatic conversion from/to pybind11::object to/from arb::util::optional using pybind11's optional_caster

```
namespace pybind11 { namespace detail {
    template <typename T>
    struct type_caster<arb::util::optional<T>> : optional_caster<arb::util::optional<T>> {};
}}
```
 (see PR #738), we can now use arbor's `arb::util::optional<T>` as argument type in the python wrapper which then gets automatically converted to `pybind11::object`. 

With the function `pyarb::assert_predicate(var, predicate, msg)` in `exception.hpp` the predicate is tested and an exception is thrown.

python/error.hpp
```diff
@@ -18,18 +17,16 @@ namespace pybind11 { namespace detail {
...
+struct pyarb_error: std::runtime_error {
+    pyarb_error(const std::string& what_msg):
+        std::runtime_error(what_msg) {}
 };
 
-template <typename T, typename F>
-T&& assert_predicate(T&& t, F&& f, const char* msg) {
-    if (!f(t)) throw std::runtime_error(msg);
-    return std::forward<T>(t);
+static
```
bcumming: Simplify this by not passing the predicate, because the test does not need to be performed in the function.
The predicate is passed to the py2optional function, because it needs to be applied after the cast.

python/event_generator.cpp
```diff
@@ -64,27 +72,29 @@ struct regular_schedule_shim {
...
 
+    explicit_schedule_shim(std::vector<time_type> t) {
+        set_times(t);
     }
 
     // getter and setter (in order to assert when being set)
-    void set_times(const std::list<time_type> t0) {
-        for (auto& t: t0) {
-            times.push_back(pyarb::assert_predicate(t, is_nonneg, "time must be a non-negative number."));
+    void set_times(std::vector<time_type> t) {
```
bcumming: use `std::vector` instead of `std::list`, because it is faster. Particularly applying `sort`.

python/event_generator.cpp
```diff
@@ -100,22 +110,28 @@ struct poisson_schedule_shim {
...
-    {}
+    poisson_schedule_shim() = default;
+
+    poisson_schedule_shim(time_type ts, time_type f, rng_type::result_type s) {
+        set_tstart(ts);
+        set_freq(f);
+        seed = s;
+    }
 
-    void set_tstart(const time_type t0) {
```
bcumming: don't need the const

PR #735, opened 2019-04-15
Add conductivity to implicit voltage solve step for stability.

halfflat:
Changes implicit solve step from:
> solve (c/δt + L) v' = c/δt v - J

to
> solve (c/δt + g + L)v' = (c/δt +g) v - J

where _c_ is capacitance, _g_ is membrane conductance, _J_ is membrane current.

* Compute conductivity contribution for mechanisms from symbolic d/dv of current contribution (extracted from linearity test.)
* Add new modcc 'source kind' for conductivity; tie to `vec_g_`.
* Add conductivity field to fvm shared state.
* Include conductivity in matrix assemblies for solution.

### DIFF
--- a/arbor/backends/gpu/matrix_state_interleaved.hpp
+++ b/arbor/backends/gpu/matrix_state_interleaved.hpp
@@ -23,6 +23,7 @@ void assemble_matrix_interleaved(
     const fvm_value_type* invariant_d,
     const fvm_value_type* voltage,
     const fvm_value_type* current,
+    const fvm_value_type* conductivitt,
     const fvm_value_type* cv_capacitance,
     const fvm_value_type* area,
     const fvm_index_type* sizes,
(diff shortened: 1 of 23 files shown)


arbor/backends/gpu/matrix_state_interleaved.hpp
```diff
@@ -267,10 +268,10 @@ struct matrix_state_interleaved {
     //   dt_intdom         [ms]     (per integration domain)
     //   voltage           [mV]     (per compartment)
     //   current density   [A.m^-2] (per compartment)
```
bcumming: Add comment with units for codunctivtt

arbor/backends/gpu/matrix_state_interleaved.hpp
```diff
@@ -267,10 +268,10 @@ struct matrix_state_interleaved {
...
-    void assemble(const_view dt_intdom, const_view voltage, const_view current) {
+    void assemble(const_view dt_intdom, const_view voltage, const_view current, const_view conductivitt) {
         assemble_matrix_interleaved
             (d.data(), rhs.data(), invariant_d.data(),
-             voltage.data(), current.data(), cv_capacitance.data(), cv_area.data(),
+             voltage.data(), current.data(), conductivitt.data(), cv_capacitance.data(), cv_area.data(),
```
bcumming: `conductivtt` or `conductivity`?

halfflat: Yeah, that was an interesting typo that somehow was correctly propagated and compiled.

halfflat:
Ring bench with medium configuration on nsuite appears to be 2% to 7% slower on both cpu and gpu (based on a couple of runs on `ault09`).

PR #737, opened 2019-04-24
Fix reduce-by-key CUDA

### LINKED ISSUE #736: Hang with sequential use of `arb::gpu::reduce_by_key` on V100/CUDA 10.1
In at least one circumstance, calling `reduce_by_key` twice in succession within a kernel causes the kernel to hang. This behaviour is exhibited on a V100 with CUDA 10.1.

This is triggered in the GPU code with the stability correction.

* Try to isolate this in a unit test.
* Fix it.

bcumming:
Fix bug in CUDA `reduce_by_key` implementation on V100 or later GPUs.

The bug was not triggered for current use cases of the algorithm in Arbor, though it will be a problem when more than one reduction is to be performed in a single kernel invocation, which is required for accumulating both current and conductance values.

Fixes  #736.

### DIFF
--- a/arbor/backends/gpu/reduce_by_key.hpp
+++ b/arbor/backends/gpu/reduce_by_key.hpp
@@ -11,12 +11,10 @@ namespace impl{
 
 constexpr unsigned mask_all = 0xFFFFFFFF;
 
-// return the first power of 2 that is less than or equal to i
+// Return the largest power of 2 that is less than i
+// Caveat: i must be non-zero.
 __device__ __inline__
 unsigned rounddown_power_of_2(std::uint32_t i) {
-    // handle power of 2 and zero
-    if(__popc(i)<2) return i;
-
     return 1u<<(31u - __clz(i));
 }
 
@@ -33,6 +31,8 @@ unsigned roundup_power_of_2(std::uint32_t i) {
 
     return 1u<<(32u - __clz(i));
 }
+
+} // namespace impl
 
 // run_length Stores information about a run length.
 //
@@ -59,20 +59,18 @@ struct run_length {
     }
 
     __device__
-    run_length(int idx) {
-        lane_id = threadIdx.x%threads_per_warp();
-        __syncwarp();
-        // TODO: calculate key mask directly from array sizes (outside main loop)
-        key_mask = __activemask();
-        unsigned num_lanes = threads_per_warp()-__clz(key_mask);
+    run_length(int idx, unsigned mask) {
+        key_mask = mask;
+        lane_id = threadIdx.x%impl::threads_per_warp();
+        unsigned num_lanes = impl::threads_per_warp()-__clz(key_mask);
 
         auto right_limit = [num_lanes] (unsigned roots, unsigned shift) {
             unsigned zeros_right  = __ffs(roots>>shift);
             return zeros_right ? shift -1 + zeros_right : num_lanes;
         };
 
-        // determine if this thread is the root
-        int left_idx  = __shfl_up_sync(key_mask, idx, 1);
+        // determine if this thread is the root (i.e. first thread with this key)
+        int left_idx  = __shfl_up_sync(key_mask, idx, lane_id? 1: 0);
         int is_root = 1;
         if(lane_id>0) {
             is_root = (left_idx != idx);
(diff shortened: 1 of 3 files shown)


arbor/backends/gpu/reduce_by_key.hpp
```diff
@@ -96,74 +46,73 @@ float get_from_lane(float value, unsigned lane) {
...
 
-    while (width) {
-        unsigned source_lane = run.lane_id + width;
+    while (__any_sync(run.key_mask, shift)) {
+        const unsigned w = participate? shift: 0;
+        const unsigned source_lane = run.lane_id + w;
 
-        auto source_value = impl::get_from_lane(contribution, source_lane);
-        if (source_lane<rhs) {
+        T source_value = __shfl_sync(run.key_mask, contribution, source_lane);
```
halfflat: It probably makes little difference, but I believe you could just use `__shfl_down_sync(..., w)` rather than compute `source_lane` explicitly.

arbor/backends/gpu/reduce_by_key.hpp
```diff
@@ -11,80 +11,30 @@ namespace impl{
...
-#if __CUDACC_VER_MAJOR__ < 9
-    return __ballot(pred);
-#else
-    return __ballot_sync(mask_all, pred);
-#endif
-}
-
-// return the power of 2 that is less than or equal to i
+// Return the largest power of 2 that is less than i
+// Caveat: i must be non-zero.
```
halfflat: Just regarding the comment: this returns the largest power of 2 no greater than i, right?

arbor/backends/gpu/reduce_by_key.hpp
```diff
@@ -11,80 +11,30 @@ namespace impl{
...
-// TODO: CUDA 9 provides a double precision version of __shfl that
-// implements this hack. When we make CUDA 9 the minimum version, use
-// the CUDA implementation instead.
-
+// Return the first power of 2 that is larger than or equal to i
+// input i is in the closed interval [0, 2^31]
+// The result for i>2^31 is invalid, because it can't be stored in an
+// unsigned 32 bit integer.
+// This isn't a problem for the use case that we will use it for, because
+// it will be used for values in the range [0, threads_per_warp]
```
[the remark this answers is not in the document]
bcumming: This function should be removed: it isn't needed.

modcc/printer/cudaprinter.cpp
```diff
@@ -445,7 +449,7 @@ void emit_state_update_cu(std::ostream& out, Symbol* from,
         out << "arb::gpu::reduce_by_key(";
         is_minus && out << "-";
         out << from->name()
-            << ", params_." << d.data_var << ", " << index_i_name(d.index_var) << ");\n";
+            << ", params_." << d.data_var << ", " << index_i_name(d.index_var) << ", lane_mask_);\n";
```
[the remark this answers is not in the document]
bcumming: Yes, it is. Before doing the full conversion to calculating run-length outside `reduce_by_key`, more tests need to be added to the cuda printer to ensure that it is only called on `APIMethod`s that perform reductions.

I was not comfortable using the `__activemask` intrinsic to find the mask inside `reduce_by_key`, because the docs didn't convince me that there wouldn't be divergent threads, even directly after `__syncwarp`.

arbor/backends/gpu/reduce_by_key.hpp
bcumming: I think you are right.

arbor/backends/gpu/reduce_by_key.hpp
bcumming: Yes.
Due to other simplifications, this is no longer needed, and has been removed.

arbor/backends/gpu/reduce_by_key.hpp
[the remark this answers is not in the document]
bcumming: Removed.

bcumming:
Righteos, the algorithm has been simplified further, in a way that addresses all of the review comments (I think).

I also made modcc only generate a thread_mask for API Methods that write indexed variables. Not strictly neccesary, because I doubt that there are any performance overheads, but it makes for cleaner generated code.

PR #738, opened 2019-04-25
Prepare arb::util::optional to be used for automatic conversion from/to pybind11::object

akuesters:
In `optional.hpp`, added 
- `value_type`
- `emplace`
in order to be used for pybind11's automatic conversion from/to `pybind11::object` to/from `arb::util::optional` using pybind11's `optional_caster`

```
namespace pybind11 { namespace detail {
    template <typename T>
    struct type_caster<arb::util::optional<T>> : optional_caster<arb::util::optional<T>> {};
}}
```


arbor/include/arbor/util/optional.hpp
```diff
@@ -165,6 +166,14 @@ namespace detail {
             }
             set = false;
         }
+
+        template <typename Y>
+        void emplace(Y&& y) {
+            if (set)
+                data.destruct();
```
halfflat: Just in terms of the Arbor code style, we should have some braces around the body of the `if`.

akuesters: Thanks. Just pushed the adjustments.

PR #744, opened 2019-05-14
Py feature context

akuesters:
+ Included conversion from `pybind11::object()` to `arb::util::optional` for `arbor.proc_allocation().gpu_id` and exception handling for `arbor.context(mpi=comm)`.

+ cleaned-up `test_event_generator.py`

+ for discussion: How do we want to handle the strings? Include them in appropriate `.cpp` or collect them in `strings.c/hpp`?


python/context.cpp
```diff
@@ -16,67 +18,158 @@
...
 void register_contexts(pybind11::module& m) {
     using namespace std::string_literals;
     using namespace pybind11::literals;
+    using opt_int = arb::util::optional<int>;
+#ifdef ARB_MPI_ENABLED
+    using opt_mpi_comm = arb::util::optional<mpi_comm_shim>;
+#endif
 
-    pybind11::class_<arb::proc_allocation> proc_allocation(m, "proc_allocation");
+// proc_allocation
```
bcumming: indentation.

python/context.cpp
```diff
@@ -16,67 +18,158 @@
...
-        .def(pybind11::init<int, int>(), "threads"_a, "gpu"_a=-1,
-             "Arguments:\n"
-             "  threads: The number of threads available locally for execution.\n"
-             "  gpu:     The index of the GPU to use, defaults to -1 for no GPU.\n")
-        .def_readwrite("threads", &arb::proc_allocation::num_threads,
+        .def(pybind11::init<int, pybind11::object>(),
+            "threads"_a=1, "gpu_id"_a=pybind11::none(),
+            "Construct an allocation with arguments:\n"
+            "  threads: The number of threads available locally for execution (defaults to 1).\n"
+            "  gpu_id:  The index of the GPU to use (defaults to None for no GPU).\n")
```
bcumming: It might be clearer to change the default text to `(default 1)` and `(default None)`.

python/context.cpp
```diff
@@ -16,67 +18,158 @@
...
 #ifdef ARB_MPI_ENABLED
         .def(pybind11::init(
-            [](const arb::proc_allocation& alloc, mpi_comm_shim c){return context_shim(arb::make_context(alloc, c.comm));}),
-             "alloc"_a, "c"_a,
-             "Arguments:\n"
+            [](const proc_allocation_shim& alloc, pybind11::object mpi){
+                if (mpi.is_none()) {
+                    return context_shim(make_context_shim(alloc));
+                }
+                opt_mpi_comm c = py2optional<mpi_comm_shim>(mpi,
```
bcumming: You don't need the `opt_mpi_comm` type if you use `auto` here, i.e.
```C++
auto c = py2optional<mpi_comm_shim>(...);
```

python/context.cpp
```diff
@@ -16,67 +18,158 @@
...
+template <typename Comm>
+arb::context make_context_shim(const proc_allocation_shim& Alloc, Comm comm) {
+    return arb::make_context(Alloc.allocation(), comm);
+}
+
 void register_contexts(pybind11::module& m) {
     using namespace std::string_literals;
     using namespace pybind11::literals;
+    using opt_int = arb::util::optional<int>;
+#ifdef ARB_MPI_ENABLED
```
bcumming: The `opt_mpi_comm` type can be removed if you use `auto`, see comments below.

python/test/unit/test_contexts.py
```diff
@@ -20,48 +20,72 @@
...
         alloc = arb.proc_allocation()
 
-        # test that by default proc_allocation has 1 thread and no GPU, no MPI
+        # test that by default proc_allocation has 1 thread and no GPU
         self.assertEqual(alloc.threads, 1)
+        self.assertEqual(alloc.gpu_id, None)
         self.assertFalse(alloc.has_gpu)
-        self.assertEqual(alloc.gpu_id, -1)
 
+    def test_set_alloc(self):
```
bcumming: for consistency with the first test, use the full word `allocation` in the title, i.e. `test_set_allocation`. Same goes for the tests below.

python/strings.cpp
```diff
@@ -21,23 +21,10 @@ std::string context_string(const arb::context& c) {
...
     s << "<context: threads " << arb::num_threads(c)
       << ", gpu " << (gpu? "yes": "None")
-      << ", distributed " << (mpi? "MPI": "Local")
+      << ", distributed " << (mpi? "MPI": "local")
       << " ranks " << arb::num_ranks(c)
       << ">";
     return s.str();
 }
 
-std::string proc_allocation_string(const arb::proc_allocation& a) {
```
bcumming: Regarding converting types to strings, we could consider using a string formatting library, like fmt (see https://github.com/fmtlib/fmt), or use our own `pprintf` implementation that is in `arbor/util/strprintf.hpp`.

The benefit of such a function is that it would make it easy to write lambdas that generate strings, doing away with the need for stand alone functions. For now, don't worry about it.

python/test/unit_distributed/test_contexts_mpi4py.py
```diff
@@ -42,14 +42,31 @@ def test_communicator_mpi4py(self):
     def test_context_mpi4py(self):
         comm = arb.mpi_comm_from_mpi4py(mpi.COMM_WORLD)
 
-        # test context with mpi usage
+        # test context with mpi
+        ctx = arb.context(mpi=comm)
```
bcumming: A more user-friendly approach would be to support mpi4py directly, to that this
```python
comm = arb.mpi_comm_from_mpi4py(mpi.COMM_WORLD)
ctx = arb.context(mpi=comm)
```

would become
```python
ctx = arb.context(mpi.COMM_WORLD)
```
How do you feel about this?

python/context.cpp
```diff
@@ -16,67 +18,158 @@
 
 namespace pyarb {
 
+namespace {
+auto is_nonneg_int = [](auto&& t){ return ((t==int(t) and t>=0)); };
```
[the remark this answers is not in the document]
akuesters: Done.

python/context.cpp
```diff
@@ -16,67 +18,158 @@
...
+        set_gpu_id(gpu);
+    }
+
+    // getter and setter (in order to assert when being set)
+    void set_gpu_id(pybind11::object gpu) {
+        gpu_id = py2optional<int>(gpu, "gpu id must be None, or a non-negative integer.", is_nonneg_int);
+    };
+
+    void set_num_threads(int threads) {
+        pyarb::assert_throw(is_nonneg_int(threads), "threads must be a non-negative integer.");
```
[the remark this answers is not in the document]
akuesters: Done.

python/context.cpp
akuesters: Done.

python/context.cpp
akuesters: Done.

python/context.cpp
akuesters: Done.

python/context.cpp
akuesters: Done.

python/test/unit/test_contexts.py
akuesters: Done.

python/strings.cpp
akuesters: Lleft open for now.

python/test/unit_distributed/test_contexts_mpi4py.py
akuesters: I think this would be good, but using `ctx = arb.context(mpi=mpi.COMM_WORLD)` as the code is right now gives (the correct) `RuntimeError: mpi must be None, or an MPI communicator.`, since it is not a `mpi_comm_shim`.

bcumming: Indeed. We would need to extend the interface to support this.

Issue #748, opened 2019-05-21
Arbitrary ions in modcc and FVM cells.


halfflat:
We need to support arbitrary ion names that may be described in user-supplied NMODL files.

1. Replace `ionKind` enums with a string containing the corresponding ion name.
2. Re-jig as required the ion channel handling in modcc and ion setup in the FVM implementations.

PR #749, opened 2019-05-21
Generalized ion names

halfflat:
Implements #748.

* Replace ionKind enumerations with corresponding string values.
* Rename `ion.hpp` as `ion_info.hpp`.
* Re-jig 'nonspecific' current handling in modcc: make explicit the external source of the corresponding current variable, and treat as non-indexed those locals with an external variable that has no data source, as opposed to special-casing the `ionKind::nonspecific` value.
* Add another field in ion dependency information that captures a write to an ionic reversal potential.


modcc/debug-CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+# The modcc compiler is split into a staric library and driver;
```
bcumming: Should this file be deleted before merging? It looks like a backup copy.

halfflat: It totally should. Mismatch between my source tree and the commit tree; I'll remove it.

halfflat:
> I have one question that wasn't answered when I was reading the proposed changes: do we now check the "spelling" of variables that are marked as `READ` or `WRITE` in a USEION statement? That is, if an ion is named `mg`, to we ensure that variables have valid names like `mgi`, `img`, `mgo`, etc?

Yup; the original code that does this is retained, which compared the names against `to_string(ionKind)`. Now it just compares against the name of the ion directly.

PR #752, opened 2019-05-22
fix spurious compile time assertionsin GPU matrix solver

bcumming:
There were run time assertions thrown when compiled with `ARB_WITH_ASSERTIONS` and `ARB_WITH_GPU` set. These were spurious, and it was clear that it has been a long time since the GPU version was last run with assertions enabled :worried:


bcumming:
Passed all tests on a V100 with CUDA 10 and GCC 7.3.0

PR #755, opened 2019-05-27
Fixes managed memory issue on K80 (Jureca)

akuesters:
*Problem:* 
Arbor for GPU wouldn't run on K80 (Jureca) getting 

```
Thread 1 "ring" received signal CUDA_EXCEPTION_15, Invalid Managed Memory Access.
0x000000000049ec87 in arb::gpu::matrix_state_fine<double, int>::matrix_state_fine 
```

from cuda-gdb and 

```
========= Error: process didn't terminate successfully
========= Fatal UVM CPU fault due to invalid operation
=========     during read access to address 0x2303964000
```

from cuda-memcheck. 

*Fix*: 
Stop accessing managed memory every time we want the single scalar value `data_size = data_partition.back()`.


arbor/backends/gpu/matrix_state_fine.hpp
```diff
@@ -345,12 +344,15 @@ struct matrix_state_fine {
...
             levels_start.push_back(prev_end + branch_map.size());
-            data_size.push_back(pos);
+            data_partition.push_back(pos);
         }
 
         // set matrix state
         matrix_size = p.size();
 
+	// set data partition size
+	data_size = data_partition.back();
```
bcumming: Instead you can set `data_size = pos;` immediately after the loop on line349.

arbor/backends/gpu/matrix_state_fine.hpp
```diff
@@ -475,14 +477,14 @@ struct matrix_state_fine {
     template <typename VFrom, typename VTo>
     void flat_to_packed(const VFrom& from, VTo& to ) {
         arb_assert(from.size()==matrix_size);
-        arb_assert(to.size()==data_size.back());
+        arb_assert(to.size()==data_size);
```
bcumming: I think that it could have been these calls here that were causing the original problem. They could have been called when a kernel was running on the GPU, accessing the same memory.

arbor/backends/gpu/matrix_state_fine.hpp
akuesters: Done.

PR #759, opened 2019-06-04
Fix build with libunwind (fix up for ad1c78ab)

muffgaga:
* `cmake/FindUnwind.cmake` used the wrong variable (caused by commit ad1c78ab)
* `CMakeLists.txt` seems to use wrong syntax for providing a compiler define
  of `WITH_UNWIND` (used in `arbor/util/unwind.cpp`)

I'm no expert in cmake, so please have a deeper look into this → I only verified that my build used `-DWITH_UNWIND` for the g++ commands but nothing else (i.e. I did not try to get stack traces).

@w-klijn That's what I mentioned yesterday (when I tried to build arbor using spack).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -281,7 +281,7 @@ endif()
 find_package(Unwind)
 if(Unwind_FOUND)
     target_link_libraries(arbor-private-deps INTERFACE Unwind::unwind)
-    target_compile_definitions(arbor-private-deps ARB_WITH_UNWIND)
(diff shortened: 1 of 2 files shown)


bcumming:
Thanks very much @muffgaga!
I would merge it immediately, except we have to get you to transfer copyright to ETHZ or Juelich.
You are the first person from outside those two organisations to contribute, so we have to get a form ready.
Would you be able/willing to assign copyright?

PR #768, opened 2019-06-04
Py feature recipe wo probes

### LINKED ISSUE #760: Python wrapper: recipe
Does not include probe interface.

akuesters:
Fixes #760

Wraps arbor recipe (without probes, i.e. num_probes, probe_info, get_probe) including 

- cell_connection, 
- gap_junction_connection, 
- recipe with
  - num_cells
  - cell_description
  - cell_kind
  - num_sources
  - num_targets
  - num_gap_junctions_sites
  - event_generators
  - connections_on
  - gap_junctions_on
  - global_properties
- enum cell_kind in `identifiers.cpp`


python/context.cpp
```diff
@@ -153,8 +153,8 @@ void register_contexts(pybind11::module& m) {
 #else
         .def(pybind11::init(
             [](int threads, pybind11::object gpu){
-                auto gpu_id = py2optional<int>(gpu,
-                        "gpu id must be None, or a non-negative integer.", is_nonneg_int);
+                const char* gpu_err_str = "gpu_id must be None, or a non-negative integer";
+                auto gpu_id = py2optional<int>(gpu, gpu_err_str, is_nonneg_int);
```
bcumming: having the error string inline here doesn't make the code harder to read.

python/recipe.cpp
```diff
@@ -0,0 +1,234 @@
...
+            std::stringstream s;
+            s << "recipe supplied an invalid event generator for gid "
+            << gid << ": " << pybind11::str(g);
+            throw pyarb_error(s.str());
+        }
+        // get a reference to the python event_generator
+        auto& p = cast<const pyarb::event_generator_shim&>(g);
+
+        // convert the event_generator to an arb::event_generator
+        gens.push_back(
```
bcumming: The indentation here doesn't shorten the lines.

python/context.cpp
akuesters: Changed back.

python/recipe.cpp
akuesters: One line now.

PR #771, opened 2019-06-04
Py feature domain decomp

### LINKED ISSUE #761: Python wrapper: domain decomposition
* Functions: `partition_load_balance`
* Types:  `group_description` and `domain_decomposition`

akuesters:
Fixes #761

Adds function `partition_load_balance` and types `group_description` and `domain_decomposition`. 

No test so far (waiting for recipe+cells to be completed)


python/domain_decomposition.cpp
```diff
@@ -0,0 +1,92 @@
...
+}
+
+void register_domain_decomposition(pybind11::module& m) {
+    using namespace pybind11::literals;
+
+    // Group description
+    pybind11::class_<arb::group_description> group_description(m, "group_description",
+        "The indexes of a set of cells of the same kind that are grouped together in a cell group.");
+    group_description
+        .def(pybind11::init<arb::cell_kind, std::vector<arb::cell_gid_type>, arb::backend_kind>(),
```
bcumming: I see what you were talking about yesterday: if we provide an interface for users to construct a `group_description`, that interface must enforce preconditions, e.g. "all gid must be sorted".
I will check whether that precondition still applies: I suspect that we actually removed that precondition at some point (I can remember changing all the cell group implementations to use a hash table for looking up gid information, which doesn't require sorted input).

python/recipe.cpp
```diff
@@ -110,9 +110,31 @@ std::vector<arb::event_generator> py_recipe_shim::event_generators(arb::cell_gid
     return gens;
 }
 
+struct cell_connection_shim {
+    arb::cell_member_type source;
+    arb::cell_member_type dest;
+    float weight;
+    float delay;
```
bcumming: For consistency we should use `arb::time_type` for `delay`. I see that in the main Arbor library we use `float`, instead of `time_type`. First fix that, then use `time_type` here.

python/recipe.cpp
```diff
@@ -110,9 +110,31 @@ std::vector<arb::event_generator> py_recipe_shim::event_generators(arb::cell_gid
...
+
+    cell_connection_shim(arb::cell_member_type s, arb::cell_member_type d, float w, float t) {
+        source = s;
+        dest   = d;
+        weight = w;
+        set_delay(t);
+    }
+
+    // getter and setter (in order to assert when being set)
+    void set_delay(float t) {
```
bcumming: use `double` on this interface, to match Python underlying real type.

python/recipe.cpp
```diff
@@ -110,9 +110,31 @@ std::vector<arb::event_generator> py_recipe_shim::event_generators(arb::cell_gid
...
+        pyarb::assert_throw([](auto&& f){ return f>0.; }(t), "delay must be positive");
```
bcumming: We know that the type of `delay` is `float`, so it is clearer to use `[](float f)...` instead of `[](auto&& f)...`.

python/recipe.cpp
```diff
@@ -126,31 +148,24 @@ void register_recipe(pybind11::module& m) {
     using namespace pybind11::literals;
 
     // Connections
-    pybind11::class_<arb::cell_connection> cell_connection(m, "cell_connection",
+    pybind11::class_<cell_connection_shim> cell_connection(m, "cell_connection",
```
bcumming: We need a method for converting a `pyarb::cell_connection_shim` to a `arb::cell_connection`, otherwise pybind11 will have runtime errors when calling converting connections provided in user-defined `connections_on`.

PR #778, opened 2019-06-07
Py feature simulation

### LINKED ISSUE #763: Python wrapper: simulation
Wrap `arb::simulation` type and interface for running, resetting, etc simulations.

akuesters:
Fixes #763

- Wrap `arb::simulation` type and interface for running `run()`, resetting `reset()`, and `set_binning_policy(policy, bin_interval)`. 
- Wrap `enum binning_kind`


python/identifiers.cpp
```diff
@@ -59,6 +59,15 @@ void register_identifiers(pybind11::module& m) {
             "Use GPU backend.")
         .value("multicore", arb::backend_kind::multicore,
             "Use multicore backend.");
+
+    pybind11::enum_<arb::binning_kind>(m, "binning_kind",
```
bcumming: I am wondering if we should remove the `_kind` postfix in Python:
```
binning_kind::regular   -> binning::regular
backend_kind::multicore -> backend::multicore
```
It is a bit less verbose.

Thoughts?

akuesters: Sure, why not. Will do.

PR #779, opened 2019-06-07
Python wrappers for cell types

### LINKED ISSUE #767: Python wrapper: cells
Wrap `cable_cell.hpp`, `benchmark_cell.hpp`, `lif_cell.hpp`, `spike_source_cell.hpp`

bcumming:
Add full wrappers for
* `lif_cell`
* `spike_source_cell`
* `benchmark_cell`

And an opaque wrapper for `cable_cell` with a helper function that generates a random morphology identical to the ones in the nsuite benchmarks.

Fixes #767.

### DIFF
--- a/python/cells.cpp
+++ b/python/cells.cpp
@@ -33,11 +33,9 @@ arb::util::unique_any convert_cell(pybind11::object o) {
     if (isinstance<arb::lif_cell>(o)) {
         return arb::util::unique_any(cast<arb::lif_cell>(o));
     }
-    /*
     if (isinstance<arb::cable_cell>(o)) {
         return arb::util::unique_any(cast<arb::cable_cell>(o));
     }
-    */
 
     throw pyarb_error("recipe.cell_description returned \""
                       + std::string(pybind11::str(o))
--- a/python/cells.cpp
+++ b/python/cells.cpp
@@ -1,0 +1,228 @@
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <arbor/benchmark_cell.hpp>
+#include <arbor/cable_cell.hpp>
+#include <arbor/lif_cell.hpp>
+#include <arbor/schedule.hpp>
+#include <arbor/spike_source_cell.hpp>
+#include <arbor/util/unique_any.hpp>
+
+#include "error.hpp"
+#include "schedule.hpp"
+#include "strprintf.hpp"
+
+namespace pyarb {
+
+// Convert a cell description inside a Python object to a cell
+// description in a unique_any, as required by the recipe interface.
+//
+// Warning: requires that the GIL has been acquired before calling,
+// if there is a segmentation fault in the cast or isinstance calls,
+// check that the caller has the GIL.
+arb::util::unique_any convert_cell(pybind11::object o) {
+    using pybind11::isinstance;
+    using pybind11::cast;
+
+    if (isinstance<arb::spike_source_cell>(o)) {
+        return arb::util::unique_any(cast<arb::spike_source_cell>(o));
+    }
+    if (isinstance<arb::benchmark_cell>(o)) {
+        return arb::util::unique_any(cast<arb::benchmark_cell>(o));
+    }
+    if (isinstance<arb::lif_cell>(o)) {
+        return arb::util::unique_any(cast<arb::lif_cell>(o));
+    }
+    /*
+    if (isinstance<arb::cable_cell>(o)) {
+        return arb::util::unique_any(cast<arb::cable_cell>(o));
+    }
+    */
+
+    throw pyarb_error("recipe.cell_description returned \""
+                      + std::string(pybind11::str(o))
+                      + "\" which does not describe a known Arbor cell type");
+}
+
+//
+// Somewhat hacky bit of code for generating cells with random morphologies.
+//
+
+struct foo {
+    std::array<int, 2> vals;
+};
+
+// Parameters used to generate the random cell morphologies.
+struct cell_parameters {
+    cell_parameters() = default;
+
+    //  Maximum number of levels in the cell (not including the soma)
+    unsigned max_depth = 5;
+
+    // The following parameters are described as ranges.
+    // The first value is at the soma, and the last value is used on the last level.
+    // Values at levels in between are found by linear interpolation.
+    std::array<double,2> branch_probs = {1.0, 0.5}; //  Probability of a branch occuring.
+    std::array<unsigned,2> compartments = {20, 2};  //  Compartment count on a branch.
+    std::array<double,2> lengths = {200, 20};       //  Length of branch in μm.
+
+    // The number of synapses per cell.
+    unsigned synapses = 1;
+
+    friend std::ostream& operator<<(std::ostream& o, const cell_parameters& p) {
+        return
+            o << "<cell_parameters: depth " << p.max_depth
+              << ", synapses " << p.synapses
+              << ", branch_probs [" << p.branch_probs[0] << ":" << p.branch_probs[1] << "]"
+              << ", compartments [" << p.compartments[0] << ":" << p.compartments[1] << "]"
+              << ", lengths ["      << p.lengths[0]      << ":" << p.lengths[1] << "]>";
+    }
+};
+
+// Helper used to interpolate in branch_cell.
+template <typename T>
+double interp(const std::array<T,2>& r, unsigned i, unsigned n) {
+    double p = i * 1./(n-1);
+    double r0 = r[0];
+    double r1 = r[1];
+    return r[0] + p*(r1-r0);
+}
+
+arb::cable_cell branch_cell(arb::cell_gid_type gid, const cell_parameters& params) {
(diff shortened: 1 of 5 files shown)


python/cells.cpp
```diff
@@ -0,0 +1,226 @@
...
+        .def(pybind11::init<>(
+            [](const explicit_schedule_shim& sched){
+                return arb::spike_source_cell{sched.schedule()};}),
+            "schedule"_a, "Construct a spike source cell that generates spikes at a sequence of user-defined times")
+        .def(pybind11::init<>(
+            [](const poisson_schedule_shim& sched){
+                return arb::spike_source_cell{sched.schedule()};}),
+            "schedule"_a, "Construct a spike source cell that generates spikes at times defined by a Poisson sequence")
+        .def("__repr__", [](const arb::spike_source_cell&){return "<spike_source_cell>";})
+        .def("__str__",  [](const arb::spike_source_cell&){return "<spike_source_cell>";});
```
akuesters: - Add full stops in constructor descriptions. 
- Add `arbor: ` in `__str__` and `__repr__` (return strings)

python/cells.cpp
```diff
@@ -0,0 +1,226 @@
...
+        .def(pybind11::init<>(
+            [](const explicit_schedule_shim& sched, double ratio){
+                return arb::benchmark_cell{sched.schedule(), ratio};}),
+            "schedule"_a, "realtime_ratio"_a)
+        .def(pybind11::init<>(
+            [](const poisson_schedule_shim& sched, double ratio){
+                return arb::benchmark_cell{sched.schedule(), ratio};}),
+            "schedule"_a, "realtime_ratio"_a)
+        .def("__repr__", [](const arb::spike_source_cell&){return "<benchmark_cell>";})
+        .def("__str__",  [](const arb::spike_source_cell&){return "<benchmark_cell>";});
```
akuesters: - Add descriptions in constructors. 
- Correct `arb::benchmark_cell` in `__str__` and `__repr__` and add `arbor: ` in return string
- Do we want/need a default for the realtime ratio?

python/cells.cpp
```diff
@@ -0,0 +1,226 @@
...
+        .def(pybind11::init<>())
+        .def_readwrite("tau_m", &arb::lif_cell::tau_m,  "Membrane potential decaying constant [ms].")
+        .def_readwrite("V_th",  &arb::lif_cell::V_th,   "Firing threshold [mV].")
+        .def_readwrite("C_m",   &arb::lif_cell::C_m,    "Membrane capacitance [pF].")
+        .def_readwrite("E_L",   &arb::lif_cell::E_L,    "Resting potential [mV].")
+        .def_readwrite("V_m",   &arb::lif_cell::V_m,    "Initial value of the Membrane potential [mV].")
+        .def_readwrite("t_ref", &arb::lif_cell::t_ref,  "Refractory period [ms].")
+        .def_readwrite("V_reset", &arb::lif_cell::V_reset, "Reset potential [mV].")
+        .def("__repr__", [](const arb::lif_cell&){return "<lif_cell>";})
+        .def("__str__",  [](const arb::lif_cell&){return "<lif_cell>";});
```
akuesters: - Add description in constructor. 
- Add `arbor: ` in return string for `__str__` and `__repr__`
- Do we want/need defaults for the parameters?

python/cells.cpp
```diff
@@ -0,0 +1,226 @@
...
+    pybind11::class_<cell_parameters> cell_params(m, "cell_parameters", "Parameters used to generate the random cell morphologies.");
+    cell_params
+        .def(pybind11::init<>())
+        .def_readwrite("depth", &cell_parameters::max_depth,"The maximum depth of the branch structure.")
+        .def_readwrite("lengths",   &cell_parameters::lengths,  "Length of branch in μm [range].")
+        .def_readwrite("synapses",  &cell_parameters::synapses, "The number of randomly generated synapses on the cell.")
+        .def_readwrite("branch_probs", &cell_parameters::branch_probs, "Probability of a branch occuring [range].")
+        .def_readwrite("compartments", &cell_parameters::compartments, "Compartment count on a branch [range].")
+        .def("__repr__", util::to_string<cell_parameters>)
+        .def("__str__",  util::to_string<cell_parameters>);
```
akuesters: - Do we want/need defaults for the parameters?

python/cells.cpp
```diff
@@ -0,0 +1,226 @@
...
+    // Cable cells are going to be replaced with a saner API, so we don't go
+    // adding much in the way of interface. Instead we just provide a helper
+    // that will generate random cell morphologies for benchmarking.
+    pybind11::class_<arb::cable_cell> cablecell(m, "cable_cell");
+    m.def("branch_cell", &branch_cell,
+        "Construct a branching cell with a random morphology and synapse end points locations described by params.\n"
+        "gid is an integral value used to seed the random number generator. The unique global identifier of a cell is a good value to use.",
+        "gid"_a,
+        "params"_a=cell_parameters());
+}
```
akuesters: - Shouldn't we name the type `cable_cell` instead of `cablecell`?

python/cells.cpp
bcumming: :+1:

python/cells.cpp
bcumming: 1 seems like a reasonable default for the `realtime_ratio`.

python/cells.cpp
bcumming: Do we need comments for the default constructor, which does not do anything noteworthy.

It uses the defaults for the underlying C++ `lif_cell`.
I have updated the `__str__` and `__repr__` to print out the parameter values:
```python
import arbor
print(arbor.lif_cell()) 
```
generates:
```
<arbor.lif_cell: tau_m 10, V_th 10, C_m 20, E_L 0, V_m 0, t_ref 2, V_reset 0>
```

python/cells.cpp
akuesters: 👍

python/cells.cpp
akuesters: I would add a comment to the default constructor (to keep consistency).

python/cells.cpp
akuesters: Here are some whitespaces misplaced.

bcumming: It uses the defaults for the underlying C++ type.

python/cells.cpp
bcumming: What would the message be? The user can see that there is a default constructor by calling `help(arbor.lif_cell)`. Printing a message like `"Construct a lif cell"` is quite redundant to me.

Are we being consistent with other default constructors in the library?

python/cells.cpp
bcumming: yes, we should.

python/cells.cpp
akuesters: Sorry, I was not clear on this: I meant again a description for the constructor with the defaults as in `(default X)`.

python/cells.cpp
bcumming: Though this makes no difference: `cablecell` is just the name of the local variable used in this C++ function: the name in Python is `cable_cell`:
```C++
pybind11::class_<arb::cable_cell> cablecell(m, "cable_cell"); 
                                                                                       ^ name in Python
```

python/cells.cpp
akuesters: So far, I always added a description. We can add this check to issue #782 Python wrapper: uniform naming and printing of types

python/cells.cpp
akuesters: It shouldn't, just for consistency again :)

python/cells.cpp
bcumming: This is a different issue to the printing ticket IMO (it is a documentation issue).

Some types in the wrapper at the moment have a help message for the default constructor, but many don't.

Below is an example of the `help` output for `explicit_schedule`. In this case it is reasonable, to disambiguate from the other constructor, and because it is worth mentioning that the default is an empty schedule.

```
 |  __init__(...)
 |      __init__(*args, **kwargs)
 |      Overloaded function.
 |      
 |      1. __init__(self: arbor.explicit_schedule) -> None
 |      
 |      Construct an empty explicit schedule.
 |      
 |      
 |      2. __init__(self: arbor.explicit_schedule, times: List[float]) -> None
 |      
 |      Construct an explicit schedule with argument:
 |        times: A list of times (in ms, default []).
```

akuesters: Ok, sure. We can only add a description if worth mentioning.

python/cells.cpp
bcumming: If we use the defaults provided by the underlying C++ type, then we have to be careful documenting those values in the Python wrapper, because if that value is changed in the C++, it probably won't get updated in the Python doc strings. We should avoid using default values in the docstrings when the value is easy to check and set using accessors.

akuesters: True.

akuesters: We might want to include the attributes into the con`tor in case the C++ default values change (and users rely on them in their scripts).

Issue #780, opened 2019-06-07
Support valence reading, verification


halfflat:
We currently parse but ignore the `USEION VALENCE` parameter. We also need a way to pass _in_ the charge associated with an ion species for the purpose of writing generic reversal potential calculation mechanisms (i.e. Nernst implementations.)

The proposal is to extend the `VALENCE` syntax so it can take an identifier as its argument. If present, this will be treated as an external global scalar, with a value taken from the global ion settings.

When `VALENCE` is given with a (signed) integer argument, we should confirm that the value matches when we start discretizing the model.

halfflat:
Implemented in #781.

PR #781, opened 2019-06-07
Extend and implement extended VALENCE semantics

halfflat:
Extended USEION VALENCE support.

* Augment modcc parser to accept identifier or signed integer after VALENCE.
* Treat identifer after VALENCE as global scalar, tied to ionic charge.
* Add ionic charge field to mechanism ion parameter pointers, replacing scalar in shared state with a length-1 array.
* Verify presence of ion name and check ion charge against numeric valence in mechanism info when discretizing in `fvm_build_mechanism_data`.
* Pass ion info, coalesce_synapse global state to `fvm_build_mechanism_data` just via cable cell global properties.
* Add more unit-test only mechanisms (test_ca_read_valence, test_cl_valence).
* Add unit tests for valence verification and reading into mechanism state (multicore only).
* Update existing unit tests for new interfaces.

Implements #780.


arbor/backends/multicore/mechanism.cpp
```diff
@@ -95,6 +95,7 @@ void mechanism::instantiate(unsigned id, backend::shared_state& shared, const la
         ion_view.reversal_potential = oion->eX_.data();
         ion_view.internal_concentration = oion->Xi_.data();
         ion_view.external_concentration = oion->Xo_.data();
+        ion_view.ionic_charge = oion->Xo_.data();
```
noraabiakar: I guess this should be ```oion->charge.data()```?

arbor/backends/gpu/shared_state.hpp
```diff
@@ -34,7 +34,7 @@ struct ion_state {
     array weight_Xi_;   // (1) concentration weight internal
     array weight_Xo_;   // (1) concentration weight external
 
-    int charge;    // charge of ionic species
+    array charge;       // charge of ionic species (global, length 1)
```
noraabiakar: Why do we need this to be an array?

arbor/backends/multicore/mechanism.cpp
halfflat: It should, thank you!

arbor/backends/gpu/shared_state.hpp
halfflat: Mainly because `array` handles automatically all the GPU memory management for us. A pointer would suffice, but would complicate the allocation and destruction somewhat.

PR #786, opened 2019-06-12
Implement mechanism ion rebinding.

halfflat:
* Make global parameters and ion rebindings part of the instantiate interface, rather than insist that all concrete mechanisms implement these as methods.
* Mechanism catalogue instance() returns a pair, comprising the concrete mechanism for the requested backend, together with the override data.
* Extend catalgoue derive() method to take a list of old ion name -> new ion name remappings for a mechanism.
* Add exceptions for ion remapping errors, and check for these errors.
* Add convenience function for reparameterizing a mechanism with a single ion dependency over other ions. (This will be used for the future nernst pre-supplied mechanism.)
* Add unit tests for: chained renamings of ion names across multiple derivations; correct shared state ion assignment after renamings; ion remapping exceptions; parameterize_over_ion.

### DIFF
--- a/arbor/mechcat.cpp
+++ b/arbor/mechcat.cpp
@@ -7,6 +7,7 @@
 #include <arbor/mechcat.hpp>
 
 #include "util/maputil.hpp"
+#include "io/trace.hpp"
 
 namespace arb {
 
@@ -45,7 +46,10 @@ const mechanism_fingerprint& mechanism_catalogue::fingerprint(const std::string&
     throw no_such_mechanism(name);
 }
 
-void mechanism_catalogue::derive(const std::string& name, const std::string& parent, const std::vector<std::pair<std::string, double>>& global_params) {
+void mechanism_catalogue::derive(const std::string& name, const std::string& parent,
+    const std::vector<std::pair<std::string, double>>& global_params,
+    const std::vector<std::pair<std::string, std::string>>& ion_remap_vec)
+{
     if (has(name)) {
         throw duplicate_mechanism(name);
     }
@@ -54,7 +58,8 @@ void mechanism_catalogue::derive(const std::string& name, const std::string& par
         throw no_such_mechanism(parent);
     }
 
-    derivation deriv = {parent, {}, nullptr};
+    std::unordered_map<std::string, std::string> ion_remap_map(ion_remap_vec.begin(), ion_remap_vec.end());
+    derivation deriv = {parent, {}, ion_remap_map, nullptr};
     mechanism_info_ptr info = mechanism_info_ptr(new mechanism_info((*this)[deriv.parent]));
 
     for (const auto& kv: global_params) {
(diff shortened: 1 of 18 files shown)


arbor/mechcat.cpp
```diff
@@ -7,6 +7,7 @@
 #include <arbor/mechcat.hpp>
 
 #include "util/maputil.hpp"
+#include "io/trace.hpp"
```
bcumming: Remove?

arbor/mechcat.cpp
```diff
@@ -74,6 +79,33 @@ void mechanism_catalogue::derive(const std::string& name, const std::string& par
...
     }
 
+    for (const auto& kv: ion_remap_vec) {
+        if (!info->ions.count(kv.first)) {
+            throw invalid_ion_remap(name, kv.first, kv.second);
+        }
+    }
+
+    std::unordered_map<std::string, ion_dependency> new_ions;
+    for (const auto& kv: info->ions) {
```
bcumming: Loops like this would benefit with some comments that explain what they aim to acheive. It isn't immediately obvious, particularly if the variables (e.g. `ion_remap_map` are not documented either).

arbor/mechcat.cpp
```diff
@@ -131,18 +166,34 @@ std::unique_ptr<mechanism> mechanism_catalogue::instance_impl(std::type_index ti
         }
     }
 
-    std::unique_ptr<mechanism> mech = prototype->clone();
+    mech.first = prototype->clone();
 
-    auto apply_globals = [this](auto& self, const std::string& name, mechanism* mptr) -> void {
+    auto apply_globals = [this](auto& self, const std::string& name, mechanism_overrides& over) -> void {
```
bcumming: This lambda has overflowed its boundaries somewhat. Might be worth making it a member function.

arbor/mechcat.cpp
```diff
@@ -179,4 +230,14 @@ void mechanism_catalogue::copy_impl(const mechanism_catalogue& other) {
     }
 }
 
+void parameterize_over_ion(mechanism_catalogue& cat, const std::string& name, const std::string& ion) {
```
bcumming: A brief comment documenting the changes this function makes to the catalogue would help understanding.

test/simple_recipes.hpp
```diff
@@ -50,6 +50,10 @@ class simple_recipe_base: public recipe {
         return catalogue_;
     }
 
+    void add_ion(const char* name, int charge, double iconc, double econc) {
+        cell_gprop_.ion_default[name] = {charge, iconc, econc};
```
bcumming: Here add an ions species and set its charge and concentrations. How do we set the default reversal potential?

arbor/mechcat.cpp
halfflat: Will do.

arbor/mechcat.cpp
halfflat: While `ion_remap_map` I thought was pretty clear from it's declaration and initialization, I'll add a purpose-of-this-code comment to this for loop.

arbor/mechcat.cpp
halfflat: Doing so just feels like I'm splitting a method in two; will do it if you insist, but I don't think it's going to be a big gain in clarity.

arbor/mechcat.cpp
halfflat: The comment in the header file is not enough? I'd prefer to expand that comment rather than add another here, if it is required.

test/simple_recipes.hpp
halfflat: While this is just a helper for the simple recipe class, in the current version of the cable cell global properties, there is not a mechanism for providing this in the ion table. This facility though is part of the big pending reversal potential PR.

arbor/mechcat.cpp
bcumming: I see now that this was done to enable recursion.
A little bit of documentation that explains the algorithm or approach being taken here would make it much easier to understand.

halfflat: This could be replaced with e.g. a stack/vector and a for loop, if that would make more sense.

halfflat:
Righto; added some more commentary, and fixed the implementation error in `arb::optional<X&>::value_or()` that was breaking the unit tests.

PR #788, opened 2019-06-13
Python spikes

### LINKED ISSUE #764: Python wrapper: spikes
Wrapper for
* `arb::spike` type
* helper for recording spike trains
* example of recording and printing/plotting spikes

bcumming:
Support for recording spikes generated by a simulation in the Python wrapper
* Implement a `spike_recorder` that holds a shared pointer to a `std::vector` of spikes, and a callback for the `arb::simulation` spike recording API.
* Add `python/example/ring.py` that creates a ring network, then records and prints spikes.
* Some fixes to get the full `recipe` -> `domain_decomposition` -> `simulation` -> `spikes` workflow to work
  * always use default `global_parameters`: user customization of global parameters for cable cells can wait until the ion species interface is finished.
  * change the Python recipe interface for `recipe::connections_on` to use `pybind11::objects` because of shim.
* Some small improvements to error and help messages.

Fixes #764

### DIFF
--- a/python/recipe.cpp
+++ b/python/recipe.cpp
@@ -43,10 +43,9 @@ std::vector<arb::event_generator> py_recipe_shim::event_generators(arb::cell_gid
     for (auto& g: pygens) {
         // check that a valid Python event_generator was passed.
         if (!isinstance<pyarb::event_generator_shim>(g)) {
-            std::stringstream s;
-            s << "recipe supplied an invalid event generator for gid "
-              << gid << ": " << pybind11::str(g);
-            throw pyarb_error(s.str());
+            throw pyarb_error(
+                util::pprintf(
+                    "recipe supplied an invalid event generator for gid {}: {}", gid, pybind11::str(g)));
         }
         // get a reference to the python event_generator
         auto& p = cast<const pyarb::event_generator_shim&>(g);
--- a/python/recipe.cpp
+++ b/python/recipe.cpp
@@ -22,41 +22,8 @@ namespace pyarb {
 // The py::recipe::cell_decription returns a pybind11::object, that is
 // unwrapped and copied into a arb::util::unique_any.
 arb::util::unique_any py_recipe_shim::get_cell_description(arb::cell_gid_type gid) const {
-    // Aquire the GIL because it must be held when calling isinstance and cast.
-    auto guard = pybind11::gil_scoped_acquire();
-
-    // Get the python object pyarb::cell_description from the python front end
     return convert_cell(impl_->cell_description(gid));
 }
-
-// The py::recipe::global_properties returns a pybind11::object, that is
-// unwrapped and copied into a arb::util::any.
-/*
-arb::util::any py_recipe_shim::get_global_properties(arb::cell_kind kind) const {
-    // TODO: handle none case
-    using pybind11::cast;
-
-    // Aquire the GIL because it must be held when calling cast.
-    auto guard = pybind11::gil_scoped_acquire();
-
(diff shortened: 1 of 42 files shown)


python/recipe.cpp
```diff
@@ -112,6 +118,32 @@ struct cell_connection_shim {
...
+    auto pycons = impl_->connections_on(gid);
+    std::vector<arb::cell_connection> cons;
+    cons.reserve(pycons.size());
+    for (unsigned i=0; i<pycons.size(); ++i) {
+        const auto& c = pycons[i];
+        if (isinstance<cell_connection_shim>(c)) {
+            cons.push_back(cast<cell_connection_shim>(c));
+        }
+        else {
+            throw std::runtime_error(
```
akuesters: Why not use `pyarb_error` as in `event_generators`?

python/recipe.cpp
```diff
@@ -133,7 +165,7 @@ void register_recipe(pybind11::module& m) {
         "  Defined by source and destination end points (that is pre-synaptic and post-synaptic respectively), a connection weight and a delay time.");
     cell_connection
         .def(pybind11::init<arb::cell_member_type, arb::cell_member_type, float, arb::time_type>(),
-            "source"_a = arb::cell_member_type{0,0}, "dest"_a = arb::cell_member_type{0,0}, "weight"_a = 0.f, "delay"_a,
+            "source"_a, "dest"_a, "weight"_a, "delay"_a,
             "Construct a connection with arguments:\n"
             "  source:      The source end point of the connection (default (0,0)).\n"
```
akuesters: We should also delete the defaults in descriptions and make unit description consistent (e.g. in lif_cell only [] are used, here unit: ).

python/simulation.cpp
```diff
@@ -33,9 +33,9 @@ void register_simulation(pybind11::module& m) {
         .def("run", &arb::simulation::run,
             pybind11::call_guard<pybind11::gil_scoped_release>(),
             "Run the simulation from current simulation time to tfinal, with maximum time step size dt.",
-            "tfinal"_a, "dt"_a)
+            "tfinal"_a, "dt"_a=0.025)
```
akuesters: Maybe add default and units (of dt and tfinal) to description?

python/recipe.hpp
```diff
@@ -39,17 +39,16 @@ class py_recipe {
...
-    virtual std::vector<pybind11::object> event_generators(arb::cell_gid_type gid) const {
-        auto guard = pybind11::gil_scoped_acquire();
-        return {};
-    }
+    virtual std::vector<pybind11::object> event_generators(arb::cell_gid_type gid) const { return {}; }
 
-    virtual std::vector<arb::cell_connection> connections_on(arb::cell_gid_type gid) const { return {}; }
+    //virtual std::vector<arb::cell_connection> connections_on(arb::cell_gid_type gid) const { return {}; }
+    virtual std::vector<pybind11::object> connections_on(arb::cell_gid_type gid) const { return {}; }
     virtual std::vector<arb::gap_junction_connection> gap_junctions_on(arb::cell_gid_type) const { return {}; }
```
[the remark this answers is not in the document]
bcumming: I have removed the connection shim, to simplify conversion from Python to C++ code.

python/recipe.cpp
bcumming: Fixed.

python/simulation.cpp
bcumming: I have added the unit here, however these further changes are beyond the scope of this PR.

Issue #789, opened 2019-06-18
Optional non-thread-safe recipes


bcumming:
The consumers of `arb::recipe` require that calls to some recipe methods (e.g.  `get_cell_description()`) are thread-safe, that is, multiple threads can call the method simultaneously.

This is a proposal to extend the `recipe` interface to allow for non-thread-safe recipes. It is motivated by work on the Python wrappers, which allow users to implement recipe methods in Python. These call backs can't be thread-safe because only one thread can hold the GIL at a time. Managing the GIL between multiple C++ threads that call into Python, from a call that originated in Python, is complicated and fragile.

There will likely be other use cases where ensuring thread safety is a large burden on users, and this option could also be used to debug potential threading bugs.

I propose extending the `arb::recipe` interface with a `thread_safe` method that defaults to `true`, and can be overridden by non-thread-safe derivations:

```C++
class recipe {    
  // ...
  virtual bool thread_safe() const {
      return true;
  }
};
```

bcumming:
If we are not willing to do this, the Python recipe could have a mutex that a thread must manually acquire on each callback. However, this would still allow threads other than the main Python thread to execute callbacks, which is a potential minefield.

halfflat:
If the mutex-in-recipe doesn't fix the Python problem, then I think the proposed extension to the recipe interface is a good one, if it's clear that it means that the recipe can only be safely accessed from the same thread that invoked the corresponding arbor library function — we can bicker about the name!

bcumming:
I found a fix to the Python problem, by means of a mutex and some additional state for error checking that allows the recipe to be called by multiple threads.
Since we still don't have a use case for this, let's put it on the back burner.

PR #790, opened 2019-06-19
move implementation of communicator to separate TU

bcumming:


arbor/communication/communicator.cpp
```diff
@@ -0,0 +1,229 @@
...
+    }
+    // Build the connection information for local cells in parallel.
+    std::vector<gid_info> gid_infos;
+    gid_infos.resize(num_local_cells_);
+    for (unsigned i=0; i<gids.size(); ++i) {
+        auto gid = gids[i];
+        gid_infos[i] = gid_info(gid, i, rec.connections_on(gid));
+    }
+    /*
+    threading::parallel_for::apply(0, gids.size(), thread_pool_.get(),
```
noraabiakar: Why remove the threaded parallel_for implementation here? Is it related to issue #789?

bcumming: Oh, this shouldn't have been removed. Good spot!

PR #796, opened 2019-06-25
Python wrapper: profiler

### LINKED ISSUE #765: Python wrapper: profiler
Because we like to time it.

akuesters:
Wraps the `meter_manager` with
- the constructor
- `start(ctx)`
- `checkpoint(name, ctx)`
- `checkpoint_names`
- `times`

and the `meter_report` with
- `make_meter_report(meter_manager, ctx)`

Fixes #765


python/example/ring.py
```diff
@@ -48,18 +48,32 @@ def event_generators(self, gid):
 context = arbor.context(threads=4, gpu_id=None)
 print(context)
 
+meters = arbor.meter_manager()
+print(meters)
```
bcumming: The meters don't need to be printed here because they are currently empty.

python/example/ring.py
```diff
@@ -48,18 +48,32 @@ def event_generators(self, gid):
...
+
+meters.start(context)
+
 recipe = ring_recipe(100)
 print(recipe)
 
+meters.checkpoint('recipe create', context)
```
bcumming: It is best to avoid using spaces in checkpoint names, so that the text output is easier to parse (e.g. using grep or awk).
For example use `recipe-create`, `load-balance`, `simulation-init`, etc.

python/profiler.cpp
```diff
@@ -0,0 +1,59 @@
...
+             and restarts the timer for the next checkpoint.")
+        .def_property_readonly("checkpoint_names", &arb::profile::meter_manager::checkpoint_names,
+            "A list of all metering checkpoint names.")
+        .def_property_readonly("times", &arb::profile::meter_manager::times,
+            "A list of all metering times.")
+        .def("__str__",  [](const arb::profile::meter_manager&){return "<arbor.meter_manager>";})
+        .def("__repr__", [](const arb::profile::meter_manager&){return "<arbor.meter_manager>";});
+
+    // meter report
+    pybind11::class_<arb::profile::meter_report> meter_report(m, "meter_report", "Gather distributed meter information.");
```
bcumming: The summary string doesn't describe what the report is for, how about:                                   
```
Summarises the performance meter results, used to print a report to screen or file.
If a distributed context is used, the report will contain a summary of results from all MPI ranks.
```

For printing, the `__str__`` representation we could remove `"<arbor.meter_report>"` from the start
to make the output cleaner. I would make the `__repr__` just indicate that this is a meter report, but not give all of the details:
```C++
        .def("__str__",  [](arb::profile::meter_report& r){return pprintf("{}", r);})
        .def("__repr__", [](arb::profile::meter_report& r){return "<arbor.meter_report>";});
```

python/example/ring.py
akuesters: Done.

python/example/ring.py
akuesters: Done.

python/profiler.cpp
akuesters: Changed summary string and `__str__`/`__repr__` strings. Also moved `make_meter_report` to `meter_report` constructor.

python/example/ring.py
akuesters: Also changed prints to python3 format printing standard.

Issue #793, opened 2019-06-21
Ionic currents in NMODL are presented with inconsistent units


halfflat:
The Arbor internal representation of a current density is in A/m², while for NEURON modules, the value is in mA/cm². This difference is handled correctly when currents are accumulated, as density mechanisms have their weight contribution scaled by 10.

When a mechanism reads an ionic current, however, they are presented with the value that Arbor stores internally, which is implicitly in Arbor's units, not NEURONs. When these currents are used to compute e.g. concentration updates in a mod file, the rates computed by that mod file will be out by a factor of 10.

Possible fixes:
* We change the internal representation of currents to mA/cm² — this is a right pain, as it requires the correct conversion factor to be inserted in all the different voltage computation matrix implementations, not to mention all the documentation.
* `modcc` adds a 0.1 scaling term to reads of external ion current variables, to match NEURON units. The downside to this is that it is asymmetric: the generated code will read values in A/m² but write them in mA/cm².
* All ionic current values are scaled by 0.1 in the back-end before any mechanism state updates are called. Mechanisms then see consistent (NEURON) units, but it's another touch-all-the-data processing step inside the integration loop.
* We make mechanisms see consistent Arbor units (i.e. A/m²), and have the generated code from `modcc` scale by 0.1 on read and by 10 on update. This adds essentially another multiplication to each mechanism kernel described by NMODL, but has the benefit of consistent units across both the internal representation and the (implicit) mechanism API.
* We throw out readable ionic currents and make everyone use species fluxes instead. This allows people to deal with flows of uncharged species! But also is much more invasive, with almost every mod file needing to be adjusted. We'd also, for good or ill, be baking in values for the Faraday constant; this though, at least, could be specified as a run-time parameter and applied consistently across mechanisms, avoiding the situation where different mechanisms may have differing values for this constant.
* Or  as a variation on the above, we: make the internal state and mechanism interface deal in fluxes and currents; make `modcc` convert fluxes to and from currents automatically in generated code, using a fixed value for the Faraday constant.

halfflat:
Happy to debate this, but I'm going to push the option that mechanisms see Arbor units (i.e A/m²) , and NMODL sees NEURON units (mA/cm²), with modcc adding the conversions when generating mechanism code.

bcumming:
Fixed in #800

Issue #798, opened 2019-06-25
NMODL mechanism-initialized ionic currents are not weighted


halfflat:
Demo with unit test: https://github.com/halfflat/arbor/tree/bugdemo/ionic-current-init

Initialization of ionic currents iX for an ion X do not get multiplied by the mechanism weights. This gives a factor of 10 error in scalar and GPU implementations, and a SIMD-width x 10 fold error in vector implementations.

halfflat:
Oops, forgot to commit before push *ahem*. Now the link should be correct.

bcumming:
#800

PR #800, opened 2019-06-26
Rejig weight and conversions for modcc

halfflat:
Address inconsistencies in unit conversions between current updates, initialization, ionic currents etc.

Fixes issues #793 and #798.

* Correct units in comments for ion state iX_ to a current density.
* Remove conversion factor for NMODL current densities [mA/cm²] to Arbor current densities [A/m²] from layout weight: from the point of view of Arbor, mechanisms now write in units that Arbor uses natively.
* Remove weight_ as a magic modcc variable; applying weight_ is now performed directly by code generated by printers.
* Change IndexedVariable interface: remove redundant index_name; remove 'op'. Whether external data is accumulated or assigned is now a property of the corresponding sourceKind.
* Conversion factors for NMODL <-> Arbor are collated in printerutils.cpp, determined by the external data sourceKind.
* Permit direct assignment of external data, as opposed to a weighted sum. This behaviour will be required for reversal potential handling.
* `current_` and `conducitivity_` get a different sourceKind in point mechanisms, allowing us to distinguish current contributions from current density contributions (and applying correct unit scaling accordingly.)
* Add unit test for correct initialization and scaling of ionic currents, together with two corresponding unit test mechanisms.


halfflat:
I'll rebase against master, and update with correct handling of ion current writing by point processes.

PR #821, opened 2019-07-19
Removed the use of CUDA Managed Memory from GPU-Stack

kanduri:
Tried to retain the API of the `arb::gpu::stack` class while removing the usage of `managed_ptr`. Now it manually refreshes the host copy of the data structure.

All the 622 unit tests seem to pass.

### DIFF
--- a/arbor/backends/gpu/stack.hpp
+++ b/arbor/backends/gpu/stack.hpp
@@ -49,18 +49,12 @@ private:
     std::vector<T> data_;
 
     void create_storage(unsigned n) {
-        storage_type p;
-        p.capacity = n;
-        p.stores = 0u;
-        p.data = n>0u ? allocator<value_type>().allocate(n): nullptr;
+       host_storage_.capacity = n;
+       host_storage_.stores = 0u;
+       host_storage_.data = n>0u ? allocator<value_type>().allocate(n): nullptr;
 
-        auto p_gpu = allocator<storage_type>().allocate(1);
-
-        memory::cuda_memcpy_h2d(p_gpu, &p, sizeof(storage_type));
-
-        host_storage_ = p;
-        
-        device_storage_ = p_gpu;
+       device_storage_ = allocator<storage_type>().allocate(1);
+       memory::cuda_memcpy_h2d(device_storage_, &host_storage_, sizeof(storage_type));
     }
 
     void update_data() {
@@ -123,6 +117,7 @@ public:
     void clear() {
         host_storage_.stores = 0u;
         memory::cuda_memcpy_h2d(device_storage_, &host_storage_, sizeof(storage_type));
+        data_.clear();
     }
 
     storage_type get_storage_copy() const {
--- a/arbor/backends/gpu/stack.hpp
+++ b/arbor/backends/gpu/stack.hpp
@@ -43,7 +43,7 @@ private:
     storage_type host_copy_;
     gpu_context_handle gpu_context_;
 
-    storage_type* create_storage(unsigned n) {
+    void create_storage(unsigned n) {
         storage_type p;
         p.capacity = n;
         p.stores = 0u;
@@ -55,7 +55,7 @@ private:
 
         host_copy_ = p;
         
-        return p_gpu;
+        device_storage_ = p_gpu;
     }
 
 public:
@@ -65,10 +65,12 @@ public:
     stack& operator=(const stack& other) = delete;
     stack(const stack& other) = delete;
 
-    stack(const gpu_context_handle& gpu_ctx):
-        device_storage_(create_storage(0)), gpu_context_(gpu_ctx) {}
+    stack(const gpu_context_handle& gpu_ctx): gpu_context_(gpu_ctx) {
+            create_storage(0);
+    }
 
-    stack(stack&& other): device_storage_(create_storage(0)), gpu_context_(other.gpu_context_) {
+    stack(stack&& other): gpu_context_(other.gpu_context_) {
+        create_storage(0);
         std::swap(device_storage_, other.device_storage_);
     }
 
@@ -77,8 +79,9 @@ public:
         return *this;
     }
 
-    explicit stack(unsigned capacity, const gpu_context_handle& gpu_ctx):
-        device_storage_(create_storage(capacity)), gpu_context_(gpu_ctx) {}
+    explicit stack(unsigned capacity, const gpu_context_handle& gpu_ctx): gpu_context_(gpu_ctx) {
+        create_storage(capacity);
+    }
 
     ~stack() {
         update_host_storage();
@@ -86,6 +89,8 @@ public:
         if (st.data) {
             allocator<value_type>().deallocate(st.data, st.capacity);
         }
+        allocator<storage_type>().deallocate(device_storage_, sizeof(storage_type));
+
     }
 
     std::vector<T> get_data() const {
--- a/arbor/backends/gpu/stack.hpp
+++ b/arbor/backends/gpu/stack.hpp
@@ -91,10 +91,12 @@ public:
     stack(stack&& other): gpu_context_(other.gpu_context_) {
         create_storage(0);
         std::swap(device_storage_, other.device_storage_);
+        update_host();
     }
 
     stack& operator=(stack&& other) {
         std::swap(device_storage_, other.device_storage_);
+        update_host();
         return *this;
     }
 
@@ -104,9 +106,8 @@ public:
 
     ~stack() {
         update_host_storage();
-        auto st = get_storage_copy();
-        if (st.data) {
-            allocator<value_type>().deallocate(st.data, st.capacity);
+        if (host_storage_.data) {
+            allocator<value_type>().deallocate(host_storage_.data, host_storage_.capacity);
         }
         allocator<storage_type>().deallocate(device_storage_, sizeof(storage_type));
 
--- a/arbor/backends/gpu/stack.hpp
+++ b/arbor/backends/gpu/stack.hpp
@@ -65,7 +65,10 @@ public:
     stack(const stack& other) = delete;
     stack() = delete;
 
-    stack(gpu_context_handle h): gpu_context_(h) {}
+    stack(gpu_context_handle h): gpu_context_(h) {
+        host_storage_.data = nullptr;
+        device_storage_ = nullptr;
+    }
 
     stack& operator=(stack&& other) {
         gpu_context_ = other.gpu_context_;
--- a/arbor/backends/gpu/stack.hpp
+++ b/arbor/backends/gpu/stack.hpp
@@ -81,6 +81,7 @@ public:
         device_storage_(create_storage(capacity)), gpu_context_(gpu_ctx) {}
 
     ~stack() {
+        update_host_storage();
         auto st = get_storage_copy();
         if (st.data) {
             allocator<value_type>().deallocate(st.data, st.capacity);
@@ -107,9 +108,12 @@ public:
         return host_copy_;
     }
 
+    void update_host_storage() {
+	    memory::cuda_memcpy_d2h(&host_copy_, device_storage_, sizeof(storage_type));
+    }
 
     void update_host() {
-        host_copy_ = get_storage_copy();
+        update_host_storage();
         data_ = get_data();
     }
 
--- a/arbor/backends/gpu/stack.hpp
+++ b/arbor/backends/gpu/stack.hpp
@@ -49,49 +49,34 @@ private:
     std::vector<T> data_;
 
     void create_storage(unsigned n) {
-       host_storage_.capacity = n;
-       host_storage_.stores = 0u;
-       host_storage_.data = n>0u ? allocator<value_type>().allocate(n): nullptr;
+        data_.reserve(n);
 
-       device_storage_ = allocator<storage_type>().allocate(1);
-       memory::cuda_memcpy_h2d(device_storage_, &host_storage_, sizeof(storage_type));
-    }
+        host_storage_.capacity = n;
+        host_storage_.stores = 0u;
+        host_storage_.data = n>0u ? allocator<value_type>().allocate(n): nullptr;
 
-    void update_data() {
-        if (capacity() != 0u) {
-
-            auto num = size();
-            std::vector<T> buf(num);
-            auto bytes = num*sizeof(T);
-            memory::cuda_memcpy_d2h(buf.data(), host_storage_.data, bytes);
-            
-            data_ = buf;
-        }
-    }
-
-    void update_host_storage() {
-	    memory::cuda_memcpy_d2h(&host_storage_, device_storage_, sizeof(storage_type));
+        device_storage_ = allocator<storage_type>().allocate(1);
+        memory::cuda_memcpy_h2d(device_storage_, &host_storage_, sizeof(storage_type));
     }
 
 public:
 
     stack& operator=(const stack& other) = delete;
     stack(const stack& other) = delete;
+    stack() = delete;
 
-    stack(const gpu_context_handle& gpu_ctx): gpu_context_(gpu_ctx) {
-            create_storage(0);
-    }
-
-    stack(stack&& other): gpu_context_(other.gpu_context_) {
-        create_storage(0);
-        std::swap(device_storage_, other.device_storage_);
-        update_host();
-    }
+    stack(gpu_context_handle h): gpu_context_(h) {}
 
     stack& operator=(stack&& other) {
+        gpu_context_ = other.gpu_context_;
         std::swap(device_storage_, other.device_storage_);
-        update_host();
+        std::swap(host_storage_, other.host_storage_);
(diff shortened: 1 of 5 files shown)


arbor/backends/gpu/stack.hpp
```diff
@@ -29,25 +29,39 @@ namespace gpu {
...
+    storage_type host_copy_;
     gpu_context_handle gpu_context_;
 
-    managed_ptr<storage_type> create_storage(unsigned n) {
-        auto p = make_managed_ptr<storage_type>();
-        p->capacity = n;
-        p->stores = 0;
-        p->data = n? allocator<value_type>().allocate(n): nullptr;
-        return p;
+    storage_type* create_storage(unsigned n) {
```
bcumming: This isn't functional after the changes, i.e. it returns a pointer, and mutates state at the same time, so it doesn't make sense to keep it.
Do this inside the constructor directly, and use the `update_device` member to perform the copy to device after initializing `host_storage_` instead of manually performing the memcpy.

arbor/memory/cuda_wrappers.cpp
```diff
@@ -71,6 +71,10 @@ void cuda_free(void* ptr) {
     }
 }
 
+void cuda_sync() {
```
bcumming: This isn't needed (see comment where it is called)

arbor/backends/gpu/stack.hpp
```diff
@@ -67,41 +81,67 @@ class stack {
         storage_(create_storage(capacity)), gpu_context_(gpu_ctx) {}
 
     ~stack() {
-        storage_.synchronize();
-        if (storage_->data) {
-            allocator<value_type>().deallocate(storage_->data, storage_->capacity);
+        memory::cuda_sync();
```
bcumming: this synchronization is not needed (why perform it here, and not at other times before performing copy from device to host?)

arbor/backends/gpu/stack.hpp
```diff
@@ -67,41 +81,67 @@ class stack {
...
+        auto st = get_storage_copy();
```
bcumming: We don't need to update the host copy of the data to do the deallocation (we have the pointer already on the host side).

arbor/backends/gpu/stack.hpp
```diff
@@ -67,41 +81,67 @@ class stack {
...
+        if (st.data) {
+            allocator<value_type>().deallocate(st.data, st.capacity);
```
bcumming: Don't forget to also deallocate `device_storage_`

arbor/backends/gpu/stack.hpp
```diff
@@ -67,41 +81,67 @@ class stack {
...
+
+    storage_type get_storage_copy() const {
+        storage_type st;
+        memory::cuda_memcpy_d2h(&st, storage_, sizeof(storage_type));
+
+        return st;
+    }
+
+
+    void refresh_host_copy() {
```
bcumming: You can replace `refresh_host_copy` and `get_storage_copy` with a single call `update_host`.

arbor/backends/gpu/stack.hpp
```diff
@@ -67,41 +81,67 @@ class stack {
...
+        if (st.data) {
+            allocator<value_type>().deallocate(st.data, st.capacity);
         }
     }
 
-    // Perform any required synchronization if concurrent host-side access is not supported.
-    // (Correctness still requires that GPU operations on this stack are complete.)
-    void host_access() const {
-        gpu_context_->synchronize_for_managed_access();
+    std::vector<T> get_data() const {
```
bcumming: If `data_` is updated by `update_host`, then this can return a `const std::vector<T>&`

arbor/backends/gpu/stack.hpp
```diff
@@ -67,41 +81,67 @@ class stack {
...
+
+    void refresh_host_copy() {
+        host_copy_ = get_storage_copy();
+        data_ = get_data();
     }
 
     void clear() {
-        storage_->stores = 0u;
+        host_copy_.stores = 0u;
+        memory::cuda_memcpy_h2d(storage_, &host_copy_, sizeof(storage_type));
```
bcumming: Don't perform the copy here: instead make it only adjust the host values, then the user can manually call an `update_device` member when they want to synchronize.

arbor/backends/gpu/stack.hpp
```diff
@@ -110,29 +150,32 @@ class stack {
 
     value_type& operator[](unsigned i) {
```
bcumming: Maybe we can deprecate all of these `operator[]` and `begin, end` if we are now providing a function that returns a `const std::vector<T>&` with the values.

arbor/backends/gpu/threshold_watcher.hpp
```diff
@@ -70,7 +70,7 @@ class threshold_watcher {
 
     /// Remove all stored crossings that were detected in previous calls to test()
     void clear_crossings() {
-        stack_.host_access();
+        stack_.refresh_host_copy();
```
bcumming: This call to `refresh_host_copy` wasn't necessary.
Now it should be 
```C++
stack_.clear();
stack_.update_device();
```

arbor/backends/gpu/threshold_watcher.hpp
```diff
@@ -90,7 +90,7 @@ class threshold_watcher {
     }
 
     const std::vector<threshold_crossing>& crossings() const {
-        stack_.host_access();
+        stack_.refresh_host_copy();
```
bcumming: `stack_.update_host();`

arbor/backends/gpu/threshold_watcher.hpp
```diff
@@ -141,7 +141,7 @@ class threshold_watcher {
     array v_prev_;              // Values at previous sample time.
 
     // Hybrid host/gpu data structure for accumulating threshold crossings.
-    stack_type stack_;
+    mutable stack_type stack_;
```
bcumming: Why is this mutable?

arbor/memory/cuda_wrappers.cpp
```diff
@@ -117,6 +121,10 @@ void cuda_free(void* ptr) {
     NOCUDA;
 }
 
+void cuda_sync() {
```
bcumming: Remove.

arbor/memory/cuda_wrappers.hpp
```diff
@@ -11,6 +11,7 @@ void cuda_host_unregister(void* ptr);
 void* cuda_malloc(std::size_t n);
 void* cuda_malloc_managed(std::size_t n);
 void cuda_free(void* ptr);
+void cuda_sync();
```
bcumming: Remove

halfflat:
Looking at the code, and at @bcumming's comments, I think it would be useful to clarify what operations are required and what their semantics are.

In terms of functionality, from the host side, I think we need:
* A reset or clear operation that resets the GPU stack state.
* A read operation, that brings the current stack state to the host side.

The former could simply be the current `clear()` function: I think requiring a separate push to GPU is unnecessary.

For the read operations, there are two major choices:
* Management of a host-side cached copy:
    * There is one read-only cache associated with the stack object.
    * Methods such as `data()`, `size()`, `begin()` and `end()` can interrogate the cached stack, or else the cache is made directly visible via a const reference returning `values()` (or whatever the name choice is). `overflow()` is required, too.
    * An explicit `update` method updates the cache to match the GPU state, _or_ implicit cache update via a synchronizing CUDA call upon access. There could of course be an efficiency cost with the latter, depending on the host-side use case. If implicit, the number of synchronizing access methods ought to be minimized, i.e. just `values()` and `overflow()`.
* Host-side data copy on request:
    * A member function `get()` returns the data and overflow state dynamically, and no host-side cache is maintained by the stack. It is then the user code's responsibility to manage that data.

The cached version has the benefit of easily avoiding dynamic memory allocation on state inquiry. Without the cache, a `get()` method which just returned a state comprising a pair of `std::vector` and `bool` would be dynamically allocating; it could alternatively take a reference to a host-side stack state object that could be updated in-place.

If we clarify these decisions first, then any implementation concerns can be easily rectified.

bcumming:
The host implementation is responsible for
    - initializing and freeing:
        - device memory (the stack struct and value buffer)
        - host memory (if we have a CPU side cache that requires initialization)
    - reseting the stack to empty (setting the `stores_` variable to zero.)
    - copying the stack state back to host (i.e. the '(store count, values[])' tuple.)
    - providing an interface for the host side to access `values[]`.
    - synchronizing kernels that will modify the stack state.

Given that the only time we update  `update_device` function if the only time that we want to copy to the device (outside of stack construction) is when calling `stack::clear`, even though I would prefer one because I like to have explicit interfaces (and I can forsee a future in which you might want to provide an explicit CUDA stream to use for host<->device updates).
However, for reading stack side state on the host, I feel very strongly about having an explicit `update_host` call that copies the contents the `(stores, values[])` informtion back to host, and not leave that up to internal logic.

Caching the values in a `std::vector` of length `capacity`, then providing begin/end and operator[] that use that vector, would be my prefered approach then for providing that information to users.

halfflat:
> [...] (and I can forsee a future in which you might want to provide an explicit CUDA stream to use for host<->device updates).

Unless the stack is going to be shared between different streams, I think the stream should constitute part of the stack state, and we still then would not need to have a two-stage clear operation.

bcumming:
I know we are walking down a garden path here....

Kernels in different streams might want to use the stack, or you might want to synchronize gpu kernels in a computation stream on h2d and d2h streams?

Even if that wasn't the case, the stream is not part of the stack state. Having it hidden in the stack makes things harder to reason about.

halfflat:
True, we can make this another issue, or take it to the wiki (IKR). That said: we get to choose where to draw the circle around the abstraction. If a stack is owned by a stream, then we have to store that ownership somewhere. If a stack is shared across streams, then we need synchronization operations on top, or at the very least, pass an execution context to anything that interacts with the GPU side.

halfflat:
Just to elaborate, we could just as easily say that the host-side cache is also not part of the stream state, and should be kept separate. Which we could! It's a perfectly reasonable design choice.

My preference is typically for low-magic minimal interfaces, where safety is guaranteed by functional semantics, or for maximum magic, where everything is handled by the object and the calling code doesn't have to worry about safety. When it's a mix of both, I believe, is where we have the most opportunity for error.

bcumming:
I think that if the only issue is whether there should be an explicit `update_device` call, or have the `clear` call make the copy, then we can make it author's call (I reckon I can get to him sooner ;) ).

Are there other issues?

halfflat:
I strongly feel that `clear()` semantics should not be split up into two calls. To make my case explicit:
* What would `clear()` mean without an `update_device()`? Clearing the host side cache without updating the GPU side has no utility.
* An `update_device()` has the implicature that state can be propagated from host to device, when really the only mutation allowed is to reset state.
* It complicates the state that has to be understood by calling code: with a unified `clear()` method, the state is either: cache reflects GPU state; or cache reflects GPU state prior to more GPU-side pushes. With separated methods, a third state is: the stack is marked for erasure; calling `update_host` will then repopulate the cache with values that may have already been consumed by client-side code. There is no way of programmatically distinguishing this state from an empty stack, without adding even more state.
* It's consequently error prone: leaving out an `update_device()` after a `clear()` would almost always be a bug. The situation where a stack might be marked for erasure later is much better addressed by making that mark explicit in the user code.

arbor/memory/cuda_wrappers.cpp
kanduri: Removed based on discussion that basically concluded with the observation that `cudaDeviceSynchronize()` is only called when dealing with managed memory code. With total removal of managed memory, there should be no need for such a function wrapper.

arbor/backends/gpu/stack.hpp
kanduri: Sure. But it doesn't seem to be a maintenance overhead as that bit is unlikely to change.

arbor/backends/gpu/stack.hpp
kanduri: This would mean that the user really reads the manual VERY carefully. A user may assume that clearing the stack implies that it is also cleared on the GPU. But would get burnt after which they explicitly call the `update_device()` function.

Is there a use case where the `clear()` and the `update_device()` calls won't happen together?

arbor/backends/gpu/stack.hpp
kanduri: Removed.

arbor/backends/gpu/threshold_watcher.hpp
kanduri: The unit test `mc_cell_group.gpu_test` fails when I don't refresh host copy. It looks like there is a query that happens on the collected spikes after the clear. I essentially did a `get_and_clear()` call there.

arbor/backends/gpu/threshold_watcher.hpp
kanduri: Done.

arbor/memory/cuda_wrappers.cpp
kanduri: Done.

arbor/backends/gpu/threshold_watcher.hpp
kanduri: Because of the `const` functions that involve copying back and forth from GPU memory into a container that's a member of `stack_type`.

arbor/memory/cuda_wrappers.hpp
kanduri: Done.

arbor/backends/gpu/stack.hpp
kanduri: Removed.

arbor/backends/gpu/stack.hpp
```diff
@@ -29,25 +29,39 @@ namespace gpu {
...
 
     using storage_type = stack_storage<value_type>;
 
     using gpu_context_handle = std::shared_ptr<arb::gpu_context>;
 
-    managed_ptr<storage_type> storage_;
+private:
+    // pointer in GPU memory
+    storage_type* storage_;
+    storage_type host_copy_;
```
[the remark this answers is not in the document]
kanduri: Done.

arbor/backends/gpu/stack.hpp
kanduri: Correct. It's not in the constructor because there are several variants including copy constructor that make use of the function. For the first implementation, I skipped it. I will look into the utility of the various constructors first.

arbor/backends/gpu/stack.hpp
```diff
@@ -29,110 +29,158 @@ namespace gpu {
...
-    stack(stack&& other): storage_(create_storage(0)), gpu_context_(other.gpu_context_) {
-        std::swap(storage_, other.storage_);
+    stack(stack&& other): gpu_context_(other.gpu_context_) {
+        create_storage(0);
+        std::swap(device_storage_, other.device_storage_);
     }
 
     stack& operator=(stack&& other) {
-        std::swap(storage_, other.storage_);
+        std::swap(device_storage_, other.device_storage_);
```
bcumming: You also have to swap `data_` and `host_copy_`.

arbor/backends/gpu/stack.hpp
```diff
@@ -29,110 +29,158 @@ namespace gpu {
...
-        storage_(create_storage(0)), gpu_context_(gpu_ctx) {}
+    stack(const gpu_context_handle& gpu_ctx): gpu_context_(gpu_ctx) {
+            create_storage(0);
+    }
 
-    stack(stack&& other): storage_(create_storage(0)), gpu_context_(other.gpu_context_) {
-        std::swap(storage_, other.storage_);
+    stack(stack&& other): gpu_context_(other.gpu_context_) {
+        create_storage(0);
+        std::swap(device_storage_, other.device_storage_);
```
bcumming: You also have to swap `data_` and `host_copy_`.

arbor/backends/gpu/stack.hpp
bcumming: Again, you don't need to update the host copy.

arbor/backends/gpu/stack.hpp
bcumming: The copy of the data should be performed in `update_host`, and this function should just return a const reference.

bcumming:
```C++
// give the device and host storage similar names (not host_copy_ and device_storage_)
storage_type* device_storage_;
storage_type host_storage_;

void update_host() {
    memory::cuda_memcpy_d2h(&host_storage_, device_storage_, sizeof(storage_type));
    const auto n = size();
    data_.resize(n);
    memory::cuda_memcpy_d2h(data_.data(), host_storage_.data, n*sizeof(T));
}

// replace get_data() with a much simpler data() call that returns a reference to whatever is on the host
const std::vector<T>& data() const {
    return data_;
}
```

Issue #756, opened 2019-05-29
Clarify and resolve cell and mechanism default values.


halfflat:
Some electrical, ion,  and mechanism parameter defaults differ from e.g. NEURON, which leads both to surprise and despair, as we try and track down any discrepancies between Arbor and NEURON simulations.

Proposal:
* Bundled mechanisms (e.g. expsyn) have parameter defaults that match those bundled with NEURON.
* Electrical properties (rl, cm) of cable cells are inherited from global or per-cell defaults: if they are not explicitly set, they will take those values.
* Per-cell defaults are optional; they will act as overrides applied to any global defaults.
* Global defaults are set in the global `cable_cell_global_properties` structure provided by a recipe with cable cells.
* We provide a bunch of defaults wrapped up in a bundle that match NEURON, but aren't applied by default. Instead, the default global defaults are in fact unset or dummy NAN values that can be tested at recipe instantiation. Users have to set the defaults explicitly, choosing to use the handy NEURON values or to supply their own.

The values/properties in question are:
* Global temperature.
* Default ion set (Na+, K+, Ca++), with default initial interior and exterior concentrations and associated reversal potential.
* Initial membrane voltage potential.
* Cellular electrical properties: axial resistivity and membrane areal capacitance.

halfflat:
This work is being rolled in together with the user-supplied reversal potential work, which also requires substantial changes to the same data structures and discretization code.

Issue #751, opened 2019-05-21
Support user-defined ionic reversal potentials


halfflat:
Currently we just Nernst everything.

1. Determine and implement an API for describing desired reversal potential evaluation in the absence of a mechanism supplying the dynamics explicitly, i.e. constant or Nernst. This could be a good opportunity also for supplying other ion data — specifically initial interior or exterior concentrations — at a finer scale than the current global defaults allow.

2. Vectorized and GPU implementations for the Nernst updates that respect a subset of ion CV indices and incorporate weight data accordingly. These indices and weights will be determined by the settings from (1) above and the presence of any explicit dynamics.

halfflat:
There is an additional issue: it's not safe to area-average multiple values for a reversal potential across the same control volume, unless the dynamics of that ion are the same across that c.v.

For example, if a control volume has weight α of one ionic current given by g₁(v-E₁), and 1-α of a current given by g₂(v-E₂), the weight averaged flux will be computed as (αg₁+(1-α)g₂)(v-(αE₁+(1-α)E₂)) instead of the correct flux of αg₁(v-E₁)+(1-α)g₂(v-E₂).

How to ameliorate?
1. We may adjust our control volume boundaries to match discontinuities in density mechanism dynamics.
2. We adjust the way that reversal potentials are described: the use of a common reversal potential across different ion channel models for the same species is essentially an optimization — the (for example) Nernst-derived potential is computed only once per control volume. Instead, we can require that models explicitly specify the form of the computed reversal potential they use to determine their ionic current contributions. Nernst-derived values can be provided by the library if required, avoiding some redundant computation, but other non-constant forms of the reversal potential would be specified within the ion current model as a function of membrane voltage, ion concentrations, and any other required parameters.

halfflat:
Current on-going work goes with (2), more or less. WIP, but reversal potential calculations can be specified for an ion at the cellular level, but not more finely, avoiding the potential nonlinearities described above.

For generic Nerst equation implementations, we need a way of feeding in ionic valence through to a mechanism (among other things). This was becoming a huger ball of changes, so moving that into another issue.

halfflat:
Ionic valence reading in mechanisms was made as #780, implemented in #781. Will make another issue and PR for re-parameterizing mechanisms over their ion channels, so that we can have, for example, a single generic Nernst mechanism that is specialized by the user for any ion they wish.

PR #823, opened 2019-07-22
New revpot and per-cell/segment parameters.

### LINKED ISSUE #750: Ion mechanism evaluation ordering in FVM.
For consistent treatment of user-defined ion concentration and reversal potential mechanisms, we need to:
* Update all such mechanisms strictly after any mechanisms that contribute to the corresponding ion currents.
* Check for and exclude mechanisms that write both ion currents and other ion variables.
* For the (currently in-progress) second-order FVM solver, treat ion concentration updates as split in the same way as voltages currently are; this may require also explicitly computing the ionic diffusivity in an implicit solution of the corresponding ODEs for stability.

halfflat:
* Collect cable cell parameter setting structures into `include/cable_cell_param.hpp`.
* Restructure electrical specifications (axial resistance, membrane capacitance) and ionic concentrations and reversal potentials on cable cells, so that these can be specified with a global default, per-cell defaults, and per-segment values.
* Allow reversal potentials to be set by a mechanism of a new kind 'revpot', which are prohibited from maintaining state or writing to any shared state other than ionic reversal potentials.
* Specify reversal potential mechanisms as global defaults or per-cell. Reversal potential mechanisms may not be specified at the level of a segment in order to avoid non-linearities arising from the discretization.
* Supply default cable cell parameter data that matches NEURON values (this is _not_ used by default).
* Replace the d_lambda calculation with one that approximates more faithfully the effect of tapered segments, and which will use the electrical values inherited by cell or global defaults.
* Supply a bundled mechanism 'nernst' that replaces the previous hard-coded Nernst potential calculation, for use (together with ion rebinding) in the new ion description scheme.
* All ions used in a cable_cell model must be present in the top level global parameter table, together with their charge.
* Extend semantics of external variables in modcc, to permit direct assignment (as opposed to just += or -=.)
* Extend `util::binary_search_index` to allow for a projection functional analagous to other range utilites.
* Add documentation on the cable cell API, sketching over many of the details, but describing in particular the interface changes for default parameters and reversal potential behaviour.

Fixes #750, #751, and #756.


arbor/include/arbor/cable_cell.hpp
```diff
@@ -251,6 +219,20 @@ class cable_cell {
...
 
+    // Approximate per-segment mean attenuation b(f) at given frequency f,
+    // ignoring membrane resistance [1/µm].
+    value_type segment_mean_attenuation(value_type frequency, index_type segidx,
+        const cable_cell_parameter_set& global_defaults) const;
+
+    // Estimate of length constant λ(f) = 1/2 · 1/b(f), following
+    // Hines and Carnevale (2001), "NEURON: A Tool for Neuroscientists",
+    // Neuroscientist 7, pp. 123-135.
+    value_type segment_length_constant(value_type frequency, index_type segidx,
```
noraabiakar: Do we use this function anywhere?

arbor/fvm_layout.cpp
```diff
@@ -59,6 +70,73 @@ namespace {
...
+
+        index_type seg_index = 0;
+        for (const auto& seg: segment_indices) {
+            const segment_info& seg_info = D.segments[seg];
+
+            if (seg_info.has_parent()) {
+                index_type cv = seg_info.parent_cv;
+
+                size_type i = parent_cv_indices.insert({cv, cv_index}).first->second;
+                if (i==cv_index) {
```
noraabiakar: Are we just checking that insertion is successful here? Why not just check `parent_cv_indices.insert({cv, cv_index}).second`?

arbor/include/arbor/cable_cell.hpp
halfflat: This is for use in (external) user code that uses the d_lambda measure for determining the number of CVs to use. The previous method just needed a segment, but now with the inheritable electrical properties, we need to move it out of the segment member functions and make it a cell method.

It's not currently used by any of our examples, but it probably does deserve a unit test.

arbor/fvm_layout.cpp
halfflat: In the instance that we've already seen this CV, we want to retrieve the previously saved index value for passing on to the action; by grabbing the value from the first field (the iterator), we can avoid a second look-up.

Issue #824, opened 2019-07-23
modcc `CONSTANT` block


noraabiakar:
Add support for `CONSTANT` block in .mod files: 
```
CONSTANT {
    temp0 = 0.1 (S/cm2)
    temp1 = -2
}
PARAMETER {
    g = temp0 (S/cm2)
}
BREAKPOINT {
    i = g*(v - temp1)
}
```
* The `CONSTANT` block should be declared before any other block that uses any of the variables declared inside it 
* Units in the `CONSTANT` block are parsed but not processed
* Variables from the constant block can be used anywhere after the point they are declared in the `CONSTANT` block, including within the block itself
* Inside the `CONSTANT` block, arithmetic operations are not allowed

PR #825, opened 2019-07-23
parse constants block and save (key, value) pairs in an unordered_map

noraabiakar:
Addresses #824 
Adds support for `CONSTANT` block in modcc files:
- The `CONSTANT` block should be declared before any other block that uses any of the variables declared inside it 
- Units in the `CONSTANT` block are parsed but not processed
- Variables from the constant block can be used anywhere after the point they are declared in the `CONSTANT` block, including within the block itself
- Inside the `CONSTANT` block, arithmetic operations are not allowed, only negation


modcc/parser.cpp
```diff
@@ -558,6 +561,68 @@ void Parser::parse_parameter_block() {
...
+
+    get_token(); // consume closing brace
+
+    parm_exit:
+    // only write error message if one hasn't already been logged by the lexer
+    if(!success && status_==lexerStatus::happy) {
+        error(pprintf("CONSTANT block unexpected symbol '%s'", token_.spelling));
+    }
+    return;
+}
```
halfflat: The error handling here is a bit convoluted. I think it could be simplified by:
* There are only three points, at lines 569, 582, and 612, where a non-lexer error condition is handled. Each has a different message, so there's nothing really to be gained by jumping on to `parm_exit` for the last two. I suggest using `error(...)` and return in each of the three instances.
* When there is a lexer error, no further error will be logged, so one can just return directly (lines 593, 602).
* This in turn means there is no need form the `parm_exit:` code at all, I think.

noraabiakar: I admit I don't understand the whole picture when it comes to the error handling code in the parser. I'll implement your suggestions, thanks!

PR #826, opened 2019-07-24
Python wrapper: complete schedule by adding events(t0,t1)

### LINKED ISSUE #770: Python wrapper: complete event_generators
add `reset()`, `events(t0,t1)` and `empty_schedule`

akuesters:
- Add `events(t0, t1)` to schedules
- No need for `reset()` anymore, since new interface asks for values between a range
- No need for `empty_schedule`, since of no use for python user

Fixes #770
Addresses #799

### DIFF
--- a/python/schedule.hpp
+++ b/python/schedule.hpp
@@ -36,7 +36,7 @@ struct regular_schedule_shim {
 
     arb::schedule schedule() const;
 
-    std::list<arb::time_type> events(pybind11::object t0, pybind11::object t1);
+    std::vector<arb::time_type> events(pybind11::object t0, pybind11::object t1);
 };
 
 // A Python shim for arb::explicit_schedule.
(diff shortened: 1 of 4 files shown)


python/schedule.hpp
```diff
@@ -34,6 +35,8 @@ struct regular_schedule_shim {
     opt_time_type get_tstop()  const;
 
     arb::schedule schedule() const;
+
+    std::vector<arb::time_type> events(pybind11::object t0, pybind11::object t1);
```
bcumming: Why not just use `arb::time_type` instead of `pybind11::object` here, and let pybind11 do the conversion?

akuesters: Since we allowed `None` in the regular schedule as inputs, so I would assume as a user that `None` would also be possible to get the list of events.

bcumming: This interface explicitly requests a start time and end time for the interval. What would `None` mean in this context?

akuesters: Same as for the `regular_schedule` - no start time, no end time. Thus, no events, or all events. Happy to change it to `arb::time_type` if these are no user cases.

bcumming: I think that all the schedules should have the same interface of `(t0, t1)` being valid `time_type`s.

akuesters: Ok.

akuesters: Done.

Issue #769, opened 2019-06-04
Python wrapper: tests


akuesters:
Add tests to `python/test`:

- [ ] test_recipe
- [x] test_domain_decomposition
- [ ] test_cells
  - [ ] test_cable_cell
    - [ ] test_segment
    - [ ] test_synapses
  - [ ] test_lif_cell
  - [ ] test_benchmark_cell
  - [ ] test_spike_source_cell
- [x] test_probes
- [x] test_spikes

thorstenhater:
Done.

PR #827, opened 2019-07-31
Python wrapper: add hint_map to domain decomposition

### LINKED ISSUE #776: Python wrapper: add hint_map to domain decomposition

akuesters:
- wraps `partition_hint` struct for python
- `partition_hint`struct in `load_balance.hpp` is adjusted in a way that in case of zero unsigned int `cpu_group_size`/`gpu_group_size` it is set to the default value (via setter/getter)
- adds documentation for `partition_hint`  
- adds testing for `domain_decomposition` including `partition_hint` (in unit and unit_distributed)
- adds `partition_hint` in example `ring.py`
- corrects `config.cpp` (and doc) to test for `ARB_GPU_ENABLED` instead of `ARB_WITH_GPU`

Fixes #776 
Addresses #799, #769


arbor/include/arbor/load_balance.hpp
```diff
@@ -7,11 +7,54 @@
 namespace arb {
 
 struct partition_hint {
-    constexpr static std::size_t max_size = -1;
 
-    std::size_t cpu_group_size = 1;
-    std::size_t gpu_group_size = max_size;
-    bool prefer_gpu = true;
+    partition_hint() :  cpu_group_size{cpu_group_size_default},
```
bcumming: The reason that these changes were made here was to simplify the Python wrapper?

Instead, I think that this interface should be left as it is (it is much simpler), and the Python wrapper might have to use a shim to implement the partition hint.

python/domain_decomposition.cpp
```diff
@@ -72,12 +106,13 @@ void register_domain_decomposition(pybind11::module& m) {
     // Partition load balancer
     // The Python recipe has to be shimmed for passing to the function that takes a C++ recipe.
     m.def("partition_load_balance",
-        [](std::shared_ptr<py_recipe>& recipe, const context_shim& ctx) {
-            return arb::partition_load_balance(py_recipe_shim(recipe), ctx.context);
+        [](std::shared_ptr<py_recipe>& recipe, const context_shim& ctx, std::unordered_map<arb::cell_kind, arb::partition_hint>& hint_map) {
```
bcumming: This can be simplified a bit, by passing the `hint_map` by value, using the convenience type `arb::partition_hint_map`, and using the C++ type as the default value (no need for pybind11 to convert a Python dictionary to a C++ hash table).

```C++
    m.def("partition_load_balance",
        [](std::shared_ptr<py_recipe>& recipe,
           const context_shim& ctx,
           arb::partition_hint_map hint_map)
        {                                                                                                
            return arb::partition_load_balance(py_recipe_shim(recipe),
                                               ctx.context,
                                               std::move(hint_map));
        },
        "Construct a domain_decomposition that distributes the cells in the model described by recipe\n"
        "over the distributed and local hardware resources described by context.\n"
        "Optionally, provide a dictionary of partition hints for certain cell kinds, by default empty.",
        "recipe"_a, "context"_a, "hints"_a=arb::partition_hint_map{});
```

arbor/include/arbor/load_balance.hpp
akuesters: Changed it back and included exception handling in `partition_load_balance`

python/domain_decomposition.cpp
akuesters: Done.

python/example/ring.py
```diff
@@ -59,10 +59,19 @@ def event_generators(self, gid):
 decomp = arbor.partition_load_balance(recipe, context)
 print(f'{decomp}')
 
+hint = arbor.partition_hint()
+hint.prefer_gpu = False
+hint.gpu_group_size = 1
```
bcumming: It is a bit odd to set gpu_group_size after setting `prefer_gpu` to false.
It also goes against the recommended approach of using many cells per group on the GPU (and an example should illustrate best practices.)

Maybe set it to `True`, then set `gpu_group_size=1000`?

python/test/unit_distributed/runner.py
```diff
@@ -21,16 +21,22 @@
     import options
     import test_contexts_arbmpi
     import test_contexts_mpi4py
+    import test_domain_decompositions_mpi4py
```
bcumming: Can you refactor the tests so that you can test either arbmpi or mpi4py with the same test?
Or maybe only test all the features using arbmpi, and have separate test that verifies that mpi4py can be used to initialize an Arbor context (but don't then verify all the features working with mpi4py)?

python/example/ring.py
akuesters: True. Done.

python/test/unit_distributed/runner.py
akuesters: I think I can refactor.

akuesters: I would refactor `test_domain_decomposition` to test either arbmpi or mpi4py (dependend on which is used/set up), but I would keep the context tests seperated to force arbmpi/mpi4py being tested individually. Is that fine?

bcumming: Yes. The context tests should be separate, to ensure that we can use both.

Can we just run all the rest of the MPI tests with `arbmpi`?

akuesters: Do you force the rest of the MPI to be tested with `arbmpi`?

akuesters: Done.

Issue #828, opened 2019-08-07
modcc: process `CONSERVE` statement in `KINETIC` block 


noraabiakar:
`CONSERVE` statements in the `KINETIC` block are currently being parsed without being processed. 

We can add support for processing conserve expressions, by replacing one of the differential equations generated from the kinetic block with the conserve statement. 

example: 
```
KINETIC state {
    ~ s <-> h (alpha1, beta1)
    ~ d <-> s (alpha2, beta2)
}
```
which translates into:

```
s' = (-alpha1 -beta2)*s + beta1*h + alpha2*d
h' = alpha1*s - beta1*h
d' = beta2*s - alpha2*d
```
results in the following system of equations: 

```
[1 -  dt*(-alpha1 -beta2)]*s -     (dt*beta1)*h -     (dt*alpha2)*d = s_prev
              (dt*-alpha1)*s - (1 + dt*beta1)*h +             (0)*d = h_prev
               (dt*-beta2)*s +            (0)*h - (1 + dt*alpha2)*d = d_prev
```

Changing the kinetic block to: 
```
KINETIC state {
    ~ s <-> h (alpha1, beta1)
    ~ d <-> s (alpha2, beta2)
    CONSERVE s + d + h = 1
}
```
should result in the same set of differential equations, but the system of equations would have one of the equations replaced with the conserve statement, for example:  

```
(dt*-alpha1)*s - (1 + dt*beta1)*h +             (0)*d = h_prev
 (dt*-beta2)*s +            (0)*h - (1 + dt*alpha2)*d = d_prev
         (1)*s +            (1)*h +             (1)*d = 1
```

PR #829, opened 2019-08-07
modcc: process `CONSERVE` statements in `KINETIC` block

noraabiakar:
Addresses #828 and #830 
Adds support for processing `CONSERVE` statements in `KINETIC` block. 

The `KineticRewriter` remains unchanged. The `SparseSolverVisitor` is modified. 
If one or more `CONSERVE` statements are present; corresponding rows in the symbolic matrix, which would otherwise represent differential equations, are replaced by the content of the conserve statements.

### DIFF
--- a/modcc/solvers.cpp
+++ b/modcc/solvers.cpp
@@ -190,7 +190,6 @@ void SparseSolverVisitor::visit(AssignmentExpression *e) {
         return;
     }
 
-    std::cout << "visit " << e->to_string() << std::endl;
     if (A_.empty()) {
         unsigned n = dvars_.size();
         A_ = symge::sym_matrix(n, n);
@@ -264,20 +263,15 @@ void SparseSolverVisitor::visit(AssignmentExpression *e) {
         auto local_a_term = make_unique_local_assign(scope, expr.get(), "a_");
         auto a_ = local_a_term.id->is_identifier()->spelling();
 
-        std::cout << "?  : " << local_a_term.local_decl->to_string() << std::endl;
-        std::cout << "?? : " << local_a_term.assignment->to_string() << std::endl;
-
         statements_.push_back(std::move(local_a_term.local_decl));
         statements_.push_back(std::move(local_a_term.assignment));
 
-        std::cout << "AT " << deq_index_ << " " << j << std::endl;
         A_[deq_index_].push_back({j, symtbl_.define(a_)});
     }
     ++deq_index_;
 }
 
 void SparseSolverVisitor::visit(ConserveExpression *e) {
-    std::cout << "visit_con " << e->to_string() << std::endl;
     if (A_.empty()) {
         unsigned n = dvars_.size();
         A_ = symge::sym_matrix(n, n);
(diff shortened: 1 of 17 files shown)


modcc/solvers.cpp
```diff
@@ -265,11 +270,81 @@ void SparseSolverVisitor::visit(AssignmentExpression *e) {
...
+        }
+    }
+
+    if (row_idx != -1) {
+        A_[row_idx].clear();
+
+        for (unsigned j = 0; j < dvars_.size(); ++j) {
+            expression_ptr expr;
+
+            for (auto &l: e->lhs()->is_stoich()->terms()) {
```
bcumming: can this be replaced with `find_if`/`find`?

modcc/solvers.hpp
```diff
@@ -19,6 +19,10 @@ class SolverVisitorBase: public BlockRewriterBase {
     // list of identifier names appearing in derivatives on lhs
     std::vector<std::string> dvars_;
 
+    // rhs of conserve statement
```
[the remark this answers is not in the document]
bcumming: Also reset `conserve_`, i.e. `conserve_=false` in `::reset().

modcc/solvers.cpp
```diff
@@ -185,6 +185,11 @@ void SparseSolverVisitor::visit(BlockExpression* e) {
 }
 
 void SparseSolverVisitor::visit(AssignmentExpression *e) {
+    if (conserve_ && !A_[deq_index_].empty()) {
```
bcumming: Double check:
* that `conserve_==true` implies that `A_` has been initialized.
* that putting CONSERVE before or in the middle of the "assignment expressions" won't (a) break things (b) give different linear system.

bcumming: Can `A_` be initialized to `n` rows on construction, when `dvars_` is provided? The only problem might be if `::reset()` is called?

noraabiakar: * `conserve_==true` does imply that `A_` has been initialized, but I will check the initialization of `A_` before checking `conserve_` anyway
* Putting `CONSERVE` before or in the middle of the assignment expressions won't break things and will give the same linear system; changing the order of the variables in the conserve statement will give a different linear system
* I think it should be okay to initialize `A_` once `dvars_` is filled, I assume the reason it wasn't done like that originally is because of the `::reset()`

modcc/solvers.cpp
```diff
@@ -265,11 +270,81 @@ void SparseSolverVisitor::visit(AssignmentExpression *e) {
...
+    }
+
+    conserve_ = true;
+
+    auto loc = e->location();
+    scope_ptr scope = e->scope();
+
+    int row_idx = -1;
+
+    auto& l = e->lhs()->is_stoich()->terms().front();
```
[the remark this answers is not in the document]
noraabiakar: That looks cleaner, I will change it

modcc/solvers.cpp
noraabiakar: Yes, I will update it

modcc/solvers.hpp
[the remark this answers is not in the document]
noraabiakar: True, I will move them

PR #842, opened 2019-08-16
New analysis for passive cable equation

halfflat:
* Replace derivation of cylindrical cable equation formula with S–L approach.
* Relate analysis of cylindrical form to Rallpack 1 in a separate section.
* Add analysis of tapered cable case.
* Unify discussion of cable with constant electrical properties but varying geometry.
* Add section describing how one can well-estimate the inter-cell flux given point samples or CV means for the membrane voltage.


halfflat:
I'm almost certain I have missed something in the proof reading.

PR #846, opened 2019-08-21
Modcc compartment

### LINKED ISSUE #838: modcc: parse and process COMPARTMENT statements
Compartment statements are only relevant in KINETIC blocks and have the following format: 
```
COMPARTMENT volume {state_1 state_2 . . . }
```
According to the Neuron Book: 
`volume` only multiplies the `d(state_x)/dt` of the equivalent differential equations.

noraabiakar:
Adds modcc support for COMPARTMENT statements of the form:
`COMPARTMENT v {state_0, state_1, ..., state_n}`
Which can be used in KINETIC blocks to multiply the derivative of state variables `state_0, state_1, ..., state_n` by a factor `v`

Fixes #838


modcc/parser.cpp
```diff
@@ -1304,13 +1306,13 @@ expression_ptr Parser::parse_conserve_expression() {
     return make_expression<ConserveExpression>(here, std::move(lhs), std::move(rhs));
 }
 
-expression_ptr Parser::parse_expression(int prec) {
+expression_ptr Parser::parse_expression(int prec, tok t) {
     auto lhs = parse_unaryop();
```
halfflat: A comment explaining what `t` means here, or a more explicit parameter name (e.g. `stop_token`?) would be a help to explain its function.

modcc/solvers.cpp
```diff
@@ -180,10 +180,25 @@ void SparseSolverVisitor::visit(BlockExpression* e) {
...
     BlockRewriterBase::visit(e);
 }
 
+void SparseSolverVisitor::visit(CompartmentExpression *e) {
+    auto loc = e->location();
+
+    for (auto& s: e->is_compartment()->state_vars()) {
+        auto it = std::find(dvars_.begin(), dvars_.end(), s->is_identifier()->spelling());
+        if (it == dvars_.end()) {
+            error({"COMPARTMENT variable is not a state variable", loc});
```
halfflat: I think strictly speaking, we can have a state variable that is not present as a derivative or participant in a kinetic scheme. Maybe an error message such as "COMPARTMENT variable is not used" would be more accurate?

modcc/parser.hpp
```diff
@@ -21,8 +21,9 @@ class Parser : public Lexer {
     expression_ptr parse_integer();
     expression_ptr parse_real();
     expression_ptr parse_call();
-    expression_ptr parse_expression(int prec);
+    expression_ptr parse_expression(int prec, tok t=tok::reserved);
```
halfflat: Defaulting a different stop-token than `tok::eq` means that assignments in subexpressions won't be caught as parse errors; is there a reason why we can't use that default rather than `tok::reserved`?

test/unit-modcc/test_parser.cpp
```diff
@@ -323,7 +323,6 @@ TEST(Parser, parse_parenthesis_expression) {
         "(x             ",
         "((x+3)         ",
         "(x+ +)         ",
-        "(x=3)          ",  // assignment inside parenthesis isn't allowed
         "(a + (b*2^(x)) ",  // missing closing parenthesis
```
halfflat: If we have `tok::eq` as the default stop token in `parse_expression`, we could keep this test case…

modcc/solvers.cpp
noraabiakar: Okay, I will update the comment

modcc/parser.hpp
noraabiakar: I think using `tok::eq` works too, I will change it.

test/unit-modcc/test_parser.cpp
noraabiakar: Will keep the test.

modcc/parser.cpp
noraabiakar: I will rename it to `stop_token`

PR #848, opened 2019-08-22
Consistent source naming and layout for examples

### LINKED ISSUE #847: Consistent source naming and layout for examples
Recently we have settled on the following pattern for examples, illustrated for `ring` here:

Example code path: `examples/ring`
Single C++ source file: `examples/ring/ring.cpp`
Executable: `bin/ring`

None of the examples is particularly complex, so they could all be moved to this organisation, which would fix some low-level aggravation whenever working with changes that touch examples.

bcumming:
Refactor the example code so that all examples have a single C++ source file, and the source file, source path and executable all have the same name.

Fixes  #847.


example/gap_junctions/gap_junctions.cpp
```diff
@@ -26,13 +27,20 @@
...
+    double sim_duration = 100;
+    bool print_all = true;
+};
 
-#include "parameters.hpp"
-
-#ifdef ARB_MPI_ENABLED
-#include <mpi.h>
-#include <arborenv/with_mpi.hpp>
-#endif
```
noraabiakar: I think these were removed by mistake?

bcumming: They were. I fixed them.

bcumming:
I added the lines that were accidentally deleted, and fixed a CMake problem that stopped Travis.

PR #860, opened 2019-08-30
Morphology part 2.

### LINKED ISSUE #845: Morphology Pt 2: abstract representation of regions and points on cells.
Add support for:
* Writing expressions with primitive location and region descriptions.
* Labeling expressions and storing in dictionaries.
* Generating concrete location lists and cable lists by applying the expressions to a morphology.
* Painting `(mechanism, parameter)` tuples to named regions

The `make_cable_cell` method that converts a `morphology` into a `cable_cell` would be extended to perform the "Painting" operation by adding mechanisms to whole sections that correspond to whole cables in a cable list.

bcumming:
Towards constructing from regions.

New "embedded morphology" `em_morphology`:
- constructed from a `morphology`
- caches information like terminal points, locations of sample points, branch lengths.
- interface for converting sample ids to locations and canonicalizing locations
  - will be extended to implement integration, interpolation etc.
- calculation of lengths, interpolation, etc currently rely on implicit piecewise linear method
  - other schemes could be implemented in embedded morphology, separating the interpretation of the morphology from the underlying description of a connected set of samples.

Simple DSL for describing locsets and regions, which are abstract descriptions of sets of locations and regions on morphologies
- use type erasure to allow for easy extension of the DSL, and hiding of implementation
- each locset and region provides method for generating a concrete set of locations of cables for a given morphology.
- operators for performing union and intersection of locations and regions.

New `label_dict`:
- store and look up named "regions"
- no smarts beyond ensuring that labels are unique across locsets and regions.
#845
`make_cable_cell` now takes an optional `label_dict` argument, and the `cable_cell` interface has been extended with a `cable_cell::paint` method for associating mechanisms with named regions.
- no changes to how `cable_cell` stores values underneath.

Fixes #845.


arbor/morph/region.cpp
```diff
@@ -0,0 +1,255 @@
...
+bool is_disjoint_union(const mcable& a, const mcable& b) {
+    if (a.branch!=b.branch) return true;
+    return a<b? a.dist_pos<b.prox_pos: b.dist_pos<a.prox_pos;
+}
+
+bool is_nonnull_intersection(const mcable& a, const mcable& b) {
+    if (a==b) return true; // handles special case of spherical branch
+    if (a.branch!=b.branch) return false;
+    return a<b? a.dist_pos>b.prox_pos: b.dist_pos>a.prox_pos;
+}
```
halfflat: This will return false if a<b but a.dist_pos==b.prox_pos; but I believe mcables represent closed intervals, and so this ought to be permissible?

More fundamentally, how does (or should) this differ from just testing `!is_disjoint_union(a, b)`?

arbor/morph/region.cpp
```diff
@@ -0,0 +1,255 @@
...
+    }
+    return region(cable_{c});
+}
+
+region branch(msize_t bid) {
+    return cable({bid, 0, 1});
+}
+
+mcable_list concretise_(const cable_& reg, const em_morphology& em) {
+    auto& m = em.morph();
```
halfflat: I'm happy with `concretise` as a name, though it is a bit of a mouthful. If we want to consider something a bit easier, `reify` may be an alternative.

arbor/include/arbor/cable_cell.hpp
```diff
@@ -273,6 +273,8 @@ cable_segment* cable_cell::add_cable(cable_cell::index_type parent, Args&&... ar
 // If compartments_from_discretization is true, set number of compartments
 // in each segment to be the number of piecewise linear sections in the
 // corresponding section of the morphology.
-cable_cell make_cable_cell(const morphology& morph, bool compartments_from_discretization);
+cable_cell make_cable_cell(const morphology& morph,
+                           label_dict labels={},
```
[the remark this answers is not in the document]
bcumming: Yep.

arbor/include/arbor/morph/label_dict.hpp
```diff
@@ -0,0 +1,38 @@
...
+public:
+    label_dict();
+    label_dict(const label_dict&);
+    label_dict(label_dict&&);
+
+    void set(const std::string& name, locset p);
+    void set(const std::string& name, region r);
+
+    util::optional<arb::region&>       region(const std::string& name);
+    util::optional<const arb::region&> region(const std::string& name) const;
```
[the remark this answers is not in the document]
bcumming: It is my error that the `set` method silently drops insertion on an (ignorance on my part regarding how `std::unordered_map::instert` behaves when inserting with a key that is already in the table).

I agree with your proposed changes (because you are right, it is icky design).

arbor/include/arbor/morph/locset.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+namespace arb {
+
+// Forward declarations of locset, region and their respective dictionaries,
+// which are required to define the interface for locset.
+class locset;
+class region;
+using locset_dictionary = std::unordered_map<std::string, locset>;
+using region_dictionary = std::unordered_map<std::string, region>;
+
```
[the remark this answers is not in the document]
bcumming: They needed to be forward declared, before I removed the code that required it before making the PR.
Consider them deleted.

arbor/cable_cell.cpp
```diff
@@ -218,6 +237,24 @@ cable_cell make_cable_cell(const morphology& morph, bool compartments_from_discr
...
+    std::unordered_map<std::string, std::vector<msize_t>> regions;
+    for (auto r: dictionary.regions()) {
+        mcable_list L = concretise(r.second, emorph);
+        std::vector<msize_t> bids;
+        for (auto c: L) {
+            if (c.prox_pos!=0 || c.dist_pos!=1) {
+                std::stringstream s;
+                s << "cable_cell does not support regions that contain partial branches (yet): \"" << r.first << "\": " << c;
+                throw cable_cell_error(s.str());
+            }
```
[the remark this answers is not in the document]
bcumming: Of course. When I wrote this I thought I was working with front end code, which can't see `pprintf`.

arbor/include/arbor/morph/locset.hpp
```diff
@@ -0,0 +1,133 @@
...
+        return p.impl_->print(o);
+    }
+
+    // The union of two location sets.
+    friend locset join(locset, locset);
+
+    template <typename ...Args>
+    friend locset join(locset l, locset r, Args... args) {
+        return join(join(std::move(l), std::move(r), std::move(args)...));
+    }
```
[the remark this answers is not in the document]
bcumming: Good catch, this was a typo. Moving the closing parenthesis to the right spot fixes the problem:

```C++
    template <typename ...Args>
    friend locset join(locset l, locset r, Args... args) {
        return join(join(std::move(l), std::move(r)), std::move(args)...);
    }
```

This was already fixed and unit tested for the region version. I have added tests for the `locset` implementation.

arbor/include/arbor/morph/primitives.hpp
```diff
@@ -48,6 +48,46 @@ struct msample {
...
 double distance(const msample& a, const msample& b);
 
+// Describe a specific location on a morpholology.
+struct mlocation {
+    // The id of the branch.
+    msize_t branch;
+    // The relative position on the branch ∈ [0,1].
+    double pos;
+
+    friend bool test_invariants(const mlocation&);
```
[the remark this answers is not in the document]
bcumming: :+1:

arbor/include/arbor/morph/label_dict.hpp
```diff
@@ -0,0 +1,38 @@
...
+    using ps_map = std::unordered_map<std::string, arb::locset>;
+    using reg_map = std::unordered_map<std::string, arb::region>;
+    ps_map locsets_;
+    reg_map regions_;
+
+public:
+    label_dict();
+    label_dict(const label_dict&);
+    label_dict(label_dict&&);
+
```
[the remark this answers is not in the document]
bcumming: Agree completely.
I wonder why I moved them to the implementation in the first place?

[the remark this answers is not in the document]
bcumming: I just removed them completely, seeing as how they are all default.

arbor/include/arbor/morph/label_dict.hpp
```diff
@@ -0,0 +1,38 @@
...
+    util::optional<arb::region&>       region(const std::string& name);
+    util::optional<const arb::region&> region(const std::string& name) const;
+
+    util::optional<arb::locset&>       locset(const std::string& name);
+    util::optional<const arb::locset&> locset(const std::string& name) const;
+
+    const ps_map& locsets() const;
+    const reg_map& regions() const;
+
+    size_t size() const;
```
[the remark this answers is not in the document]
bcumming: :+1:

arbor/morph/morphology_impl.hpp
```diff
@@ -0,0 +1,38 @@
+#pragma once
```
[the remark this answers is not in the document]
bcumming: The `mbranch` type is used internally in the morpholgy implementation, but I needed it for unit testing. Turns out it was the only thing I needed, so you are right that `mbranch.hpp` is a more appropriate name.
Furthermore, the implementation should go in `mbranch.cpp`, not `morphology.cpp`.

arbor/morph/region.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+mcable_list merge(const mcable_list& lhs, const mcable_list& rhs) {
+    mcable_list v;
+    v.resize(lhs.size() + rhs.size());
+    std::merge(lhs.begin(), lhs.end(), rhs.begin(), rhs.end(), v.begin());
+    return v;
+}
+
+bool is_disjoint_union(const mcable& a, const mcable& b) {
+    if (a.branch!=b.branch) return true;
```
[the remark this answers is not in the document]
bcumming: How about calling it `is_disjoint`, with an explicit comment that outlines these restrictions? It is only used internally, not exported for consumption outside this file.

arbor/morph/region.cpp
```diff
@@ -0,0 +1,255 @@
...
+    return a<b? a.dist_pos<b.prox_pos: b.dist_pos<a.prox_pos;
+}
+
+bool is_nonnull_intersection(const mcable& a, const mcable& b) {
+    if (a==b) return true; // handles special case of spherical branch
```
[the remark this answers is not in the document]
bcumming: It is! And no longer needed.
In older versions of this code `(cable 0 0 0)` was a perfectly acceptable way to refer to the whole sphere (because a sphere has inside and outside, but no ends).
Happily deleting that line keeps the unit tests happy.

arbor/morph/region.cpp
bcumming: The problem is that I am effectively ignoring regions that contain a single point.
Given two cables: `a=(cable 1 0 0.5)` and `b=(cable 1 0.5 1)`, the current implementation gives

`(join a b) -> (cable 1 0 1)` ... nothing controversial there
`(intersection a b) -> {}` when it should be `(cable 1 0.5 0.5)` if we treat cable segments as closed intervals.

Using `!is_disjoint(a,b)` makes unit tests fail where we get these interesctions that define a location.

If we allow zero-length regions like this, we would still be inconsistent:
Take `(join (cable p 0 1) (cable c 0 1))` where branch `p` is the parent of child branch `c`: this would still be empty, because the code that performs the intersection performs intersections on cables on the same branch, ignoring parent-child relationships.

I  propose either ignoring zero-length regions, or using half open intervals to keep the current results. Outside of formal completeness, do we have a practical use for compact finite regions?

bcumming: We can define a locset as a multiset on the cover, where we define `m_A(x)` is the multiplicity of location `x` in locset `A`.

The rationale behind using a multiset is:
- If a locset is used to describe the locations of synapses, it has to permit multiple synapses at the same location.
- To describe CV discretisations uniquely with only a set of locations.

How do we define the operations of union (`join`) and intersection (`intersection`)?

To get the "expected" property of intersection distributing over union, and vice versa, i.e.
`A ∪ B) ∩ C = (A ∩ C) ∪ (B ∩ C)` and `A ∩ B) ∪ C = (A ∪ C) ∩ (B ∪ C)`, the following definitions work:
- `C =  A ∪ B`: `m_C(x) = max(m_A(x), m_B(x))`
- `C =  A ∩ B`: `m_C(x) = min(m_A(x), m_B(x))`

However, when a user "combines" two sets of locations to define where to place synapses, they the union above would not provide the behaviour that they expect.
We could implement a `sum` operator:
- `C = A + B`: `m_C(x) = m_A(x) + m_B(x)`

With this, the following "painting" would work logically for a user `cell.paint("expsyn", "sum(A, B)")`.

halfflat: Yup, definitely. For multisets, + and ∪ do have to be distinct. We should support both of these primitive operations, I believe.

halfflat: Yup, definitely. For multisets, + and ∪ do have to be distinct. We should support both of these primitive operations, I believe.

arbor/morph/region.cpp
bcumming: It isn't that much of a mouthful! `reify` is a bit obscure: it wasn't in my active vocabulary.

We could just use `concrete`?

halfflat: Or `realize`?

halfflat: `concretise` is fine though.

arbor/include/arbor/morph/label_dict.hpp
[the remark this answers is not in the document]
bcumming: Fixed.

arbor/cable_cell.cpp
```diff
@@ -164,10 +180,13 @@ value_type cable_cell::segment_mean_attenuation(
     return 2*std::sqrt(math::pi<double>*tau_per_um*frequency)*length_factor; // [1/µm]
 }
 
-cable_cell make_cable_cell(const morphology& morph, bool compartments_from_discretization) {
+cable_cell make_cable_cell(const morphology& m, label_dict dictionary, bool compartments_from_discretization) {
     using point3d = cable_cell::point_type;
     cable_cell newcell;
 
+    auto emorph = em_morphology(m);
+    auto& morph = emorph.morph();
```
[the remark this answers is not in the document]
bcumming: Done. Moved declaration of `emorph` to the bottom of the function, where it is used, while I was at it.

arbor/morph/region.cpp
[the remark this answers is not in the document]
bcumming: Done.

arbor/morph/region.cpp
bcumming: Done. I removed `join` and `intersect` from the public API for locset, because they are not used anywhere currently (and I am still trying to wrap my head around where and when we might want to use them).
I have kept their implementation in the back end, and wrote comprehensive tests for them.

arbor/morph/region.cpp
bcumming: Right, then, `thingify` it is!

bcumming:
Righteos, I have made the changes. Improved things for sure.

arbor/morph/em_morphology.cpp
```diff
@@ -0,0 +1,150 @@
...
+    else if (loc.pos==0) {
+        // The location is at the start of a branch: at the location at
+        // the end of the parent branch, and locations at the start of
+        // each child of the parent branch.
+        auto p = morph_.branch_parent(loc.branch);
+        if (p!=mnpos) L.push_back({p, 1});
+        for (auto b: morph_.branch_children(p)) {
+            if (b!=loc.branch) L.push_back({b, 0});
+        }
+    }
```
halfflat: This case and the `loc==mlocation{0,0}` case for non-spherical root could be combined, which would avoid a bit of code duplication (and of course make things a little tighter.). You could e.g. change the root special case code to just
```
if (loc==mlocation{0,0} && morph_.spherical_root()) return L;
```

arbor/morph/em_morphology.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+const morphology& em_morphology::morph() const {
+    return morph_;
+}
+
+mlocation em_morphology::root() const {
+    return {0,0};
+}
```
halfflat: This could be re-examined with the upcoming `minset` code perhaps, but should the root be `{0,0}` in the non-spherical soma case, rather than e.g. `{mnpos, 0}`?

arbor/morph/em_morphology.hpp
```diff
@@ -0,0 +1,39 @@
...
+public:
+    em_morphology(const morphology& m);
+
+    const morphology& morph() const;
+
+    mlocation_list terminals() const;
+    mlocation root() const;
+
+    mlocation sample2loc(msize_t sid) const;
+
```
halfflat: (This is just a personal preference, but I'm really not fond of `foo2bar` type names for conversions. Can I convince you of e.g. `from_sample_id` or `sample_to_loc` or `sample_location` …?)

arbor/morph/label_dict.cpp
```diff
@@ -0,0 +1,76 @@
...
+        it->second = std::move(reg);
+    }
+    else {
+        regions_.emplace(name, std::move(reg));
+    }
+}
+
+util::optional<const region&> label_dict::region(const std::string& name) const {
+    auto it = regions_.find(name);
+    if (it==regions_.end() ) return {};
```
halfflat: Cosmetic only, but there's an unusual space in the condition here and on line 64.

arbor/morph/locset.cpp
```diff
@@ -0,0 +1,248 @@
...
+    auto r    = rhs.begin();
+    auto rend = rhs.end();
+
+    auto at_end = [&]() { return l==lend || r==rend; };
+    while (!at_end()) {
+        auto x = (*l<*r) ? *l: *r;
+        auto count = (*l<*r)? multiplicity(l, lend):
+                     (*r<*l)? multiplicity(r, rend):
+                     std::max(multiplicity(l, lend), multiplicity(r, rend));
+        while (count--) L.push_back(x);
```
halfflat: Here and below (line 88), an alternative is to use `L.insert(L.end(), count, x)`. I'm sure it makes negligible performance difference.

arbor/morph/region.cpp
```diff
@@ -0,0 +1,306 @@
...
+// Returns true iff cable sections a and b:
+//  1. are on the same branch
+//  2. overlap, i.e. their union is not empty.
+bool is_disjoint(const mcable& a, const mcable& b) {
+    if (a.branch!=b.branch) return true;
+    return a<b? a.dist_pos<b.prox_pos: b.dist_pos<a.prox_pos;
+}
+
+bool is_nonnull_intersection(const mcable& a, const mcable& b) {
+    //if (a==b) return true; // handles special case of spherical branch
```
halfflat: This helper is no longer being used.

arbor/morph/em_morphology.cpp
bcumming: This had occurred to me.

arbor/morph/region.cpp
```diff
@@ -0,0 +1,306 @@
...
+region tagged(int id) {
+    return region(tagged_{id});
+}
+
+mcable_list thingify_(const tagged_& reg, const em_morphology& em) {
+    auto& m = em.morph();
+    size_t nb = m.num_branches();
+
+    std::vector<mcable> L;
+    L.reserve(nb);
```
halfflat: In many cases, the number of branches will be a significant over-estimate of the size of the cable list; is it worth reserving? Or, alternatively, is it worth doing a `shrink_to_fit` before returning it?

arbor/morph/region.cpp
```diff
@@ -0,0 +1,306 @@
...
+    return branches;
+}
+
+std::ostream& operator<<(std::ostream& o, const all_& t) {
+    return o << "all";
+}
+
+//
+// intersection of two point sets
+//
```
halfflat: Copy-pasto on the comment here and on line 269: should be regions rather than point sets.

arbor/morph/em_morphology.cpp
bcumming: Done.

arbor/morph/em_morphology.cpp
bcumming: But nowhere in the current code do we distinguish between this and other uses. And it would require hand implementing the lexographical ordering for `mlocation`, to special case that `mnpos` less than all other values.

test/unit/test_morphology.cpp
```diff
@@ -19,21 +17,38 @@
...
 
+// Test basic functions on properties of mpoint
+TEST(morphology, mpoint) {
+    using mp = arb::mpoint;
+
+    EXPECT_EQ(arb::distance(mp{0,0,0},mp{0   , 0   , 0   }), 0.);
+    EXPECT_EQ(arb::distance(mp{0,0,0},mp{3.14, 0   , 0   }), 3.14);
+    EXPECT_EQ(arb::distance(mp{0,0,0},mp{0   , 3.14, 0   }), 3.14);
+    EXPECT_EQ(arb::distance(mp{0,0,0},mp{0   , 0   , 3.14}), 3.14);
+    EXPECT_EQ(arb::distance(mp{0,0,0},mp{1   , 2   , 3   }), std::sqrt(14.));
```
halfflat: This clearly works, but using a Pythagorean quadruple here would make the test less contingent on choice of fp precision used in the implementation,  or other floating point weirdosity.  e.g., `mp{7, 14, 22}` vs. `27`.

arbor/morph/em_morphology.cpp
halfflat: Yep; might need to do that anyway depending on decisions later, but we can worry about such a choice down the track and let things stand.

arbor/morph/region.cpp
bcumming: For they typical soma case, it grossly underestimates. For dendrite not. On the balance of things, I can't tell what is worse: over allocating on regions like soma, or making many small allocations when building dendrites? I would say pick an approach, and see if it is an issue in production.

bcumming: Added a shrink to fit.

arbor/morph/locset.cpp
bcumming: Done.

Issue #742, opened 2019-05-12
modcc: support calling functions inside functions


bcumming:
Currently functions can't be called within a function, because it messes with our function inlining. So, for example, the following would not work

```
FUNCTION sum(x, y) {
  sum = x+y
}
FUNCTION foo(x,y) {
  foo = x / sum(x,y)
}
```

Implementing this requires first implementing #741.

Then the function inlining in `module::semantic_func_proc` will have to become recursive, which is a bit more intrusive.

PR #861, opened 2019-09-04
Modcc: Inline functions containing if/else statements

### LINKED ISSUE #741: modcc: support inlining of multi-line functions
Modcc only supports inlining of single-statement functions, for example:
```
FUNCTION foo(x, y) {
    foo = x/(x+y)
}
```
The following isn't currently supported:
```
FUNCTION foo(x, y) {
  LOCAL denom
  denom = x+y
  foo = x/denom
}
```
This would require implementing two additional steps
1. renaming the local variables in the function so that they are unique w.r.t. variables in the caller.
2. inserting function statements before the final assignment into the calling site.

noraabiakar:
- Implement inlining functions with multiple statements
- Implement inlining nested function calls
- Implement inlining if/else statements. Compilation fails if a function fails to set the return value.

Fixes #741, #742, #862

### DIFF
--- a/modcc/expression.cpp
+++ b/modcc/expression.cpp
@@ -199,7 +199,6 @@ bool LocalDeclaration::add_variable(Token tok) {
 }
 
 void LocalDeclaration::semantic(scope_ptr scp) {
-//    std::cout << this->to_string() << std::endl;
     scope_ = scp;
 
     // loop over the variables declared in this LOCAL statement
--- a/modcc/expression.cpp
+++ b/modcc/expression.cpp
@@ -701,30 +701,32 @@ void FunctionExpression::semantic(scope_type::symbol_map &global_symbols) {
         if(lhs && lhs->name()==name()) {
             last_expr_is_assign = true;
         }
-    } else if(body()->back()->is_if()) {
+    } else if(auto e = body()->back()->is_if()) {
         last_expr_is_assign = true;
-        std::stack<expression_ptr> if_exp;
-        if_exp.push(body()->back()->is_if()->clone());
+        std::stack<IfExpression*> stack;
+        stack.push(e);
 
-        while (!if_exp.empty()) {
-            auto tail = std::move(if_exp.top());
-            if_exp.pop();
-            if (auto true_branch = tail->is_if()->true_branch()) {
-                if (auto last = true_branch->is_block()->statements().back()->is_assignment()) {
+        while (!stack.empty()) {
+            auto tail = stack.top();
+            stack.pop();
+            if (auto true_branch = tail->true_branch()) {
+                auto& last = true_branch->is_block()->statements().back();
+                if (last->is_assignment()){
                     auto lhs = last->is_assignment()->lhs()->is_identifier();
                     last_expr_is_assign &= (lhs && lhs->spelling()==name());
                 }
-                else if (auto last = true_branch->is_block()->statements().back()->is_if()) {
-                    if_exp.push(last->clone());
+                else if (last->is_if()) {
+                    stack.push(last->is_if());
                 }
             }
-            if (auto false_branch = tail->is_if()->false_branch()) {
-                if (auto last = false_branch->is_block()->statements().back()->is_assignment()) {
+            if (auto false_branch = tail->false_branch()) {
+                auto& last = false_branch->is_block()->statements().back();
+                if (last->is_assignment()) {
                     auto lhs = last->is_assignment()->lhs()->is_identifier();
                     last_expr_is_assign &= (lhs && lhs->spelling()==name());
                 }
-                else if (auto last = false_branch->is_block()->statements().back()->is_if()) {
-                    if_exp.push(last->clone());
+                else if (last->is_if()) {
+                    stack.push(last->is_if());
                 }
             }
         }
--- a/modcc/expression.cpp
+++ b/modcc/expression.cpp
@@ -1,5 +1,7 @@
 #include <cstring>
 
+#include <stack>
+
 #include "expression.hpp"
 #include "identifier.hpp"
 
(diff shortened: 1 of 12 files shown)


modcc/expression.cpp
```diff
@@ -690,18 +691,44 @@ void FunctionExpression::semantic(scope_type::symbol_map &global_symbols) {
...
         // we know that the tail is an assignment expression
         auto lhs = tail->lhs()->is_identifier();
         // use nullptr check followed by lazy name lookup
         if(lhs && lhs->name()==name()) {
             last_expr_is_assign = true;
         }
-    }
+    } else if(body()->back()->is_if()) {
+        last_expr_is_assign = true;
+        std::stack<expression_ptr> if_exp;
```
bcumming: I think that you can avoid the need for all of the clone and move below by using `std::stack<IfExpression*>`, which would also simplify later expressions...

```C++
else if (auto e = body()->back()->is_if()) {
  std::stack<IfExpression*> stack;
  stack.push(e);
  while (!stack.empty()) {
    auto tail = stack.top();
    stack.pop();
    if (auto branch = tail->true_branch()) {
        auto last = &branch->is_block->statements().back();
        if (last->is_assignment()) {
          // ...
        else if (last->is_if()) {
          stack.push(last->is_if());
        }
      // ...
    }
    ...
```

modcc/module.cpp
```diff
@@ -791,10 +791,10 @@ int Module::semantic_func_proc() {
                 //      ll1_ = 2+x
                 //      ll0_ = ll1_*(y + 1)
                 //      a = 2 + ll0_
-                for(auto e=b.begin(); e!=b.end(); ++e) {
-                    if(auto ass = (*e)->is_assignment()) {
+                for(auto& e: b) {
+                    if(auto ass = e->is_assignment()) {
                         if(ass->rhs()->is_function_call()) {
-                            ass->replace_rhs(inline_function_call(ass->rhs()));
+                            e = inline_function_call(e);
```
bcumming: `inline_function_call` will return either
* an assigment expression; or
* an if-else expression

Right?

modcc/functioninliner.cpp
```diff
@@ -76,9 +59,78 @@ expression_ptr inline_function_call(Expression* e)
...
+        } else if (false_branch.front()->is_assignment()) {
+            // Modify the false assignment
+            fix_expression(false_branch.front());
+        } else {
+            throw compiler_exception(
+                    "can only inline assignment expressions and if expressions containing single assignment expressions", exp->location()
+            );
+        }
+    }
+    else if (exp->is_assignment()) {
```
bcumming: How about putting the `is_assignement` case first, i.e. as the "true branch", because it is simpler and hopefully the more common case.

modcc/functioninliner.cpp
```diff
@@ -76,9 +59,78 @@ expression_ptr inline_function_call(Expression* e)
...
+        }
+    };
+
+    if (exp->is_if()) {
+        if (!exp->is_if()->false_branch()) {
+            throw compiler_exception(
+                    "can only inline if statements with associated else", exp->location()
+            );
+        }
+        // Modify the condition
```
bcumming: What does this comment mean?

modcc/functioninliner.cpp
```diff
@@ -76,9 +59,78 @@ expression_ptr inline_function_call(Expression* e)
...
+            // Modify the true assignment
+            fix_expression(true_branch.front());
+        } else {
+            throw compiler_exception(
+                    "can only inline assignment expressions and if expressions containing single assignment expressions", exp->location()
+            );
+        }
+
+        if (false_branch.front()->is_if()) {
+            replace_and_inline(false_branch.front(), lhs, fargs, cargs, scope);
```
bcumming: Does this recursively handle the case of multiple `else if` statements?

Are both of these cases now handled?

```
if (x) {
}
elif (y) {
}
elif (z) {
}
else {
}
```

```
if (x) {
  if (y) {
  }
  else {
  }
}
else {
}
```

modcc/expression.cpp
noraabiakar: Yes, that does look better. I'll change it.

modcc/module.cpp
noraabiakar: Right. It will take whatever expression is in the function (if more than one expression is in the body of the function it will return an error), it checks if the expression is an if expression or an assignment expression: if yes, it will modify the expression and return the inlined version; if no it will return an error.

modcc/functioninliner.cpp
noraabiakar: Done

modcc/functioninliner.cpp
noraabiakar: It means we are inlining the condition of the if condition (replacing the function arguments with the call arguments)
I can fix to : Modify the condition of the if statement / Substitute arguments in if condition ?

modcc/functioninliner.cpp
noraabiakar: The second case is handled. The first is not.

modcc/functioninliner.cpp
```diff
@@ -1,85 +1,157 @@
...
-        std::cout << "inline_function_call for statement " << f->to_string()
-                  << " with body" << func->body()->to_string() << "\n";
-#endif
-        auto& body = func->body()->statements();
-        if(body.size() != 1) {
-            throw compiler_exception(
-                "can only inline functions with one statement", func->location()
-            );
+    auto assign_to_func = e->is_assignment();
+    auto ret_name = assign_to_func->lhs()->is_identifier()->clone();
```
bcumming: Does `ret_name` need to be cloned? Is it modified below, or just used as an input below?

modcc/functioninliner.cpp
```diff
@@ -1,85 +1,157 @@
...
+            throw compiler_exception(pprintf("return variable of function % not set", f->name()), e->location());
         }
+        return body;
+    }
+    return {};
+}
+///////////////////////////////////////////////////////////////////////////////
+//  function inliner
+///////////////////////////////////////////////////////////////////////////////
+void FunctionInliner::replace_with_args(Expression* e) {
```
bcumming: Please add a comment above this function definition that describes what it does.

modcc/module.cpp
```diff
@@ -699,114 +699,135 @@ void Module::add_variables_to_symbols() {
...
+            auto& s = e.second;
+
+            if(s->kind() == symbolKind::procedure)
+            {
+                if(errors==0) {
+                    auto &b = s->kind()==symbolKind::function ?
+                        s->is_function()->body()->statements() :
+                        s->is_procedure()->body()->statements();
+
+                    // Do the inlining, which currently only works for functions
```
bcumming: I think that this comment is out of date after this PR.

modcc/functioninliner.cpp
noraabiakar: Done

modcc/functioninliner.cpp
noraabiakar: It's just used as an input, I will pass it as `Expression*`

modcc/module.cpp
noraabiakar: I updated it.

PR #865, opened 2019-09-13
Cable cell

bcumming:
First step of refactoring `cable_cell` API.

* Hide implementation and state using PIMPL
* Consistent `place` methods for adding synapse, stimuls, gap junction site and spike detectors to `cable_cell`.
* Add default constructor for `region` and `locset` that create `nil` instances.

This is very much "work in progress", with another follow up PR to:
* Remove non-const interface for querying and adding soma/cable to a cell
* Move `make_cable_cell` functionality into `cable_cell` constructor such that all "morphology description" is performed before instantiation of a cable cell
* Add missing region/locations DSL features to support all examples currently implemented in Arbor.

### DIFF
--- a/arbor/cable_cell.cpp
+++ b/arbor/cable_cell.cpp
@@ -31,56 +31,59 @@ struct cable_cell_impl {
     using locset_map = cable_cell::locset_map;
 
     cable_cell_impl() {
-        segments_.push_back(make_segment<placeholder_segment>());
-        parents_.push_back(0);
+        segments.push_back(make_segment<placeholder_segment>());
+        parents.push_back(0);
     }
 
     cable_cell_impl(const cable_cell_impl& other) {
-        parents_ = other.parents_;
-        stimuli_ = other.stimuli_;
-        synapses_ = other.synapses_;
-        gap_junction_sites_ = other.gap_junction_sites_;
+        parents = other.parents;
+        stimuli = other.stimuli;
+        synapses = other.synapses;
+        gap_junction_sites = other.gap_junction_sites;
         spike_detectors_ = other.spike_detectors_;
-        regions_ = other.regions_;
+        regions = other.regions;
 
         // unique_ptr's cannot be copy constructed, do a manual assignment
-        segments_.reserve(other.segments_.size());
-        for (const auto& s: other.segments_) {
-            segments_.push_back(s->clone());
+        segments.reserve(other.segments.size());
+        for (const auto& s: other.segments) {
+            segments.push_back(s->clone());
         }
     }
 
     cable_cell_impl(cable_cell_impl&& other) = default;
 
     // storage for connections
-    std::vector<index_type> parents_;
+    std::vector<index_type> parents;
 
     // the segments
-    std::vector<segment_ptr> segments_;
+    std::vector<segment_ptr> segments;
 
     // the stimuli
-    std::vector<stimulus_instance> stimuli_;
+    std::vector<stimulus_instance> stimuli;
 
     // the synapses
-    std::vector<synapse_instance> synapses_;
+    std::vector<synapse_instance> synapses;
 
     // the gap_junctions
-    std::vector<gap_junction_instance> gap_junction_sites_;
+    std::vector<gap_junction_instance> gap_junction_sites;
 
     // the sensors
     std::vector<detector_instance> spike_detectors_;
 
     // Named regions
-    region_map regions_;
+    region_map regions;
 
     // Named location sets
-    locset_map locsets_;
+    locset_map locations;
+
+    // Underlying embedded morphology
+    em_morphology morph;
 
     void assert_valid_segment(index_type i) const {
-        if (i>=segments_.size()) {
+        if (i>=segments.size()) {
             throw cable_cell_error("no such segment");
+        }
     }
-}
 };
 
 using impl_ptr = std::unique_ptr<cable_cell_impl, void (*)(cable_cell_impl*)>;
@@ -98,7 +101,7 @@ cable_cell::cable_cell(const cable_cell& other):
 {}
 
 size_type cable_cell::num_segments() const {
-    return impl_->segments_.size();
+    return impl_->segments.size();
 }
 
 //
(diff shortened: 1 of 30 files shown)


arbor/cable_cell.cpp
```diff
@@ -54,41 +148,210 @@ cable_segment* cable_cell::add_cable(index_type parent, segment_ptr&& cable) {
...
+// Synapses
+//
+
+locrange cable_cell::place(const std::string& target, const mechanism_desc& desc) {
+    const auto first = impl_->synapses.size();
+
+    const auto it = impl_->locations.find(target);
+    if (it==impl_->locations.end()) return locrange(first, first);
+
+    return impl_->place(it->second, desc, impl_->synapses);
```
halfflat: Some boiler plate code here and below could be removed by defaulting to an empty `mlocation_list` if `target` is not found, e.g.
```
    const auto it = impl_->locations.find(target);
    return impl_->place(it->first!=impl_->locations.end()? it->second: mlocation_list{}, desc, impl_->synapses);
```
Or using `util::value_by_key` from `util/maputil.hpp`
```
    return impl_->place(util::value_by_key(impl_->locations, target).value_or({}), desc, impl_->synapses);
```

arbor/cable_cell.cpp
```diff
@@ -54,41 +148,210 @@ cable_segment* cable_cell::add_cable(index_type parent, segment_ptr&& cable) {
...
+    const auto locs = thingify(ls, impl_->morph);
+    return impl_->place(locs, desc, impl_->synapses);
+}
+*/
+
+locrange cable_cell::place(const mlocation& loc, const mechanism_desc& desc) {
+    if (!test_invariants(loc) || loc.branch>=num_segments()) {
+        throw cable_cell_error(util::pprintf(
+            "Attempt to add synapse at invalid location: \"{}\"", loc));
+    }
```
halfflat: Until there's always an embedded morphology to check against, could these location validity tests be provided by a single routine in `impl_`, again to avoid some boilerplate? e.g. via a new member function `assert_valid_location`.

arbor/include/arbor/cable_cell.hpp
```diff
@@ -16,22 +17,17 @@
...
-    {}
+// Pair of indexes that describe range of local indices.
+// Returned by cable_cell::place() calls, so that the caller can
+// refer to targets, detectors, etc on the cell.
+struct locrange {
+    cell_lid_type first;
+    cell_lid_type last;
+    locrange(cell_lid_type first, cell_lid_type last):
+        first(first), last(last) {}
 };
```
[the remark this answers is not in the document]
bcumming: I don't like using `std::pair` for public API code. `second` is even more misleading than `last`.

I have changed it to `lid_range`, and using `begin` and `end` instead of `first` and `last`.

arbor/include/arbor/cable_cell_param.hpp
```diff
@@ -16,6 +16,30 @@ struct cable_cell_error: arbor_exception {
...
+    i_clamp() = default;
+
+    i_clamp(value_type delay, value_type duration, value_type amplitude):
+        delay(delay), duration(duration), amplitude(amplitude)
+    {}
+};
+
+// Threshold detector description.
+struct detector {
+    double threshold;
```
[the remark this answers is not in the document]
bcumming: I went with `threshold_detector`.

arbor/cable_cell.cpp
```diff
@@ -17,20 +17,114 @@ using value_type = cable_cell::value_type;
...
-    if (i>=num_segments()) {
-        throw cable_cell_error("no such segment");
+    cable_cell_impl(const cable_cell_impl& other) {
+        parents = other.parents;
+        stimuli = other.stimuli;
+        synapses = other.synapses;
+        gap_junction_sites = other.gap_junction_sites;
+        spike_detectors = other.spike_detectors;
+        regions = other.regions;
+
```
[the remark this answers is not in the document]
bcumming: I just do it manually for now, and then when we remove the `segments` we will be able to use the default copy constructor.

arbor/cable_cell.cpp
```diff
@@ -17,20 +17,114 @@ using value_type = cable_cell::value_type;
...
+    locset_map locations;
+
+    // Underlying embedded morphology
+    em_morphology morph;
+
+    template <typename Desc, typename List>
+    locrange place(const mlocation_list& locs, const Desc& desc, List& list) {
+        const auto first = list.size();
+
+        stimuli.reserve(first+locs.size());
```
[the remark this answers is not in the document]
bcumming: Ha! The type of cut and paste error I was trying to avoid with this helper.
Done.

arbor/cable_cell.cpp
```diff
@@ -17,20 +17,114 @@ using value_type = cable_cell::value_type;
...
+    region_map regions;
+
+    // Named location sets
+    locset_map locations;
+
+    // Underlying embedded morphology
+    em_morphology morph;
+
+    template <typename Desc, typename List>
+    locrange place(const mlocation_list& locs, const Desc& desc, List& list) {
```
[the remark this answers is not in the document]
bcumming: Sure.

arbor/cable_cell.cpp
```diff
@@ -17,20 +17,114 @@ using value_type = cable_cell::value_type;
...
+    }
+
+    locrange place_gj(const mlocation_list& locs) {
+        const auto first = gap_junction_sites.size();
+
+        gap_junction_sites.reserve(first+locs.size());
+        for (auto loc: locs) {
+            gap_junction_sites.push_back(loc);
+        }
+
```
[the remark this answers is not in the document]
bcumming: Indeed. Done.

arbor/cable_cell.cpp
bcumming: It doesn't save much, because we have a custom message for each `place` call.
I have added a `bool valid_location(mlocation)` function to perform the test.

arbor/cable_cell.cpp
bcumming: I find the current, slightly more verbose version much easier to read. And faster (not that that matters in this context).

arbor/cable_cell.cpp
halfflat: I don't think we need a different message for each type of place error though.

arbor/cable_cell.cpp
halfflat: Note for the record: had a chat in person re: `loc_range(first, first)` semantics as distinct from a pure empty range. Meaning will be clarified in documentation to come later.

arbor/cable_cell.cpp
halfflat: Another note: all of this code will change later with generic expressions for place calls, so no need to sweat details about error messages now.

PR #878, opened 2019-09-27
CV policies with unit tests.

halfflat:
* Add new `cv_policy` class that wraps classes derived from `cv_policy_base`, which in turn provide a facility for producing CV boundaries on morphology trees from a `cable_cell`.
* Implement fixed-CVs-per-branch and maximum-CV-length policies, with flags that allow for single-CV oots and for CVs that terminate at fork points or which are centred on fork points.
* Expose locset `join` expressions.
* Add `empty()` test method on `morphology`.
* Expose branch length on `em_morphology` via member function.
* Account for the length of a spherical root branch in `em_morphology`.
* Test for validity but do not canonicalize locations when thingifying.
* Rename `root_sample_has_same_tag_as_child` to `root_sample_tag_differs_from_children` to better reflect sense of test.
* Add unit tests for CV policies, branch_length method.

### DIFF
--- a/arbor/cable_cell_param.cpp
+++ b/arbor/cable_cell_param.cpp
@@ -1,8 +1,13 @@
 #include <cfloat>
 #include <cmath>
+#include <numeric>
+#include <vector>
 
+#include <arbor/cable_cell.hpp>
 #include <arbor/cable_cell_param.hpp>
+#include <arbor/morph/locset.hpp>
 
+#include "morph/em_morphology.hpp"
 #include "util/maputil.hpp"
 
 namespace arb {
@@ -63,5 +68,75 @@ cable_cell_local_parameter_set neuron_parameter_defaults = {
     // membrane capacitance [F/m²]
     0.01
 };
+
+// Discretization policy implementations:
+
+locset cv_policy_max_extent::cv_boundary_points(const cable_cell& cell) const {
+    const auto& emorph = *cell.morphology();
+    const unsigned nbranch = emorph.morph().num_branches();
+    if (!nbranch || max_extent_<=0) return ls::nil();
+
+    std::vector<mlocation> points;
+
+    unsigned bidx = 0;
+    if (flags_&cv_policy_flag::single_root_cv) {
+        points.push_back({0, 0.});
+        points.push_back({0, 1.});
+        bidx = 1;
+    }
+
+    const double oomax_extent = 1./max_extent_;
+    while (bidx<nbranch) {
+        unsigned ncv = std::ceil(emorph.branch_length(bidx)*oomax_extent);
+        double ooncv = 1./ncv;
+
+        if (flags_&cv_policy_flag::interior_forks) {
+            for (unsigned i = 0; i<ncv; ++i) {
+                points.push_back({bidx, (1+2*i)*ooncv/2});
+            }
+        }
+        else {
+            for (unsigned i = 0; i<ncv; ++i) {
+                points.push_back({bidx, i*ooncv});
+            }
+            points.push_back({bidx, 1.});
+        }
+        ++bidx;
+    }
+
+    return std::accumulate(points.begin(), points.end(), ls::nil(), [](auto& l, auto& p) { return join(l, ls::location(p)); });
+}
+
+locset cv_policy_fixed_per_branch::cv_boundary_points(const cable_cell& cell) const {
+    const unsigned nbranch = cell.morphology()->morph().num_branches();
+    if (!nbranch) return ls::nil();
+
+    std::vector<mlocation> points;
+
+    unsigned bidx = 0;
+    if (flags_&cv_policy_flag::single_root_cv) {
+        points.push_back({0, 0.});
+        points.push_back({0, 1.});
+        bidx = 1;
+    }
+
+    double ooncv = 1./cv_per_branch_;
+    while (bidx<nbranch) {
+        if (flags_&cv_policy_flag::interior_forks) {
+            for (unsigned i = 0; i<cv_per_branch_; ++i) {
+                points.push_back({bidx, (1+2*i)*ooncv/2});
+            }
+        }
+        else {
+            for (unsigned i = 0; i<cv_per_branch_; ++i) {
+                points.push_back({bidx, i*ooncv});
+            }
+            points.push_back({bidx, 1.});
+        }
+        ++bidx;
+    }
+
+    return std::accumulate(points.begin(), points.end(), ls::nil(), [](auto& l, auto& p) { return join(l, ls::location(p)); });
+}
 
 } // namespace arb
--- a/arbor/cable_cell_param.cpp
+++ b/arbor/cable_cell_param.cpp
@@ -73,7 +73,7 @@ cable_cell_local_parameter_set neuron_parameter_defaults = {
 
 locset cv_policy_max_extent::cv_boundary_points(const cable_cell& cell) const {
     const auto& emorph = *cell.morphology();
-    const unsigned nbranch = emorph.morph().num_branches();
+    const unsigned nbranch = emorph.num_branches();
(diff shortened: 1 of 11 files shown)


arbor/morph/em_morphology.cpp
```diff
@@ -40,6 +40,9 @@ em_morphology::em_morphology(const morphology& m):
         auto idx = util::make_range(morph_.branch_indexes(i));
         branch_lengths_.push_back(dist2root_[idx.back()]- dist2root_[idx.front()]);
     }
+    if (morph_.spherical_root()) {
```
bcumming: I am treating it as the distance from the center, where `0,0` is the centre of the sphere, not an edge.

arbor/cable_cell_param.cpp
```diff
@@ -64,4 +69,74 @@ cable_cell_local_parameter_set neuron_parameter_defaults = {
...
+
+locset cv_policy_max_extent::cv_boundary_points(const cable_cell& cell) const {
+    const auto& emorph = *cell.morphology();
+    const unsigned nbranch = emorph.morph().num_branches();
+    if (!nbranch || max_extent_<=0) return ls::nil();
+
+    std::vector<mlocation> points;
+
+    unsigned bidx = 0;
+    if (flags_&cv_policy_flag::single_root_cv) {
```
bcumming: Would this generate a correct discretisation if `single_root_cv` was set without a spherical root?

arbor/include/arbor/cable_cell_param.hpp
```diff
@@ -127,11 +220,18 @@ struct cable_cell_local_parameter_set {
...
+    cv_policy discretization = default_cv_policy();
 
     // We'll need something like this until C++17, for sane initialization syntax.
     cable_cell_parameter_set() = default;
-    cable_cell_parameter_set(cable_cell_local_parameter_set p, std::unordered_map<std::string, mechanism_desc> m = {}):
-        cable_cell_local_parameter_set(std::move(p)), reversal_potential_method(std::move(m))
+    cable_cell_parameter_set(
+        cable_cell_local_parameter_set p,
+        std::unordered_map<std::string, mechanism_desc> m = {},
+        cv_policy d = default_cv_policy()
```
bcumming: Can we apply a policy to a cell, then specialise it on a subset of branches?

arbor/cable_cell_param.cpp
```diff
@@ -64,4 +69,74 @@ cable_cell_local_parameter_set neuron_parameter_defaults = {
...
+}
+
+locset cv_policy_fixed_per_branch::cv_boundary_points(const cable_cell& cell) const {
+    const unsigned nbranch = cell.morphology()->morph().num_branches();
+    if (!nbranch) return ls::nil();
+
+    std::vector<mlocation> points;
+
+    unsigned bidx = 0;
+    if (flags_&cv_policy_flag::single_root_cv) {
```
bcumming: Same here with the spherical root.

arbor/include/arbor/cable_cell_param.hpp
```diff
@@ -103,6 +104,98 @@ struct mechanism_desc {
...
+namespace cv_policy_flag {
+    using value = unsigned;
+    enum : unsigned {
+        none = 0,
+        interior_forks = 1<<0,
+        single_root_cv = 1<<1
+    };
+}
+
+struct cv_policy_explicit: cv_policy_base {
```
bcumming: A short description of what each policy does would be good.

arbor/include/arbor/cable_cell_param.hpp
```diff
@@ -103,6 +104,98 @@ struct mechanism_desc {
...
+    }
+
+private:
+    cv_policy_base_ptr policy_ptr;
+};
+
+// Common flags for CV policies; bitwise composable.
+namespace cv_policy_flag {
+    using value = unsigned;
+    enum : unsigned {
```
bcumming: missing documentation for how each flag is interpreted.

arbor/cable_cell_param.cpp
```diff
@@ -64,4 +69,74 @@ cable_cell_local_parameter_set neuron_parameter_defaults = {
     0.01
 };
 
+// Discretization policy implementations:
+
+locset cv_policy_max_extent::cv_boundary_points(const cable_cell& cell) const {
+    const auto& emorph = *cell.morphology();
+    const unsigned nbranch = emorph.morph().num_branches();
```
bcumming: The current `em_morp` vs. `morph` interface can be a bit verbose and confusing, e.g. `emporhp.morph().num_branches()`.
Any ideas how to improve this?

arbor/morph/em_morphology.cpp
halfflat: Yup, but it's inconsistent: the branch length of a non-spherical root is indeed the full length of the cable. We have other mechanisms for getting the radius of the root node, and the branch length should consistently return the length of the branch.

arbor/cable_cell_param.cpp
halfflat: The issue of whether or not to deal specially with a spherical root CV at discretization has to be pushed down to the actual discretizer, as an explicit discretization, for example, may include points on the root branch that are interior also.

So the locsets produced by CV policies should be agnostic when it comes to spherical or cable root branches, and the decision regarding indivisible spherical somata deferred to later.

arbor/include/arbor/cable_cell_param.hpp
halfflat: While in principle we could make something like that, it would be quite complex in terms of semantics, given that CVs may cross one or more fork points. I'd recommend that we don't offer such a functionality unless there is a clear need.

As the result of the application of a CV policy is a locset, that set can be amended by user code to cover some special cases. That is, the user code can apply a policty, edit the result, then supply that as an explicit set of CV boundaries.

arbor/include/arbor/cable_cell_param.hpp
halfflat: Sure; I'll add some comments for policies and flags; I'll supply full docs later when the whole thing is done.

arbor/cable_cell_param.cpp
halfflat: We could have `em_morphology` derive from `morphology`, but I don't like the strong coupling that entails. Or have a number of methods that just forward queries through to the `em_morphology`.

arbor/morph/em_morphology.cpp
bcumming: The spherical root is a special case, and I feel uncomfortable with this approach because it attempts to "orient" the branch.
You can also construct inconsistent interpretations from this, e.g. the distance between the sample 0 and sample 1, which is at the start of the next branch, is now `length(branch 0)/2`.

arbor/include/arbor/cable_cell_param.hpp
bcumming: Right.

arbor/cable_cell_param.cpp
bcumming: I also prefer the second option.

arbor/morph/em_morphology.cpp
halfflat: On the second point first, the samples define the geometry, not the other way around: there is nothing inconsistent about the distance between the sample 0 and sample 1 not being the length of the branch, if the branch is a sphere.

Spherical somata *are* oriented: one end has all the other branches sticking out of it, and the other end doesn't. Those branches share a common end point, just as they would if they were attached to a non-spherical cable root with the same topology.

halfflat: But none of that changes the fact that geometrically, the length of a cable shaped like a sphere should be its diameter, not its radius. If I were to substitute the spherical root with an equivalent cable, the length should not change.

Issue #837, opened 2019-08-14
modcc: Implement solve for kinectic schemes at steady-state


noraabiakar:
Currently, we only allow solve statements with the following formats: 
```
SOLVE x METHOD method
or 
SOLVE x
```
For kinetic schemes, we can choose to solve the system at steady state, provided there is a conserve statement. This is done using: 
```
SOLVE x STEADYSTATE sparse
```
Example:
 ```
KINETIC init {
    ~ s <-> h (alpha1, beta1)
    ~ d <-> s (alpha2, beta2)
    CONSERVE s + d + h = 1
}
```
solve using **METHOD**:
```
(dt*-alpha1)*s + (1 + dt*beta1)*h +             (0)*d = h_prev
 (dt*-beta2)*s +            (0)*h + (1 + dt*alpha2)*d = d_prev
         (1)*s +            (1)*h +             (1)*d = 1
```
solve using **STEADYSTATE**:
```
alpha1*s - beta1*h +    (0)*d = 0
 beta2*s +   (0)*h - alpha2*d = 0
       s +       h +        d = 1
```

halfflat:
Implemented in #879.

PR #879, opened 2019-10-01
Modcc: Steady State

noraabiakar:
Modifies `SparseSolverVisitor` to allow solving kinetic equations at steady state. 
Adds new enum class `SolverVariant` to differentiate between regular and steady state solutions.
Includes `CONSERVE` statement support.

Addresses #837


modcc/solvers.cpp
```diff
@@ -254,27 +264,26 @@ void SparseSolverVisitor::visit(AssignmentExpression *e) {
         // Otherwise, for non-zero coefficient c, the entry is -c*dt.
 
         if (r.coef.count(dvars_[j])) {
-            expr = make_expression<MulBinaryExpression>(loc,
-                       r.coef[dvars_[j]]->clone(),
-                       dt_expr->clone());
+            expr = steadystate_ ? r.coef[dvars_[j]]->clone() :
+                    make_expression<MulBinaryExpression>(loc, r.coef[dvars_[j]]->clone(), dt_expr->clone());
```
halfflat: Could you expand the comment above (lines 262–264) to describe the steady-state case?

noraabiakar: Yep. Thanks for pointing it out.

Issue #850, opened 2019-08-26
Magic 'diam' variable for NMODL support


halfflat:
Until we have a better option for supporting NMODL-based mechanisms with explicit radial diffusion, make a 'diam' variable visible to NMODL mechanisms that represents the mean diameter across a CV.

PR #880, opened 2019-10-02
Expose CV diam to mechanisms

noraabiakar:
Sets the diameter of a CV as the diameter at the middle of a CV
Exposes the diameter in `shared_state` and `mechanism`
Adds support in modcc to treat `diam` as a special parameter 
Implements unit tests

Addresses #850


test/unit/mod/diam_test.mod
```diff
@@ -0,0 +1,27 @@
+NEURON {
+    SUFFIX diam_test
+}
+
+PARAMETER {
+    diam
+}
```
bcumming: Should we permit `diam` as a `PARAMETER`? I see that `diam` is removed from the parameter list in `modcc/module.cpp`.

What does NMODL do if a user creates an ASSIGNED or PARAMETER called `diam` and sets  its value explicitly? Will that override the real value? If NEURON doesn't need these magic variables to be explicitly declared, then we could enforce that they are not without forcing users to write NMODL that is incompatible with NEURON.

Issue #881, opened 2019-10-02
Modcc: non-linear kinetic schemes


noraabiakar:
Currently, we only have support for linear kinetic schemes. We can expand our sparse solver to handle non-linear systems. 

We can solve this using backward Euler and Newton's method: 

Let *x* be the vector of state variables. Let *G(x)* be a non-linear function of *x*

The kinetic scheme can be represented as: 
`x' = G(x)`

Backward Euler gives: 
<code>F(x<sub>t+1</sub>) = x<sub>t+1</sub> - x<sub>t</sub> - dt * G(x<sub>t+1</sub>) = 0
</code>

The Jacobian *J(x)* of *F(x)*: 
<code>J(x) = I - dt * &#8706;G/&#8706;x 
</code>

We can then iterate using Newton's method: 
<code>x<sub>n+1</sub> = x<sub>n</sub> - J(x<sub>n</sub>)<sup>-1</sup>*F(x<sub>n</sub>)
</code>

We can hard-code the number of Newton's iterations for the time being. 
And generate the code for every iteration already unrolled as opposed to a while/for loop.

noraabiakar:
`COMPARTMENT` statements `c[x]` scale the `F(x)` and the `J(x)` functions:

<code>F(x<sub>t+1</sub>) = x<sub>t+1</sub> - x<sub>t</sub> - (dt * G(x<sub>t+1</sub>))/c[x] = 0
</code>

<code>J(x) = I - dt/c[x] * &#8706;G/&#8706;x 
</code>

PR #883, opened 2019-10-02
Modcc: nonlinear kinetic schemes

noraabiakar:
- Adds support for solving non-linear kinetic schemes.
- Implements new solver: `SparseNonlinearSolverVisitor` that uses backward Euler and 3 iterations of Newton's method to update the state variables.
- Includes support for `COMPARTMENT` statements.
- Does not include support for steady-state solutions
- Implements unit tests

Addresses #881


modcc/solvers.cpp
```diff
@@ -528,6 +529,258 @@ void LinearSolverVisitor::finalize() {
...
+
+    // Form and save F(x) = x(t+1) - x(t) - dt(G(x(t+1))
+    // x(t+1) are stored in dvar_temp_ and are updated at every iteration of Newton's method
+    // x(t)   are stored in dvar_init_ and are constant across iterations of Newton's method
+    // G(x)   is the rhs of the derivative assignment expression
+
+    expression_ptr F_x;
+    F_x = make_expression<MulBinaryExpression>(loc, expanded_rhs->clone(), dt_expr->clone());
+    if (scale_factor_[deq_index_]) {
+        F_x = make_expression<DivBinaryExpression>(loc, std::move(F_x), scale_factor_[deq_index_]->clone());
```
halfflat: There's the possibility of a small improvement in performance if the reciprocals of the scale factors are assigned to associated variables, which are multiplied here, rather than taking the division each time.

modcc/solvers.cpp
```diff
@@ -528,6 +529,258 @@ void LinearSolverVisitor::finalize() {
...
+    if (s!=dvars_[deq_index_]) {
+        error({"ICE: inconsistent ordering of derivative assignments", loc});
+        return;
+    }
+
+    auto dt_expr = make_expression<IdentifierExpression>(loc, "dt");
+    auto one_expr = make_expression<NumberExpression>(loc, 1.0);
+
+    // Form and save F(x) = x(t+1) - x(t) - dt(G(x(t+1))
+    // x(t+1) are stored in dvar_temp_ and are updated at every iteration of Newton's method
```
halfflat: Definition of `F` in the comment is just a little unclear; should probable be
```
// […] F(x(t+1)) = x(t+1) - x(t) - dt G(x(t+1))
```
or e.g.
```
//  […] F(y) = y - x(t) - dt G(y)
```
with the other comments amended to indicate that the y are the `dvar_temp_` variables etc., and Newton's method is used to find x(t+1) = y s.t. F(y) = 0.

halfflat: Oh, and given the tricky nature of the scale factors, they should probably also be mentioned in the comment.

modcc/solvers.cpp
```diff
@@ -528,6 +529,258 @@ void LinearSolverVisitor::finalize() {
...
+    for (unsigned i = 0; i < dvars_.size(); ++i) {
+        auto expr = make_expression<AssignmentExpression>(loc,
+                        make_expression<IdentifierExpression>(loc, dvars_[i]),
+                            make_expression<IdentifierExpression>(loc, dvar_temp_[i]));
+        statements_.push_back(std::move(expr));
+    }
+
+    BlockRewriterBase::finalize();
+}
+
```
halfflat: Perhaps for a different PR, but what do you think about using an explicit step y₀ ← x(t) + δt G(x(t)) for the first approximation in the Newton process?

test/unit/test_kinetic_linear.cpp
```diff
@@ -88,89 +89,111 @@ void run_test(std::string mech_name,
...
+
+TEST(mech_kinetic, kintetic_nonlinear) {
+    std::vector<std::string> state_variables = {"a", "b", "c"};
+    std::vector<fvm_value_type> t0_values = {0.2, 0.3, 0.5};
+    std::vector<fvm_value_type> t1_0_values = {0.222881, 0.31144, 0.48856};
+    std::vector<fvm_value_type> t1_1_values = {0.2078873133, 0.34222075, 0.45777925};
+
+    run_test<gpu::backend>("test2_kin_diff", state_variables, {}, t0_values, t1_0_values, 0.025);
+    run_test<gpu::backend>("test3_kin_diff", state_variables, {}, t0_values, t1_1_values, 0.025);
 }
```
halfflat: Again, given the grief from COMPARTMENT scaling up to now, can you add a test for a kinetic scheme with two different, interacting, COMPARTMENTs?

modcc/solvers.cpp
noraabiakar: I agree, I had initially assumed these would typically be identifier expressions, but that's not necessarily true.

modcc/solvers.cpp
noraabiakar: I will modify the comments.

modcc/solvers.cpp
noraabiakar: I don't mind that. I could add it now or we can keep it in mind if we need it later?

test/unit/test_kinetic_linear.cpp
noraabiakar: Yes, I think that's a good idea, I will add a new test

modcc/solvers.cpp
halfflat: Let's just keep it in mind — it might be a way of keeping a lower bound on the number of Newton iterations, but it's something to be experimented with in the future IMO.

PR #891, opened 2019-10-23
Modcc: normalize Gaussian elimiation in solvers for large systems

### LINKED ISSUE #890: Modcc: normalize after gaussian elimination
When systems get too large, the gaussian reduction method used in the solvers starts producing `nan` because the products become too large. 
Fix: normalize the elements in a row by multiplying by the (safe) inverse of the largest absolute value of the elements in the row.

noraabiakar:
Modifies `SparseSolverVisitor`, `SparseNonlinearSolverVisitor`, and `LinearSolverVisitor`
Adds option to normalize rows in a matrix by multiplying by the (safe) inverse of the largest absolute value of the elements in a given row. 
This option kicks in when there are more than 5 rows in a matrix.

Fixes #890


modcc/solvers.cpp
```diff
@@ -499,6 +536,43 @@ void LinearSolverVisitor::finalize() {
...
+
+            // Update the row elements
+            for (auto &elem: row_elems) {
+                auto ratio = make_expression<MulBinaryExpression>(elem->location(), elem->clone(), std::move(local_inv_term.id));
+                auto assign = make_expression<AssignmentExpression>(elem->location(), std::move(elem), std::move(ratio));
+                statements_.push_back(std::move(assign));
+            }
+        }
+    }
+
```
halfflat: Now that the Gaussian elimination, divide and assign code in the finalize() methods for  LinearSolverVisitor,  SparseSolverVisitor and SparseNonlinearSolverVisitor are getting quite large, and share a lot of common code, could that code be factored out into its own routine?

modcc/symge.cpp
```diff
@@ -123,12 +125,14 @@ void gj_reduce(sym_matrix& A, symbol_table& table) {
...
             if (i==p.row || A[i].index(p.col)==msparse::row_npos) continue;
             A[i] = row_reduce(p.col, A[i], A[p.row], define_sym);
+            row_partition.push_back(row_partition.back() + A[i].size());
         }
 
         if (remaining_rows.empty()) {
             break;
         }
     }
+    return row_partition;
```
halfflat: `row_partition` content seems to depend rather crucially on the details of how row `row_reduce` makes new symbols, and so it feels a bit brittle. Is it possible for the normalization code to just walk the rows of the matrix `A_` itself? I think it will make the code both clearer and more robust.

arbor/include/arbor/math.hpp
```diff
@@ -116,6 +116,14 @@ C round_up(T v, U b) {
     return v-m+signum(m)*impl::abs_if_signed(b, Signed{});
 }
 
+// Returns 1/x if x != 0; 0 otherwise
+template <typename T>
+inline
+T safeinv(T x) {
+    // If abs(x) is less than epsilon return 1, else calculate the result directly.
```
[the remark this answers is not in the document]
noraabiakar: Done

modcc/solvers.cpp
```diff
@@ -407,6 +407,43 @@ void SparseSolverVisitor::finalize() {
...
+            statements_.push_back(std::move(local_inv_term.local_decl));
+            statements_.push_back(std::move(local_inv_term.assignment));
+
+            // Update the row elements
+            for (auto &elem: row_elems) {
+                auto ratio = make_expression<MulBinaryExpression>(elem->location(), elem->clone(), std::move(local_inv_term.id));
+                auto assign = make_expression<AssignmentExpression>(elem->location(), std::move(elem), std::move(ratio));
+                statements_.push_back(std::move(assign));
+            }
```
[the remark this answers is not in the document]
noraabiakar: Done

modcc/solvers.cpp
noraabiakar: I was thinking about how to do that. Let me know what you think of the implementation.

modcc/symge.cpp
noraabiakar: Yes, I agree. It would be clearer.

modcc/solvers.hpp
```diff
@@ -83,6 +82,28 @@ class SparseSolverVisitor : public SolverVisitorBase {
...
+        local_expr_.clear();
+        A_.clear();
+        symtbl_.clear();
+        SolverVisitorBase::reset();
+    }
+
+    virtual std::vector<expression_ptr> GenerateReduceStatements();
+    virtual std::vector<expression_ptr> GenerateUpdateStatements(std::vector<std::string>);
+};
+
```
halfflat: I think this would be better implemented as a class outside of the Visitor hierarchy: with the exception of `reset()`, it's not providing any implementations of Visitor or SolverVisitorBase methods, and neither are the `Generate...()` overridden by its derived classes.

It could be implemented as a separate class, used in the solvers as a member, or as a mix-in, which is inherited by the solvers.

In terms of isolation or interface, there a few issues:
* `deq_index_` and `local_expr_` are only used in the solvers, and so shouldn't be in this class.
* By convention in the modcc code, method names are given in lower_case_with_underscores; this convention should be adopted for `GenerateReduceStatements()` etc.
* `GenerateReduceStatements()` both writes to `statements_` _and_ returns a list of expressions to be used by the caller — this makes for a more complex and difficult to understand interface. Given that the temporaries have to be declared as local once, but assigned to multiple times in the nonlinear solver, it is probably best to have a method for generating the local declarations (without assignments), and another which returns the list of statements for the reduction.
*  `GenerateUpdateStatements()` — I'm being a bit picky here, but it's not really clear to me from the name of the function or the name of the parameter what it actually does. How about something like `generate_solution_assignments(std::vector<std::string> lhs_vars)`?

In terms of the implementation (and I didn't see this before), aren't the rescalings happening too late? The temporary fill-ins will grow in magnitude through the series of assignments generated from the symbol table (`modcc/solvers.cpp`:185). It may still be that the easiest way to get the scaling happening at the right place is to extend the functionality of symge, allowing symbols to be bound to `symbol_term_diff` expressions _or_ a new symbol reciprocal expression _or_ explicitly 1. Before pivoting, we could optionally scale the row: make a new sym for the reciprocal of the pivot element, and replace the row with terms that are the product of the old elements and this reciprocal.

modcc/symge.cpp
```diff
@@ -80,7 +80,10 @@ double estimate_cost(const sym_matrix& A, pivot p) {
 // The reduction is division-free: the result will have non-zero terms
 // that are symbols that are either primitive, or defined (in the symbol
 // table) as products or differences of products of other symbols.
-void gj_reduce(sym_matrix& A, symbol_table& table) {
+// Returns a vector of vectors of symbols, partitioned by row of the matrix
+std::vector<std::vector<symge::symbol>> gj_reduce(sym_matrix& A, symbol_table& table) {
```
halfflat: Just to clarify, is it that each element of the returned vector corresponds to one row reduction operation, and contains the symbols in the reduced row?

modcc/solvers.cpp
```diff
@@ -171,6 +170,84 @@ static expression_ptr as_expression(symge::symbol_term_diff diff) {
...
+            max = make_expression<MaxBinaryExpression>(elem->location(), max->clone(), std::move(abs));
+        }
+    }
+    // Safely invert max
+    auto inv = make_expression<SafeInvUnaryExpression>(max->location(), std::move(max));
+
+    // Create a symbol for inv
+    auto local_inv_term = make_unique_local_assign(scope, inv.get(), "inv_");
+
+    return std::move(local_inv_term);
```
halfflat: gcc just warned me that the `std::move` here prevents copy-elision; we can just say `return local_inv_term;`.

modcc/symge.cpp
noraabiakar: Correct.

modcc/solvers.cpp
noraabiakar: Okay, I'll do that

PR #893, opened 2019-10-23
Modcc: Allow solving kinetic schemes at steady-state from intial block

### LINKED ISSUE #892: Modcc: Allow solving Kinetic schemes at steady-state from the `INITIAL` block in nmodl
Currently, we only allow solving Linear systems from the `INTIAL` block.

noraabiakar:
Fixes #892

### DIFF
--- a/modcc/module.cpp
+++ b/modcc/module.cpp
@@ -277,6 +277,9 @@ bool Module::semantic() {
     auto initial_api = make_empty_api_method("nrn_init", "initial");
     auto api_init  = initial_api.first;
     auto proc_init = initial_api.second;
+
+    scope_ptr nrn_init_scope = api_init->scope();
+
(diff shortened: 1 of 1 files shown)


modcc/module.cpp
```diff
@@ -277,24 +277,32 @@ bool Module::semantic() {
     auto initial_api = make_empty_api_method("nrn_init", "initial");
     auto api_init  = initial_api.first;
     auto proc_init = initial_api.second;
+
+    scope_ptr nrn_init_scope = api_init->scope();
```
bcumming: It doesn't look like `nrn_init_scope` is used anywhere. Are you creating it because you need side-effects of the call to `api_init->scope()`?

noraabiakar: I think it was a mistake. Will remove it.

PR #897, opened 2019-10-23
Reinitialize mechanisms after updating the ion state

### LINKED ISSUE #896: Reintialize mechanisms after ion write
**Scenario 1:**
Assume mechanism `A` updates ion concentrations in its `INITIAL` block, while another mechanism `B` reads these ion concentrations in its `INITIAL` block.
Currently, mechanisms are initialized prior to writing any ion concentrations. So, mechanism `B` ends up reading incorrect values of the ion concentrations in its `INITIAL` block. To circumvent that, we can initialize mechanisms twice: 
- First initialize: `A` updates ion concentrations, `B` reads incorrect ion concentrations
- Write ions: `A`'s updated ion concentrations are written in arbor
- Second initialize: `B` reads correct ion concentrations. 

**Scenario 2:**
Assume mechanism `A` updates ion concentrations in its `BREAKPOINT` block, while another mechanism `B` reads these ion concentrations in its `INTIAL` block.
This will cause an error. `B` will read incorrect ion concentrations even if we initialize the mechanisms twice similar to scenario 1. 
We will have an error even if the ion concentrations are given an initial value in the arbor simulation.
There is no _simple_ way around this.

noraabiakar:
If ion concentrations are read and written in the `INITIAL` blocks of mechanisms, this ensures that the reading mechanisms will receive the correct values. 
Fixes #896

### DIFF
--- a/arbor/fvm_lowered_cell_impl.hpp
+++ b/arbor/fvm_lowered_cell_impl.hpp
@@ -433,7 +433,7 @@ void fvm_lowered_cell_impl<B>::initialize(
         const std::string& ion_name = i.first;
 
         if (auto charge = value_by_key(global_props.ion_species, ion_name)) {
-            state_->add_ion(ion_name, *charge, i.second.cv, i.second.init_iconc, i.second.init_econc, i.second.init_revpot);
+            state_->add_ion(ion_name, *charge, i.second);
         }
         else {
             throw cable_cell_error("unrecognized ion '"+ion_name+"' in mechanism");
--- a/arbor/fvm_lowered_cell_impl.hpp
+++ b/arbor/fvm_lowered_cell_impl.hpp
@@ -164,6 +164,8 @@ void fvm_lowered_cell_impl<Backend>::reset() {
 
     update_ion_state();
 
+    state_->zero_currents();
+
     // Call initialize again
     for (auto& m: revpot_mechanisms_) {
         m->initialize();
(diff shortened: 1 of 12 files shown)


arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -164,6 +164,15 @@ void fvm_lowered_cell_impl<Backend>::reset() {
 
     update_ion_state();
 
+    // Call initialize again
```
bcumming: This comment needs to be expanded to explain why we are double-initializing the ion species and mechanisms. See the comment below that explains why `threshold_watcher_` is reset at that point.

noraabiakar: I will update the comment.
However, there's another issue in ticket #896 (Scenario 2) that is not addressed in this PR yet. Initially we were going to not allow that scenario to occur, but after re-discussing it with Sam we agreed it's better to add support for it. So I will update this PR in the next few days.

test/unit/test_fvm_lowered.cpp
```diff
@@ -555,6 +556,68 @@ TEST(fvm_lowered, read_valence) {
...
+
+    layout.weight.assign(ncv, 1.);
+    for (fvm_size_type i = 0; i<ncv; ++i) {
+        layout.cv.push_back(i);
+        ion_config.cv.push_back(i);
+    }
+    ion_config.init_revpot.assign(ncv, 0.);
+    ion_config.init_econc.assign(ncv, 0.);
+    ion_config.init_iconc.assign(ncv, 0.);
+    ion_config.reset_iconc.assign(ncv, 0.);
```
halfflat: It doesn't invalidate the test, but I think this should be `rest_econc`.

noraabiakar: Thanks! I'll fix it

PR #899, opened 2019-10-31
Improve spike handling times

apeyser:
topic: feature/double prefetch

Improves spike handling times by 1) mutlithreading spike handling and 2) fixing how spikes are walked over with smaller number of spikes.

For 2, after a spike has had all dependent connections handled, those connections need not be looked at again (see make_queues_by_spike).

The improvement in behavior: even for a single thread, these changes cut overall time in half for fairly large simulations in dry-run mode. For multithread, this is then scaled again 2x or more depending on the architecture. The details of where the latency is completely hidden still need to be fully investigated on several architectures and for very large scales.


brenthuisman:
Shall we close this issue? The branch will remain present in the main repo for future reference, until someone accidentally or intentionally deletes it.

thorstenhater:
Hi,

I am going to close this. It has diverged quite a bit and we have no test case demonstrating the issue and/or
benchmark measuring the speed-up. @apeyser feel free to re-open this if needed.

T

PR #902, opened 2019-11-22
Remove managed memory from cuda backend

### LINKED ISSUE #841: Remove CUDA managed memory
I introduced CUDA managed memory to the Arbor library. For that I am very sorry.

It has been removed from the stack implementation. The only remaining uses are in the construction of the fine-grained matrix solver, and in the unit tests.

It is actually quite useful in tests, so we could move the managed memory support into only the unit tests.

This would simplify the gpu context, and remove a whole class of awful GPU synchronization bugs.

noraabiakar:
Completely remove managed memory from the gpu backend.
`managed_vector<level>` is split into:
- A device vector of structs containing meta data about each level 
- A device vector containing concatenated lengths of each branch in a level for all levels
- A device vector containing concatenated parents of each branch in a level for all levels

All other instances of managed memory in `matrix_fine` are easily replaced by device vectors  
Managed memory is also removed from the unit tests. 

Fixes #841


arbor/backends/gpu/matrix_fine.cu
```diff
@@ -100,30 +100,40 @@ void solve_matrix_fine(
...
     for (unsigned l=0; l<num_levels-1; ++l) {
-        const auto& lvl = block_levels[l];
+        // Metadata for this level and the next level
+        const auto& lvl_meta = block_levels_meta[l];
+        const auto& next_lvl_meta = block_levels_meta[l+1];
+
+        // Addresses of the first elements of levels_lengths and levels_parents
+        // that belong to this level
+        const auto lvl_lengths = &levels_lengths[lvl_meta.level_data_index];
+        const auto lvl_parents = &levels_parents[lvl_meta.level_data_index];
```
bcumming: For me writing this using pointer arithmetic as `lvl_parents = levels_parents + lvl_meta.level_data_index;` makes more sense.

arbor/backends/gpu/matrix_fine.cu
```diff
@@ -200,23 +212,29 @@ void solve_matrix_fine(
 
     // take great care with loop limits decrementing unsigned counter l
     for (unsigned l=num_levels-1; l>0; --l) {
-        const auto& lvl = block_levels[l-1];
+        const auto& lvl_meta = block_levels_meta[l-1];
```
bcumming: This definition shadows the one on line 175.

The `next_lvel_meta` variable is used only in one place, on line 224, so you might as well inline it.

arbor/backends/gpu/matrix_state_fine.hpp
```diff
@@ -83,52 +83,60 @@ struct matrix_state_fine {
...
+    iarray data_partition;
     std::size_t data_size;
 
-    // the meta data for each level for each block layed out linearly in memory
-    managed_vector<level> levels;
-    // the start of the levels of each block
-    // block b owns { leves[level_start[b]], ..., leves[level_start[b+1] - 1] }
+    // Metadata for each level
+    // Includes indices into d, u, rhs and levels_lengths, levels_parents
+    metadata_array levels_meta;
```
bcumming: The plural on `levels` is a bit confusing, I would rename these to `level_meta`, `level_lengths` etc.

arbor/backends/gpu/matrix_state_fine.hpp
```diff
@@ -300,53 +312,63 @@ struct matrix_state_fine {
...
                 }
 
-                pos += lvl.max_length*lvl.num_branches;
+                data_start+= lvl_meta.num_branches;
+                pos += lvl_meta.max_length*lvl_meta.num_branches;
 
-                levels.push_back(std::move(lvl));
+                temp_meta.push_back(std::move(lvl_meta));
+                std::move(lvl_lengths.begin(), lvl_lengths.end(), std::back_inserter(temp_lengths));
+                std::move(lvl_parents.begin(), lvl_parents.end(), std::back_inserter(temp_parents));
```
bcumming: I didn't know that `std::move` was also an algorithm! Because we are copying integers, `std::copy` would be just as efficient.

Or you could use `util::append(temp_parents, lvl_parents)` from `util/rangeutil.hpp`.

arbor/backends/gpu/threshold_watcher.hpp
```diff
@@ -115,9 +117,7 @@ class threshold_watcher {
                 cv_index_.data(), values_, thresholds_.data());
 
             // Check that the number of spikes has not exceeded capacity.
-            // ATTENTION: requires cudaDeviceSynchronize to avoid simultaneous
-            // host-device managed memory access.
-            arb_assert((cudaDeviceSynchronize(), !stack_.overflow()));
+            arb_assert(!stack_.overflow());
```
bcumming: Yay! Remove the last synchronizations!

test/unit/test_intrin.cu
```diff
@@ -46,34 +45,42 @@ namespace kernels {
 TEST(gpu_intrinsics, cuda_atomic_add) {
     int expected = (128*129)/2;
 
-    auto f = arb::gpu::make_managed_ptr<float>(0.f);
-    kernels::test_atomic_add<<<1, 128>>>(f.get());
+    arb::memory::device_vector<float> f(1);
+    f[0] = 0.f;
+
+    kernels::test_atomic_add<<<1, 128>>>(f.data());
     cudaDeviceSynchronize();
```
bcumming: we don't need the `cudaDeviceSynchronize` calls now that we don't use managed memory (it was required to stop the host from accessing the memory before the kernel had finished modifying the memory).
The access below (i.e. `f[0]` on line 54) uses `cudaMemcpy` under the hood, which will be synchronized with the kernel launch.

arbor/backends/gpu/matrix_fine.cu
noraabiakar: Done.

arbor/backends/gpu/matrix_fine.cu
noraabiakar: Will change the previous one to `last_lvl_meta` and inline `next_lvl_meta`

arbor/backends/gpu/matrix_state_fine.hpp
noraabiakar: Done.

arbor/backends/gpu/matrix_state_fine.hpp
noraabiakar: I'll use `util::append`, thanks!

PR #905, opened 2019-11-27
Modcc: Add support for `and` and `or` operators

### LINKED ISSUE #25: modcc: && not recognized as a boolean operator
Using the && to chain two boolean statement leads to an error:

error: unexpected character '&' at (line 55,col 22)
./offset_current.mod:(line 55,col 22)
  unexpected token &

Codeblock:

```
    if (t >=  delay  && t <  duration  +  delay) {
        i = amplitude : standard OnCondition
    }
```

noraabiakar:
Also renumber operator precedence.  

Fixes #25


modcc/lexer.cpp
```diff
@@ -195,6 +195,20 @@ Token Lexer::parse() {
                 }
                 return t;
             }
+            case '&':
+                t.spelling += character();
+                if(*current_=='&') {
+                    t.spelling += character();
+                    t.type = tok::and_t;
+                    return t;
+                }
```
bcumming: What happens if there is a single `&` or `|`, e.g. `if (foo & bar)` ?

In that case it should be an error, i.e. every `&` should be followed by another `&`, right?

```c++
t.spelling += character();
if(*current_=='&') {
    t.spelling += character();
    t.type = tok::and_t;
    return t;
}
else {
    error_string_ = pprintf("Unknown symbol '&': did you mean '&&'?");
    status_ = lexerStatus::error;
    t.type = tok::reserved;
    return t;
}
```

modcc/expression.cpp
```diff
@@ -1138,6 +1138,8 @@ expression_ptr binary_expression(Location loc,
         case tok::lte      :
         case tok::gt       :
         case tok::gte      :
+        case tok::and_t    :
```
bcumming: What was the logic behind calling them `_t` (other than `and` and `or` being reserved keywords)?
How about `tok::land` and `tok::lor` for "logical and" and "logical or" respectively?

noraabiakar: That works.

modcc/lexer.cpp
noraabiakar: True, I'll add tests and make sure errors are thrown when needed.

PR #901, opened 2019-11-13
NESTIO coupling with Arbor

apeyser:
Topic: Feature/nestio
Goes with: https://github.com/jougs/nest-simulator/pull/43

Proxy external cells can be created and coupled with a running nest simulation.

PR #913, opened 2019-12-06
NEST as External spike source

w-klijn:
Arbor changes and example codes to connect Arbor and NEST3 a single way co-simulation.

Spikes are generated in NEST send over to Arbor, with a shared COM world. Split into two using colors.

The example directory contains proxy application simulating the different simulators. This includes a pure Python arbor fake to be included in the NEST repo.


halfflat:
Closing for now, pending co-sim prototype work. See comment on #901 for slightly more detail.

Issue #920, opened 2019-12-11
Architecture support: build failures on architectures other than x86_64


sanjayankur31:
While [building for Fedora](https://koji.fedoraproject.org/koji/taskinfo?taskID=39482950), I've noticed that arbor fails to build on all architectures other than x86_64. These other architectures we currently build in Fedora for are: armv7hl, i686, aarch64, ppc64le, s390x.

The crash log says:

```
c++: warning: '-mcpu=' is deprecated; use '-mtune=' or '-march=' instead
In file included from /builddir/build/BUILD/arbor-0.2.2/arbor/include/arbor/arbexcept.hpp:6,
                 from /builddir/build/BUILD/arbor-0.2.2/arbor/arbexcept.cpp:4:
/builddir/build/BUILD/arbor-0.2.2/arbor/include/arbor/common_types.hpp: In member function 'std::size_t std::hash<arb::cell_member_type>::operator()(const arb::cell_member_type&) const':
/builddir/build/BUILD/arbor-0.2.2/arbor/include/arbor/common_types.hpp:102:46: error: static assertion failed: invalid size assumptions for hash of cell_member_type
  102 |             static_assert(sizeof(std::size_t)>sizeof(cell_gid_type), "invalid size assumptions for hash of cell_member_type");
      |                           ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
/builddir/build/BUILD/arbor-0.2.2/arbor/include/arbor/common_types.hpp:104:76: warning: left shift count >= width of type [-Wshift-count-overflow]
  104 |             std::size_t k = ((std::size_t)m.gid << (8*sizeof(cell_gid_type))) + m.index;
      |                                                                            ^
make[2]: *** [arbor/CMakeFiles/arbor.dir/build.make:66: arbor/CMakeFiles/arbor.dir/arbexcept.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Entering directory '/builddir/build/BUILD/arbor-0.2.2/build-serial'
[ 32%] Building CXX object arbor/CMakeFiles/arbor.dir/backends/multicore/stimulus.cpp.o
[...]

PR #925, opened 2019-12-13
Replace `isnan()` with fp equality test in neon code

halfflat:
* Replace serialized isnan(x) calls with cmp_eq(x,x) (changing sense of
associated tests accordingly).

While we could simply replace `isnan(x)` with `std::isnan(x)` to resolve the issue on the Fedora aarch64 build noted in #920, we should be able to make the code a bit more vector-ish with `cmp_eq(x, x)` (ultimately, `vceqq_f64(x, x)`.)


halfflat:
Note that we don't have an ARM box available for testing at the moment, so we should put off merging until the PR has been properly validated.

PR #932, opened 2020-01-07
Modcc: modify semantic passes

### LINKED ISSUE #931: Semantic pass not propagating errors
For example, `BinaryExpression::semantic(scp)` will execute 
```
lhs_->semantic(scp);
rhs_->semantic(scp);

if(rhs_->is_procedure_call() || lhs_->is_procedure_call()) {
    error("procedure calls can't be made in an expression");
}
```
If `lhs_->semantic(scp)` has some error, `lhs_->is_procedure_call()` could cause a segmentation fault. To fix this, we should check `lhs_->has_error()` and `rhs_->has_error()` after calling their semantic passes. 

In addition, the errors in a semantic pass don't always need to be raised; for example, when inlining function iteratively, we only need to check that there are no errors once all functions have been inlined successfully. We should make sure we have these `ErrorVisitor`s in the right places.

noraabiakar:
- Reset the error of the semantic pass of an expression at the start of the function call. 
- Check the errors of sub-expressions after applying to them their semantic passes, and report errors if found. If the error could disrupt the rest of the original semantic pass, return.
- Check errors in functions/procedures before and after inlining, but not during. 

Fixes #931


modcc/expression.cpp
```diff
@@ -324,8 +337,13 @@ expression_ptr StoichTermExpression::clone() const {
 }
 
 void StoichTermExpression::semantic(scope_ptr scp) {
+    error_ = false;
     scope_ = scp;
+
     ident()->semantic(scp);
+    if(ident()->has_error()) {
```
bcumming: Similar comment as above: can you use the error string that was set when calling `ident()->semantic(scp)`?

modcc/expression.cpp
```diff
@@ -303,12 +308,20 @@ expression_ptr ReactionExpression::clone() const {
...
+    error_ = false;
     scope_ = scp;
+
     lhs()->semantic(scp);
     rhs()->semantic(scp);
 
     fwd_rate()->semantic(scp);
     rev_rate()->semantic(scp);
+
+    if(lhs_->has_error() || rhs_->has_error() || fwd_rate_->has_error() || rev_rate_->has_error()) {
```
bcumming: Do we need to set the error state here? The error visitor will collect the errors set in `lhs`, `rhs`, etc. An additional error message here won't make much sense?

modcc/expression.cpp
noraabiakar: Yes, I can just set the error to true

modcc/expression.cpp
noraabiakar: We can just set the error flag and keep the same error message.

noraabiakar: The error visitor will only collect the errors of the main expression, in this case the `ReactionExpression`. The errors of `lhs`, `rhs`, `fwd_rate` and `rev_rate` will not be collected by the error visitor unless they are propagated to the main `ReactionExpression` expression.

The `ErrorVisitor` has visitors for each of: 
* `Expression`
* `ProcedureExpression`
* `FunctionExpression`
* `UnaryExpression`
* `BinaryExpression`
* `CallExpression`
* `BlockExpression`
* `InitialBlock`
* `IfExpression`

If we don't set the error state of the `ReactionExpression`: when the `ErrorVisitor` gets to it, it will call the default `Expression` visitor which won't print any errors, since the `ReactionExpression` itself has no errors.

If we don't want to propagate the errors of subexpressions (`lhs`, `rhs` etc) to their calling expressions, then we can extend the error visitor to visit a wider set of expressions and recursively check all subexpressions for errors.

modcc/expression.cpp
noraabiakar: We can copy the error of `ident` into the error string of the `StoichTermExpression`. I think that would be a more helpful message. I'll fix that.

PR #934, opened 2020-01-17
Remove unnecessary points from intersection results

noraabiakar:
Results of an intersection are sometimes returning extra zero length `cover` cables that shouldn't be included. 
Remove these extra cables when they are part of another cable in the returned list. 
Example:
- Assuming branch 1 is the parent of branch 2: `{1,1,1},{2,0,0.5}` should become `{2,0,0.5}`
- Assuming branch 1 is not the parent of branch 2: `{1,1,1},{2,0,0.5}` remains the same


arbor/morph/region.cpp
```diff
@@ -132,6 +132,43 @@ mcable_list remove_cover(mcable_list cables, const morphology& m) {
     return merge(cables);
 }
 
+mcable_list remove_redundancy(mcable_list cables, const morphology& m) {
```
bcumming: Would `remove_covered_points` be a better name for this function?

How about this for a comment above the function?
```C++
// Remove zero-length regions from a sorted cable_list that are in the cover of another region in the list.
```

noraabiakar: Okay, I'll modify it

PR #938, opened 2020-01-20
C++ library changes for python-cells

bcumming:
Changes to the C++ core library and associated tests that were performed during python-cell wrapping.

These are separated from the main Python wrapper PR, for separate review according to C++ standards.

### DIFF
--- a/arbor/include/arbor/morph/locset.hpp
+++ b/arbor/include/arbor/morph/locset.hpp
@@ -121,9 +121,6 @@ namespace ls {
 // Explicit location on morphology.
 locset location(mlocation);
 
-// Explicit location on morphology.
-locset location(msize_t, double);
-
 // Location of a sample.
 locset sample(msize_t);
 
--- a/arbor/include/arbor/morph/locset.hpp
+++ b/arbor/include/arbor/morph/locset.hpp
@@ -118,8 +118,11 @@ private:
 
 namespace ls {
 
-// Location of a sample.
+// Explicit location on morphology.
 locset location(mlocation);
+
+// Explicit location on morphology.
+locset location(msize_t, double);
 
 // Location of a sample.
 locset sample(msize_t);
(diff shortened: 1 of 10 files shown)


arbor/include/arbor/morph/locset.hpp
```diff
@@ -138,4 +141,9 @@ locset nil();
 
 } // namespace ls
 
+// union of two locsets
+locset join(locset, locset);
+// multiset sum of two locsets
+locset sum(locset, locset);
+
```
halfflat: I reckon it'd be cleaner to put lambdas in the S-expression parser function table, than to add these extra constructors and functions: it saves doubling up on the interfaces.

arbor/include/arbor/recipe.hpp
```diff
@@ -21,7 +21,7 @@ struct probe_info {
 };
 
 /* Recipe descriptions are cell-oriented: in order that the building
- * phase can be done distributedly and in order that the recipe
+ * phase can be done distributed, and in order that the recipe
```
halfflat: 'distributedly', while awkward, is actually correct :)

arbor/include/arbor/morph/locset.hpp
bcumming: The functions that you have hilighted here are not actually new calls.
I added their prototypes here so that the can be referred to (the only place their prototype appeared before was as `friend` calls inside the `locset`/`region` class definition.)

I think it is cleaner to directly refer to an existing function if it exists, and restrict use of lambdas to situations where no calls exist.

I have, however, removed the `location(msize_t, double)` prototype and its implementation, and replaced it with a lambda in the s-expresson table.

arbor/include/arbor/recipe.hpp
bcumming: You are right. I fixed it by also removing the `done`, because `distributedly` is indeed both correct and funny sounding.

bcumming:
Can we make that a todo? It isn't so much in the scope of this PR.

halfflat:
> Can we make that a todo? It isn't so much in the scope of this PR.

Sure. I approved the PR, but didn't merge in case you thought it's worth amending. If you think it's in principle a good thing to do, I can whip up a small patch as a separate PR.

halfflat:
Changes got a little involved, so I've made a separate PR.

PR #943, opened 2020-01-22
Make region/locset interface more consistent.

halfflat:
* Add implicit conversions from `mlocation` and `mlocation_list` to `locset`.
* Add implicit conversions from `mcable` and `mcable_list` to `region`.
* Change arguments for `ls::location` to `(branch, position)`.
* Extend unit tests to suit.
* Make locset and region wrapping less promiscuous by demanding that expression implementation classes derive from `locset_tag` or `region_tag` respectively.
* Remove dubious chunk of variable declarations in `cable_cell_param.cpp`.

### DIFF
--- a/arbor/morph/locset.cpp
+++ b/arbor/morph/locset.cpp
@@ -216,7 +216,7 @@ locset::locset(mlocation loc) {
 
 locset::locset(const mlocation_list& ll) {
     *this = std::accumulate(ll.begin(), ll.end(), ls::nil(),
-        [](auto& ls, auto& p) { return join(ls, locset(p)); });
+        [](auto& ls, auto& p) { return sum(ls, locset(p)); });
 }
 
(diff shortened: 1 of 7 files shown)


arbor/morph/locset.cpp
```diff
@@ -200,12 +204,19 @@ locset sum(locset lhs, locset rhs) {
...
 }
 
 locset::locset(mlocation loc) {
-    *this = ls::location(loc);
+    *this = ls::location(loc.branch, loc.pos);
+}
+
+locset::locset(const mlocation_list& ll) {
+    *this = std::accumulate(ll.begin(), ll.end(), ls::nil(),
+        [](auto& ls, auto& p) { return join(ls, locset(p)); });
```
bcumming: `sum`, so as not to remove non-unique points from the input.

PR #949, opened 2020-01-29
Address poor fvm_cv_discretize performance.

### LINKED ISSUE #946: Slow discretization
Very slow, superlinear discretization is observed with the new discretizer — on a Purkinje cell model with circa 700 branches, discretization and mechanism layout with 36 CVs per branch is taking over a minute.

Find and fix!

halfflat:
* Provide (internal) locset wrapper for mlocation_list so as to avoid quadratic mlocation sum(...) behaviour.
* Add ubench for fvm_discretize, with lmorpho-generated test SWC.
* Rewire `cv_policy_every_sample` to produce a locset that thingifies faster.

Fixes #946 .


arbor/morph/locset.cpp
```diff
@@ -68,6 +68,32 @@ std::ostream& operator<<(std::ostream& o, const location_& x) {
...
+    for (mlocation loc: x.ll) {
+        if (loc.branch>=n_branch) {
+            throw no_such_branch(loc.branch);
+        }
+    }
+    return x.ll;
+}
+
+std::ostream& operator<<(std::ostream& o, const location_list_& x) {
+    o << "(list";
```
bcumming: Should this print `(sum (location x y) (location a b) ....)` to would make the round trip to string and back again valid?

halfflat: Seems reasonable! Will edit.

arbor/morph/locset.cpp
```diff
@@ -362,9 +388,8 @@ locset::locset(mlocation loc) {
     *this = ls::location(loc.branch, loc.pos);
 }
 
-locset::locset(const mlocation_list& ll) {
-    *this = std::accumulate(ll.begin(), ll.end(), ls::nil(),
-        [](auto& ls, auto& p) { return sum(ls, locset(p)); });
+locset::locset(mlocation_list ll) {
+    *this = ls::location_list(std::move(ll));
```
noraabiakar: Can we also have a move constructor?

PR #950, opened 2020-01-30
Replace unnecessary cv_policy copy in fvm_discretize.

halfflat:
Also: add a couple more cases to fvm_discretize ubench.


halfflat:
I haven't been able to reproduce the explicit `locset::locset(mlocation_list&&)` speed difference in the simple ubench test though :(

PR #955, opened 2020-02-07
enable locset names for probe sites in single cell model.

bcumming:
Use locset names to set probe sites in single cell model.

1. Extend C++ cable_cell interface to provide a concrete list of
   locations associated with a locset on the cell.
2. Extend Python single_cell_model to allow placement of probes
   on locsets.
3. Remove "debug" output in Python ring example.

The single cell model maintains a trace callback for each probe
location, and needs to know the explicit locations before
starting the simulation. The first step is required to enumerate
and record the locations associated with a locset when attaching probes.

### DIFF
--- a/arbor/cable_cell.cpp
+++ b/arbor/cable_cell.cpp
@@ -103,6 +103,10 @@ struct cable_cell_impl {
             }
         }
     }
+
+    mlocation_list locations(const locset& l) const {
+        return thingify(l, provider);
+    }
 };
 
 using impl_ptr = std::unique_ptr<cable_cell_impl, void (*)(cable_cell_impl*)>;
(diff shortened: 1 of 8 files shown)


bcumming:
After discussion with Sam, there are the following issues that he alluded to above

1. The extension to `cable_cell` won't sufficiently support locsets that rely on random number generators with global seeds (it would require passing additional meta data required to fully instantiate the locset)
2. Enhancing the probe recipe interface to be able to associate one `probe_id` with all locations in a locset (the number of which is not specified at the time the `probe_id` is created) would support the interface I have implemented for the `single_cell_model`.

My position on this is that with this in place, the Python interface for `single_cell_model` would be unchanged. The current approach is a valid way to expose this functionality in a timely manner, and I think we can proceed, with a note that the richer interface on the C++ side will be implemented in time.

bcumming:
Just waiting in case @halfflat has further comments.

bcumming:
I have made an issue for extending the probe interface, to make the back end implementation of this feature more consistent #957

PR #963, opened 2020-02-23
Include required header

sanjayankur31:


bcumming:
Was this breaking the build for you?

I will make a separate PR that fixes this, because for us to merge your changes you first need to sign a form assigning copyright of your contributions to us, which is a bit of administrative overhead for such a small change.

sanjayankur31:
Yep, this broke the build on Fedora here for me. I forget what architecture it was though. It could be related to the flags we use for our builds, but I'll have to check.

Sure. No worries.

PR #967, opened 2020-02-24
Add missing semi-colon

sanjayankur31:
Tiny syntax fix.

### DIFF
--- a/arbor/include/arbor/simd/neon.hpp
+++ b/arbor/include/arbor/simd/neon.hpp
@@ -626,7 +626,7 @@ struct neon_double2 : implbase<neon_double2> {
     }
 };
 
-}  // namespace detail
+};  // namespace detail
 
 namespace simd_abi {
 template <typename T, unsigned N>


arbor/include/arbor/simd/neon.hpp
```diff
@@ -626,7 +626,7 @@ struct neon_double2 : implbase<neon_double2> {
     }
 };
 
-}  // namespace detail
+};  // namespace detail
```
bcumming: Semi-colons are not required at the end of namespace definitions. To the extent of my knowledge, the only time a semi-colon is required after a `}` is at the end of a `class`, `struct`, or `union` definition.

sanjayankur31: It's a struct definition. The matching brace is on line 190 at  ```struct neon_double2 : implbase<neon_double2> {```

sanjayankur31: Either the comments are off or matching brackets are missing: vim tells me that `namespace detail` ends on line 646 which says `// namespace simd`, and nothing matches the bracket for `// namespace arb` on line 647.

noraabiakar: A closing bracket was accidentally deleted in #925 line 489

sanjayankur31: Ah, then putting that back would be the correct fix. I'll close this and let you fix it up properly.

PR #971, opened 2020-02-24
More robust Python installation

bcumming:
Improve the Python wrapper generation and installation:
  - install a proper module that can be extended with Python code;
  - give the user more control over where to install the module (e.g. as a user package or in a virtualenv).

During building, the following sub-directory is built in the build director (`CMAKE_BINARY_DIR`)
```
└── python
    └── arbor
            ├── __init__.py
            ├── arbor.so
            └── VERSION
```
This path can then be copied VERBATIM to the target installation path. By default this will be in `CMAKE_INSTALL_PREFIX/lib/python%d.%d/site-packages`.
An additional CMake parameter `ARB_PYTHON_PREFIX` can be used to specify an alternative destination for installing the Python module.

The Python part of the wrapper, implemented in `__init__.py` is currently very limited, only providing `__version__` and `__config__` variables.

The installation guide was updated to cover the Python installation.

### DIFF
--- a/doc/install.rst
+++ b/doc/install.rst
@@ -117,10 +117,11 @@ More information on building with MPI is in the `HPC cluster section <cluster_>`
 Python
 ~~~~~~
 
-Arbor has a Python frontend, for which Python 3.6 is required.
+Arbor has a Python frontend, for which a minimum of Python 3.6 is required.
 In order to use MPI in combination with the python frontend the
 `mpi4py <https://mpi4py.readthedocs.io/en/stable/install.html#>`_
-Python package is recommended.
+Python package is recommended. See :ref:`pythonfrontend` for more information.
+
 
 Documentation
 ~~~~~~~~~~~~~~
@@ -354,6 +355,8 @@ example:
     Arbor supports and has been tested on the Kepler (K20 & K80), Pascal (P100) and Volta (V100) GPUs
 
 
+.. _pythonfrontend:
+
 Python Frontend
 ----------------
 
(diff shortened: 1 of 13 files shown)


Helveg:
Would the whole installation be duplicated if you repeated the Cmake with different executables and prefixes? If so, specifying multiple install locations could really benefit users with many virtual environments or users of pyenv

Helveg:
> * install a proper module that can be extended with Python code;

I'm a big fan of this. This should allow the module to be freely monkey patched as well >:)

bcumming:
If you re-run CMake with a different `PYTHON_EXECUTABLE`, then only the wrapper is rebuilt, and it will be installed in the right location (I checked).

doc/install.rst
```diff
@@ -362,13 +365,37 @@ CMake ``ARB_WITH_PYTHON`` option:
...
+
+.. code-block:: bash
+
+    cmake .. -DARB_WITH_PYTHON=ON                   \
+             -DARB_PYTHON_PREFIX=${HOME}/.local/lib \
+             -DPYTHON_EXECUTABLE=/user/bin/python3.8
 
-By default ``ARB_WITH_PYTHON=OFF``. When this option is turned on, a python module called :py:mod:`arbor` is built.
+On the target LINUX system, the Arbor package was installed in
+``/home/$USER/.local/lib/python3.8/site-packages``.
```
Helveg: On Windows the path should be `{prefix}\Lib\site-packages` according to https://docs.python.org/3/library/site.html. Some sources say capital L some don't, not sure if it matters.

doc/install.rst
```diff
@@ -362,13 +365,37 @@ CMake ``ARB_WITH_PYTHON`` option:
...
+By default the Python module will be installed in the standard ``CMAKE_INSTALL_PREFIX``
+location. To install the module in a different location, for example as a
+user module or in a virtual environment, set ``ARB_PYTHON_PREFIX``.
+For example, the CMake configuration for targetting Python 3.8 and install as a
+user site package might look like the following:
+
+.. code-block:: bash
+
+    cmake .. -DARB_WITH_PYTHON=ON                   \
+             -DARB_PYTHON_PREFIX=${HOME}/.local/lib \
```
Helveg: Wouldn't it make more sense to `use ${HOME}/.local` as prefix and append the "lib", to be more in line with `sys.prefix`?

doc/install.rst
bcumming: Arbor only has official support for Linux and OS X, so we don't consider Windows. This is an easy one to fix when we do support Windows.
FWIW, from what I can tell, case sensitive path names are turned off by default in Windows, and have to be turned on for individual paths.

doc/install.rst
bcumming: Yes, probably. If we want to install something in `bin` or `share` this would also be logical. I will update this.

Helveg: Wait are you sure? Because with my proposal it's going to be `bin/lib/python3.8/site-packages` is that ok?

bcumming:
I updated the docs to have a little blurb about sudo and `CMAKE_INSTALL_PREFIX`.
I also made the `ARB_PYTHON_PREFIX` consistent with `sys.prefix`.

Issue #969, opened 2020-02-24
More sophisticated domain decomposition


noraabiakar:
Allow users to provide extra hints to `partition_load_balance` in the form of a set of `gid` ranges of cells with the same complexity. 

`partition_hint_map` can be replaced by `partition_hint` 
```
struct cell_group_hint {
    constexpr static std::size_t max_size = -1;

    std::size_t cpu_group_size = 1;
    std::size_t gpu_group_size = max_size;
    bool prefer_gpu = true;
};

using gid_range_hint =  std::pair<cell_gid_type, cell_gid_type>;

using cell_group_hint_map = std::unordered_map<cell_kind, cell_group_hint>;
using gid_range_hint_set  = std::set<gid_range_hint, custom_compare>;

struct partition_hint {
    cell_group_hint_map cell_group_map;
    gid_range_hint_set gid_range_set;
}

domain_decomposition partition_load_balance(
    const recipe& rec,
    const context& ctx,
    partition_hint hint_map = partition_hint());
}
```
`partition_load_balance` can then distribute each `gid_range` in `gid_range_set` equally among sets.

noraabiakar:
Instead of modifying `partition_load_balance`, add a new partitioning algorithm that allows users to decide on their own cell-groupings.

thorstenhater:
Fixed by
https://github.com/arbor-sim/arbor/pull/1788

PR #974, opened 2020-02-25
Partition cells by complexity

noraabiakar:
Allow users to group cells into ranges of `gid`s which have the same complexity, and pass the set of ranges as a hint to `partition_load_balance`. 
`partition_load_balance` can then place an equal number of cells of the same complexity on each rank. 

Addresses #969


noraabiakar:
We can revisit this issue in the future, when the need for a more expressive domain decomposition re-arises

Issue #962, opened 2020-02-20
Build fails for the ring example  


Kaveh01:
make fails for the ring example (only for mpi):   
```
[ 91%] Built target arborenv
[ 94%] Building CXX object sup/CMakeFiles/arbor-sup.dir/json_meter.cpp.o
[ 97%] Linking CXX static library ../lib/libarborsup.a
[ 97%] Built target arbor-sup
Scanning dependencies of target ring
[100%] Building CXX object example/ring/CMakeFiles/ring.dir/ring.cpp.o
   .../arbor/example/ring/ring.cpp: In constructor ‘cell_stats::cell_stats(arb::recipe&)’:
   .../arbor/example/ring/ring.cpp:156:28: error: ‘class arb::cable_cell’ has no member named ‘num_branches’
```
I guess line 156 of ring.cpp should be something like
 `nsegs_tmp += c.morphology().num_branches();`
 instead of   
`nsegs_tmp += c.num_branches();`

bcumming:
Thanks for that. Indeed, it should be
```C++
nsegs_tmp += c.morphology().num_branches();
```
This slipped through because we don't compile the examples in our CI.

bcumming:
To fix this, I suggest just scrapping the `cell_stats` step.

Kaveh01:
Thanks for the replay.

bcumming:
Let's keep this open until we have fixed the problem.

PR #979, opened 2020-03-11
Remove cell stats 

noraabiakar:
Remove `cell_stats` from gap junction, dry run and ring examples.

Addresses #962


bcumming:
Have you verified that `make examples` works with MPI?

noraabiakar:
> Have you verified that `make examples` works with MPI?

Yes, I compiled and ran the examples with MPI

Issue #634, opened 2018-10-24
Parallelise construction of cell groups with many cells


bcumming:
Model building parallelises cell group construction over cell groups.
Models with few cell groups with many cells in each group don't benefit from this optimization.
The most extreme example of this is a model on the GPU with a single cell group with thousands of cells.

There are two parts to the fvm lowered cell construction:
* 60% of time is in getting the vector of cell groups from the recipe
* 40% of time is then spent building the lowered back end cell group data structures

The 60% in recipe querying is trivial to parallelise, however the remaining 40% is considerably more difficult.
So at least we should parallelise the recipe querying, which would require making the task system available to the lowered fvm cell.

halfflat:
The discretization step (part of the 40%) is serialized, but I think making this more parallel will be fairly straightforward if we make a concatenation function for `fvm_mechanism_data` and `fvm_discretization` objects — this is mostly appending vectors and offsetting segment or cv indices accordingly.

PR #980, opened 2020-03-12
Parallelize parts of model initialization 

noraabiakar:
Use the task system to parallelize parts of the model initialization phase, namely cell building,`fvm_cv_discretize` and `fvm_build_mechanism_data`. 

These changes speed up large cell group builds.

Addresses #634

### DIFF
--- a/arbor/fvm_layout.cpp
+++ b/arbor/fvm_layout.cpp
@@ -450,13 +450,13 @@ fvm_cv_discretization fvm_cv_discretize(const std::vector<cable_cell>& cells,
     const cable_cell_parameter_set& global_defaults,
     const arb::execution_context& ctx)
 {
-    std::vector<fvm_cv_discretization> temp(cells.size());
+    std::vector<fvm_cv_discretization> cell_disc(cells.size());
     threading::parallel_for::apply(0, cells.size(), ctx.thread_pool.get(),
-          [&] (int i) { temp[i]=fvm_cv_discretize(cells[i], global_defaults);});
+          [&] (int i) { cell_disc[i]=fvm_cv_discretize(cells[i], global_defaults);});
 
     fvm_cv_discretization combined;
     for (auto cell_idx: count_along(cells)) {
-        append(combined, temp[cell_idx]);
+        append(combined, cell_disc[cell_idx]);
     }
     return combined;
 }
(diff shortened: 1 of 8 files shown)


bcumming:
The speedup of 1.9x is with how many cores?

arbor/fvm_layout.cpp
```diff
@@ -446,12 +447,16 @@ fvm_cv_discretization fvm_cv_discretize(const cable_cell& cell, const cable_cell
...
+    std::vector<fvm_cv_discretization> cell_disc(cells.size());
+    threading::parallel_for::apply(0, cells.size(), ctx.thread_pool.get(),
+          [&] (int i) { cell_disc[i]=fvm_cv_discretize(cells[i], global_defaults);});
 
-    for (const auto& c: cells) {
-        append(combined, fvm_cv_discretize(c, global_defaults));
+    fvm_cv_discretization combined;
+    for (auto cell_idx: count_along(cells)) {
+        append(combined, cell_disc[cell_idx]);
     }
```
halfflat: Would there be any material benefit in implementing a parallel reduction?

arbor/fvm_layout.hpp
```diff
@@ -112,7 +113,7 @@ fvm_cv_discretization& append(fvm_cv_discretization&, const fvm_cv_discretizatio
 // Construct fvm_cv_discretization from one or more cells.
 
 fvm_cv_discretization fvm_cv_discretize(const cable_cell& cell, const cable_cell_parameter_set& global_dflt);
-fvm_cv_discretization fvm_cv_discretize(const std::vector<cable_cell>& cells, const cable_cell_parameter_set& global_defaults);
+fvm_cv_discretization fvm_cv_discretize(const std::vector<cable_cell>& cells, const cable_cell_parameter_set& global_defaults, const arb::execution_context& ctx);
```
halfflat: Here and below: if we give the context parameter a default value (e.g. a default-constructed `execution_context`) I believe we could avoid the modifications to the unit tests (save for any parallel tests of course!)

arbor/fvm_layout.cpp
halfflat: On thinking about this, I realise there can be no gain here. Sorry!

noraabiakar:
I tested this on my laptop (4 cores; i7-7600U): Prior to these changes, building a cell_group with 10000 cells takes around 8s; 60% in `fvm_build_mechanism_data`; 25% in `fvm_cv_discretize` and 10% in building the actual cells (`get_cell_description`). After the changes, the time goes down to 4.2s. The `append` functions aren't the main remaining hotspots, for further improvement we'd need to look at the content of `fvm_build_mechanism_data` and `fvm_cv_discretize`. I've already achieved some minor improvements there for a test case with many synapses, and I'm continuing to look at other opportunities for optimization.

arbor/fvm_layout.hpp
noraabiakar: Okay, I will make the change :)

arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -356,15 +356,17 @@ void fvm_lowered_cell_impl<B>::initialize(
...
-        catch (util::bad_any_cast&) {
-            throw bad_cell_description(rec.get_cell_kind(gid), gid);
-        }
-    }
+    cells.resize(ncell);
+    threading::parallel_for::apply(0, gids.size(), context_.thread_pool.get(),
+           [&](cell_size_type i) {
+               auto gid = gids[i];
+               try {
+                   cells[i] = std::move(any_cast<cable_cell>(rec.get_cell_description(gid)));
```
halfflat: The `std::move` here is redundant, but in order to remove one stage of move-construction, one can use `any_cast<cable_cell&&>`.

noraabiakar: Thanks! I hadn't looked closely at what `any_cast` was doing.

halfflat:
Note: travis failure was spurious; accepting.

Issue #803, opened 2019-06-28
Update HBP acknowledgement to follow new guidelines


bcumming:
We received new guidance from EC on how to acknowledge their funding. We currently acknowledge it correctly in our Zenodo releases, however we should add it to one of ACKNOWLEDGEMENTS/README and the online docs.

```
The EC Project Officer has informed the HBP Coordinator regarding the proper
way to acknowledge EC funding*:

So far we acknowledge EC funding in publications and in communication activities
(e.g. on project websites). Our project officer now informed us that in addition to that
also any infrastructure, equipment and major results (co-) funded by the HBP grants
must include such an acknowledgement. This in particular applies to open source
software as well. 

What do you need to do?
For all open source software (co-) funded by the HBP, add a new release as soon
as possible to include the following or a similar acknowledgement: 

This open source software code was developed in part or in whole in the Human
Brain Project, funded from the European Union’s Horizon 2020 Framework
Programme for Research and Innovation under the Specific Grant Agreement
No. 785907 (Human Brain Project SGA2).

If the software was already HBP-funded in Ramp-up Phase and/or SGA1, please
adjust the acknowledgement accordingly. You can find all grant numbers in
our SP7 Coordination Collab:
https://collab.humanbrainproject.eu/#/collab/7065/nav/107015

If the acknowledgment can be accompanied by the EU flag (i.e. if it’s not in a
plain text file), the flag should be included as well. In source code the
acknowledgment can be included either in a comment close to the top of
source files or in a separate README or ACKNOWLEDGEMENTS file.

Please also check if the funding statement is included in other (publicly available)
websites, on social media channels... about HBP funded results. 
```

halfflat:
Resolved by #981.

Issue #695, opened 2019-02-26
Substitute tinyopt for TCLAP in modcc and Brunel example.


halfflat:
While a less tiny tinyopt is still a goal (#557), for 0.2 release use the existing tinyopt to replace the option parsing in modcc and the Brunel network example.

Then remove TCLAP from the Arbor repo and build.

halfflat:
One problem: even if we separate tinyopt from arbor-sup, it depends upon `arb::util::optional`. If `modcc` uses `tinyopt`, then it indirectly depends upon the `arbor` library (though not for linking — `util::optional` is header-only — but in terms of targets); the `arbor` library depends upon `modcc` for built-in mechanism generation.

The option parser being developed in #557 uses its own `optional`-like class `to::maybe<V>` which will allow us to side-step this issue.

Consequently, I'd like to get #557 done first, and migrate our TCLAP- and tinyopt-using code to use that.

halfflat:
Missed 0.2 for this one; will resolve for next point release.

halfflat:
Resolved by #981.

Issue #557, opened 2018-07-31
Replace TCLAP


halfflat:
TCLAP is basically awful, but the other command-line parsing code in the repo (`tinyopt`) is too feature-poor to make a satisfying replacement.

Make a less tiny `tinyopt` that supports separation from option specification and parsing, which is still compact in source, simple to use and extend, and remains less awful than TCLAP.

halfflat:
Note: goal is to have TCLAP out as part of the planned 0.2 release.

halfflat:
Current goal is to copy-paste a replacement from (a subset of) https://github.com/halfflat/tinyopt.

halfflat:
Resolved by #981.

PR #981, opened 2020-03-13
Remove TCLAP

noraabiakar:
- Replace TCLAP with tinyopt
- Remove `sup/tinyopt.hpp` 
- Fix `modcc` to use smolopt
- Fix `brunel` to use smolopt
- Fix the unit tests to use correct tinyopt
- Fix our attributions and add HBP acknowledgement

Addresses #557 #695 #803


example/single/single.cpp
```diff
@@ -117,19 +117,19 @@ options parse_options(int argc, char** argv) {
...
             opt.syn_weight = weight.value();
         }
-        else if (auto swc = parse_opt<std::string>(arg, 'm', "morphology")) {
+        else if (auto swc = parse<std::string>(arg, 'm', "morphology")) {
             opt.swc_file = swc.value();
         }
-        else if (auto nseg = parse_opt<unsigned>(arg, 'n', "cv-per-branch")) {
+        else if (auto nseg = parse<unsigned>(arg, 'n', "cv-per-branch")) {
             opt.policy = arb::cv_policy_fixed_per_branch(nseg.value(), arb::cv_policy_flag::single_root_cv);
         }
```
halfflat: While this is the most direct adaption, an alternative would be to use the `<<` operator, for something like the following:
```
    auto set_policy = [&opt](int n) {
        opt.policy = arb::cv_policy_fixed_per_branch(n, arb::cv_policy_flag::single_root_cv);
    };

    char** arg = argv+1;
    while (*arg) {
        bool ok =
            opt.dt << parse<double>(arg, 'd', "dt") ||
            opt.t_end << parse<double>(arg, 't', "t-end") ||
            opt.syn_weight << parse<float>(arg, 'w', "weight") ||
            opt.swc_file << parse<std::string>(arg, 'm', "morphology") ||
            set_policy << parse<unsigned>(arg, 'n', "cv-per-branch");

        if (!ok) […] // usage error etc.
   }
```
The code you have though is somewhat less magical looking, and may be preferable!

noraabiakar: While I like this implementation (especially how compact it is), I think the original code is slightly more readable which may be preferable in the example code. However, if you disagree with that I have no problem using this version :)

modcc/modcc.cpp
```diff
@@ -141,84 +127,61 @@ simd_spec parse_simd_spec(std::string spec) {
...
+                { opt.modfile,  to::mandatory},
+                { opt.outprefix,                    "-o", "--output" },
+                { to::set(opt.verbose),  to::flag,  "-V", "--verbose" },
+                { to::set(opt.analysis), to::flag,  "-A", "--analyse" },
+                { opt.modulename,                   "-m", "--module" },
+                { to::set(popt.profile), to::flag,  "-P", "--profile" },
+                { popt.cpp_namespace,               "-N", "--namespace" },
+                { to::set(simd_enabled), to::flag,  "-s", "--simd" },
+                { spec,                             "-S", "--simd-abi" },
+                { {targets, to::delimited<std::string>()}, "-t", "--target" },
```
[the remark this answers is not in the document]
noraabiakar: I used the third option proposed here, along with the suggestions for parsing the simd options below. I like parsing directly into `opt` and `popt`.

PR #985, opened 2020-03-20
Optimize model-init runtime

noraabiakar:
`model-init` run-times have been observed to be too high when running systems with many synapses.
 
The main culprit is the `sort_by` function (fvm_layout:723). When testing on a model with 1024 cells and 10000 synapses, 65% of model-init is spent in this function. 

Comparing two `std::map<std::string, double>` is really expensive. We can replace the string map `synapse_instance.param_value` with a sorted `std::vector<std::pair<unsigned, double>>`, and have an additional `std::map<std::string, unsigned> param_map`  to keep track of all the param_name -> unsigned mappings.

`model-init` run-time is 2.5x faster (tested on my 4 core  i7-7600U laptop):
```
10000 synapses/cell 
--------------------------------------------------------------------
|  num cells       | model-init before  (s) | model-init after (s) |
--------------------------------------------------------------------
|       1024       |          12            |       5              |
|       2048       |          25.5          |       10.8           |
|       4096       |          55.5          |       23.2           |
|       8192       |          121.8         |       49             |
--------------------------------------------------------------------
```

### DIFF
--- a/arbor/fvm_layout.cpp
+++ b/arbor/fvm_layout.cpp
@@ -684,29 +684,38 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties& 
 
     struct synapse_instance {
         size_type cv;
-        std::map<std::string, double> param_value; // uses ordering of std::map
+        std::vector<std::pair<unsigned, double>> param_value; // sorted vector of <param_id,value> pairs
         size_type target_index;
     };
 
     for (const auto& entry: cell.synapses()) {
         const std::string& name = entry.first;
         mechanism_info info = catalogue[name];
-        std::map<std::string, double> default_param_value;
         std::vector<synapse_instance> sl;
 
+        // Map from a param string identifier to a unique unsigned int identifier
+        std::map<std::string, unsigned> param_map;
+
+        // Map from a param unsigned int identifier to a default value
+        std::map<unsigned, double> default_param_value;
+
+        unsigned id=0;
         for (const auto& kv: info.parameters) {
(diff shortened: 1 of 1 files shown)


arbor/fvm_layout.cpp
```diff
@@ -684,29 +684,38 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties&
...
     };
 
     for (const auto& entry: cell.synapses()) {
         const std::string& name = entry.first;
         mechanism_info info = catalogue[name];
-        std::map<std::string, double> default_param_value;
         std::vector<synapse_instance> sl;
 
+        // Map from a param string identifier to a unique unsigned int identifier
+        std::map<std::string, unsigned> param_map;
```
bcumming: A hash table (`std::unordered_map`) might be a better choice here, because below on line 715 the `map::at` call will do a binary search using string comparison, whereas with a hash table it will use a hash of the string.

bcumming: A pretty minor optimization, given that it isn't the pain point that you are optimizing for here, so let me know if you want to ignore it.

noraabiakar: No, that's a good point. I'll fix it :)

noraabiakar: Actually, I remembered that it was originally `unordered_map` but that broke the assertions at `fvm_layout.cpp:500` and `fvm_layout.cpp:501`. I don't think those assertions are necessary but I'm not comfortable removing them.

bcumming: :+1:

PR #987, opened 2020-03-23
Fix mech_vec

noraabiakar:
Fixed `mech_vec` to work with new API and re-enabled the benchmark. Had to disable some of the tests because of long run-times (function responsible for slow-down is `cv_geometry::location_cv`, called from `fvm_layout.cpp:721` and is probably related to unrealistically large cv counts).

### DIFF
--- a/test/ubench/mech_vec.cpp
+++ b/test/ubench/mech_vec.cpp
@@ -34,172 +34,266 @@ mechanism_ptr& find_mechanism(const std::string& name, fvm_cell& cell) {
 class recipe_expsyn_1_branch: public recipe {
     unsigned num_comp_;
     unsigned num_synapse_;
+    arb::cable_cell_global_properties gprop_;
+
 public:
     recipe_expsyn_1_branch(unsigned num_comp, unsigned num_synapse):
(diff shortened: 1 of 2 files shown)


test/ubench/mech_vec.cpp
```diff
@@ -34,173 +34,267 @@ mechanism_ptr& find_mechanism(const std::string& name, fvm_cell& cell) {
...
+        tree.append(2, {{0          ,dend_length,soma_radius+dend_length, dend_radius}, 3});
+        tree.append(2, {{dend_length,0          ,soma_radius+dend_length, dend_radius}, 3});
+
+        d.set("soma",      tagged(1));
+        d.set("dend", tagged(3));
+
+        arb::cable_cell cell(arb::morphology(tree, true), d);
+
+        cell.paint("soma", "pas");
+        cell.paint("dend", "pas");
```
bcumming: Because we are applying the same mechanism to the whole cell here (and in the other recipes), you could just
```C++
cell.paint(arb::reg::all(), 'pas')
```
and initialise the cell with an empty dictionary.

PR #990, opened 2020-03-26
Re-write pow in terms of exp and log.

thorstenhater:
Re-write `pow(b, e) = exp(log(b) e)`, a common optimisation.

In my tests (nsuite busyring with 'hh' and 'pas' mechanism swapped) it
yields a significant speed-up 11.5s vs 7.7s on the largee/run_128_2.json
input.


thorstenhater:
We agreed on not doing this on grounds of precision.

PR #994, opened 2020-03-30
New cable representation for CVs.

halfflat:
* Replace CV cable representation with one which makes it clear to which CV a fork point 'belongs': every fork point belongs to exactly one CV; if a CV representation has cables on two different branches at a fork point, it must have cables on all branches at that fork point, and the fork point is associated with that CV.
* Rework (and simplify) `cv_geometry_from_ends` to suit new representation.
* Add CV child information to `cv_geometry`.
* Add computed pw-constant axial resistivity over cell branches to `fvm_cv_discretization`, so it is available for voltage/current interpolation.
* Extend `location_cv` method: add a new parameter of enum `cv_prefer::type` which instructs `location_cv` on how to resolve locations on CV boundaries.
* Provide pw-constant over cable integration methods for `embed_pwlin`.
* Split out `unique_in_place` to its own header, with unit tests.
* Unit tests to suit.

### DIFF
--- a/arbor/fvm_layout.cpp
+++ b/arbor/fvm_layout.cpp
@@ -1,3 +1,4 @@
+#include <algorithm>
 #include <set>
 #include <stdexcept>
 #include <unordered_set>
@@ -19,6 +20,7 @@
 #include "util/piecewise.hpp"
 #include "util/rangeutil.hpp"
 #include "util/transform.hpp"
+#include "util/unique.hpp"
 
 namespace arb {
 
@@ -124,43 +126,52 @@ pw_elements<U> pw_over_cable(const mcable_map<T>& mm, mcable cable, U dflt_value
 
 // Construct cv_geometry for cell from locset describing CV boundary points.
 cv_geometry cv_geometry_from_ends(const cable_cell& cell, const locset& lset) {
-    struct mloc_hash {
-        std::size_t operator()(const mlocation& loc) const { return loc.branch ^ std::hash<double>()(loc.pos); }
-    };
-    using mlocation_map = std::unordered_map<mlocation, fvm_size_type, mloc_hash>;
     auto pop = [](auto& vec) { auto h = vec.back(); return vec.pop_back(), h; };
 
     cv_geometry geom;
     const auto& mp = cell.provider();
     const auto& m = mp.morphology();
 
-    if (mp.morphology().empty()) {
+    if (m.empty()) {
         geom.cell_cv_divs = {0, 0};
         return geom;
     }
 
-    auto canon =    [&m](mlocation loc)  { return canonical(m, loc); };
-    auto origin =   [&canon](mcable cab) { return canon(mlocation{cab.branch, cab.prox_pos}); };
-    auto terminus = [&canon](mcable cab) { return canon(mlocation{cab.branch, cab.dist_pos}); };
-
     mlocation_list locs = thingify(lset, mp);
 
-    std::vector<msize_t> n_cv_cables;
-    std::vector<mlocation> next_cv_head;
-    next_cv_head.push_back({mnpos, 0});
+    // Filter out root, terminal locations and repeated locations so as to
+    // avoid trivial CVs outside of fork points. (This is not necessary for
+    // correctness, but is for the convenience of specification by lset.)
 
-    mcable_list cables, all_cables;
+    auto neither_root_nor_terminal = [&m](mlocation x) {
+        return !(x.pos==0 && x.branch==(m.branch_children(mnpos).size()>1u? mnpos: 0)) // root?
+            && !(x.pos==1 && m.branch_children(x.branch).empty()); // terminal?
+    };
+    locs.erase(std::partition(locs.begin(), locs.end(), neither_root_nor_terminal), locs.end());
+    util::sort(locs);
+    util::unique_in_place(locs);
+
+    // Collect cables constituting each CV, maintaining a stack of CV
+    // proximal 'head' points, and recursing down branches in the morphology
+    // within each CV.
+
+    constexpr fvm_index_type no_parent = -1;
+    std::vector<std::pair<mlocation, fvm_index_type>> next_cv_head; // head loc, parent cv index
+    next_cv_head.emplace_back(mlocation{mnpos, 0}, no_parent);
+
+    mcable_list cables;
     std::vector<msize_t> branches;
-
-    mlocation_map head_count;
-    unsigned extra_cv_count = 0;
+    geom.cv_cables_divs.push_back(0);
+    fvm_index_type cv_index = 0;
 
     while (!next_cv_head.empty()) {
-        mlocation h = pop(next_cv_head);
+        auto next = pop(next_cv_head);
+        mlocation h = next.first;
 
         cables.clear();
         branches.clear();
         branches.push_back(h.branch);
+        geom.cv_parent.push_back(next.second);
 
         while (!branches.empty()) {
             msize_t b = pop(branches);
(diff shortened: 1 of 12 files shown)


arbor/fvm_layout.hpp
```diff
@@ -68,8 +110,43 @@ struct cv_geometry {
...
+
+        using namespace cv_prefer;
+        switch (prefer) {
+        case cv_distal:
+            break;
+        case cv_proximal:
+            if (loc.pos==cv_prox && i>0) --i;
+            break;
+        case cv_nonempty:
+            if (zero_extent(i)) {
```
bcumming, line 138: Can we had adjacent `zero_extent` CVs? If so, what happens if `i` has zero-length cvs either side?

arbor/fvm_layout.hpp
```diff
@@ -19,15 +19,52 @@
...
+//     a cable in the CV for each branch of the fork, that
+//     includes the fork point.
+//
+//   * If a CV has more than one cable covering a fork point, then
+//     that fork point must belong to that CV.
+//
+// These requirements are a consequence of the CV geometry being determined
+// by a collection of CV boundarary locations.
+
+namespace cv_prefer {
```
bcumming, line 39: Why not use a scoped enum?

arbor/fvm_layout.cpp
```diff
@@ -124,43 +126,52 @@ pw_elements<U> pw_over_cable(const mcable_map<T>& mm, mcable cable, U dflt_value
...
 
-    mlocation_map head_count;
-    unsigned extra_cv_count = 0;
+    // Collect cables constituting each CV, maintaining a stack of CV
+    // proximal 'head' points, and recursing down branches in the morphology
+    // within each CV.
+
+    constexpr fvm_index_type no_parent = -1;
+    std::vector<std::pair<mlocation, fvm_index_type>> next_cv_head; // head loc, parent cv index
+    next_cv_head.emplace_back(mlocation{mnpos, 0}, no_parent);
```
bcumming, line 160: `{mnpos, 0}` instead of `{mnpos, 1}` ?

arbor/fvm_layout.cpp
```diff
@@ -190,69 +201,33 @@ cv_geometry cv_geometry_from_ends(const cable_cell& cell, const locset& lset) {
...
+    arb_assert(n_cv>0);
+    arb_assert(geom.cv_parent.front()==-1);
+    arb_assert(util::all_of(util::subrange_view(geom.cv_parent, 1, n_cv),
+            [](auto v) { return v!=no_parent; }));
 
-        auto cables = util::subrange_view(all_cables, all_cables_index, all_cables_index+n_cables);
-        std::copy(cables.begin(), cables.end(), std::back_inserter(geom.cv_cables));
+    // Construct CV children mapping by sorting CV indices by parent.
+    util::assign(geom.cv_children, util::make_span(1, n_cv));
+    util::sort_by(geom.cv_children, [&geom](auto cv) { return geom.cv_parent[cv]; });
```
bcumming, line 218: What if two CVs have the same parent index? Does the order of their indexes in the sorted list matter?

arbor/morph/embed_pwlin.cpp
```diff
@@ -53,6 +53,32 @@ double integrate(const branch_pw_ratpoly<p, q>& f, unsigned bid, const pw_consta
     return accum;
 }
 
+template <unsigned p, unsigned q>
+double integrate(const branch_pw_ratpoly<p, q>& f, mcable c, const pw_constant_fn& g) {
+    msize_t bid = c.branch;
+    double accum = 0;
+
```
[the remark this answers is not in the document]
halfflat: Yep, good point. I was following a keep it simple for now strategy, but we can certainly revisit it. Would you like a comment so it's marked for later?

arbor/fvm_layout.cpp
```diff
@@ -124,43 +126,52 @@ pw_elements<U> pw_over_cable(const mcable_map<T>& mm, mcable cable, U dflt_value
 
 // Construct cv_geometry for cell from locset describing CV boundary points.
 cv_geometry cv_geometry_from_ends(const cable_cell& cell, const locset& lset) {
-    struct mloc_hash {
-        std::size_t operator()(const mlocation& loc) const { return loc.branch ^ std::hash<double>()(loc.pos); }
-    };
-    using mlocation_map = std::unordered_map<mlocation, fvm_size_type, mloc_hash>;
     auto pop = [](auto& vec) { auto h = vec.back(); return vec.pop_back(), h; };
```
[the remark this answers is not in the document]
halfflat: Ha! I mean, would you like me to change it? It's of course a trivial edit. That said, I don't really understand the widespread mistrust of the comma operator :)

arbor/fvm_layout.cpp
halfflat: It's definitely relying upon the sort putting CV indices together which have the same parent. There is no guarantee of sorting within the CV child list — but only because that wasn't in my use case. Making them sorted is just a matter of using a stable sort instead, and the extra cost is small.

arbor/fvm_layout.hpp
halfflat: Mainly because we can then `use namespace cv_prefer`.

halfflat: At least until we get the ability to bring class enum names into scope with `using`, this seems a more convenient way of doing things — we have the option of bringing them in or not, and it's still not polluting the current namespace.

arbor/fvm_layout.cpp
halfflat: Here, it doesn't matter. But if we move to allowing `{mnpos, 1}` to represent the more-proximal-than-root point, we should use that.

arbor/fvm_layout.hpp
halfflat: This can only occur if we have multiple zero-length CVs at the end of a cable. The CV building algorithm will not make such discretizations, but the `location_cv` code's heuristic is: give the nonempty CV if there is one, otherwise what we have at hand.

arbor/morph/embed_pwlin.cpp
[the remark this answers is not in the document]
bcumming: Agreed, let's not over complicate things. I think a comment would be good, so that one of us is spelunking through there looking for the cause of a performance regression we can save some time.

arbor/fvm_layout.cpp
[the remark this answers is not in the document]
bcumming: Nah, no need to change.

Of course they are mistrusted, look at it! It is winking! Up to no good.

arbor/fvm_layout.hpp
bcumming: Not a pattern I have seen before, and feels like a bit of a long winded way to avoid the verbosity of a scoped enum. When reading the code I see the `using namespace` and these symbols and think "probably an enum in a namespace, maybe?"

arbor/fvm_layout.cpp
bcumming: Agreed, it doesn't matter here, but "parent is on the distal end of parent branch" feels like a good pattern to follow.

arbor/fvm_layout.cpp
bcumming: If the only requirement is that they are sorted by parent, no problem. I was just making sure that this wouldn't cause subtle bugs, or different results if the sort algorithm gives a different result.

arbor/fvm_layout.hpp
halfflat: I disagree with it being long-winded; it's one or two lines of code, that can save egregiously verbose enum value names later. I would like to keep this usage pattern, pending something like [P1099r5](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1099r5.html) getting in.

bcumming: Not long winded, but elaborate. But I understand that you have a strange aversion to scoped enums.

PR #996, opened 2020-04-01
Add Gitpod setup

havogt:


.gitpod.yml
```diff
@@ -0,0 +1,36 @@
...
+tasks:
+  - init: |
+      git submodule update --init --recursive
+      mkdir build
+      cmake -DCMAKE_EXPORT_COMPILE_COMMANDS:BOOL=TRUE -H/workspace/arbor -B/workspace/arbor/build \
+      -DCMAKE_INSTALL_PREFIX:STRING=/workspace/arbor/install -DCMAKE_BUILD_TYPE:STRING=Debug -G "Unix Makefiles" \
+      -DARB_WITH_PYTHON:BOOL=ON
+      ln -s /workspace/arbor/build/compile_commands.json /workspace/arbor
+    prebuild: |
+      cmake --build /workspace/arbor/build/ --parallel $(nproc)
```
bcumming, line 17: Can we make this also `make install`, so that Arbor is installed in the `CMAKE_INSTALL_PREFIX`?

Then we can set the folloing path when the environment starts:
```
export PYTHONPATH=/workspace/arbor/install/lib/python3.7/site-packages
```
So users can start the python module straight away.

havogt: done

PR #998, opened 2020-04-01
Extents don't include the cover of fork points by default.

bcumming:
This PR removes the requirement that all cover points are included in a region.
The motivation is to allow more flexible definion of regions, particularly the proximal and distal sets thereof.
The other motivation is that the author finds it much simpler to reason about, however others find the existing approach easier to reason about.

The changes:
 * `mextent` does not always include points on the cover at fork points.
   * it still enforces that its constituent cables have no intersections and are ordered.
 * a `super` region expression adds the cover points to a region.
 * update `most_proximal` to return the `minset` of the proximal points in a region's cables.
 * fix some cut and paste errors in comments and printing of locset expressions.

Open questions:
 * rename `mextent` because it has moved away from the definition of an extent as cables + cover
 * should we still pass the morphology to extent constructor to check that the extent contains
   only branches on the morphology? This wasn't tested before, which might have been a bug.

Discussion continue about how to represent regions in a way that:
  * Is robust.
  * Handles connections between branches consistently.
  * Allows for economical expression of "all the regions" in user code.
  * That @bcumming and @halfflat really like ;)

This PR aims to a functional stopgap now, while we work on a long term solution, which isn't too far off


arbor/morph/morphology.cpp
```diff
@@ -365,35 +328,6 @@ bool mextent::test_invariants(const morphology& m) const {
     // Too many branches?
     if (!empty() && cables_.back().branch>=m.num_branches()) return false;
```
noraabiakar, line 329: Is this worth an extra `test_invariants` variant?

arbor/morph/region.cpp
```diff
@@ -551,6 +552,73 @@ std::ostream& operator<<(std::ostream& o, const named_& x) {
     return o << "(region \"" << x.name << "\")";
 }
 
+// Region with all segments with projection less than val
```
noraabiakar, line 555: This comment was probably copied from another region

arbor/morph/morphology.cpp
bcumming: I think so, because it asserts that the extent is valid for a given morphology.

arbor/morph/region.cpp
bcumming: Yes. There is also a cut and paste comment  2 lines lower.
I will fix it.

halfflat:
I know this is merged, but I want to note publicly that I don't think is a good idea, and that it makes the semantics of what a region is more complex and less clear. Neither do I think this addresses an urgent need.

PR #999, opened 2020-04-02
Move q10 calculation to initialisation.

thorstenhater:
Pull out a constant computation from the update function to the initialisation step in the HH mechanism. In the busyring benchmark with hh/pas swapped this results in an overall speed-up
of roughly 20%.

### DIFF
--- a/mechanisms/mod/hh.mod
+++ b/mechanisms/mod/hh.mod
@@ -64,7 +64,7 @@ DERIVATIVE states {
 
 PROCEDURE rates(v)
 {
-    LOCAL  alpha, beta, sum, q10
+    LOCAL  alpha, beta, sum
 
     :"m" sodium activation system
     alpha = .1 * vtrap(-(v+40),10)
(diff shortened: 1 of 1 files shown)


mechanisms/mod/hh.mod
```diff
@@ -47,25 +48,24 @@ BREAKPOINT {
...
+    rates(v)
     m' = (minf-m)/mtau
     h' = (hinf-h)/htau
     n' = (ninf-n)/ntau
 }
 
-PROCEDURE rates(v, celsius)
+PROCEDURE rates(v)
 {
     LOCAL  alpha, beta, sum, q10
```
bcumming, line 67: This local variable `q10` is hiding the new assigned variable, and is used instead.
As a result, there is a warning during compilation, and one of the unit tests is not passing.

thorstenhater: Fixed, tests are ok.

PR #1009, opened 2020-04-08
Fix bad param values in partial CV coverage.

### LINKED ISSUE #1008: Incorrect parameter determination for density mechanisms partially present in a CV
The offending code is here:
https://github.com/arbor-sim/arbor/blob/1889a285df185161251c1873f1ea0edffc3c01b3/arbor/fvm_layout.cpp#L603-L643

Essentially, there are two bugs in the code:
* Normalization is performed on the wrong area (CV area instead of mechanism coverage area).
* Parameter values over a CV are integrated by integrating the specified values over the CV with the default value filled in where unspecified, instead of integrating the specified-or-default values over the CV, using the value zero where the mechanism is not present.

halfflat:
* Replace code that determines density mechanism parameter values across a CV with code that (hopefully) does it correctly. New code has the advantage also of being a bit simpler.
* Add unit test `fvm_layout.density_norm_area_partial` that catches the bug, and which validates the fix.
* Change old 'segment' terminology in `fvm_layout` test comments to 'branch'.

Fixes #1008.

### DIFF
--- a/arbor/fvm_layout.cpp
+++ b/arbor/fvm_layout.cpp
@@ -593,6 +593,8 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties& 
         sort(param_names);
 
         std::size_t n_param = param_names.size();
+        param_dflt.reserve(n_param);
+        config.param_values.reserve(n_param);
 
         for (std::size_t i = 0; i<n_param; ++i) {
             const auto& p = param_names[i];
@@ -603,7 +605,7 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties& 
         mcable_map<double> support;
         std::vector<mcable_map<double>> param_maps;
 
-        param_maps.resize(param_names.size());
+        param_maps.resize(n_param);
 
         for (auto& on_cable: entry.second) {
             verify_mechanism(info, on_cable.second);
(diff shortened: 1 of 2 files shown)


arbor/fvm_layout.cpp
```diff
@@ -592,31 +592,32 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties&
         assign(param_names, util::keys(info.parameters));
         sort(param_names);
 
-        for (auto& p: param_names) {
+        std::size_t n_param = param_names.size();
+
+        for (std::size_t i = 0; i<n_param; ++i) {
+            const auto& p = param_names[i];
```
bcumming, line 598: The original `for (const auto& p: param_names)` (with added `const`) is a bit clearer methinks.

The `n_param` could be used to `reserve` capacity in `config.param_values` and `param_dflt`.

arbor/fvm_layout.cpp
```diff
@@ -592,31 +592,32 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties&
...
-                mcable cable = on_cable.first;
-
-                support.insert(cable, 1.);
-                for (auto param_assign: on_cable.second.values()) {
-                    keyed_param_maps[param_assign.first].insert(cable, param_assign.second);
-                }
-            }
-            for (auto& p: param_names) {
-                param_maps.push_back(std::move(keyed_param_maps[p]));
+        param_maps.resize(param_names.size());
```
bcumming, line 606: Use `n_param` now we have it.

halfflat: 👍

arbor/fvm_layout.cpp
halfflat: I'll keep the `n_param` (because of parallel iteration through multiple containers of the same size). The reserve is definitely warranted; will add now.

PR #1011, opened 2020-04-09
Extend repertoire of cable cell probes.

halfflat:
* Add the following scalar (i.e. corresponding to just one raw value)
probe address types for cable cells:
   - `cell_probe_membrane_voltage`, replacing `cell_probe_address::membrane_voltage`;
   - `cell_probe_total_ionic_current_density`, replaing `cell_probe_address::membrane_current`;
   - `cell_probe_density_state`;
   - `cell_probe_point_state`;
   - `cell_probe_ion_current_density`;
   - `cell_prone_ion_int_concentration`;
   - `cell_prone_ion_ext_concentration`.
* Extend backend class interface with `mechanism_field_data` method, that returns a raw (device) pointer to the per-instance values of a mechanism state variable.
* Make unit test for probes backend-generic.
* Add unit tests for new probes.
* Modify `fixed_ica_current.mod` unit test NMODL module to be more generic in naming, so it can be better used in probe unit tests.
* Update examples to suit.

### DIFF
--- a/arbor/backends/gpu/fvm.cpp
+++ b/arbor/backends/gpu/fvm.cpp
@@ -1,0 +1,18 @@
+#include <string>
+
+#include <arbor/mechanism.hpp>
+#include "fvm.hpp"
+#include "mechanism.hpp"
+
+// Provides implementation of backend::mechanism_field_at.
+
+namespace arb {
+namespace gpu {
+
+fvm_value_type* backend::mechanism_field_data(arb::mechanism* mptr, const std::string& field) {
+    arb::gpu::mechanism* m = dynamic_cast<arb::gpu::mechanism*>(mptr);
+    return m? m->field_data(field): nullptr;
+}
+
+} // namespace gpu
+} // namespace arb
--- a/arbor/backends/gpu/fvm.cpp
+++ b/arbor/backends/gpu/fvm.cpp
@@ -4,7 +4,7 @@
 #include "fvm.hpp"
 #include "mechanism.hpp"
 
-// Provides implementation of backend::mechanism_field_at.
+// Provides implementation of backend::mechanism_field_data.
 
 namespace arb {
 namespace gpu {
--- a/arbor/fvm_lowered_cell_impl.hpp
+++ b/arbor/fvm_lowered_cell_impl.hpp
@@ -136,6 +136,15 @@ private:
     void set_gpu() {
         if (context_.gpu->has_gpu()) context_.gpu->set_gpu();
     }
+
+    // Translate cell probe descriptions into probe handles etc.
+    fvm_probe_info resolve_probe_address(
+        std::size_t cell_idx,
+        const util::any& paddr,
+        const fvm_cv_discretization& D,
+        const fvm_mechanism_data& M,
+        const std::vector<target_handle>& handles,
+        const std::unordered_map<std::string, mechanism*>& mech_tbl);
 };
 
 template <typename Backend>
@@ -444,6 +453,9 @@ void fvm_lowered_cell_impl<B>::initialize(
 
     target_handles.resize(mech_data.n_target);
 
+    // Keep track of mechanisms by name for probe lookup.
+    std::unordered_map<std::string, mechanism*> mechptr_by_name;
+
     unsigned mech_id = 0;
     for (auto& m: mech_data.mechanisms) {
         auto& name = m.first;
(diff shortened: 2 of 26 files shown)


arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -136,6 +136,15 @@ class fvm_lowered_cell_impl: public fvm_lowered_cell {
     void set_gpu() {
         if (context_.gpu->has_gpu()) context_.gpu->set_gpu();
     }
+
+    // Translate cell probe descriptions into probe handles etc.
```
bcumming, line 140: Can the comment document the arguments.

arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -649,4 +648,111 @@ fvm_size_type fvm_lowered_cell_impl<B>::fvm_intdom(
...
+
+    struct mechanism_data {
+        const fvm_value_type* state_data = nullptr;
+        unsigned id;
+        explicit operator bool() const { return state_data; }
+    };
+
+    auto lookup_mechanism_data = [&](auto* p) -> mechanism_data {
+        if (mechanism* m = value_by_key(mech_tbl, p->mechanism).value_or(nullptr)) {
+            const fvm_value_type* data = B::mechanism_field_data(m, p->state);
```
bcumming, line 684: What is `B`? This is too far from the definition to make for a really useful single letter name.

arbor/include/arbor/cable_cell.hpp
```diff
@@ -29,14 +29,49 @@ struct lid_range {
...
+};
+
+// Value of state variable `key` in point mechanism `source` at target `target`.
+struct cell_probe_point_state {
+    cell_lid_type target;
+    std::string mechanism;
+    std::string state;
+};
+
+// Current density [A/m²] across membrane attributed to the ion `source` at `location`.
```
bcumming, line 59: What about ion reversal potential? Would that also be handled as a separate type?
What about a `(location, ion, key)` tuple like we have for density mech. state variables?

example/single/single.cpp
```diff
@@ -39,7 +39,7 @@ struct single_recipe: public arb::recipe {
 
     arb::probe_info get_probe(arb::cell_member_type probe_id) const override {
         arb::mlocation mid_soma = {0, 0.5};
-        arb::cell_probe_address probe = {mid_soma, arb::cell_probe_address::membrane_voltage};
+        arb::cell_probe_membrane_voltage probe = {mid_soma};
```
bcumming, line 42: I think I prefer the new interface.

arbor/fvm_lowered_cell_impl.hpp
halfflat: I think the only unclear one would be `mech_tbl`, right? If so, I can just change the name of the parameter. We otherwise only ever have the one `fvm_cv_discretization`, `fvm_mechanism_data`, and vector of `target_handle`.

arbor/backends/gpu/fvm.cpp
```diff
@@ -0,0 +1,18 @@
+#include <string>
+
+#include <arbor/mechanism.hpp>
+#include "fvm.hpp"
+#include "mechanism.hpp"
+
+// Provides implementation of backend::mechanism_field_at.
```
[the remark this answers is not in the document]
halfflat: Yup, stale comment.

arbor/fvm_lowered_cell_impl.hpp
halfflat: All the other member functions defined around here also use `B` for the type parameter. Change them all to `Backend`?

arbor/include/arbor/cable_cell.hpp
halfflat: I prefer keeping the semantics in the type; the sort of data we can pluck from an ion say don't form an open class, so I would prefer not to switch on a value (i.e. enum kind).

example/single/single.cpp
halfflat: 👍

halfflat:
Regarding the use of a value rather than a type for ion attribute selection, I would prefer to keep the selectors as different types, because the sorts of attribute we can select constitute a fixed, finite class, and the values have different semantics: an ionic current density is not commensurable with an ionic concentration.

arbor/include/arbor/cable_cell.hpp
halfflat: I want to exclude revpot, only because the whole precomputed reversal potential thing is a bit of a hack. We can come back to that later if there is a use case of course — adding it is not much work.

PR #1015, opened 2020-04-15
Fix ODR in GPU unit tests and memory bugs.

### LINKED ISSUE #1013: ODR violation un unit tests
The `arbor-config-defs` used to compile the library are used to configure the header file `arbor/backends/gpu/fvm.hpp` with the `ARB_HAVE_GPU_FINE_MATRIX` preprocessor variable.

These definitions are not seen when compiling the unit tests, so we end up with two definitions of `arb::gpu::backend` floating around, and hence very confusing ODR violations in the unit tests.

To be honest, the `backends/gpu./fvm.hpp` could be hidden with pimpl to reduce compilation and recompilation times, but that is a more serious bit of refactoring.

### LINKED ISSUE #1014: Out of bounds access when assembling matrix with more than one cell
When initializing the invariant parts of the Hines matrix, there is an out of bounds access when there is more than one cell in a cell group on both CPU and GPU backends. The cause is the values of -1 in `p[i]` when `i` corresponds to the root of a matrix, e.g.:

```C++
        for (auto i: util::make_span(1u, n)) {
            auto gij = face_conductance[i];

            u[i] = -gij;
            invariant_d[i] += gij;
            invariant_d[p[i]] += gij;
        }
```
here https://github.com/arbor-sim/arbor/blob/master/arbor/backends/multicore/matrix_state.hpp#L64

This has been a letent bug for a while now, and was only caught by a valgrind sweep when investigating an unrelated issue.

bcumming:
* Remove interleaved matrix solver on GPU.
* Fix out of bound memory bug when initialising matrix for all back ends

Fixes #1014 
Fixes #1013


bcumming:
bors try

PR #1016, opened 2020-04-15
Voltage interpolation from CV discretization.

halfflat:
* Add functions for determining CVs and weights for membrane voltage interpolation and axial current determination.
* Add unit tests for voltage interpolation and axial current interpolation.
* Accommodate mextent changes.


bcumming:
bors try

PR #1023, opened 2020-04-29
clean up of region/locset implementation

bcumming:
Add `maxset` function that returns the most distal set of locations in a location list, similar to existing `minset` function.

Use `minset` and `maxset` consistently in `most_proximal` and `most_distal` locset expressions respectively.

### DIFF
--- a/arbor/morph/locset.cpp
+++ b/arbor/morph/locset.cpp
@@ -205,33 +205,12 @@ locset most_distal(region reg) {
 }
 
 mlocation_list thingify_(const most_distal_& n, const mprovider& p) {
+    // Make a list of the distal ends of each cable segment.
     mlocation_list L;
-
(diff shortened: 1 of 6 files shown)


arbor/morph/locset.cpp
```diff
@@ -250,16 +229,13 @@ locset most_proximal(region reg) {
 }
 
 mlocation_list thingify_(const most_proximal_& n, const mprovider& p) {
-    auto extent = thingify(n.reg, p);
```
noraabiakar, line 253: Why remove the assertions here?

bcumming: Because it isn't asserting preconditions on the output of this function: it was asserting the output of the `thingify` code. This wasn't done anywhere else in the region or locset code, and we have unit tests that verify that `thingify` calls are valid.

And it made the code more complex!

PR #1028, opened 2020-05-04
Gpu/optimise matrix fine

thorstenhater:
Memory access optimisations on matrix solver.

Here is the effect on the busyring benchmark (swapped pas -> hh) with 8192 cells on a
V100 GPU (time for model-run in seconds).
```
|--------+-------|
| Before | After |
|--------+-------|
|  2.242 | 2.141 |
|  2.233 | 2.127 |
|  2.218 | 2.144 |
|  2.243 | 2.115 |
|  2.228 | 2.120 |
|--------+-------|
|  2.218 | 2.115 |
|--------+-------|
```


thorstenhater:
bors try

bcumming:
Can you please describe which optimizations were made in the PR description?

thorstenhater:
Sure.

*Assembly*

- converted to using the ternary operator, which results in `fsel` instructions over conditional branches.
- The double indirection `cv -> cell -> intdom` has been squashed into one `cv -> intdom`, saving memory, allocations and data movement to GPU as the indirection arrays are not used elsewhere.

*Solve*
- No FMA instructions where generated for operator `-=`; fixed by using `fma` and factoring out the negation. This saves some instructions.
- induction variables `i` and modifying `pos` has been rolled into one (`pos`)
- Some invariant variables have been extracted from loops and conditionals
- In the first pass, `factor` and `pos` were kept alive between loops, this has been removed.
The latter changes save live registers, allowing for higher occupancy, but are mainly clean-up

arbor/backends/gpu/matrix_fine.cu
```diff
@@ -143,66 +142,67 @@ void solve_matrix_fine(
...
-                const unsigned parent_index = next_lvl_meta.matrix_data_index;
-                const unsigned p = parent_index + lvl_parents[tid];
-                //d[p]   -= factor * u[pos];
-                gpu_atomic_add(d  +p, -factor*u[pos]);
-                //rhs[p] -= factor * rhs[pos];
-                gpu_atomic_add(rhs+p, -factor*rhs[pos]);
+                const auto u_ = u[posN];
+                const auto d_ = d[posN];
+                const auto rhs_ = rhs[posN];
+                const T factor = -u[posN] / d[posN];
```
bcumming, line 165: shouldn't you use `u_` and `d_` in this expression?

arbor/backends/gpu/matrix_fine.cu
```diff
@@ -143,66 +142,67 @@ void solve_matrix_fine(
...
+            if (d[pos0]!=0) {
                 // backward
-                for (unsigned i=0; i<len-1; ++i) {
-                    T factor = u[pos] / d[pos];
-                    const unsigned next_pos = pos + width;
-                    d[next_pos]   -= factor * u[pos];
-                    rhs[next_pos] -= factor * rhs[pos];
-
-                    pos = next_pos;
+                for (int pos = pos0, pos_next = pos0 + width; pos < posN; pos += width, pos_next += width) {
```
bcumming, line 188: The changes to indexing make the code much more difficult to understand, for no obvious benefit.

arbor/backends/gpu/matrix_fine.cu
```diff
@@ -143,66 +142,67 @@ void solve_matrix_fine(
...
+                    const auto u_ = u[pos];
+                    const auto d_ = d[pos];
+                    const auto rhs_ = rhs[pos];
+                    const T factor = -u_ / d_;
+                    const auto rhs_next = rhs[pos_next];
+                    const auto d_next = d[pos_next];
+                    d[pos_next]   = fma(factor, u_, d_next);
```
bcumming, line 195: This feels overly complex.
You could pre load `u[pos]`, `d[pos]` and `rhs[pos]` at the top with a comment saying why.

arbor/backends/gpu/matrix_fine.cu
```diff
@@ -143,66 +142,67 @@ void solve_matrix_fine(
...
 
         const unsigned width = num_matrix[bid];
 
         if (tid < width) {
-            const unsigned len = lvl_lengths[tid];
-            unsigned pos = last_lvl_meta.matrix_data_index + tid;
-
-            if (d[pos]!=0) {
-
+            const int len = lvl_lengths[tid];
```
bcumming, line 183: The changes in the loop below have made the code more complex and difficult to understand.
There might be some value in "pre-loading" some values to ensure that the compiler doesn't have to worry about aliasing.

arbor/backends/gpu/matrix_fine.cu
```diff
@@ -45,43 +45,39 @@ void scatter(const T* from, T* to, const I* p, unsigned n) {
...
-        const T* voltage,
-        const T* current,
-        const T* conductivity,
-        const T* cv_capacitance,
-        const T* area,
-        const I* cv_to_cell,
-        const T* dt_intdom,
-        const I* cell_to_intdom,
-        const I* perm,
+        T* __restrict__ d,
```
bcumming, line 48: We have these `__restrict__` pointers here, and then "pre-load" the values all the same in the loops.

I would remove these `__restrict__` keywords, and write the loops in such a way that the compiler can be certain that there is no aliasing.

arbor/backends/gpu/matrix_fine.cu
```diff
@@ -233,15 +233,17 @@ void solve_matrix_fine(
 
             // Find the index of the first node in this branch.
             const unsigned len = lvl_lengths[tid];
-            unsigned pos = lvl_meta.matrix_data_index + (len-1)*width + tid;
+            const int pos0 = lvl_data_index + tid;
+            const int posN = pos0 + (len - 1)*width;
```
bcumming, line 237: This new approach to indexing makes the for loop logic more complex. I prefer the old approach which explicitly loops over the number of branches. It is clearer and less likely to be buggy.

arbor/backends/gpu/matrix_fine.cu
thorstenhater: As https://github.com/arbor-sim/arbor/pull/1026 has been merged, it might be better to go the other route of keeping the restrict.

arbor/backends/gpu/matrix_fine.cu
thorstenhater: yes

arbor/backends/gpu/matrix_fine.cu
thorstenhater: While refactoring this I felt the exact opposite as it's hard to keep track of the current state of `pos`. This way it is always a straight counting loop and `pos` is not kept alive in between loops. That said, I wholeheartedly agree that the loop boundaries muddle the waters and could be better motivated.

thorstenhater:
bors try

thorstenhater:
bors try

thorstenhater:
bors try

arbor/backends/gpu/matrix_fine.cu
bcumming: I am not a big fan of relying on `__restrict__` when better written versions of kernels (many of which I wrote!) would remove the need for them. In this PR we are adding `__restrict__` and also rewriting loops to avoid reading a value twice that could have been aliased: we should pick one approach and stick to it.

The other big danger of `__restrict__` is that we use it when it is not valid, and create all manner of nasty aliasing bugs.

thorstenhater: Hi,

`const __restrict__` in particular does a bit more for us here, it also enables use of constant memory (quoting a conversation with NVIDIA here). But it only takes effect if all pointers are qualified as `__restrict__`. Also, #1026 has been merged, adding `__restrict__` qualifiers to most kernels. About the invalid use: Most of these kernels are single use, ie there are very few callers and most often only one. That would make abuse pretty obvious and also quite unlikely. I also like that the qualifier comminucates our intend.

That said, there is is only a minor impact, if any and I can remove the qualifiers if needed.

```
|-------+-------|
|  with |   w/o |
|-------+-------|
| 1.996 | 2.005 |
| 2.006 | 2.013 |
| 2.011 | 2.014 |
| 2.014 | 2.014 |
| 2.015 | 2.017 |
| 2.022 | 2.024 |
| 2.024 | 2.024 |
| 2.025 | 2.027 |
| 2.026 | 2.029 |
| 2.042 | 2.046 |
|-------+-------|
```

thorstenhater:
bors try

thorstenhater:
bors try

bcumming (approved):

PR #1030, opened 2020-05-06
Use the safe container based jobs for ci

haampie:
The difference is that there's now one job per slurm setup. CI runs `salloc` and `srun` in the background, pulls in the container and then runs the `script` section inside the container(s).

In case you need more flexibility with multiple build configurations, see https://github.com/eth-cscs/COSMA/blob/master/.gitlab-ci.yml for an example.


haampie:
bors try

haampie:
bors try

Issue #730, opened 2019-04-03
Correct current sampling for FVM cells


halfflat:
Current sampling implementation just takes the mechanism contribution to current, but does not account for capacitive membrane currents.

By virtue of current balance, a better approach might just be to compute the weighted Laplacian of the voltage at the given CV. More generally, the trans-membrane current can be computed as the weighted sum of a number of voltage samples, with the weights fixed at discretization.

This requires a more elaborate representation for sample points in the lowered cell, but we will need that regardless for e.g. interpolated voltage measurements.

halfflat:
Fixed by #1034

Issue #988, opened 2020-03-24
Default environment support in arbor-env


halfflat:
Goal: allow simple environment-based determination of resources for an execution context from `arborenv`.

The motivations are: reduce bolierplate for querying environment variables and threading environment; provide consistent environment variables for user code that takes advantage of this functionality; ease unit testing in multithreaded and GPU contexts.

Proposal: 
* We keep the current arbenv resource functions.
* We remove `OMP_NUM_THREADS` from the environment check.
* `ARB_NUM_THREADS` renamed to `ARBENV_NUM_THREADS` to make it clear it is a functionality from the arbenv library.
* Add `arbenv::default_concurrency` that wraps the environment-check-or-else-thread-count-from-system code.
* Add support for an environment variable `ARBENV_GPU_ID`; if set, and we have GPU support, and it’s within the device count, we use that value in `arbenv::default_gpu()`. A value <0 would effectively mean: do not use the GPU.
* We use this in the unit tests and maybe the examples. As we fix up our GPU test coverage, we can work towards tests being GPU-agnostic, but which will try to use the GPU if available and we run tests multiple times in CI with multithreading and/or GPU enabled via environment.

In the future we can add a facility for marking unit tests that are specific to GPU or multithreaded functionality, so we can filter for them at invocation time.

Related: #982, #983

brenthuisman:
1. Python builds seem to ignore `ARB_NUM_THREADS`. Is that a bug?
2. Does point 4 mean a change from 1-thread by default to `std::thread::hardware_concurrency()` by default?

bcumming:
I would like to remove support for `OMP_NUM_THREADS`, having pushed for it multiple times in the past and been shot down.

Using environment variables to set GPU and thread count should be opt-in, like it is in the proposal, and accessed via something like `arbenv::default_gpu()`, `arbenv::default_concurrency`, etc.

We originally set the default thread count that according to the environment/OS, and removed it.

brenthuisman:
What was the reason for moving back to a default thread count of 1?

`std::thread::hardware_concurrency()` is something we can trust, outside of MPI contexts.

brenthuisman:
Strongly held opinion: by default I expect `context()` to default to the number of threads (on a local machine).

halfflat:
To answer the question regarding `std::thread::hardware_concurrency()`, it's not as trustworthy as it looks, sadly.

[...]

PR #1034, opened 2020-05-12
Extend cable cell probe variety.

### LINKED ISSUE #589: Full-cell sampling for mc_cell
LFP use cases look like they will want to sample membrane voltage or current at the same spatial resolution as the simulation. Setting up probes one for each compartment will comprise a lot of redundant data and make for very long sampler event queues.

Proposed solution:
1. Extend mc_cell probe address to also admit a cell-wide 'vector' sampling.
2. Provide a new sample representation for vector samples that contain the multiple values per time-point.
3. Provide a mechanism for associating the indexed values in the returned vectors with morphological locations (segment index, normalized position) — one option is to have this be supplied directly in the sample representation provided to sampler callbacks.
4. Implement wide-sampling for cpu and gpu back-ends.

I think the only unclear bit is part 3. As we're trying to keep separate the mechanics of the FVM discretization from the morphology specification, we're faced with a trade-off:
* Do we bake in knowledge of the discretization into the `mc_cell` class, allowing this information to be queried from the user code when recipes construct them, for later reference by samplers? This is unattractive from a design point of view, as it increases both coupling in arbor and complexity in user code.
* Do we add sample location information to the information passed to the sampler from `mc_cell_group`? A slightly more generic approach could replace the `tag` data sent to samplers with a pointer/reference to probe meta-data, comprising per-probe constant data including the tag and per-cell-group metadata such as cell gid and morphology info in the `mc_cell` case, wrapped in an `arb::any`.

I'm currently leaning towards the second option, especially as it would simplify sampler usage from the user side (they would no longer have to maintain per-probe metadata themselves for the benefit of samplers).

### LINKED ISSUE #822: Sampling in space vs. interpolation
A membrane voltage probe will give the sampler callback the voltage value from the corresponding CV, but from the user point of view, this comprises a piece-wise constant in space interpolation of the membrane potential. See issue #725 

In order to support user-side interpolation, metadata regarding the discretization will need to be available to user code. We've reserved the possibility of changing the discretization over time, so any such metadata will have to be tied to the sample results that get passed to the callback. This could be an alternative to arbor-side interpolation, or an extension of it.

halfflat:
Fixes #589 and #730, providing cell-wide probes and a correct total trans-membrane current.
Fixes #822, providing spatial interpolation of membrane voltages as governed by cable resistivity and diameter.

* Add partition of point mechanism target vector by cell index to `fvm_mechanism_config`.
* Fix sign error in axial current interpolation.
* Replace `fvm_probe_info` with class that wraps a variant type capturing one of a set of backend-cellgroup probe translation representations: `fvm_probe_scalar`, `fvm_probe_interpolated`, `fvm_probe_multi`, `fvm_probe_weighted_multi`, `fvm_probe_membrance_currents`.
* Refactor `fvm_lowered_cell_impl::resolve_probe_address` so that it avoids the long chain of if-else tests on wrapped `any` type with an invocation of `util::any_visitor`. Each specific cable cell probe address representation is translated into one of the intermediate `fvm_probe_info` variants via an overload of `resolve_probe`.
* Add new non-scalar cell probe address types for all-of-cell values, including point mechanism state, density mechanism state, ion currents and concentrations, membrane voltage by CV cable, and total trans-mebrane currents. The latter are ultimately calculated by computing the net axial current fluxes in each CV.
* Add `util::any_ptr` field to the sampling function interface, pointing to constant-per-simulation-run metadata associated with a probe address. Underlying metadata type varies between probe types.
* Extend `simple_sampler` so that it optionally captures sampler metadata.
* Add special case to `simple_sampler` for vector-valued sample data, allowing it to translate from a pointer-pair range to a std::vector.
* Fix minor bugs in `util::variant`.
* Add to `mc_cell_group` the logic for translating each intermediate probe representation to a sampler callback invocation, reusing temporary scratch vectors across calls.
* Add unit comments to `embed_pwlin_data`.
* Add probe variety demonstration `probe-demo` to examples, covering most of the new cable cell probe classes.
* Add `util::any_visitor` runtime-type dependent dispatcher class.
* Add `util::overload` helper for generating overloaded functional objects from individual functions, lambdas or functionals.
* Extend probe unit tests in `test_probe.cpp` to cover new probe address types.
* Add unit tests for `any_visitor` and `overload`.


halfflat:
I'll make documentation edits in a separate PR, once this PR has been resolved, in case changes to the API are required.

arbor/include/arbor/cable_cell.hpp
```diff
@@ -29,51 +29,153 @@ struct lid_range {
...
 struct cell_probe_ion_int_concentration {
     mlocation location;
     std::string ion;
 };
 
-// Ionic external concentration [mmol/L] of ion `source` at `location`, possibly interpolated.
+// Ionic internal concentration [mmol/L] of ion `source` across components of the cell.
+// Sample value type: `cell_sample_range`
+// Sample metadata type: `mcable_list`
+struct cell_probe_ion_int_concentration_cell {
```
bcumming, line 160: These `cell_X_cell` types are a little difficult to parse.

`cable_X_cell` and `cable_X` to indicate that we are talking about a cable cell specifically?

arbor/include/arbor/cable_cell.hpp
```diff
@@ -29,51 +29,153 @@ struct lid_range {
...
 struct cell_probe_ion_current_density {
     mlocation location;
     std::string ion;
 };
 
-// Ionic internal concentration [mmol/L] of ion `source` at `location`, possibly interpolated.
+// Total ionic current [nA] attributed to the ion `source` across components of the cell.
+// Sample value type: `cell_sample_range`
+// Sample metadata type: `mcable_list`
+struct cell_probe_ion_current_cell {
```
bcumming, line 145: The explosion of types makes the interface unwieldy (according to my personal tastes).

The challenge is to make documentation and an interface that users can quickly get to what they want (particularly the python wrapper) without spending much time having to read docs and look for previous examples.

I think a "stringy" interface might fit the bill. Do you think it should be implemented in C++, on top of these types, or only in the Python wrapper (I would still write the string->probe-type code in C++, but keep it in `arbor/python` to be called directly by python wrapper calls).

arbor/mc_cell_group.cpp
```diff
@@ -79,6 +80,245 @@ void mc_cell_group::set_binning_policy(binning_kind policy, time_type bin_interv
...
+// Working space for computing and collating data for samplers.
+using fvm_probe_scratch = std::tuple<std::vector<double>, std::vector<cell_sample_range>>;
+
+template <typename VoidFn, typename... A>
+void tuple_foreach(VoidFn&& f, std::tuple<A...>& t) {
+    (void)(int []){(f(std::get<A>(t)), 0)...};
+}
+
+void reserve_scratch(fvm_probe_scratch& scratch, std::size_t n) {
+    tuple_foreach([n](auto& v) { v.reserve(n); }, scratch);
```
bcumming, line 104: Given that `tuple_foreach` is used only once, right below where it is defined, I think it would be easier to just say:
```C++
scratch.get<0>().reserve(n);
scratch.get<1>().reserve(n);
```

Looking at `fvm_probe_scratch` and how it is used, it looks like it could be a simple struct. Is there reason it is a `tuple`?

example/probe-demo/probe-demo.cpp
```diff
@@ -0,0 +1,247 @@
...
+    argv0 = argv[0];
+    try {
+        options opt;
+        if (!parse_options(opt, argc, argv)) {
+            return 0;
+        }
+
+        cable_recipe R(opt.probe_addr, opt.n_cv);
+
+        auto context = arb::make_context();
```
bcumming, line 138: Do we explicitly all the new probe features on GPU?

Can we add a flag to "turn on" running the sim on the GPU if one is available?

test/unit/test_probe.cpp
```diff
@@ -472,28 +1099,59 @@ TEST(probe, gpu_v_i) {
     }
 }
 
+TEST(probe, gpu_v_cell) {
+    context ctx = make_context(proc_allocation{1, arbenv::default_gpu()});
+    if (has_gpu(ctx)) {
+        run_v_cell_probe_test<gpu::backend>(ctx);
```
bcumming, line 1105: Well, that answers my question about whether the GPU backend was being tested.

arbor/fvm_layout.cpp
```diff
@@ -433,6 +436,12 @@ fvm_cv_discretization fvm_cv_discretize(const cable_cell& cell, const cable_cell
             D.init_membrane_potential[i] /= D.cv_area[i];
             D.temperature_K[i] /= D.cv_area[i];
         }
+        else if (p!=-1) {
```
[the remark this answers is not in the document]
halfflat: This will only turn up if there is a trivial CV at the root with multiple top level branches. There isn't really a consistent choice of how to pick a starting membrane potential or temperature here, though some heuristic could be picked. Pick the first branch values, perhaps?

arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -648,111 +653,367 @@ fvm_size_type fvm_lowered_cell_impl<Backend>::fvm_intdom(
...
-        }
+    r.shrink_to_fit();
+    return r;
+}
+
+template <typename B>
+fvm_probe_info resolve_probe(const cell_probe_total_current_cell& p, probe_resolution_data<B>& R) {
+    fvm_probe_membrane_currents r;
+
+    auto cell_cv_ival = R.D.geometry.cell_cv_interval(R.cell_idx);
```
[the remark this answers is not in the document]
halfflat: `D` for the `fvm_discretiazation` is used consistently through `fvm_layout.cpp` and the rest of `fvm_lowered_cell_impl.hpp`. Using a longer name makes some of these expressions even more unwieldy than they are, and I'm pretty sure anyone with their hands in the code will quickly see how `D` is used.

We can make it longer, but I don't think it will consequently aid readability.

arbor/include/arbor/cable_cell.hpp
halfflat: Happy to rename `cell_probe_FOO` to `cable_probe_FOO`.

arbor/include/arbor/cable_cell.hpp
halfflat: For the C++ API, it's this, or an enum together with a bunch of enum-contigent variant values. The type interface is, I am certain, clearer and easier to follow than the alternative.

What we do in Python though is another story.

arbor/mc_cell_group.cpp
halfflat: At one point, there were more vectors in scratch, and a struct was getting unwieldy and copy/pastey.

With just the two, a struct is not bad. But the tuple approach has the benefit of being easy to generalize. It doesn't strike me as being particularly obscure as a tuple, but that's probably a matter of taste.

example/probe-demo/probe-demo.cpp
halfflat: I'm not sure I follow what you'd prefer here ... as things stand, the CPU ones are always tested, and the GPU versions are tested if there is one available and we've built GPU support in.

test/unit/test_probe.cpp
halfflat: Ah, yes :)

example/probe-demo/probe-demo.cpp
halfflat: Regarding the probe example, this comes back to the ongoing debate about the easy context stuff, #988.

arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -648,111 +653,367 @@ fvm_size_type fvm_lowered_cell_impl<Backend>::fvm_intdom(
...
+// `resolve_probe_address` collates this data into a `probe_resolution_data`
+// struct which is then passed on to the specific resolution procedure
+// determined by the type of the user-supplied probe address.
+
+template <typename Backend>
+struct probe_resolution_data {
+    typename Backend::shared_state* state;
+    const cable_cell& cell;
+    const std::size_t cell_idx;
+    const fvm_cv_discretization D;
```
halfflat, line 669: This should be a const ref.

halfflat:
Updated PR addressing review (see commit comments on 57d8bcce2e27d5bc1ae59cddf52dd76b5e8fbd5f), plus added documentation for the sampling API changes and new probe types.

PR #1042, opened 2020-05-15
Add optimised HH model.

thorstenhater:
This adds an optimised version of the HH mechanism. 

Speeds up the busyring benchmark with 1024 on my laptop by roughly 20%.

All procedures have been inlined, and many algebraic optimisations have been
made. Saves most of the memory traffic for `*inf` etc.

### DIFF
--- a/mechanisms/mod/hh.mod
+++ b/mechanisms/mod/hh.mod
@@ -63,30 +63,30 @@ INITIAL {
 }
 
 DERIVATIVE states {
-    LOCAL alpha, beta, sum, minf, ninf, hinf, mtau, ntau, htau
+    LOCAL alpha, beta, sum, minf, ninf, hinf, mrate, nrate, hrate
 
     :"m" sodium activation system
     alpha = exprelr(-(v+40)*0.1)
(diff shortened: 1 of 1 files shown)


mechanisms/mod/hh.mod
```diff
@@ -26,70 +25,68 @@ STATE {
...
+    LOCAL alpha, beta, sum, minf, ninf, hinf, mtau, ntau, htau
 
     :"m" sodium activation system
-    alpha = .1 * vtrap(-(v+40),10)
-    beta =  4 * exp(-(v+65)/18)
+    alpha = exprelr(-(v+40)*0.1)
+    beta =  4 * exp(-(v+65)*0.05555555555555555)
     sum = alpha + beta
-    mtau = 1/(q10*sum)
+    mtau = q10*sum
```
halfflat, line 72: As these are now reciprocals of a time constant rather than a time constant, the names `mtau` etc. are misleading. Could they be renamed? e.g. `oo_mtau` or `mrate`?

thorstenhater: Good idea.

Issue #1021, opened 2020-04-27
SVE Backend


noraabiakar:
Add an SVE simd backend. 
The intrinsics are fairly similar to neon, however the vector types are _sizeless_. This means that our current method of doing things using the `sim_impl` class doesn't work. 

Proposed solution: 
- Continue using `simd_impl` for the already supported architectures. 
- For `sizeless` vector types, skip the `simd_impl` wrapper and implement the needed functions as free functions.
- Choose between `simd_impl` and free functions based on the abi `width` of the vectors. Consider vectors of `width = 0` to be sizeless vectors. 
- Implement an `indirect` struct that mirrors our current `indirect_expression` struct used for accessing memory according to an external index. The `indirect` struct has several purposes 
    1) It creates a consistent interface alongside the current `indirect_expression` indexed struct. 
    2) It's a good replacement for `simd_impl` constructors.
    3) Arguably the most important use of the `indirect` struct is the getting access to the width of the vector we are operating on (used to read/write the correct amount of data from memory using masks).
- Changes to the API will have to be reflected in the generated code from modcc.
- This may include some refactoring of `simd.hpp`.
- The actual width of the SVE vectors can be set at runtime via a `prctl(...)` call. This can be used both to our disadvantage and detriment, so we have to be careful not to make any assumptions about the width of the SVE vectors.

PR #1044, opened 2020-05-19
Refactor simd API and add SVE backend 

noraabiakar:
The simd wrapper `simd_impl` is incompatible with the new sizeless SVE vectors. As a workaround, we can add a second simd API that can work for both sizeless vectors and regular SVE vectors. 
The new API has less features, but is complete enough to compile all our `nmodl` files as well as other more complex `nmodl` files external to Arbor. 

Changes:
- `indirect_expressions`, and `indirect_indexed_expressions` control memory reads/writes. `where_expressions` control masked access to simd vectors.
- For the new API, conversion into simd vector types from other types has to go through `simd_cast<simd_vector_type>`
- For  the new API overloading operators is not possible, function calls need to be used instead. 
- For SVE, the mechanism is responsible for the selecting the vector width. Arbor checks the validity of the vector width at the beginning of each API call. The selected vector width can be less than or equal to the maximum supported vector width. 
- Modcc vectorizes code using the new API
- Unit tests for the old API are updated and new unit tests for the new API are added 
- SVE backend is added and has been tested.

Addresses #1021


noraabiakar:
bors try

bcumming:
bors try

bcumming:
bors try

arbor/include/arbor/simd/sve.hpp
```diff
@@ -0,0 +1,900 @@
...
+template <> struct sve<int, 0> {using type = detail::sve_int;};
+
+template <typename T> struct reg_type;
+template <> struct reg_type<int> { using type = svint64_t;};
+template <> struct reg_type<double> { using type = svfloat64_t;};
+
+template <typename T> struct mask_type;
+template <> struct mask_type<int> { using type = svbool_t;};
+template <> struct mask_type<double> { using type = svbool_t;};
+
```
halfflat, line 633: `reg_type`, `mask_type` and `type_to_impl` should live in `simd::detail` rather than `simd::simd_abi` (so that only ABI names live in `simd_abi`). That said, I don't think `reg_type` and `mask_type` are necessary: the `simd_wrap` can use the `simd_traits` directly, e.g.
```
template <typename Value>
struct simd_wrap<Value, 0u, simd_abi::sve> {
    using type = typename detail::simd_traits<typename simd_abi::sve<Value, 0u>::type>::vector_type;
};
```

arbor/include/arbor/simd/sve.hpp
```diff
@@ -0,0 +1,900 @@
...
+template<typename Type> struct type_to_impl;
+template<> struct type_to_impl<svint64_t> { using type = detail::sve_int;};
+template<> struct type_to_impl<svfloat64_t> { using type = detail::sve_double;};
+template<> struct type_to_impl<svbool_t> { using type = detail::sve_mask;};
+};  // namespace simd_abi
+
+using namespace arb::simd::simd_abi;
+using arb::simd::detail::simd_traits;
```
halfflat, line 641: These `using` statements will lift things into `arb::simd` that don't really belong there. I'd recommend renaming `type_to_impl` as `sve_type_to_impl`, putting it in the `detail` namespace, and using it qualified below.

arbor/include/arbor/simd/sve.hpp
```diff
@@ -0,0 +1,900 @@
...
+
+template <typename Value, unsigned N, template <class, unsigned> class Abi>
+struct simd_mask_wrap;
+
+template <typename Value, template <class, unsigned> class Abi>
+struct simd_mask_wrap<Value, (unsigned)0, Abi> { using type = typename simd_abi::mask_type<Value>::type; };
+
+// Math functions exposed for SVE types
+
+#define SVE_UNARY_ARITHMETIC(name)\
```
halfflat, line 657: As these macros are for internal use but populate the global namespace, I'd prefix them with `ARB_` to help avoid collisions, and put a `_` suffix on the end, to indicate that they are only used internally. They can in fact be `#undef`ed after their use is finished.

arbor/include/arbor/simd/simd.hpp
```diff
@@ -17,41 +18,336 @@ namespace detail {
...
+template <typename Impl>\
+simd_impl<Impl> name(const simd_impl<Impl>& a, typename simd_impl<Impl>::scalar_type b) {\
+    return simd_impl<Impl>::wrap(Impl::name(a.value_, Impl::broadcast(b)));\
+};\
+template <typename Impl>\
+simd_impl<Impl> name(const typename simd_impl<Impl>::scalar_type a, simd_impl<Impl> b) {\
+    return simd_impl<Impl>::wrap(Impl::name(Impl::broadcast(a), b.value_));\
+};
+
+#define BINARY_COMPARISON(name)\
```
halfflat, line 72: I made a similar comment in `sve.hpp` — these macros should probably be named along the lines of `ARB_BINARY_COMPARISON_` etc. so as to avoid collisions and to indicate (with the final '_') that they are for internal use only. They can be undefined after use.

arbor/backends/multicore/shared_state.cpp
```diff
@@ -168,27 +169,34 @@ void shared_state::ions_init_concentration() {
...
 void shared_state::update_time_to(fvm_value_type dt_step, fvm_value_type tmax) {
+    using simd::simd_cast;
+    using simd::indirect;
+    using simd::add;
+    using simd::min;
     for (fvm_size_type i = 0; i<n_intdom; i+=simd_width) {
-        simd_value_type t(time.data()+i);
-        t = min(t+dt_step, simd_value_type(tmax));
-        t.copy_to(time_to.data()+i);
+        simd_value_type t = simd_cast<simd_value_type>(indirect(time.data()+i, simd_width));
```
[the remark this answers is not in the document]
noraabiakar: I prefer the first method, I'll add an `assign` function to the API. This will require quite a few changes to integrate into the cprinter correctly. Maybe we can add an issue for that?

arbor/backends/multicore/shared_state.cpp
```diff
@@ -26,9 +26,10 @@
 namespace arb {
 namespace multicore {
 
-constexpr unsigned simd_width = simd::simd_abi::native_width<fvm_value_type>::value;
-using simd_value_type = simd::simd<fvm_value_type, simd_width>;
-using simd_index_type = simd::simd<fvm_index_type, simd_width>;
+constexpr unsigned vector_length = (unsigned) simd::simd_abi::native_width<fvm_value_type>::value;
+using simd_value_type = simd::simd<fvm_value_type, vector_length, simd::simd_abi::default_abi>;
+using simd_index_type = simd::simd<fvm_index_type, vector_length, simd::simd_abi::default_abi>;
+int simd_width = simd::width(simd::simd_cast<simd_value_type>(0));
```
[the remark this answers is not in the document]
noraabiakar: Yes, I agree it's a better way

arbor/include/arbor/simd/native.hpp
```diff
@@ -87,10 +92,14 @@ struct native_width;
 
 template <typename Value, int k>
 struct native_width {
-   static constexpr int value =
-        std::is_same<void, typename native<Value, k>::type>::value?
-        native_width<Value, k/2>::value:
-        k;
+#ifdef __ARM_FEATURE_SVE
+    static constexpr int value = 0;
+#else
```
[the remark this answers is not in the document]
noraabiakar: I agree, it looks better

arbor/include/arbor/simd/sve.hpp
noraabiakar: Okay, I agree with both points

arbor/include/arbor/simd/sve.hpp
noraabiakar: Okay

arbor/include/arbor/simd/sve.hpp
noraabiakar: Okay

arbor/include/arbor/simd/simd.hpp
noraabiakar: Okay

noraabiakar:
bors try

PR #1047, opened 2020-05-26
Modcc/from to block

thorstenhater:
Add parsing for FROM ... TO ... descriptions in parser.

### DIFF
--- a/modcc/parser.cpp
+++ b/modcc/parser.cpp
@@ -402,8 +402,13 @@ void Parser::parse_state_block() {
         get_token();
 
         if(token_.type == tok::from) {
-            // silently skip
+            // silently skips from/to
             from_to_description();
+            if (status_ == lexerStatus::error) {
+                error(pprintf("range description must be of form FROM <int> TO <int>: '%'", token_));
+                return;
+            }
+
         }
 
         // get unit parameters
@@ -767,6 +772,10 @@ std::pair<Token, Token> Parser::range_description() {
     }
 
     get_token();
+    if(token_.type != tok::integer) {
+        error(pprintf("range description must be <int, int>, found '%'", token_));
+        return {};
+    }
     lb = token_;
 
     get_token();
@@ -776,6 +785,10 @@ std::pair<Token, Token> Parser::range_description() {
     }
 
     get_token();
+    if(token_.type != tok::integer) {
+        error(pprintf("range description must be <int, int>, found '%'", token_));
+        return {};
+    }
     ub = token_;
 
     get_token();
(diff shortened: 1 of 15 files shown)


modcc/parser.cpp
```diff
@@ -782,6 +786,31 @@ std::pair<Token, Token> Parser::range_description() {
...
+std::pair<Token, Token> Parser::from_to_description() {
+    Token lb, ub;
+
+    if(token_.type != tok::from) {
+        error(pprintf("range description must start with a FROM '%'", token_));
+        return {};
+    }
+
+    get_token();
+    lb = token_;
```
bcumming, line 798: Test that `token_` is a value?

modcc/parser.cpp
```diff
@@ -399,9 +399,13 @@ void Parser::parse_state_block() {
         }
 
         parm.token = token_;
-        //state_block.state_variables.push_back(token_.spelling);
         get_token();
 
+        if(token_.type == tok::from) {
```
bcumming, line 404: Should this check for an incorrectly formed `FROM <value> To <value>` statement?
If an error was set, it should return, I think.

modcc/parser.cpp
thorstenhater: Good idea. We should add this to <lo, hi> ranges as well.

modcc/parser.cpp
bcumming: This part is right, and you should call `from_to_description()` should call `error` if it detects an error in the statement.
You can then check if an error was set, and just `return`.

modcc/parser.cpp
```diff
@@ -399,9 +399,17 @@ void Parser::parse_state_block() {
...
+            // silently skips from/to
+            from_to_description();
+            if (status_ == lexerStatus::error) {
+                error(pprintf("range description must be of form FROM <int> TO <int>: '%'", token_));
```
bcumming, line 408: I don't think you need to set the error message here: it was already set in the call to `from_to_description()`, right?

modcc/parser.cpp
```diff
@@ -820,6 +869,14 @@ expression_ptr Parser::parse_prototype(std::string name=std::string()) {
 
         get_token(); // consume the identifier
 
+        // args may have a unit attached
+        if(token_.type == tok::lparen) {
+            unit_description();
+            if(status_ == lexerStatus::error) {
```
bcumming, line 875: `unit_description` has already set the error message, so we can just return.

thorstenhater:
Done in #1050

PR #1054, opened 2020-06-03
New recipe probe API

### LINKED ISSUE #957: More generic probe location specification
Currently the probe interface associates each `probe_id` with a single `mlocation`. This isn't sufficient to support the following sampling scenarios:

1. sampling on a set of locations defined by a `locset`, particularly if the number of locations in the locset is not know before the model is instantiated (which is possible with `join`s and random number generators).
2. sampling voltage over an entire cell (or region).

halfflat:
Resolves #957.

* Make recipe return probes as a vector.
* Remove `probe_id` from `probe_info`.
* Rename `fvm_probe_info` to `fvm_probe_data` (everything was being called info, and it was getting confusing).
* Make `probe_association_map` specific to `mc_cell_group`/`fvm_lowered_cell`.
* Change probe_association_map representation: unordered map for probe_id -> tag; unordered multimap for probe_id -> fvm_probe_data. This allows multiple probes to be associated with the same probe id.
* Call sampler callback separately for each probe with the same probe_id.
* Replace location-based probes with locset equivalents.
* Add index for probes sharing a probe id.
* Bundle all probe metadata (id, tag, index, probe-specific meta) into `probe_metadata` struct for passing to sampler callbacks.
* Change simple_sampler to work on `trace_vector`, a vector of `trace_data`. The _i_th element is the data from probe with index i.
* Consolidate hash composition and `std::hash` specialization code in new header.
* Update python lib for new API.
* Update tests and examples for new recipe, internal probe, and simple_sampler APIs.
* Update docs to suit.

### DIFF
--- a/arbor/include/arbor/cable_cell.hpp
+++ b/arbor/include/arbor/cable_cell.hpp
@@ -15,6 +15,7 @@
 #include <arbor/morph/mprovider.hpp>
 #include <arbor/morph/morphology.hpp>
 #include <arbor/morph/primitives.hpp>
+#include <arbor/util/hash_def.hpp>
 #include <arbor/util/typed_map.hpp>
 
 namespace arb {
@@ -52,7 +53,10 @@ using cable_sample_range = std::pair<const double*, const double*>;
 //     * `cable_probe_point_info` for point mechanism state queries;
 //     * `mcable_list` for most vector queries;
 //     * `std::vector<cable_probe_point_info>` for cell-wide point mechanism state queries.
-
+//
+// Scalar probes which are described by a locset expression will generate multiple
+// calls to an attached sampler, one per valid location matched by the expression.
+//
 // Metadata for point process probes.
 struct cable_probe_point_info {
     cell_lid_type target;   // Target number of point process instance on cell.
@@ -64,7 +68,7 @@ struct cable_probe_point_info {
 // Sample value type: `double`
 // Sample metadata type: `mlocation`
 struct cable_probe_membrane_voltage {
-    mlocation location;
+    locset locations;
 };
 
 // Voltage estimate [mV], reported against each cable in each control volume. Not interpolated.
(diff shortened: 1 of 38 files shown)


arbor/include/arbor/cable_cell.hpp
```diff
@@ -64,7 +68,7 @@ struct cable_probe_point_info {
 // Sample value type: `double`
 // Sample metadata type: `mlocation`
```
thorstenhater, line 69: Should this read `locset` instead?

halfflat: Possibly surprisingly, no: the locset is evaluated to some collection of (thingified) locations; there will be a membrane voltage probe at each location, and that location will constitute the probe-specific metadata.

arbor/include/arbor/simple_sampler.hpp
```diff
@@ -22,16 +22,76 @@ struct trace_entry {
...
+    using base::clear;
+    using base::resize;
+    using base::push_back;
+    using base::emplace_back;
+    using base::operator[];
 };
 
 template <typename V>
-struct trace_data<V, void>: public std::vector<trace_entry<V>> {};
+struct trace_data<V, void>: public std::vector<trace_entry<V>> {
```
halfflat, line 49: Hmm. That `public` should be `private`. Will fix.

arbor/include/arbor/simple_sampler.hpp
```diff
@@ -22,16 +22,76 @@ struct trace_entry {
...
+// the ith element will correspond to the sample data obtained
+// from the probe with index i.
+//
+// The overriden method `at(i)` returns a const reference to the ith
+// element if it exists, or else to an empty trace_data value.
+
+template <typename V, typename Meta = void>
+struct trace_vector: private std::vector<trace_data<V, Meta>> {
+    const trace_data<V, Meta>& at(std::size_t i) const {
+        return i<this->size()? (*this)[i]: empty_trace;
```
bcumming, line 77: Why return an empty trace instead of throwing an exception?
`::at` methods in STL are expected to throw on an out of bounds check, so this might be a bit misleading.

halfflat: Yeah, this is a good point. I chose `at` because it is nice and short, and I don't want to make it a chore to access a trace. On the other hand, if I know a locset has 3 locations, and I want to get the trace from the 3rd one, I don't want things to blow up if it turns out there were no samples taken for that location.

Can you think of a better name for the method?

halfflat: Okay, let's use `get` (pers. comm.)

PR #1063, opened 2020-06-12
No sphere here

### LINKED ISSUE #1060: Expunge spherical somata from morphology.
What does NEURON do, when given a spherical soma and apical and basal dendrites from an SWC file. Where does it attach what, when it discretizes?

The spherical root special case in the Arbor morphology code complicates a bunch of things, and still leaves various cases open. Should apical dendrites be attached to 'the other end' of the root (soma) branch? How much special casing would we need when reading SWC files?

Spherical roots, when discretized, are treated as squat cylinders in the current Arbor implementation, and that's unlikely to change. So let's simplify things:

1. Morphologies are described by cable segments only, which are in turn defined unambiguously by pairs of parent-adjacent samples. Anything to do with spherical roots is expunged.
2. Sample trees with a single sample are ill-defined.
3. Conversion from a spherical soma to a squat cylinder cable is done in SWC parsing, if required.
4. If we want to support SWC directly in the C++ library, we can interpret it effectively as we currently do: anything attached to a spherical root is attached at the same point.
5. More elaborate interpretations or outright manglings of SWC data are relegated to client-side; whatever NEURON does, if it's not trivial, can go in the Python library.

bcumming:
Remove support for spherical segments, and by extension "spherical somas".

* remove `cv_olicy_flag::single_root_cv`, which will only produce the intended results when all all cables attached to the soma are separate branches.
* remove `arb::math::{volume/area}_sphere`, which were not being used anywhere in the code before this.
* `arb::morphology`:
  * remove constructor that auto-detects whether the cell has spherical root
  * remove `spherical_root` query from API
  * remove spherical root logic from implementation
* remove a fair bit of special casing for spheres from different locations, such as the `pw_lin` code.
* arbor exception types print parents as `mnpos` instead of numeric values where appropriate
* fixed a memory violation (out of bounds access) in the sample tree on invalid use input
* update unit tests and python wrapper to reflect the changes.

Fixes #1060

### DIFF
--- a/arbor/morph/region.cpp
+++ b/arbor/morph/region.cpp
@@ -152,15 +152,6 @@ mextent thingify_(const tagged_& reg, const mprovider& p) {
 
     for (msize_t i: util::make_span(nb)) {
         auto ids = util::make_range(m.branch_indexes(i)); // range of sample ids in branch.
-        size_t ns = util::size(ids);        // number of samples in branch.
-
-        if (ns==1) {
-            // The branch is a spherical soma
-            if (samples[0].tag==reg.tag) {
-                L.push_back({0,0,1});
-            }
-            continue;
-        }
 
         // The branch has at least 2 samples.
         // Start at begin+1 because a segment gets its tag from its distal sample.
--- a/arbor/morph/sample_tree.cpp
+++ b/arbor/morph/sample_tree.cpp
@@ -28,20 +28,13 @@ void sample_tree::reserve(msize_t n) {
 }
 
 msize_t sample_tree::append(msize_t p, const msample& s) {
-    std::cout << "+++++ " << (empty()? "empty": "initl") << " " << p << " " << size() << " " << s << std::endl;
-    //if ((empty() && p!=mnpos) || p>=size()) {
-    //   if (p!=mnpos) throw invalid_sample_parent(p, size());
-    //}
-
+    // Assert that the root (first) sample in a tree has mnpos as its parent.
     if (empty() && p!=mnpos) {
         throw invalid_sample_parent(p, size());
     }
     else if (!empty()) {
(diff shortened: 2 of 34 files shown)


bcumming:
bors try

arbor/morph/region.cpp
```diff
@@ -152,15 +152,6 @@ mextent thingify_(const tagged_& reg, const mprovider& p) {
...
-
-        if (ns==1) {
-            // The branch is a spherical soma
-            if (samples[0].tag==reg.tag) {
-                L.push_back({0,0,1});
-            }
-            continue;
-        }
 
         // The branch has at least 2 samples.
```
bcumming, line 156: Might want to replace this comment with an assertion that this is, indeed, the case.

arbor/morph/sample_tree.cpp
```diff
@@ -28,8 +28,15 @@ void sample_tree::reserve(msize_t n) {
 }
 
 msize_t sample_tree::append(msize_t p, const msample& s) {
-    if ((empty() && p!=mnpos) || p>=size()) {
-        if (p!=mnpos) throw invalid_sample_parent(p, size());
+    // Assert that the root (first) sample in a tree has mnpos as its parent.
```
bcumming, line 31: This fixes a nasty bug (seg fault) caused when trying to insert a sample with `mnpos` as its parent when it wasn't a root.

bcumming:
bors try

halfflat:
I know that the segment-based approach is in the works, and I'll finish the review when that's up, but basically this looks great.
Only concern at this point is that there may be an issue building morphologies from SWC records which follow a single-point soma convention.

bcumming:
Clonsing because these changes have been rolled into #1083 1083

PR #1069, opened 2020-06-24
Nmodl/allen

thorstenhater:
Add a new catalogue comprising optimised version of some of the mechanisms commonly used in models of the Allen Cell Types Database, eg http://celltypes.brain-map.org/mouse/experiment/electrophysiology/323535089.
This made required extending modcc to some new NMODL features, as well as to expose some optimisations.

In addition clean up the default catalogue by moving test mechanisms to their proper place in the unit tests.

PR #1071, opened 2020-06-24
Allen catalogue

thorstenhater:
Replaces #1069


mechanisms/generate_catalogue
```diff
@@ -77,7 +83,7 @@ def parse_arguments():
     return vars(parser.parse_args())
 
 
-def generate(modpfx='', arbpfx='', modules=[], backends=[], **rest):
+def generate(modpfx='', arbpfx='', modules=[], backends=[], catalogue='default', **rest):
```
bcumming, line 86: Should we make the catalogue name default `'default'`, or force the user to always specify a name.
My preference is to not have a default name.

thorstenhater: Agreed

thorstenhater:
@noraabiakar yes, fixed

thorstenhater:
bors try

PR #1074, opened 2020-06-25
Python/allen catalogue

thorstenhater:
Expose catalogues to Python, second part.

For practical use, mechanism from multiple catalogues needed to be combined.
This functionality is added and exposed to Python. 

Minor clean-up to avoid lumping all generated mechanisms together (name collisions).

The global Allen catalogue is made accessible.

### DIFF
--- a/arbor/include/arbor/mechcat.hpp
+++ b/arbor/include/arbor/mechcat.hpp
@@ -107,7 +107,7 @@ public:
     }
 
    // Copy over another catalogue's mechanism and attach a -- possibly empty -- prefix
-   void insert(const mechanism_catalogue& other, const std::string& prefix);
+   void import(const mechanism_catalogue& other, const std::string& prefix);
 
    ~mechanism_catalogue();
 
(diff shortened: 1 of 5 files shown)


halfflat:
I'd like to pin down and/or change the `insert` semantics:

Derived mechanisms refer to their parent implementations by name, and accumulate overridden settings by walking up the derivation tree; the mechanism info associated with a derivation though is cached in the map entry. If a parent gets swapped out for another entry further up the tree due to an insert, existing derived mechanisms will accumulate different data or will refer to a different implementation than before, and further the derived mechanism info will be out of sync.

One way to sidestep this is to make it an error to add a mechanism that already has an entry, as is the case with `add()` and `derive()`.  In my opinion, this offers both the simplest semantics and the simplest correct implementation. It also aligns more closely with the use of `insert` in the standard library, which won't overwrite.

I also prefer the name `import` rather than `insert`, as we're dragging in a whole other catalogue and giving it a prefix etc.

thorstenhater:
That is a really good point, I will add a check towards this.

thorstenhater:
bors try

thorstenhater:
Hey, I pulled in your suggestion, looks good to me and passes the tests here.

PR #1075, opened 2020-06-29
Bump ci

haampie:
~Turns out that the experimental inline caching feature (BUILDKIT_INLINE_CACHE) only works when subsequent builds happen on the same host / node. That's not really something we can assume to happen. Instead using `docker pull`-`docker build`-`docker push` caching should just work fine :).~ in fact this is still OK, after some testing and discussion with people, so disregard this.

Also this PR removes `resource_group` which ensures that only a single job runs on daint. I had added this because previously there were some CI failures potentially due to flooding slurm with jobs, but it seems that's currently not much of an issue anymore.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -49,6 +49,7 @@ variables:
   SLURM_JOB_NUM_NODES: 2
   SLURM_PARTITION: normal
   SLURM_TIMELIMIT: '15:00'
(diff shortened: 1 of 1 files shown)


haampie:
bors try

haampie:
bors try

haampie:
bors try

Issue #1076, opened 2020-06-30
Warning: uninitialized variables 


noraabiakar:
Compiling the generated vectorized c++ code of mechanisms can generate warnings of variables being used undefined. 
This is common when initializing a variable depends on a condition, for example:
```
if (a < 0) {
   b = 1
} else {
   b = 2
}
```
This gets translated into the following vectorized code (simplified):
```
{
    auto mask_vec = vec_a < 0;
    where(mask_vec, vec_b) = 1;
    where(mask_vec, !vec_b) = 2;
}
```
`vec_b` is originally not initialized before being used as an argument to the `where` function call, and the compiler generates a warning. The use of an uninitialized variable is not incorrect in this case, but the warning can be eliminated by initializing temporary variables to zero before their first use.

halfflat:
Unfortunately we don't have a handy attribute for this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=55288

PR #1081, opened 2020-07-03
Fix compiler warnings for simd code 

noraabiakar:
initialize variables generated when expanding and inlining functions to zero before first use. 
 
Addresses #1076


halfflat:
Just one query: does the redundant initialization make it through to the generated assembly? (I.e. are we doing more actual work in the mechanisms?)

noraabiakar:
I tried on a couple of simple examples with clang and gcc, the initialization get optimized out when using -O3

Issue #1077, opened 2020-06-30
Use `assign` in generated vectorized code


noraabiakar:
To reduce the verbosity of generated vectorized mechanism code, use the `assign` function (already available for all backends).
Replaces: 
```
simd_type t = simd_cast<simd_type>(indirect(data, simd_width));
```
with:
```
simd_type t;
assign(t, indirect(data, simd_width))
```

PR #1082, opened 2020-07-03
Use `simd::assign` in generated simd code 

noraabiakar:
Use `simd::assign` to reduce verbosity of simd variable assignments from memory.

Addresses #1077

### DIFF
--- a/modcc/printer/cprinter.cpp
+++ b/modcc/printer/cprinter.cpp
@@ -182,7 +182,8 @@ std::string emit_cpp_source(const Module& module_, const printer_options& opt) {
             "namespace S = ::arb::simd;\n"
             "using S::index_constraint;\n"
             "using S::simd_cast;\n"
-            "using S::indirect;\n";
+            "using S::indirect;\n"
+            "using S::assign;\n";
 
         out << "static constexpr unsigned vector_length_ = ";
         if (opt.simd.size == no_size) {
@@ -685,7 +686,8 @@ void emit_simd_state_read(std::ostream& out, LocalVariable* local, simd_expr_con
                 << "[0]);\n";
         }
         else if (constraint == simd_expr_constraint::contiguous) {
-            out << " = simd_cast<simd_value>(indirect(" <<  d.data_var
+            out << ";\n"
+                << "assign(" << local->name() << ",indirect(" <<  d.data_var
                 << " + " << d.index_var
                 << "[index_], simd_width_));\n";
         }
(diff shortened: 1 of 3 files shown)


modcc/printer/cprinter.cpp
```diff
@@ -615,6 +616,14 @@ void SimdPrinter::visit(AssignmentExpression* e) {
...
     else {
+        if (auto rhs = e->rhs()->is_identifier()) {
+            if (auto sym = rhs->symbol()) {
+                if (sym->is_variable() && sym->is_variable()->is_range()) {
+                    out_ << "assign(" << lhs->name() << ", indirect(" << rhs->name() << ", simd_width_))";
+                    return;
+                }
+            }
+        }
         out_ << lhs->name() << " = ";
```
halfflat, line 627: Is there ever an instance where we get an external variable on the rhs here?

More generally though. if `assign` were implemented generically in the rhs using `simd_cast` rather than making it specifically only take an indirect expression on the rhs, this clause could be written as
```
if (cast) {
    out_ << assign(" << lhs->name() << ", ";
    e->rhs()->accept(this);
    out_ << assign(" << ")";
}
else ...
```

I noticed when looking at the definition of `assign` in `simd.hpp` that the code currently pulls into the `simd` namespace `simd::detail::simd_impl` and `simd::detail::simd_mask_impl`, which sort of defeats the point of putting them in the `simd::detail` namespace to begin with. Is it possible to rejig that to keep the namespaces clean?

modcc/printer/cprinter.cpp
```diff
@@ -685,7 +694,8 @@ void emit_simd_state_read(std::ostream& out, LocalVariable* local, simd_expr_con
                 << "[0]);\n";
         }
         else if (constraint == simd_expr_constraint::contiguous) {
-            out << " = simd_cast<simd_value>(indirect(" <<  d.data_var
+            out << ";\n"
+                << "assign(" << local->name() << ",indirect(" <<  d.data_var
```
halfflat, line 698: Just for consistency, `", indirect("`

modcc/printer/cprinter.cpp
noraabiakar: Okay. I have extended `assign` to work with scalars and other simd_vectors, which will simplify the generated code even more. However, we can't use the visitor for the `rhs` when it is a range variable. This is because the visitor for range variables automatically casts the indirect expression into a simd_vector because it is necessary when they are arguments to binary expressions.
We also don't expose `sim::detail` with `using` anymore. But I'm not sure that addresses all your concerns on the matter.

Issue #1084, opened 2020-07-10
Fix issues found by static analysis tools


thorstenhater:
Hi,

I ran three static analysers against arbor
- clang analyzer
- PVS Studio with a trial license
- Facebook Infer
the results are collected here
https://gist.github.com/thorstenhater/1131960bf4a530590d92a68ee900c92b
We should triage these issues.

Cheers,
  Thorsten

thorstenhater:
As dicussed, here is the cleaned-up list of issues

```
| DONE? | File                                                |  Line | Category | Comment                                                                                                                                                                                                              |
|-------+-----------------------------------------------------+-------+----------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| [ ]   | arbor/include/arbor/simd/avx.hpp                    |   333 | warn     | V1032 The pointer 'data' is cast to a more strictly aligned pointer type.                                                                                                                                            |
| [ ]   | arbor/include/arbor/simd/simd.hpp                   |   165 | warn     | V1037 Two or more case-branches perform the same actions. Check lines: 165, 168                                                                                                                                      |
| [ ]   | arbor/include/arbor/simd/simd.hpp                   |   190 | warn     | V1037 Two or more case-branches perform the same actions. Check lines: 190, 193                                                                                                                                      |
| [ ]   | arbor/include/arbor/util/optional.hpp               |   286 | err      | V521 Such expressions using the ',' operator are dangerous. Make sure the expression 'assert_set(), ref()' is correct.                                                                                               |
| [ ]   | arbor/include/arbor/util/optional.hpp               |   286 | err      | V685 Consider inspecting the return statement. The expression contains a comma.                                                                                                                                      |
| [ ]   | arbor/include/arbor/util/optional.hpp               |   290 | err      | V521 Such expressions using the ',' operator are dangerous. Make sure the expression 'assert_set(), ref()' is correct.                                                                                               |
| [ ]   | arbor/include/arbor/util/optional.hpp               |   290 | err      | V685 Consider inspecting the return statement. The expression contains a comma.                                                                                                                                      |
| [ ]   | arbor/include/arbor/util/optional.hpp               |   294 | err      | V521 Such expressions using the ',' operator are dangerous. Make sure the expression 'assert_set(), std::move(ref())' is correct.                                                                                    |
[...]

PR #1086, opened 2020-07-15
Fix issues found in device_coordinator.hpp.

thorstenhater:
Issue found by static analyser, see #1084.

Device reference: 
- has assignment, but no copy constructor. 
- assignment not protected against self assign. 
- assignment does not return self.

Device coordinator: 
- Ternary constructs temporary.

### DIFF
--- a/arbor/memory/device_coordinator.hpp
+++ b/arbor/memory/device_coordinator.hpp
@@ -81,7 +81,14 @@ public:
     }
 
     device_reference& operator=(const device_reference& ref) {
-        gpu_memcpy_d2d(pointer_, ref.pointer_, sizeof(T));
+        if (this != &ref) {
+            gpu_memcpy_d2d(pointer_, ref.pointer_, sizeof(T));
+        }
+        return *this;
+    }
+        
+    device_reference(const device_reference& ref) {
+        *this = ref;
     }
 
     operator T() const {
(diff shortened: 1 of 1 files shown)


thorstenhater:
bors try

haampie:
failure is probably related to the cached git repo being partly removed from $SCRATCH after 30 days.

bors try

haampie:
Ah, there were two instances of gitlab-runner running after a restart, interfering with each other's files.

bors try

arbor/memory/device_coordinator.hpp
```diff
@@ -81,7 +81,14 @@ class device_reference {
...
-        gpu_memcpy_d2d(pointer_, ref.pointer_, sizeof(T));
+        if (this != &ref) {
+            gpu_memcpy_d2d(pointer_, ref.pointer_, sizeof(T));
+        }
+        return *this;
+    }
+        
+    device_reference(const device_reference& ref) {
+        *this = ref;
     }
```
halfflat, line 92: Is this not incorrect? At this point `this->pointer_` is default initialized to an indeterminate value, and then the assignment will perform a GPU memory copy to outer space.

I think the static analyzer might be mollified with `device_reference(const device_reference&) = default`.

arbor/memory/device_coordinator.hpp
```diff
@@ -115,9 +122,10 @@ class device_coordinator {
     using rebind = device_coordinator<Tother, Allocator>;
 
     view_type allocate(size_type n) {
-        Allocator allocator;
-
-        pointer ptr = n>0 ? allocator.allocate(n) : nullptr;
+        pointer ptr = nullptr;
+        if (n > 0) {
+            ptr = Allocator{}.allocate(n);
+        }
```
halfflat, line 128: This looks semantically almost identical. I honestly don't understand the original complaint from the static analyzer:
```
arbor/memory/device_coordinator.hpp:120 V623 Consider inspecting the '?:' operator. A temporary object of the 'array_view' type is being created and subsequently destroyed. Check third operand.                                                            
```
The third argument is `nullptr`.

I think it may be being confused by the final return statement (line 128 of original file).

thorstenhater: Yeah, let's go with the original here.

PR #1087, opened 2020-07-15
Fix circumvention of virtual dispatch in constructor.

thorstenhater:
Fix a potential issue when extending the class hierachy. There is no virtual dispatch in constructors, so 
the constructor might end up calling the wrong `reset_`


halfflat:
Could you describe in more detail the potential issue? Is it a scenario where someone writing a derived class of `benchmark_cell_group` might expect `benchmark_cell_group` to call the derived's version of `reset()`?

I ask only because I am not seeing the danger at the moment.

thorstenhater:
I misunderstood the issue the analyser was pointing out. Close and mark as non-issue.

PR #1090, opened 2020-07-16
Update required CMake version in the docs (3.12).

brenthuisman:


bcumming:
@brenthuisman, CMake 3.12 is the actual minimum required version, so instead this PR should update the docs in `doc/install.rst`.

PR #1091, opened 2020-07-16
Support Allen/SONATA-style single sample somata in SWC.

bcumming:
Add an `load_swc_allen(str fname, bool no_gaps=False)` function to the python wrapper that
generates a segment tree from an SWC file following the rules prescribed by                                                                                                                            
AllenDB and Sonata. Specifically:

 * The first sample (the root) is treated as the center of the soma.
 * The first morphology is translated such that the soma is centered at (0,0,0).
 * The first sample has tag 1 (soma).
 * All other samples have tags 2, 3 or 4 (axon, apic and dend respectively)

SONATA prescribes that there should be no gaps, however the models in AllenDB
have gaps between the start of sections and the soma. The flag `no_gaps` can be
used to enforce this requirement.

Arbor does not support modeling the soma as a sphere, so a cylinder with length
equal to the soma diameter is used. The cylinder is centered on the origin, and
aligned along the z axis.
Axons and apical dendrites are attached to the proximal end of the cylinder, and
dendrites to the distal end, with a gap between the start of each branch and the
end of the soma cylinder to which it is attached.

### DIFF
--- a/python/morphology.cpp
+++ b/python/morphology.cpp
@@ -21,16 +21,15 @@ arb::segment_tree load_swc_allen(const std::string& fname, bool no_gaps=false) {
         try {
             using namespace arb;
             auto records = parse_swc_file(fid);
-            swc_canonicalize(records);
 
             // Assert that the file contains at least one sample.
             if (records.empty()) {
-                throw pyarb_error("SWC file is empty");
+                throw pyarb_error("Allen SWC: empty file");
             }
 
             // assert that root sample has tag 1.
             if (records[0].tag!=1) {
-                throw pyarb_error("SONATA requires that the first swc recrod has tag 1 for soma");
+                throw pyarb_error("Allen SWC: the soma record does not have tag 1");
             }
 
             // Handle the case of a single-sample swc file.
@@ -50,7 +49,7 @@ arb::segment_tree load_swc_allen(const std::string& fname, bool no_gaps=false) {
                         [](auto& r){return r.tag<2 || r.tag>4;});
             if (it!=records.end()) {
                 throw pyarb_error(
-                        "Every SWC record must have a tag of 2, 3 or 4, except for the first which must have tag 1.");
+                        "Allen SWC: every record must have a tag of 2, 3 or 4, except for the first which must have tag 1");
             }
 
             // Translate the morphology so that the soma is centered at the origin (0,0,0)
@@ -70,6 +69,7 @@ arb::segment_tree load_swc_allen(const std::string& fname, bool no_gaps=false) {
 
             // Build branches off soma.
             std::unordered_map<msize_t, msize_t> pmap;
+            std::set<msize_t> unused_samples;
             for (unsigned i=1; i<nrec; ++i) {
                 const auto& r = records[i];
                 // If sample i has the root as its parent don't create a segment.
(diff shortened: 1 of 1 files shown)


python/morphology.cpp
```diff
@@ -13,6 +13,97 @@
...
+
+            // Build branches off soma.
+            for (unsigned i=1; i<nrec; ++i) {
+                const auto& r = records[i];
+                // If sample i has the root as its parent don't create a segment.
+                if (r.parent_id==0) {
+                    if (no_gaps) {
+                        // Assert that this branch starts on the "surface" of the spherical soma.
+                        auto d = std::fabs(soma_rad - std::sqrt(r.x*r.x + r.y*r.y + r.z*r.z));
+                        if (d<1e-6) {
```
thorstenhater, line 82: Should this be `>`?

python/morphology.cpp
```diff
@@ -13,6 +13,97 @@
...
 
+arb::segment_tree load_swc_allen(const std::string& fname, bool no_gaps=false) {
+        std::ifstream fid{fname};
+        if (!fid.good()) {
+            throw pyarb_error(util::pprintf("can't open file '{}'", fname));
+        }
+        try {
+            using namespace arb;
+            auto records = parse_swc_file(fid);
+            swc_canonicalize(records);
```
thorstenhater, line 24: this is redundant, see https://github.com/arbor-sim/arbor/blob/111104b4b6082a7768a6933cc54dbc065cffbd19/arbor/swcio.cpp#L148

python/morphology.cpp
bcumming: It should. I fixed this in another commit, but obviously failed to push it correctly.

python/morphology.cpp
bcumming: Indeed. We also canonicalize input in the normal `load_swc` function in C++/Python. I will fix that too.

python/morphology.cpp
```diff
@@ -13,6 +13,97 @@
...
+            // Handle the case of a single-sample swc file.
+            const auto nrec = records.size();
+            if (nrec==1) {
+                segment_tree tree;
+                // Model the spherical soma as a cylinder with length=2*radius.
+                // The cylinder is centred on the sample location, and extended along the z axis.
+                const auto r = records[0].r; // The radius of the soma.
+                tree.append(mnpos, {0, 0, -r, r}, {0, 0, r, r}, 1);
+                return tree;
+            }
```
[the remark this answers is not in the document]
bcumming: I removed the special case.

python/morphology.cpp
```diff
@@ -13,6 +13,102 @@
...
+            // Model the spherical soma as a cylinder with length=2*radius.
+            // The cylinder is centred on the origin, and extended along the y axis.
+            double soma_rad = sloc.radius;
+            tree.append(mnpos, {0, -soma_rad, 0, soma_rad}, {0, soma_rad, 0, soma_rad}, 1);
+
+            // Build branches off soma.
+            std::unordered_map<msize_t, msize_t> pmap;
+            std::set<msize_t> unused_samples;
+            const auto nrec = records.size();
+            for (unsigned i=1; i<nrec; ++i) {
```
thorstenhater, line 73: That will fall over if size == 1, right? Maybe just return just before the loop here?

bcumming: This isn't a problem. I add the soma associated with sample 0 above, and this loop will not execute.

thorstenhater: Right, that's true.

PR #1094, opened 2020-07-17
arb::util::variant fixes

### LINKED ISSUE #1093: Variant hash is broken
It's currently impossible to make an `unordered_map` with an `arb::util::variant` key because:
1. The `std::hash` overload is incorrect.
2. There is no `operator==` for `arb::util::variant`.

halfflat:
* Correct hash implementation.
* Add ==, != comparisons.

Fixes #1093

### DIFF
--- a/arbor/include/arbor/util/variant.hpp
+++ b/arbor/include/arbor/util/variant.hpp
@@ -181,6 +181,14 @@ struct variant_dynamic_impl<> {
     }
 
     static void destroy(std::size_t i, char* data) {}
+
+    static bool cmp_eq(std::size_t i, const char* left, const char* right) {
+        return i==std::size_t(-1)? true: throw bad_variant_access{};
+    }
+
+    static bool cmp_ne(std::size_t i, const char* left, const char* right) {
+        return i==std::size_t(-1)? false: throw bad_variant_access{};
+    }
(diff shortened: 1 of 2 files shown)


arbor/include/arbor/util/variant.hpp
```diff
@@ -584,16 +614,16 @@ namespace std {
 // Unambitious hash:
 template <typename... T>
 struct hash<::arb::util::variant<T...>> {
-    std::size_t operator()(const ::arb::util::variant<T...>& v) {
+    std::size_t operator()(const ::arb::util::variant<T...>& v) const {
         return v.index() ^
-            visit([](const auto& a) { return std::hash<std::remove_cv_t<decltype(a)>>{}(a); }, v);
+            ::arb::util::visit([](const auto& a) { return std::hash<std::remove_cv_t<std::remove_reference_t<decltype(a)>>>{}(a); }, v);
```
bcumming, line 619: Can you replace `remove_cv<remove_reference<>>` with `decay<>`?

halfflat: It's not _quite_ the same (it doesn't turn arrays into pointers), so I think it matches the semantics more precisely.

PR #1096, opened 2020-07-17
Fix python examples

noraabiakar:
Fixes typo in `single_cell_swc.py` 
Fixes bug in `flat_cell_builder.cpp` that was causing a segmentation fault in `single_cell_builder.py`. The bug is caused by returning the compartment index instead of the cable index in the `add_cable` function.

### DIFF
--- a/python/flat_cell_builder.cpp
+++ b/python/flat_cell_builder.cpp
@@ -78,9 +78,8 @@ public:
         double dr = (r2-r1)/ncomp;
         for (auto i=0; i<ncomp; ++i) {
             p = tree_.append(p, {0,0,z+i*dz, r1+i*dr}, {0,0,z+(i+1)*dz, r1+(i+1)*dr}, tag);
+            cable_distal_segs_.push_back(p);
         }
-
-        cable_distal_segs_.push_back(p);
 
         return size()-1;
     }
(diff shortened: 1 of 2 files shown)


python/flat_cell_builder.cpp
```diff
@@ -78,10 +78,9 @@ class flat_cell_builder {
         double dr = (r2-r1)/ncomp;
         for (auto i=0; i<ncomp; ++i) {
             p = tree_.append(p, {0,0,z+i*dz, r1+i*dr}, {0,0,z+(i+1)*dz, r1+(i+1)*dr}, tag);
+            cable_distal_segs_.push_back(p);
```
bcumming, line 81: `cable_distal_segs_` records the last segment in each cable added by a user, so there should be one entry for each cable, not for each segment in the tree. This should be revereted.

noraabiakar: This function returns `tree_.size() -1`, and `tree_.size()` grows by `ncomp` for each branch, but we only add one entry in `cable_distal_segs_` per branch. 
The return value of this function is then passed again to this function as the `parent`, and then the access to `cable_distal_segs_[parent]` is out of bounds. 
I should have been clearer in expailning the bug earlier, and maybe there's a better way to do this.

bcumming: What is the input that caused the bug?
This is appending the correct value, so I suspect that the bug is that the wrong value is being returned from this function, i.e. it should return `cable_distal_segs_.size()-1`

noraabiakar: The `single_cell_builder.py` example fails with a segmentation fault. Returning `cable_distal_segs_.size()-1` also fixes the problem and is the better solution. I'll make the adjustment.

PR #1099, opened 2020-07-22
Set the status on the commit from the PR

haampie:
This is copied from DLA Future. It should set the CI status on the commit from the PR instead of just on the Bors commit. This way the gitlab CI status also appears in the checks section below the PR.


haampie:
bors try

PR #1105, opened 2020-08-06
Updated all in-repo references and scripts to Python 3

brenthuisman:
bors try


haampie:
bors try-

haampie:
bors try

scripts/tsplot
```diff
@@ -84,7 +82,7 @@ def parse_clargs():
     return P.parse_args(argsbis)
 
 def isstring(s):
-    return isinstance(s,str) or isinstance(s,unicode)
+    return isinstance(s,str) or isinstance(s,str)
```
bcumming, line 85: This should simplify to just
```Python
    return isinstance(s,str)
```
To the point where you can just call `isinstance(s,str)` directly inline.

brenthuisman: I used 2to3, I removed this and some other clutter in introduced.

PR #1106, opened 2020-08-06
Add python version check github action

haampie:
Not sure how useful it is, but in spack/spack they're using `vermin` to quickly check whether the python version requirements are met in the source files. This action runs in <30s, so could be useful :)


bcumming:
Thanks for the heads up! While useful, I think that adding another action to our repo, that takes 30s to check our Python for Python 3 is a bit of overkill.

Issue #1102, opened 2020-08-03
Replace soma_cell_builder with builder suitable for NeuroML morphology support.


halfflat:
As part of the implementation of NeuroML 2 morphology reading #1088 and to fix the rather ad hoc 'some cell builder' process (see #1085), it is useful to provide in the main library an alternative cell morphology construction process in terms of labelled segments, which can then be referred to later via region expressions.

The task is to implement a corresponding builder class, where the basic unit of description corresponds to a NeuroML-style segment. These objects can have children attached to points _within_ the segment, and not just the ends, and so do not correspond directly to the `msegment` already present in Arbor morphology descriptions: we'll need a new name for these, at least as far as the Arbor API is concerned.

halfflat:
I'll finish this work when/if #1111 is merged.

thorstenhater:
Soma builder last point of use is in tests, this can be closed until needed again.

PR #1111, opened 2020-08-11
Implement 'stitch' morphology builder.

halfflat:
* Add (forward) ordered forest implementation, tests.
* Add `segment` region expression; to ease implementation, `msegment` now also knows its own id.
* Add `stitch_builder` and `stitched_morphology`. A stitch corresponds to a labelled, linearly-interpolated segment which can be attached at any point along a parent stitch. A `stitched_morphology` takes a `stitch_builder` object and constructs a segment tree and morphology, and provides a dictionary of stitch labels to segment indices and region expressions.
* Add `import` method for `label_dict`, so that the label dictionary returned by `stitched_morphology` can be merged with an existing dictionary.

Work towards #1102 — remaining work is the replacement of `soma_cell_builder` in unit tests.

### DIFF
--- a/arbor/morph/stitch.cpp
+++ b/arbor/morph/stitch.cpp
@@ -115,6 +115,9 @@ struct stitch_builder_impl {
 
 stitch_builder::stitch_builder(): impl_(new stitch_builder_impl) {}
 
+stitch_builder::stitch_builder(stitch_builder&&) = default;
+stitch_builder& stitch_builder::operator=(stitch_builder&&) = default;
+
 stitch_builder& stitch_builder::add(mstitch f, const std::string& parent_id, double along) {
     impl_->add(std::move(f), parent_id, along);
     return *this;
@@ -162,6 +165,8 @@ stitched_morphology::stitched_morphology(stitch_builder&& builder):
 stitched_morphology::stitched_morphology(const stitch_builder& builder):
     impl_(new stitched_morphology_impl(*builder.impl_))
 {}
+
+stitched_morphology::stitched_morphology(stitched_morphology&& other) = default;
 
 arb::morphology stitched_morphology::morphology() const {
     return arb::morphology(impl_->stree);
--- a/arbor/morph/stitch.cpp
+++ b/arbor/morph/stitch.cpp
@@ -51,7 +51,7 @@ struct stitch_builder_impl {
         return *this = stitch_builder_impl(other);
     }
 
-    void add(mstitch f, const std::string& parent, double along, bool infer_prox = false) {
+    void add(mstitch f, const std::string& parent, double along) {
         if (id_to_node.count(f.id)) throw duplicate_stitch_id(f.id);
 
         forest_type::iterator p;
--- a/arbor/morph/stitch.cpp
+++ b/arbor/morph/stitch.cpp
@@ -1,0 +1,209 @@
+#include <memory>
+#include <unordered_map>
+#include <vector>
+
+#include <arbor/morph/stitch.hpp>
+#include <arbor/morph/morphexcept.hpp>
+#include <arbor/morph/locset.hpp>
+#include <arbor/morph/region.hpp>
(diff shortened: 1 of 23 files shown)


halfflat:
(Travis is being incommunicado, but as of commit 9b81526, it's passing: [](https://github.com/arbor-sim/arbor/commit/98a9e85488e6fe72bdd60c51f250c6bae0f134e1).)

noraabiakar (approved):
Looks good! And I enjoyed reading through the ordered forest implementation. My only comment is that maybe we should add some documentation to the ordered forest and stitch builder. Maybe inline in the code will be enough for the ordered forest.

halfflat:
I'll write some docs! I think in source comments for the ordered forest would probably be sufficient, but I'll add something in the generated documentation for the stitch builder.

arbor/util/ordered_forest.hpp
```diff
@@ -0,0 +1,607 @@
...
+    const_sibling_iterator child_end(const iterator_base&) const { return {}; }
+
+    sibling_iterator root_begin() { return sibling_iterator{first_else_end()}; }
+    const_sibling_iterator root_begin() const { return const_sibling_iterator{first_else_end()}; }
+
+    sibling_iterator root_end() { return sibling_iterator{}; }
+    const_sibling_iterator root_end() const { return const_sibling_iterator{}; }
+
+    postorder_iterator postorder_begin() { return postorder_iterator{first_leaf()}; }
+    const_postorder_iterator postorder_begin() const { return const_postorder_iterator{first_else_end()}; }
```
halfflat, line 152: Oops, this should be `const_postorder_iterator(first_leaf())`.

test/unit/test_ordered_forest.cpp
```diff
@@ -0,0 +1,497 @@
...
+        }
+    }
+
+    for (auto i = f.begin(); i!=f.end(); ++i) {
+        auto cb = f.child_begin(i);
+        auto ce = f.child_end(i);
+
+        for (auto j = cb; j!=ce; ++j) {
+            if (j.parent() != i) {
+                auto failure = ::testing::AssertionFailure() << "child node " << *j << " of " << *i << " is node ";
```
halfflat, line 79: That should be `" has parent "` not `" is node "`.

test/unit/test_ordered_forest.cpp
```diff
@@ -0,0 +1,497 @@
...
+
+    ivector pre_cdefault{f.cbegin(), f.cend()};
+    EXPECT_EQ((ivector{1, 2, 3, 4, 5, 6, 7, 8, 9}), pre_cdefault);
+
+    ivector root{f.root_begin(), f.root_end()};
+    EXPECT_EQ((ivector{1, 4, 9}), root);
+
+    ivector post{f.postorder_begin(), f.postorder_end()};
+    EXPECT_EQ((ivector{2, 3, 1, 5, 7, 6, 8, 4, 9}), post);
+
```
halfflat, line 267: Missing tests for const root iterators and const postorder iterators, which allowed that const postorder iterator begin bug to slip by...

arbor/morph/stitch.cpp
```diff
@@ -0,0 +1,214 @@
...
+        }
+    }
+
+    stitch_builder_impl& operator=(stitch_builder_impl&&) = default;
+    stitch_builder_impl& operator=(const stitch_builder_impl& other) {
+        if (this==&other) return *this;
+        return *this = stitch_builder_impl(other);
+    }
+
+    void add(mstitch f, const std::string& parent, double along, bool infer_prox = false) {
```
noraabiakar, line 54: `infer_prox` is not actually used anywhere

halfflat: Nice catch! Yep, this is leftover from an earlier approach.

noraabiakar (approved):
Thanks for adding the documentation!

Issue #1109, opened 2020-08-10
modcc/defining a time-dependent non-specific current in NMODL


espenhgn:
Hi. I'm trying to define a non-specific current in NMODL that depends on time. The following `sinsyn.mod` which creates a sinusoid transmembrane current works fine with nrnivmodl and NEURON:
```
COMMENT
Bla bla
ENDCOMMENT

NEURON {
    POINT_PROCESS SinSyn
    RANGE del, dur, pkamp, freq, phase, bias
    NONSPECIFIC_CURRENT i
}

UNITS {
    (nA) = (nanoamp)
}

PARAMETER {
    del=5     (ms)
    dur=200   (ms)
    pkamp=1   (nA)
    freq=1    (Hz)
    phase=0
    bias=0    (nA)
    PI=3.14159265358979323846
}

ASSIGNED {
    i (nA)
}

BREAKPOINT {
    if (t < del) {
        i=0
    }else{
        if (t < del+dur) {
            i = -pkamp*sin(2*PI*freq*(t-del)*(0.001)+phase)-bias
        }else{
            i = 0
        }
    }
}

NET_RECEIVE(weight (nA)) {
}
```
`modcc` however doesn't like `i (nA)` in the ASSIGNED block:
```
$ modcc -o . -t cpu sinsyn.mod 
internal compiler error: the symbol i already exists @ (line 6,col 25)
```
Removing that results in multiple `the variable 't' is undefined` errors as I presume `t` is not globally defined in arbor as it is in NEURON. `nrnivmodl` will no longer compile the file at this point. 
The error with `modcc` can be silenced by adding `t` to RANGE and PARAMETER but `modcc` ultimately gives a warning with ` there is no SOLVE statement, required to update the state variables, in the BREAKPOINT block`. 
In this case a solve statement `SOLVE states METHOD cnexp` and `DERIVATIVE states {...}`  shouldn't be required, so surely there must be a better way to define this. Any suggestions?

bcumming:
Hello @espenhgn.

Arbor does use a slightly different dialect of NMODL, as you have discovered.

* it doesn't make sense for currents to be `ASSIGNED`, so it is disallowed.
* we don't expose `t` as a magic variable, like `v` or `celcius`, however we can expose it easily enough.
* The `there is no SOLVE statement...` statement should just be a warning, not an error. We have other mechanisms that don't have `SOLVE` statements, which just give this warning (that we should really disable).

[...]

PR #1113, opened 2020-08-12
Expose `time` to the mechanisms 

noraabiakar:
The `vec_t_` time vector is already available to the mechanisms, but not exposed. It is indexed by `cell_index` instead of directly by the `node_index` (`vec_t[vec_ci_[node_index_[i]]]`). This kind of indexing was previously unavailble. It has not been added to the printers. 
This PR also includes some cleanup in the vectorized code printer. 

Address #1109

### DIFF
--- a/modcc/printer/gpuprinter.cpp
+++ b/modcc/printer/gpuprinter.cpp
@@ -1,7 +1,7 @@
 #include <cmath>
 #include <iostream>
 #include <string>
-#include <unordered_set>
+#include <set>
 
 #include "gpuprinter.hpp"
 #include "expression.hpp"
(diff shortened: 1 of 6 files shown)


bcumming:
bors try

modcc/printer/gpuprinter.cpp
```diff
@@ -384,11 +384,23 @@ void emit_api_body_cu(std::ostream& out, APIMethod* e, bool is_point_proc) {
     auto body = e->body();
     auto indexed_vars = indexed_locals(e->scope());
 
-    std::unordered_set<std::string> indices;
+    struct index_prop {
+        std::string source_var; // array holding the indices
+        std::string index_name; // index into the array
+
+        bool operator<(const index_prop& other) const {
```
bcumming, line 391: When this is inserted in a `std::set`, if you have two indexes with the same `source_var` and different `index_name`, then only one will be stored.
Will that situation ever arise? Or will a given `source_name` only ever have one possible `index_name`? If that's the case, then a hash table might be a better option.

bcumming: Reading the code a bit more, this approach has the benefit of printing all the variables in alphabetical (deterministic) order.

noraabiakar: Actually, I have a bug here that I hadn't noticed before. Alphabetical order is not enough, because we have scenarios such as:
```
auto node_index_i_ = node_index[i_];
auto vec_ci_i_ = vec_ci_[node_index_i_];
```
We need to make sure that `node_index_i` is printed before `vec_ci_i_` regardless of the names of the vectors. Basically, all arrays that are indexed directly with `i_` should go first. I have changed `std::set` to an `std::list` and I insert either to the front or the back of the list to enforce my ordering.

noraabiakar:
I don't know whether testing this feature is easy. Testing the printing will require quite a bit of work as we don't currently test the printing of indices. We can test that we are getting the correct `time` in the mechanisms but it's pretty straight forward that we will provided the printing is correct.

noraabiakar:
bors try

bcumming (approved):

PR #1123, opened 2020-08-27
Implement 'stitch' morphology builder.

halfflat:
New PR post master rollback; squashed and rebased, but reprises #1111.

* Add (forward) ordered forest implementation, tests.
* Add `segment` region expression; to ease implementation, `msegment` now also knows its own id.
* Add `stitch_builder` and `stitched_morphology`. A stitch corresponds to a labelled, linearly-interpolated segment which can be attached at any point along a parent stitch. A `stitched_morphology` takes a `stitch_builder` object and constructs a segment tree and morphology, and provides a dictionary of stitch labels to segment indices and region expressions.
* Add `import` method for `label_dict`, so that the label dictionary returned by `stitched_morphology` can be merged with an existing dictionary.
* Add section on stitch builder etc. to cable cell docs.
* Update cable cell docs to remove out of date info and to provide some context.
* Describe ordered forest datastructure and interface in a long comment at the beginning of the header file.

### DIFF
--- a/arbor/CMakeLists.txt
+++ b/arbor/CMakeLists.txt
@@ -31,6 +31,7 @@ set(arbor_sources
     memory/gpu_wrappers.cpp
     memory/util.cpp
     morph/embed_pwlin.cpp
+    morph/stitch.cpp
     morph/label_dict.cpp
     morph/locset.cpp
     morph/morphexcept.cpp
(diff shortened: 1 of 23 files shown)


bcumming (approved):
Looks good!
This is a "re-merge" of a PR that had to be reverted to fix problems elsewhere.

PR #1127, opened 2020-09-01
Rewrite SWC reader code.

halfflat:
* Simplify scanning of SWC records.
* `parse_swc` now retains metadata read from SWC comments.
* `parse_swc` is given two modes, relaxed and strict. Strict mode enforces contiguous record numbering and rejects SWC that looks like it has a single-point soma.
* Subtype `swc_error` according to error mode.
* Remove `point` class and corresponding unit tests.
* Add a RAII stream formatting flag saver, used in SWC record output operator.
* Replace included 1-point soma SWC example with a regular SWC from NeuroMorpho.


bcumming:
I am not sure that attempts to rewrite swc files with single sample somas (i.e. a unique tag on the first sample) is a good idea, because:
  * it is ambiguous how to interpret such files if the intention was, indeed, to have a single-sample/spherical soma
  * you can't represent the user's intention using a "correct" swc representation, because you are going to need to have gaps.

halfflat:
> I am not sure that attempts to rewrite swc files with single sample somas (i.e. a unique tag on the first sample) is a good idea, because:
> 
> * it is ambiguous how to interpret such files if the intention was, indeed, to have a single-sample/spherical soma
> * you can't represent the user's intention using a "correct" swc representation, because you are going to need to have gaps.

I totally agree: that stupid script is only there to 'fix' the SWC used in the examples and unit tests.

bcumming:
Then how about we just use that script to update the swc files, but not leave it in the repo?
We could create a rewriter at a later date that provides a bunch of flags and options for folk to translate swc into sane representations like NeuroMorpho, or our own Arbor representation.
Working on NeuroLucida asc files has taught me a lot about how we should not use s-expressions in our representation!

halfflat:
So on that topic, should we nuke the script and the Allen SWC files, and replace them with something that has a real soma?

halfflat:
I note also that the Travis fail is yet another g++6ism.

halfflat:
I think we should keep the script if we're keeping the original SWC and the munged ones, just so there is a record of what the hell we're doing. Otherwise, we should pick different example SWC files that have real somata.

halfflat:
Replaced the one point soma SWC files with a sensible SWC file from NeuroMorpho.

arbor/include/arbor/swcio.hpp
```diff
@@ -57,37 +71,59 @@ class swc_record {
...
+// Note that 'one-point soma' SWC files are explicitly not supported.
+//
+// In `relaxed` mode, it will check that:
+//     * There are no duplicate record ids.
+//     * All record ids are positive.
+//     * There are no records whose parent id is not less than the record id.
+//     * Only one record has parent id -1; all other parent ids correspond to records.
+//     * There are at least two records.
+//
+// In `strict` mode, it will additionally check:
```
bcumming, line 100: This is too course grained.
E.g. Allen model insist that records on the same unbranched section should be contiguously numbers, and that the soma must be a "sphere/single sample", which can't be specified with this flag.

arbor/swcio.cpp
```diff
@@ -1,213 +1,202 @@
...
+    for (std::size_t i = 0; i<n_rec; ++i) {
+        swc_record& r = records[i];
+
+        if (r.parent_id>=r.id) {
+            throw swc_record_precedes_parent(r.id);
         }
 
-        if (is_comment(line)) {
-            continue;
+        if (mode==swc_mode::strict && r.id != (int)i+1) {
```
bcumming, line 97: If would prefer to see `if (mode&swc_mode::contiguous && ...)`, so that we can pick and choose the restrictions, as required (unfortunately) by SONATA.

arbor/swcio.cpp
```diff
@@ -1,213 +1,202 @@
...
+    for (std::size_t i = 0; i<n_rec; ++i) {
+        const swc_record& r = records[i];
+        first_tag_match |= r.parent_id==first_id && r.tag==first_tag;
+
+        if ((i==0 && r.parent_id!=-1) || (i>0 && !seen.count(r.parent_id))) {
+            throw swc_no_such_parent(r.id);
+        }
+    }
+
+    if (mode==swc_mode::strict && !first_tag_match) {
```
bcumming, line 118: See above `(mode&swc_mode::single_sample_root_ok && ..)`

arbor/swcio.cpp
```diff
@@ -1,213 +1,202 @@
...
-            else {
-                record.assert_consistent();
-                records.push_back(std::move(record));
+    while (in) {
+        auto c = in.get();
+        if (c=='#') {
+            getline(in, line, '\n');
+            auto from = line.find_first_not_of(" \t");
+            if (from != std::string::npos) {
+                // data.metadata.append(line, from); // TODO: restore when g++6 is gone.
```
bcumming, line 137: TODO.

bcumming (changes requested):

arbor/include/arbor/swcio.hpp
halfflat: It's definitely not meant to cover all cases: the intention of strict is to firstly abide by standard SWC conventions, and secondly prohibit spherical somata.

Relaxed just enforces, essentially, the bare minimum; Allen-specific checks would be imposed by whatever is dealing with their particular brand of SWC.

arbor/swcio.cpp
halfflat: Can't avoid special handling of spherical somata: I think in this special handling one should perform any additional checks required by e.g. SONATA.

arbor/include/arbor/swcio.hpp
bcumming: Then why shouldn't all these checks be enforced in the different readers? Unfortunately there is no SWC spec that mandates the rules that are being enforced here, and in practice we see that tools like NEURON are far more relaxed. So, despite these being reasonable rules, they don't appear to be 'standard', and in order to make SWC a useful format folk will nearly never use these strict defaults.

halfflat: I don't think this is true: the 'relaxed' mode is, as far as I can tell, much more relaxed than any extant SWC description, and what's more, I think it corresponds to a minimal reasonable set of requirements, as far as sensible parsing is concerned.

The strict mode adds exactly two requirements: contiguous record numbering, which again I think is common to every SWC spec, and a prohibition of single-point somata. The latter is a necessary requirement for sensible interpretation by Arbor. That extant SWC files _do_ use single point somata is a problem we can't fix: we can only provide ad hoc interpretations as required at best.

In terms of parsing SWC and generating an unambiguous morphology, 'relaxed' + no single point soma is the bare minimum. I can see that would could relax the contiguous numbering requirement, but I don't think this is the problem, right? I included that in 'strict' only because of its commonality in SWC format descriptions, and because in my own view, when reading a complete SWC file (as opposed to a fragment), non-contiguous data is likely to indicate bad data.

tl;dr: 'relaxed' is really relaxed and should accept any extant SWC; 'strict' is slightly more strict than necessary for unambiguous interpretation, but that extra strictness I don't think precludes any extant SWC.

bcumming: I think that both of these tests are valuable, but I still feel that bundling them together isn't great for the following reason.

If contiguous numbering is common enough to be a kind of standard requirement, but single point somas are commonly accepted, it doesn't make sense to bundle those two requirements. A parser of the records that handles spherical somas (e.g. one that mimics NEURON) will have to apply the numbering requirement itself.

My preference would be to make this more flexible so that such validations are only implemented once, instead in an ad-hoc manner in each swc reader (which may or may not be a problem, depending on how many we end up having to write).

I will merge, because at the end of the day we are bike shedding over a poorly-specified widely-misused file format whose use we are unable to ban.

bcumming (approved):

Issue #1129, opened 2020-09-07
[Arbor 0.3.1-dev] No add_sphere attribute in flat_cell_builder()


lungsi:
**Bug** :bug:
`flat_cell_builder()` in Arbor 0.3.1-dev does not have the attribute `add_sphere`. As a result, the [ball-and-stick example](https://arbor.readthedocs.io/en/latest/single_cell.html?) fails.

**To Reproduce**
| Arbor 0.3                   | Arbor 0.3.1-dev                                                                |
| :---------------------- | :------------ |
| `pip install arbor`       | `pip install git+https://github.com/arbor-sim/arbor.git` |
|  `import arbor`           | `import arbor`   |
| `builder = arbor.flat_cell_builder()`   | `builder = arbor.flat_cell_builder()` |
| `p = builder.add_sphere(radius=10, name='soma')`   | `p = builder.add_sphere(radius=10, name='soma')` |
| **No error**   | **Error**</br>---------------------------------------------------------------------------</br>AttributeError                            Traceback (most recent call last)</br>\<ipython-input-3-11dd81bb0702\> in \<module\></br>----> 1 p = builder.add_sphere(radius=10, name='soma')</br>AttributeError: 'arbor._arbor.flat_cell_builder' object has no attribute 'add_sphere'|

**Expected behavior**
| Arbor 0.3                   |
| :---------------------- |
| `pip install arbor`       |
|  `import arbor`           |
| `builder = arbor.flat_cell_builder()`   |
| `p = builder.add_sphere(radius=10, name='soma')`   |
| `q = builder.add_cable(parent=p, length=100, radius=4, name='dend')` |
| `p = builder.add_cable(parent=q, length=50, radius=(4,2), name='dend')` |
| `p = builder.add_cable(parent=q, length=50, radius=(4,2), name='dend')` |

**Screenshots**
![arbor_bug_add_sphere](https://user-images.githubusercontent.com/12233689/92357285-df23d300-f0e7-11ea-9581-f53d34a70392.PNG)

**System:**
 - OS: Ubuntu 20.04
 - Python 3.7
 - Arbor 0.3.1-dev (installed with `pip install git+https://github.com/arbor-sim/arbor.git`)

**Additional context**
As shown in the screenshot, `flat_cell_builder()` has attributes: `add_cable`, `add_label`, `build`, `labels`, `morphology`, and `segments`.

bcumming:
Thanks for the detailed report @lungsi.

[...]

PR #1130, opened 2020-09-07
update flat_cell_builder docs and add deprecation warning

bcumming:
Update flat_cell_builder docs and add deprecation warning.

Addresses #1129.

### DIFF
--- a/doc/single_cell.rst
+++ b/doc/single_cell.rst
@@ -99,7 +99,7 @@ Morphology
 
     The flat cell builder documented below is likely to change before
     version 0.4 is released. It is a wrapper around a segment tree that
-    simplifies morphology construction the spatial location of segments
+    simplifies morphology construction when the spatial location of segments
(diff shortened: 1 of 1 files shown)


brenthuisman:
`flat_cell_builder()` is used in `single_cell_builder.py` and `ring.py` in `python/example`. If it is certain `flat_cell_builder` is going to be removed in Arbor 0.4, let's remove them from there as well.

brenthuisman (changes requested):
flat_cell_builder() is used in single_cell_builder.py and ring.py in python/example (no spherical soma though). Is it certain flat_cell_builder is going to be removed in Arbor 0.4, or will it just be changed, and will that change be removal of spherical somas?

brenthuisman (approved):
The title of the PR confused me a bit. It's about the docs reflecting the fact that adding spherical somas have been removed from flat_cell_builder in master, which is correct.

PR #1137, opened 2020-09-07
Fix bugs in util::variant assignment implementation.

### LINKED ISSUE #1136: Copy assignment on util::variant broken
In non-trivial cases, copy assignment of a `util::variant` will wrongly throw an exception of type `bad_variant_access`.

In addition, the copy assignment operation is not exception safe, and the move assignment operation incorrectly sets the lhs to valueless in the case of an exception.

halfflat:
* Fix copy-paste error that led to legitimate copies throwing a bad_variant_access exception.
* Fix exception safety for the case when an error thrown in copy assignment takes the lhs to valueless.
* Fix wrong valueless semantics associated with an exception being thrown during move assignment.
* Add unit tests that exercise these issues.

Fixes #1136.

### DIFF
--- a/arbor/include/arbor/util/variant.hpp
+++ b/arbor/include/arbor/util/variant.hpp
@@ -218,7 +218,6 @@ struct variant_dynamic_impl<H, T...> {
         else {
             variant_dynamic_impl<T...>::assign(i-1, data, from);
         }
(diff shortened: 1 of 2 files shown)


bcumming (approved):
Looks good, thanks!

PR #1138, opened 2020-09-07
remove arb::util::either from Python interface

### LINKED ISSUE #1133: Replace arb::util::either in Python code with arb::util::variant
We will deprecate `either` as it's own type as part of implementing a `std::expected` workalike.
This will remove any dependence on `either` via the public API.

bcumming:
Switch `arb::util::either` to `arb::util::variant` in the Python wrapper.

Fixes #1133


halfflat (approved):
Nice and simple!

Issue #1022, opened 2020-04-28
C++17 migration TODO list


halfflat:
We'll migrate to C++17 soon. Let's collate here the various bits and pieces that we will update accordingly. They can be spun off into individual issues as required.

* <s>Replace `util::optional` with `std::optional` — usages of `util::optional<T&>` will need to be replaced with `std::optional<std::reference_wrapper<T>>` and/or the corresponding interfaces changed.></s> [fixed in #1158]
* <s>Replace `util::variant` with `std::variant`</s> [fixed in #1152] <s>and implement `util::either` over `std::variant`.</s> [fixed in #1142]
* Replace `make_foo` helper usage with simple constructors, adding deduction guides as appropriate.
* Remove class constructors that are present only to allow default member values — these will be able to use brace initialization.
* <s>Attend to anything with a `TODO: C++17` comment.</s> [fixed in #1159 (pending review)]
* <s>Use `std::in_place_t` where appropriate in our utility class interfaces.</s> [fixed in #1142]
* <s>Replace `util::size` with `std::size`.</s> [fixed in #1140]
* Replace `std::is_blah<X>::value` predicates with `std::is_blah_v<X>`.
* <s>Replace `util::any` with `std::any`.</s> [fixed in #1152]
* Replace `util::unique_in_place` with `std::unique` or a wrapper around `std::unique`.

thorstenhater:
Also, investigate `std::pmr::` for allocator-aware containers.

thorstenhater:
Hey,

most of this has been done by linting the codebase and will continue to improve, so I am closing this.

PR #1142, opened 2020-09-09
Replace `util::either` with `util::expected`.

### LINKED ISSUE #1135: Replace value-or-error usage of `util::either` with `util::expected`
`either` was written as a 2-slot variant type _and_ as a success-value-or-error-return type. For the former, we now have `util::variant`; for the latter, it is preferable to implement a work-a-like for the proposed `std::expected` type than to keep subclassing `util::either` to add the required functionality.

Proposal:
* Implement a `util::expected` class following (as C++14 allows) the interface proposed in http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0323r7.html .
* Replace usage of `either` in the code base with `variant` or `expected`.

When all `either` usages are removed, we can remove the class too.

halfflat:
* Implement a C++14-compatible workalike for the proposed `std::expected` class: see http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0323r9.html .
* Replace use of `either` with `expected` in `mprovider`, `mechanism_catalogue`, `util::partition_range`, and `pyarb::hopefully`.
* Replace use of `either` with `variant` in `util::sentinel_iterator`.
* Add `in_place_t` constructor for `util::optional`.
* Fix move assignment bug in `util::variant`.
* Remove `util/either.hpp` and associated tests.

Fixes #1135.


halfflat:
I'm not saying that those Travis failures are the compiler, but I think it's the compiler. I'll investigate further.

arbor/include/arbor/util/optional.hpp
```diff
@@ -45,6 +43,9 @@
 namespace arb {
 namespace util {
 
+struct in_place_t {};
+static constexpr in_place_t in_place{};
```
bcumming, line 47: Should this go to `meta.hpp`, given that in theory it can be used by other types, like `any`?

Mind, you could just use the C++17 equivalent now that we have upgraded...

arbor/util/partition.hpp
```diff
@@ -52,7 +52,7 @@ class partition_range: public range<partition_iterator<I>> {
     void validate() const {
         auto ok = is_valid();
         if (!ok) {
-            throw invalid_partition(ok.second());
+            throw invalid_partition(ok.error());
```
bcumming, line 55: I am always happy when I see ambiguous `first` and `second` get replaced with meaningful names.

bcumming (approved):
Looks good, modulo one small comment, and a suggestion that all of the C++14-specific changes herein want changing after the C++17 merge is in.

arbor/include/arbor/util/optional.hpp
halfflat: Yep, we'll get C++17 PR in first, and then I'll update this PR to just use `std::in_place_t` instead.

halfflat:
Presuming the updated PR passes and it's still accepted, we can either merge it as is and I can fix the pre-C++17isms after (as part of  #1022), or I can just patch things like `in_place_t` and update the PR again.

bcumming:
bors try.

arbor/mechcat.cpp
```diff
@@ -72,42 +71,26 @@ template <typename V>
...
 }
 
-void value(const hopefully<void>& x) {
-    if (!x) {
-        std::rethrow_exception(x.second());
-    }
-}
+// Conveniently make an unexpected exception_ptr:
+template <typename X>
+auto unexpected_exception_ptr(X x) { return unexpected(make_exception_ptr(std::move(x))); }
```
bcumming, line 92: I see that you started to run out of vertical space during the refactoring ;)

bcumming: Also, `std::make_exception_ptr` is just bad naming, because it does a lot more than the usual `make_x` construction helper functions like `make_unique_ptr`.

bcumming (approved):

arbor/mechcat.cpp
halfflat: It was just getting a bit verbose :)

halfflat:
bors try.

Issue #681, opened 2019-01-31
Sanity checking recipes


halfflat:
It's currently very easy to make a `recipe` that is inconsistent with regards to targets and connections, that will then happily crash the simulator.

Task:
* provide checks at model initialization time (and throw an exception) so that we never casually segfault.
* provide some functionality that allows a recipe to be checked for consistency in user code, independent of creating a `simulation` object.

PR #1143, opened 2020-09-15
Sanitize recipe

noraabiakar:
* Raise an exception if:
1. `recipe.num_souces(gid)` != number of detectors placed on the cell.
2. `recipe.num_targets(gid)`  != number of synapses placed on the cell.
3. `recipe.connections_on(gid)` has connections with non-existent source or target gids or lids
* Raise better exception for gap junctions
* Fix unit test 

Addresses #681

### DIFF
--- a/arbor/arbexcept.cpp
+++ b/arbor/arbexcept.cpp
@@ -58,10 +58,10 @@ gj_unsupported_domain_decomposition::gj_unsupported_domain_decomposition(cell_gi
 {}
 
 gj_connection_mismatch::gj_connection_mismatch(cell_gid_type gid, cell_member_type site_0, cell_member_type site_1):
-        arbor_exception(pprintf("recipe::gap_junctions_on(gid={}) -> ({}, {}) <-> ({}, {}): one of the sites must be on the cell with gid = {})", gid, site_0, site_1, gid)),
-        gid(gid),
-        site_0(site_0),
-        site_1(site_1)
+    arbor_exception(pprintf("recipe::gap_junctions_on(gid={}) -> ({}, {}) <-> ({}, {}): one of the sites must be on the cell with gid = {})", gid, site_0.gid, site_0.index, site_1.gid, site_1.index, gid)),
+    gid(gid),
+    site_0(site_0),
+    site_1(site_1)
 {}
 
 bad_gj_connection::bad_gj_connection(cell_gid_type gid, cell_member_type site_0, cell_member_type site_1):
--- a/arbor/arbexcept.cpp
+++ b/arbor/arbexcept.cpp
@@ -15,30 +15,34 @@ bad_cell_description::bad_cell_description(cell_kind kind, cell_gid_type gid):
     gid(gid),
     kind(kind)
 {}
-
 bad_target_description::bad_target_description(cell_gid_type gid, cell_size_type rec_val, cell_size_type cell_val):
-    arbor_exception(pprintf("recipe::num_targets(gid={}) -> {} does not match the number of synapses on the cell -> {})", gid, rec_val, cell_val)),
+    arbor_exception(pprintf("Model building error on cell {}: recipe::num_targets(gid={}) = {} is greater than the number of synapses on the cell = {}", gid, gid, rec_val, cell_val)),
     gid(gid), rec_val(rec_val), cell_val(cell_val)
 {}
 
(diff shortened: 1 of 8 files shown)


bcumming (commented):
Can we have some unit tests that verify that incorrectly formulated recipes throw the correct exceptions?

bcumming:
bors try

arbor/arbexcept.cpp
```diff
@@ -16,6 +16,31 @@ bad_cell_description::bad_cell_description(cell_kind kind, cell_gid_type gid):
     kind(kind)
 {}
 
+bad_target_description::bad_target_description(cell_gid_type gid, cell_size_type rec_val, cell_size_type cell_val):
```
bcumming, line 19: These error messages will be reported to users who might not be writing a recipe (e.g. using a tool that generates the recipe for them), so I think it is good to present the error information as generally as possible, e.g:
`Model building error on cell 23: connection target 108 is out of range. Cell 23 has 88 targets, in the range [0:87].`

arbor/communication/communicator.cpp
```diff
@@ -33,6 +34,12 @@ communicator::communicator(const recipe& rec,
     num_local_groups_ = dom_dec.groups.size();
     num_local_cells_ = dom_dec.num_local_cells;
 
+    std::vector<cell_size_type> cell_num_sources;
```
bcumming, line 37: Storing these for all cells in the model will use a lot of memory for large models. Can we just look these up at runtime, and optimize it later if need be?

arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -538,6 +537,23 @@ void fvm_lowered_cell_impl<Backend>::initialize(
     for (auto cell_idx: make_span(ncell)) {
         cell_gid_type gid = gids[cell_idx];
 
+        // Sanity check recipe
```
bcumming, line 540: This could be simplified slightly, and have the tests on the variables closer to where they are defined.
This will make opting out of specific checks easier if we decide to allow that later for performance reasons.
```C++
auto& cell = cells[cell_idx];
if (rec.num_source(gid) != cell.detectors().size()) {
    throw ...;
}
auto cell_targets = util::sum_by(cell.synapses(), [](auto& syn) {return syn.second.size();});
if (num_targets != rec.num_targets(gid)) {
    throw ...;
}
```

test/simple_recipes.hpp
```diff
@@ -117,7 +117,11 @@ class cable1d_recipe: public simple_recipe_base {
     }
 
     cell_size_type num_targets(cell_gid_type i) const override {
-        return cells_.at(i).synapses().size();
+        cell_size_type total_synapses = 0;
```
bcumming, line 120: Here you can use `util::sum_by` to avoid the for loop.

bcumming (changes requested):
Looks good, just a few comments. 
The main concern is to make uniform error messages that clearly readable to a user who isn't directly creating a recipe.

arbor/communication/communicator.cpp
noraabiakar: I thought it might make more sense to do it this way cause otherwise we're calling `num_sources(gid)` once for every connection, and there's typically many more connections  than cells. But I can undo it.

arbor/arbexcept.cpp
noraabiakar: Okay, I'll write more useful messages.

arbor/fvm_lowered_cell_impl.hpp
noraabiakar: Sure.

test/simple_recipes.hpp
noraabiakar: Right, I always forget what we have in `util`

noraabiakar:
bors try

test/simple_recipes.hpp
bcumming: Me too!

arbor/communication/communicator.cpp
bcumming: It's a trade off: and which approach is best will depend on the model.
I assumed that the vast majority of models will automatically return 1 for `num_sources`, calculating the value on the fly would be cheap... but worst case scenario they generate the cell every time they ask the question.

noraabiakar: I agree. It makes more sense to save the memory.

bcumming (approved):

PR #1152, opened 2020-09-17
Replace `util::any` and `util::variant` with std:: equivalents.

### LINKED ISSUE #1144: C++17 update: replace `util::any` and `util::variant` with `std::` versions.
As part of the C++17 update process (#1022), do the easier standard library replacements:
* Substitute `std::any` for `util::any`. Where it is being applied in a `std::any` context, replace usage of `util::any_cast` with `std::any_cast`. (Perhaps this awkwardness can be addressed with a bit of ADL magic.)
* Substitute `std::variant` for `util::variant`.

Doing `std::optional` is a bit more tricky, because `std::optional<T&>` is not a thing; this can be its own ticket.

halfflat:
* Replace usages of `util::any` with `std::any`.
* Modify `util::any_cast` so that it forwards `std::any` arguments to `std::any_cast`; move into own header.
* Prevent `std::unique_any` from implicitly encapsulating `std::any` objects.
* Replace `test_any.cpp` unit tests with `test_any_cast.cpp` unit tests.
* Replace some usages of `std::is_same<X, Y>::value` with `std::is_same_v<X, Y>` as part of general C++17ification.
* Remove `util/variant.hpp`, `test_variant.cpp`.
* Replace occurrences of `util::variant` with `std::variant`.
* Replace `util::variant::get<N>` with `std::get<N>`.
* Replace `util::get_if<N>` with `std::get_if<N>`, passing a pointer to variant instead of a ref or value.
* Add some standard #includes where they were required in the source, but only included transitively.
* Some minor comment tidying and fix-ups.
* Update C++ API docs.

Fixes #1144.


bcumming (approved):
All good.

I was fond of my `util::any`, but `std::any` appears to have implemented its interface well-enough ;)

PR #1161, opened 2020-09-25
Remove const_cast

bcumming:
Remove the only cases of `const_cast` in the library (except for those required for calls to external C APIs).

I don't know how or when these snuck into the code, but they do not appear to be necessary.


halfflat (approved):
I wonder why we needed them in the first place, too!

PR #1163, opened 2020-09-28
Add BBP catalogue

noraabiakar:


noraabiakar:
bors try

arbor/include/arbor/mechcat.hpp
```diff
@@ -122,5 +122,6 @@ class mechanism_catalogue {
 
 const mechanism_catalogue& global_default_catalogue();
 const mechanism_catalogue& global_allen_catalogue();
+const mechanism_catalogue& global_bbp_catalogue();
```
thorstenhater, line 125: This is more of a TODO item than a request for change in this PR: Shouldn't we start to have a more dynamic catalogue of catalogues here? Something like a map<string, cat> and a register_catalogues helper? However
to make this right, we would need some kind of plugin mechanism.

mechanisms/bbp/Nap_Et2.mod
```diff
@@ -0,0 +1,57 @@
...
+
+    mInf = m_inf(v)
+    mAlpha = 0.182*6*exprelr(-(v + 38)/6)
+    mBeta  = 0.124*6*exprelr( (v + 38)/6)
+    mTau   = 6.0/(mAlpha + mBeta)
+
+    hInf = h_inf(v)
+    hAlpha = 2.88e-6*4.63*exprelr( (v + 17.0)/4.63)
+    hBeta  = 6.94e-6*2.63*exprelr(-(v + 64.4)/2.63)
+    hTau   = 1.0/(hAlpha + hBeta)
```
thorstenhater, line 40: This could be hRho = (hAlpha + hBeta) to save a whole division. Similar for mTau.

mechanisms/bbp/K_Tst.mod
```diff
@@ -0,0 +1,55 @@
...
+}
+
+STATE {
+    m
+    h
+}
+
+BREAKPOINT {
+    SOLVE states METHOD cnexp
+    ik = gK_Tstbar*(m^4)*h*(v-ek)
```
thorstenhater, line 29: m^4 will call exp, which might be slower than neccessary. I would not change this here, but consider making a patch to modcc instead.

thorstenhater (changes requested):
Hi,

I have only a minor request, and two more general comments.

arbor/include/arbor/mechcat.hpp
noraabiakar: Let's bring it up on slack, maybe add an issue for it.

mechanisms/bbp/Nap_Et2.mod
noraabiakar: Right, I'll fix that

mechanisms/bbp/K_Tst.mod
noraabiakar: I think that might be more trouble than it's worth.

arbor/include/arbor/mechcat.hpp
noraabiakar: I can see that it's already there's already an issue. Thanks!

mechanisms/bbp/K_Tst.mod
thorstenhater: That might also be true

thorstenhater (approved):
Looks good.

PR #1165, opened 2020-09-29
New SWC parser broke `load_swc_allen`.

halfflat:
* Minor output formatting fix for `swc_record`.
* Modify the Python `load_swc_allen` implementation to cope with SWC record ids not necessarily being contiguous, and with SWC record parent ids corresponding to record ids, not 0-based indices.

### DIFF
--- a/arbor/swcio.cpp
+++ b/arbor/swcio.cpp
@@ -52,7 +52,7 @@ std::ostream& operator<<(std::ostream& out, const swc_record& record) {
 
     out.precision(std::numeric_limits<double>::digits10+2);
     return out << record.id << ' ' << record.tag << ' '
-               << record.x  << ' ' << record.y   << ' ' << record.z << ' ' << record.r
+               << record.x  << ' ' << record.y   << ' ' << record.z << ' ' << record.r << ' '
                << record.parent_id << '\n';
 }
 
--- a/python/morphology.cpp
+++ b/python/morphology.cpp
@@ -27,28 +27,32 @@ arb::segment_tree load_swc_allen(const std::string& fname, bool no_gaps=false) {
                 throw pyarb_error("Allen SWC: empty file");
             }
 
-            // assert that root sample has tag 1.
+            // Map of SWC record id to index in `records`.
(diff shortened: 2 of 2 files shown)


bcumming:
The Allen/SONATA spec states that record ids must be contiguous, and furthermore that ids on the same unbranched cable have to be contiguous.

brenthuisman:
Unfortunately, I get this following result for the poster example:
[arbor.pdf](https://github.com/arbor-sim/arbor/files/5298614/arbor.pdf)
We used to match the Neuron curve.

halfflat:
@bcumming We can add a check for that if you like. (But do we really need to validate the SONATA spec in the reader?)
The big problem was that record id != record index.

halfflat:
@brenthuisman I just verified that this gives the same segment tree as did master before the new SWC code. The discrepancy is coming from somewhere else.

brenthuisman:
@halfflat Can I help you find the problem? A lot changed since the last verified working commit, not sure where to begin.

brenthuisman (approved):

bcumming:
Was the problem that it crashed with non-contiguous id values?

This reader has the description `Generate a segment tree from an SWC file following the rules prescribed by AllenDB and Sonata.`, because AllenDB (at least nominally) uses SONATA. It wasn't enforcing that rule, though it should to comply with the specification.

The broader problem is that files in the AllenDB don't actually follow their own rules, hence the need for the `no_gaps` flag, so another flag would be needed. They just rely on NEURON's import3d mess.

Nora and I planned to create `load_swc_neuron` that follows the rules that NEURON follows (and documents them).

We could then remove this method, and create `load_swc_sonata` with strict compliance.

brenthuisman:
I needed this PR + PR #1169 in order to have the poster example run again. [This](https://github.com/thorstenhater/bernstein-2020/blob/master/src/cell.swc) is the `.swc` file it is using. Lines like [l110](https://github.com/thorstenhater/bernstein-2020/blob/master/src/cell.swc#L110) break the assertion of contiguity (new word!).

PR #1166, opened 2020-09-29
Update Python examples

brenthuisman:
* Python examples: Updated to use region expressions (quotes)
* Python examples: Uniform plotting
* Python examples: Removed tutorial directory (example roughly matches python/example/single_cell_model.py)
* Python examples: Rewrote flat_cell_builder() to segment_tree()


python/example/network_ring.py
```diff
@@ -1,44 +1,53 @@
...
+
+    # Soma (tag=1) with radius 6 μm, modelled as cylinder of length 2*radius
+    s = tree.append(arbor.mnpos, arbor.mpoint(-12, 0, 0, 6), arbor.mpoint(0, 0, 0, 6), tag=1)
 
-    # Soma with radius 6 μm, modelled as cylinder of length 2*radius
-    s  = b.add_cable(length=12, radius=6, name="soma", ncomp=1)
-    # Single dendrite of length 100 μm and radius 2 μm attached to soma.
-    b1 = b.add_cable(parent=s, length=100, radius=2, name="dend", ncomp=1)
-    # Attach two dendrites of length 50 μm to the end of the first dendrite.
+    # Single dendrite (tag=2) of length 100 μm and radius 2 μm attached to soma.
```
noraabiakar, line 28: tag 2 is typically associated with an axon, according to the SWC specs, tag 3 is dendrite. It doesn't matter here, but it's probably better to be consistent

python/example/single_cell_multicompartment.py
```diff
@@ -23,32 +24,42 @@
...
+
+s = tree.append(arbor.mnpos, arbor.mpoint(-12, 0, 0, 6), arbor.mpoint(0, 0, 0, 6), tag=1)
 
 # Add the dendrite cables, labelling those closest to the soma "dendn",
 # and those furthest with "dendx" because we will set different electrical
 # properties for the two regions.
-b0 = b.add_cable(parent=s,  length=100, radius=2, name="dendn", ncomp=100)
+
+labels = arbor.label_dict()
+labels['soma'] = '(tag 1)'
```
noraabiakar, line 35: I would also give different tags for `dendn` and `dendx` something >= 5 so as not to interfere with the "reserved" SWC tags

noraabiakar (commented):
Are the `single_cell_model.py` and `single_cell_multicompartment.py` examples meant to be similar except for that the first is a single branch, and the second is multiple branches? If so, maybe we can give them better names.

python/example/single_cell_multicompartment.py
bcumming: I think that this might be ok: Arbor explicitly lets you define tags how you please.

noraabiakar: We aren't calling these things SWC samples or using any SWC loaders so you're right that it doesn't matter here. I've been using too much SWC. But it might be better to stick to the convention regardless.

brenthuisman: Done

brenthuisman:
> Are the `single_cell_model.py` and `single_cell_multicompartment.py` examples meant to be similar except for that the first is a single branch, and the second is multiple branches? If so, maybe we can give them better names.

You mean the filenames? `single_cell_model`->`single_cell_single_compartment`?

noraabiakar:
@Brent, yes I mean the filenames. They're not actually single/multi-compartment, they're single/multi-branch. The compartments typically refer to the discretization which is independent of the morphology.

PR #1164, opened 2020-09-29
Fix python and C++ examples

noraabiakar:
- Update `dryrun` to generate some spikes again, add a voltage printer
- Fix `ring.py` recipe
- Fix labels in `single_cell_builder.py` and `single_cell_swc.py`

PR #1168, opened 2020-09-29
Fix assorted typos

brenthuisman:
Fix assorted typos


halfflat (approved):
I'm slightly worried that the probe method change might break the python examples, but as they're already broken on probes, I don't think that should stop us from merging the PR.

noraabiakar:
@halfflat the probe method change is part of #1164, and was changed to fix the python ring example. I think we should address that PR first.

PR #1169, opened 2020-09-29
Fix copy paste error on typeful ion painting.

halfflat:


brenthuisman (approved):
This together with #1165  fixes the poster example.

PR #1170, opened 2020-10-02
Tighten SWC requirements, comments.

halfflat:
* Remove contiguous numbering requirement from 'strict' mode SWC parsing, as it is not necessary for correct interpretation as a segment tree.
* Describe in the swcio.hpp header how (strict) swc data is interpreted as a segment tree.
* Demonstrate in unit tests that out-of-order and non-contiguously numbered records are valid for `parse_swc`.
* Add unit test for `as_segment_tree`.


noraabiakar (approved):
Looks good! thanks Sam.

PR #1173, opened 2020-10-05
bump pybind11 to v2.5

bcumming:


brenthuisman (approved):
Our poster model runs OK, docs are OK, let's wait for Travis' OK and then all is gonna be OK.

PR #1174, opened 2020-10-06
Fixes the curious case of the missing quotes in the Python single cell model example, both code and docs.

brenthuisman:
Tested all Python examples on current master successfully.


noraabiakar (approved):
lgtm!

Issue #1172, opened 2020-10-05
Use autodoc for Python API reference


brenthuisman:
Using the `sphinx.ext.autodoc` plugin to autogenerate Python Api documentation accomplishes the following:

* API documentation is never out of sync with the code.
* Less work: write documentation once!
* Consistency: `help()` and the docs will show the same documentation.

Requirements:
* `sphinx.ext.autodoc` requires an importable `arbor` module, so building the docs requires Arbor be built.
* Integrate current docs into docstrings. Format: Google style docstring + Python type annotations.
    * Python type annotation: https://docs.python.org/3/library/typing.html
    * [Napoleon extension](https://sphinxcontrib-napoleon.readthedocs.io): [Google style docstring](https://google.github.io/styleguide/pyguide.html) (more brief)

Options:
* Module, class, member function or variable documentation can be inlined anywhere in the docs to facilitate deviation in the docs w.r.t docstrings if needed. Let's do this as little as possible (creates one more thing to track!).
* Autodoc shows all members by default. Can be tuned, this removes everything I consider noise for a full module documentation page:

```
.. automodule:: arbor
    :imported-members:
    :members:
    :undoc-members:
    :special-members: __init__
    :inherited-members:
```

brenthuisman:
Regarding type signatures, things are hairy. A few ways to do it exist:

Python code:
* [Python-sanctioned annotations](https://docs.python.org/3/library/typing.html)
* These can apparently live in separate stub files as per [PEP 484](https://github.com/sphinx-doc/sphinx/issues/7630), see also [stubgen](https://mypy.readthedocs.io/en/stable/stubgen.html)
* Pybind generates docstrings from the type info it can extract from the code. [The quality of this is however in development](https://github.com/pybind/pybind11/issues/2619). Pybind itself uses stubfiles (`.pyi`).
  * TODO: The format is Sphinx style docstrings [VERIFY, DUMP reference.rst].

Python docstrings:
* A Python docstring is a string placed in `.__doc__`.
  * Classes can rely on docstrings on its member or members can be documented in the class docstring.
* [Sphinx style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#napoleon-marching-toward-legible-docstrings)
* [Numpy and Google style docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/#google-vs-numpy). This seems to convert those inputs to the Sphinx style docstring, which is then further processed.
[...]

PR #1176, opened 2020-10-06
Fix autodoc Python API reference

brenthuisman:
* Added script `doc/scripts/autodoc_cache.py`.  It generates `doc/py_reference.rst` if the Arbor Python package is importable.
  `doc/py_reference.rst` should be checked in, such that the Python API references will be generated
  even if `arbor` can't be imported (on RTD).
  On occasion, it should be updated by building the docs locally and git adding `py_reference.rst`.
* Added `doc/py_reference_generator.rst`. It is required to cause autodoc to be run.

Provides _a_ fix for #1172

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -2,13 +2,6 @@
 # -*- coding: utf-8 -*-
 import sys, os
 
-# Path to Python Binding (_arbor)
-try:
-    sys.path.append(os.path.join(os.environ['OLDPWD'],"python"))
-    import arbor
-except ImportError:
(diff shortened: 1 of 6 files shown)


bcumming (changes requested):
Can we just not include automatically generated docs?

* This is a big file, and I want to understand its impact on the size of the repository.
* Every option like increases our surface area for bugs and issues building the docs

Let's stabilise the docs first, before adding more features like this.

brenthuisman:
No longer required, Arbor builds on RTD.

PR #1177, opened 2020-10-06
Fixes imports that are no longer required and causes RTD to fail building

brenthuisman:


doc/scripts/make_images.py
```diff
@@ -12,6 +10,8 @@
 #
 
 def dataframe_line_plot(input_csv, output_svg):
+    import seaborn
```
bcumming, line 13: How does this fix the problem? It is still importing the problematic dependencies.

Instead, I thought that we had agreed to include the svg file in the documentation, and completely remove this code for generating the image.

bcumming (changes requested):
The dependency is still there and should be removed.

doc/scripts/make_images.py
brenthuisman: Yes, I left the function there just in case (it's not called). If you prefer, I can remove it entirely.

bcumming: Yes, let's remove it entirely.

bcumming (approved):

PR #1180, opened 2020-10-07
Add optimised HH.

thorstenhater:
Provide a cleaned-up and optimised version of the HH-mechanism. Locally I see a ~5% improvement on the busyring benchmark with HH on dend and soma.

These optimisations are in-line with those done on the BBP and Allen catalogues.

### DIFF
--- a/mechanisms/default/hh.mod
+++ b/mechanisms/default/hh.mod
@@ -1,10 +1,9 @@
 NEURON {
     SUFFIX hh
     USEION na READ ena WRITE ina
-    USEION k READ ek WRITE ik
+    USEION k  READ ek  WRITE ik
     NONSPECIFIC_CURRENT il
-    RANGE gnabar, gkbar, gl, el, gna, gk
-    GLOBAL minf, hinf, ninf, mtau, htau, ntau
+    RANGE gnabar, gkbar, gl, el, q10
 }
 
 UNITS {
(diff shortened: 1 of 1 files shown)


brenthuisman:
https://travis-ci.org/github/arbor-sim/arbor/jobs/733603551#L557

bcumming:
Was this tested to verify that exactly the same results are produced when run in an `examples/ring`?

brenthuisman:
Nearly exactly.

thorstenhater:
The spike count and rate are exactly the same for my test casein nsuite, feel free to verify. 
A deviation here would be too awkward.

@brenthuisman Why nearly? These should be exact.

bcumming:
"These should be exact". I think we need to see the spike times from the ring example with and without this.

@thorstenhater, can you please provide validation of the new hh mechanism using the examples in the main repository?

thorstenhater:
@bcumming My "These should be exact" was directed at the "nearly exactly" comments.

The nsuite test (256 cells) I used has this result
```
831 spikes generated at rate of 0.240674 ms between spikes
```
for the spike rates in both cases.
Does the ring example in the main repo provide more detail or is more reliable?

noraabiakar:
For an added level of verification, I usually compare the generated voltages at a location where the mechanism is active.

bcumming:
Spike count doesn't validate that the mechanism is the same: you can have small variations in the mechanism and the same number of spikes.
At a minimum, we have to inspect the times at which the spikes occur, and for a change like this, I would expect bit wise reproducibility. If not, then a bit of investigation using th approach that Nora advocated would be required to satisfy you that the results are, indeed, unchanged.

brenthuisman:
@thorstenhater Spike times were identical here, I saw _very_ slight variations in the plot. [ring.zip](https://github.com/arbor-sim/arbor/files/5363755/ring.zip)

mechanisms/default/hh.mod
```diff
@@ -13,83 +12,74 @@ UNITS {
...
 
 BREAKPOINT {
     SOLVE states METHOD cnexp
-    gna = gnabar*m*m*m*h
-    ina = gna*(v - ena)
-    gk = gkbar*n*n*n*n
-    ik = gk*(v - ek)
-    il = gl*(v - el)
+          
+    ina = gnabar*m*m*m*h*(v - ena)
```
bcumming, line 29: For these calculations, I would prefer to still calculate `gna` and `gna` as `LOCAL` variables

We evaluate the expression `ik = g*(v - ek)` to determine the conductance `g`, which gives the following in the generated C++ code for the original version:
```C++
conductivity_ = gna[i_]+gk[i_]+gl[i_]
```
What does the new version give?

bcumming (changes requested):
Looks good. I just have one question/suggestion for how conductances are calculated, which I suspect might improve performance.

mechanisms/default/hh.mod
thorstenhater: I use the SIMD version here
```
assign(conductivity_, S::add(S::add(S::mul(S::mul(S::mul(S::mul(simd_cast<simd_value>(indirect(gnabar+index_, simd_width_)), simd_cast<simd_value>(indirect(m+index_, simd_width_))), simd_cast<simd_value>(indirect(m+index_, simd_width_))), simd_cast<simd_value>(indirect(m+index_, simd_width_))), simd_cast<simd_value>(indirect(h+index_, simd_width_))), simd_cast<simd_value>(indirect(gl+index_, simd_width_))), S::mul(S::mul(S::mul(S::mul(simd_cast<simd_value>(indirect(gkbar+index_, simd_width_)), simd_cast<simd_value>(indirect(n+index_, simd_width_))), simd_cast<simd_value>(indirect(n+index_, simd_width_))), simd_cast<simd_value>(indirect(n+index_, simd_width_))), simd_cast<simd_value>(indirect(n+index_, simd_width_)))));
```
so simplified
```
conductivity_ = add(add(mul(mul(mul(mul(indirect(gnabar+index_)), indirect(m+index_))), indirect(m+index_))), indirect(m+index_))), indirect(h+index_))), indirect(gl+index_))), mul(mul(mul(mul(indirect(gkbar+index_)), indirect(n+index_))), indirect(n+index_))), indirect(n+index_))), indirect(n+index_)))));
```
and this is
```
gnabar*m^3*h + gl + gkbar * n^4
```
which is the same as in the original.

I do not see how a LOCAL would help. What I would expect to help is storing and `m` and `n` into LOCALs to avoid multiple `load`s. That is if the compiler does not figure it out on its own.

thorstenhater:
For documentation:
- spike timings are bit-identical
- added a voltage trace to the soma. This reveals some variation in the last digit(s). Max delta squared is 1e-24 for me.

thorstenhater:
bors try

mechanisms/default/hh.mod
thorstenhater: Good point, quite an improvement.

bcumming (approved):

PR #1181, opened 2020-10-07
Make python examples executable and add shebang

schmitts:

PR #1183, opened 2020-10-07
Add python examples to CI

noraabiakar:
* Add some log messages in python examples
* Add python examples to CI


scripts/travis/build.sh
```diff
@@ -110,6 +110,10 @@ if [[ "${WITH_PYTHON}" == "true" ]]; then
     make pyarb -j4                                                           || error "building pyarb"
     progress "Python unit tests"
     python$PY $python_path/test/unit/runner.py -v2                           || error "running python unit tests (serial)"
+    python$PY $python_path/example/network_ring.py                           || error "running python network_ring example"
```
schmitts: maybe make use of https://github.com/arbor-sim/arbor/pull/1181 ?

bcumming (approved):
Looks good.

scripts/travis/build.sh
noraabiakar: Is there an advantage to doing it that way?

schmitts: it would test that the examples can be used as "standalone" executables

PR #1198, opened 2020-10-19
Allow using external C++ dependencies

schmitts:


bcumming:
bors try

bcumming:
bors try.

thorstenhater:
This might be considered to be blocked by https://github.com/arbor-sim/arbor/issues/1271

bcumming (changes requested):
Looks good.

There is an issue with Travis: the build scripts need to be updated to use the packaged dependencies, and not look for installed versions

brenthuisman:
Should `docs/install/build_install.rst` be updated?

setup.py
```diff
@@ -53,6 +53,7 @@ class install_command(install):
                         'none, cuda, cuda-clang, hip'),
         ('vec',   None, 'enable vectorization'),
         ('arch=', None, 'cpu architecture, e.g. haswell, skylake, armv8-a'),
+        ('bundled_libs', None, 'use bundled/git-submoduled 3rd party libraries (ON/OFF)')
```
bcumming, line 56: For the pip build, should we use the bundled libs by default? I would go as far as not providing the option to use bundled dependencies.

If not, then we have to add pybind11 and nlohmann/json as dependencies in pip (with all the head aches that brings).

brenthuisman: Yes we should. Many users (think of the notebooks!) are not going to have these libs installed, and `pip install` not working would make Arbor quickly top my worst-pip-package top1.

As this PR stands, I understand it will use the bundled libs (they have not been removed) and switch them on for a Python package build, so :+1: 

When we get around producing pre-compiled wheels we can consider different options, but until then we cannot require anything for the pip package.

doc/install/build_install.rst
```diff
@@ -152,7 +152,8 @@ the `Github repository <https://github.com/arbor-sim/arbor>`_:
     git clone https://github.com/arbor-sim/arbor.git --recurse-submodules
 
 We recommend using a recursive checkout, because Arbor uses Git submodules for some
-of its library dependencies.
+of its library dependencies. The usage of some of these libraries can be controlled by the
```
brenthuisman, line 155: Some or all?

setup.py
bcumming: If we don't use the bundled versions we would have to make these dependencies that would be installed by pip. I agree that installing Arbor should not fail because a missing dependency isn't available.

The main reason I suggest using the bundled libs is that we would then know that every user who installs version X of Arbor using pip is using version Y of pybind11 etc. This would make dealing with bug reports somewhat easier.

One downside is that if Arbor uses a different version of pybind11 to other modules on the system, there could be strange runtime library issues. But pip isn't great at managing such dependencies, so I don't know how well we could mitigate those efforts even with 'maximum effort".

brenthuisman (approved):

brenthuisman (changes requested):
I see the Github Action is failing due to Pybind being missing (I think). Probably solved by updating the configuration of CMake with the new flag.
https://github.com/arbor-sim/arbor/pull/1198/checks?check_run_id=1655029786

brenthuisman (approved):

bcumming:
bors try

bcumming:
bors try

bcumming (approved):

PR #1199, opened 2020-10-22
Add interface for querying mechanism names.

thorstenhater:
Support functionality to querying the names of mechanisms in a mechanism catalogue.


bcumming (changes requested):
Looks good.

Before merging, we need:
- [ ] unit tests that show that all mechanisms in the catalogue are represented in the list
- [ ] python wrapping
- [ ] documentation

arbor/mechcat.cpp
```diff
@@ -495,6 +495,15 @@ mechanism_catalogue::mechanism_catalogue():
     state_(new catalogue_state)
 {}
 
+std::vector<std::string> mechanism_catalogue::mechanism_names() const {
+    std::vector<std::string> result;
+    for (const auto& [k, v]: state_->info_map_) {
+        result.push_back(k);
+    }
+    return result;
+}
```
halfflat, line 504: This could be replaced with
```
std::vector<std::string> result = util::assign_from(util::keys(state_->info_map_);
return result;
```

Ben queried if this grabs all the mechanism names, and it does miss those mechanisms which have been explicitly derived. As these look just like mechanisms added with `mechcat::add()` from a user's point of view, these should probably be returned as well.

The corresponding code using `util::keys` from `util/maputils.hpp` and the helpers from `util/rangeutils.hpp` would be:
```
std::vector<std::string> result;
util::assign(result, util::keys(state_->info_map_);
util::append(result, util::keys(state_->derived_map_);
return result;
```

That said, as it relies upon the implementation maps in `catalogue_state`, I think it would be cleaner to implement it as a method on this object, and have the `mechanism_catalogue` method forward to that.

thorstenhater: Thanks for the tip.

About the names, I am wondering whether we should make including derived mechanisms optional, either by offering
two methods eg `base_mechanism_names` and `all_mechanism_names` or via an extra flag.

bcumming: @halfflat , "explicitly derived" means somebody has created a new mechanism with a new name, like `"my_fancy_hh"` with some modified parameters?
If so, then I think it is appropriate that the `mechanism_names()` method returns those too.

thorstenhater: Yes, that's true, but what about names like `nernst/x=ca` that get added via parameter override?

halfflat: While `foo/bar=baz` does get inserted into the maps, anything of that form is _potentially_ a derived mechanism. I'll have to take a closer look at that code, but I'd like to think that the insertion of those guys should be treated as an implementation detail, and that they should be treated _as if_ they are just created out of thin air when required, and not appear on a list of mechanisms.

halfflat: @bcumming Exactly what I meant, yep.

thorstenhater: Ok, that I did not fully understand, so here's some follow questions:
- Auto-derived mechanisms (via parameter-override); are they added or not?
- If they are added, can I filter them out?
- Is there the potential for someone making a derived mechanism with a name _looking_ like an auto-derived one?

thorstenhater:
bors try

halfflat (approved):
The problem of what to do with "foo/bar=quux" names can be a topic for another PR, I think :)
Looks good.

bcumming (approved):

PR #1203, opened 2020-10-28
Correct sense of target vs synapse count test.

### LINKED ISSUE #1202: Test for targets without synapses has wrong sense.
**Describe the bug**

The `arb::bad_target_description` error is thrown when the target count is less than the number of synapses, instead of greater.

**To Reproduce**
1. Build a recipe which reports fewer targets on a gid than there are synapses on the corresponding cell description for a cable cell.
2. Initialize a simulation: an exception of type `arb::bad_target_description` will be thrown, with an error message claiming that the target count is _greater_ than the number of synapses on the cell.

**Expected behavior**

This exception should only be thrown if the recipe `num_targets(gid)` reports a larger number than there are corresponding synapses on a cable cell. This corresponds to the situation where events may potentially be delivered to non-existent synapse instances.

halfflat:
* Swap arguments in target count check in `fvm_lowered_cell_impl.hpp`.
* Update unit test to suit.

Fixes #1202.

### DIFF
--- a/arbor/fvm_lowered_cell_impl.hpp
+++ b/arbor/fvm_lowered_cell_impl.hpp
@@ -540,7 +540,7 @@ void fvm_lowered_cell_impl<Backend>::initialize(
             throw arb::bad_source_description(gid, rec.num_sources(gid), cell.detectors().size());;
         }
         auto cell_targets = util::sum_by(cell.synapses(), [](auto& syn) {return syn.second.size();});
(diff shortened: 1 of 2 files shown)


noraabiakar (approved):
I must have been confused when I wrote that. Thanks for fixing it!

halfflat:
No worries! Easy sort of error to make.

Issue #1212, opened 2020-11-03
Better mechanism errors. 


noraabiakar:
We have 2 methods to change mechanism parameters:
1. New mechanism derivation using `mechanism_desc("mech/x=...")`. This is for when parameter `x` is a `GLOBAL` parameter.
2. Parameter setting using `mechanism_desc("mech", {"x": ...})`. This is for when parameter `x` is a `RANGE` parameter. 

When these are called with the wrong type of parameter, they both generate the following warning: 
`mechanism mech has no parameter x`.
It would be clearer to specify whether the mechanism has no `GLOBAL`/`RANGE` parameter, to help users debug the error messages.

halfflat:
I'd like to solve this in a different way, maybe by calling only one of these a 'parameter' — GLOBAL and RANGE are NMODLisms that I'd like to keep out of the main code base, if possible.

For density mechanisms (if I get my way), we can distinguish between global 'parameters' used for derivation and conductance coefficients that correspond to a sort of areal density (of channels, or of conductance). The RANGE parameters could be called 'densities' or 'coefficients'.

This wouldn't work for point mechanisms though. There, unless we incorporate constant folding in run time mechanism code generation, there is nothing to be gained by using GLOBAL parameters, but the RANGE parameters _don't_ correspond to densities or coefficients.

An alternative is to remove this distinction entirely: derivation becomes only a way to change ion variables; GLOBAL parameters are disallowed (or silently treated as non-GLOBAL); parameters in density mechanisms that have a nonlinear effect on current contributions and conductance are detected at mechanism-compile time, and changes in these values automatically generate extra instances of the mechanism. When we add support for spatially-dependent parameter values, these nonlinear ones would be explicitly excluded.

Could we discuss this in #dev or at the next call?

noraabiakar:
I think we should discuss this at the next call. For now we can modify the exceptions to not mention RANGE and GLOBAL. I agree that they should probably stay out of arbor.

noraabiakar:
Closing this issue, discussion continues in #1226

PR #1215, opened 2020-11-03
Split `no_such_parameter` exception

noraabiakar:
Split the `no_such_parameter` exception such that 
* `mechanism_desc("mech/x=...")` throws `mechanism mech has no global x that can be derived` 
* `mechanism_desc("mech", {"x": ...})` throws `mechanism mech has no parameter x that can be set` 

Addresses #1212


noraabiakar:
Closing this PR and extending the scope of the issue.

PR #1216, opened 2020-11-03
Docs: various changes

noraabiakar:
Fix typos, reword some sections.

### DIFF
--- a/doc/concepts/cell.rst
+++ b/doc/concepts/cell.rst
@@ -83,6 +83,7 @@ Cell kinds
 
    LIF cells are single compartment leaky integrate and fire neurons with one **source** and one **target**.
    LIF cells do not support adding additional **sources** or **targets**. They do not support **gap junctions**.
+   They are typically used to simulate point-neuron networks.
 
 3. **Spiking Cells**
 
--- a/doc/concepts/cell.rst
+++ b/doc/concepts/cell.rst
@@ -76,23 +76,27 @@ Cell kinds
    * **Gap Junction Sites**: These refer to the sites of :ref:`gap junctions <modelgapjunctions>`.
      They are declared by specifying a location on a branch of the cell.
 
-   Because cable cells are the main cell kind in Arbor and have more properties than listed here, they have a :ref:`dedicated page <cablecell>`.
+   Because cable cells are the main cell kind in Arbor and have more properties than listed here, they have a
+   :ref:`dedicated page <cablecell>`.
 
 2. **LIF Cells**
 
-   A single compartment leaky integrate and fire neuron with one **source** and one **target**.
-   LIF cells does not support adding additional **sources** or **targets** or gap junctions.
+   LIF cells are single compartment leaky integrate and fire neurons with one **source** and one **target**.
+   LIF cells do not support adding additional **sources** or **targets**. They do not support **gap junctions**.
 
 3. **Spiking Cells**
 
-   Spike source from values inserted via a `schedule description`. It is a point neuron with one built-in **source** and no **targets**.
-   It does not support adding additional **sources** or **targets**. It does not support gap junctions.
+   Spiking cells act as spike sources from values inserted via a `schedule description`.
+   They are point neurons with one built-in **source** and no **targets**.
+   They do not support adding additional **sources** or **targets**. They do not support **gap junctions**.
 
 4. **Benchmark Cells**
 
-   Proxy cell used for benchmarking, and used by developers to benchmark the spike exchange and event delivery infrastructure.
+   Benchmark cells are proxy cells used for benchmarking, and used by developers to benchmark the spike exchange and
+   event delivery infrastructure.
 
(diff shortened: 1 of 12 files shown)


doc/concepts/cell.rst
```diff
@@ -76,23 +76,27 @@ Cell kinds
...
 
 4. **Benchmark Cells**
 
-   Proxy cell used for benchmarking, and used by developers to benchmark the spike exchange and event delivery infrastructure.
+   Benchmark cells are proxy cells used for benchmarking, and used by developers to benchmark the spike exchange and
+   event delivery infrastructure.
 
-Most Arbor users will want to use the cable cell, because it's the only cell kind that supports complex morphologies and user-defined mechanisms. See cable cells :ref:`dedicated page <cablecell>`. The LIF cell can be used to build networks with point-neurons.
+Most Arbor users will want to use the cable cell, which is the only cell kind that supports complex morphologies
+and user-defined mechanisms. You can visit the :ref:`cable cell page <cablecell>` for more information.
```
brenthuisman, line 99: You removed the hint that LIF cells can be used to make a point-neuron simulation. I can be reworded, but having this hint here would clarify that you in fact can build such a simulation in Arbor.

doc/concepts/recipe.rst
```diff
@@ -38,7 +39,8 @@ The steps of building a simulation from a recipe are:
 
 .. topic:: 1. Load balancing
 
-    First, the cells are partitioned over MPI ranks, and each rank parses
+    This can be done manually, by creating a :ref:`domain decomposition<modeldomdec>`
+    object. Or by using an Arbor-provided load balancer. First, the cells are partitioned over MPI ranks, and each rank parses
```
brenthuisman, line 43: You *either* use domain decomp *or* the Arbor load balancer, which is what is explained here, right? Maybe start the section with sth like "There are two ways to balance load, 1. domain decomp and 2. use an Arbor-provided balancer. This automatic balancer works thus: ..."

doc/install/index.rst
```diff
@@ -3,12 +3,12 @@
 Get Arbor
 =========
 
-Currently we offer three ways to get Arbor.
+Currently we offer two ways to get Arbor.
```
brenthuisman, line 6: I suppose Spack will be added shortly.

doc/tutorial/single_cell_model.rst
```diff
@@ -53,40 +53,43 @@ create the ``cable_cell`` that represents it are as follows:
...
-  channels all over the surface of the cell. :meth:`arbor.cable_cell.paint` lets us
-  instruct Arbor to use HH dynamics on the region we've labelled soma and sort the details
-  out for us.
-* Other properties should be added to the cell on a precise :class:`arbor.location`. This is
-  done using the :meth:`arbor.cable_cell.place<arbor.cable_cell.place>` method.
-  We place a current stimulus :class:`arbor.iclamp<arbor.iclamp>` with a duration of 2 ms
-  and a current of 0.8 nA, starting at 10 ms on the location we've labelled 'center'. We also
-  place a :class:`arbor.spike_detector<arbor.spike_detector>` with a threshold of -10 mV on the
-  same location.
+Step **(3)** constructs the :class:`arbor.cable_cell` from the segment tree and dictionary of labeled
```
brenthuisman, line 75: This will change with the new cable_cell, right? So this part of the example will need an update.

brenthuisman (changes requested):

doc/concepts/recipe.rst
noraabiakar: Well The arbor-provided load balancer will generate the domain decomposition object. So how about: 
```
This can be done manually by the user, through creating a domain decomposition object; 
or automatically by Arbor using a load balancer that generates the domain decomposition 
object according to some heuristics. 
```
I don't plan on going into too much detail about this in this PR. I can add more details to the domain decomposition page when I get to it later, then I can modify this page.

doc/install/index.rst
noraabiakar: Yes, but till then it's better to keep it correct :)

doc/tutorial/single_cell_model.rst
noraabiakar: Yes. Again, I'm just trying to keep the information correct for the current state of Arbor. We can change it again when the new changes are in.

doc/concepts/recipe.rst
brenthuisman: OK!

doc/tutorial/single_cell_model.rst
brenthuisman: OK.

doc/concepts/recipe.rst
noraabiakar: I'm actually going to undo these changes because I misread the section underneath. They don't make much sense here.

brenthuisman (approved):

PR #1217, opened 2020-11-05
Docs: concept/recipe

noraabiakar:
* Add an example describing the content of a recipe in the recipe concepts page in the docs.
* Mention that recipes are necessary, but can be abstracted away using the python `single_cell_model` (Hasn't been documented yet).


doc/concepts/recipe.rst
```diff
@@ -1,22 +1,77 @@
...
-  * the type of a cell;
-  * a description of a cell, e.g. with soma, synapses, detectors, stimuli;
-  * the number of spike targets;
-  * the number of spike sources;
-  * the number of gap junction sites;
-  * incoming network connections from other cells terminating on a cell;
-  * gap junction connections on a cell.
+  * The **number of cells** in the model.
+  * The **kind** of each cell.
+  * The **description** of each cell, e.g. with morphology, synapses, detectors, stimuli.
```
brenthuisman, line 11: I'm slightly confused as to what the **description** of a cell precisely refers to. Does it include spike targets, sources, etc? The [conceptual cable cell page](https://arbor-rtd-dev.readthedocs.io/en/latest/concepts/cable_cell.html) uses **description** in a broad fashion.

doc/concepts/recipe.rst
```diff
@@ -1,22 +1,77 @@
...
+  * The number of **spike targets**.
+  * The number of **spike sources**.
+  * The number of **gap junction sites**.
+  * Incoming **network connections** from other cells terminating on a cell.
+  * **Gap junction connections** on a cell.
+  * **Probes** on a cell.
+
+To better illustrate the content of a recipe, let's consider the following model of
```
brenthuisman, line 19: In case **description** includes all properties of a cable cell, I'd say sth like "To better illustrate the content of a recipe, let's consider a model of three cells. We give each of these three cells a **description**:"

doc/concepts/recipe.rst
```diff
@@ -1,22 +1,77 @@
...
+   | It's quite expensive to build complex cells, so we don't want to do this too often.
+     But when the simulation is first set up, it needs to now how cells interact with
+     one another in order to partition cells across nodes in an ideal manner. This is
+     why the number of **targets**, **sources** and **gap junction sites** is needed
+     separately from the cell description: with them, the simulation can tell that
+     ``cell 0`` has 1 **spike source** (the detector), 0 **spike targets**, and 0
+     **gap junction sites**, without having to build the cell.
+-  | ``Cell 1``: Is a soma and a single dendrite, with ``passive`` dynamics everywhere.
+     It has a single synapse at the end of the dendrite and a gap junction site in the
+     middle of the soma. This is the **description** of the cell.
```
brenthuisman, line 37: **description**

doc/concepts/recipe.rst
```diff
@@ -1,22 +1,77 @@
...
+     It's also a cable cell, which is its **cell kind**. It has 0 **spike sources**, 1
+     **spike target** (the synapse) and 1 **gap junction site**.
+-  | ``Cell 2``: Is a soma and a single dendrite, with ``passive`` dynamics everywhere.
+     It has a gap junction site in the middle of the soma. This is the **description**
```
brenthuisman, line 41: **description**

doc/concepts/recipe.rst
```diff
@@ -1,22 +1,77 @@
...
+     of the cell. It's also a cable cell, which is its **cell kind**. It has 0
+     **spike sources**, 0 **spike targets** and 1 **gap junction site**.
+
+The total **number of cells** in the model is 3. The **kind**, **description** and
+number of **spike sources**, **spike targets** and **gap junction sites** on each cell
+is known and can be registered in the recipe. Next is the cell interaction.
+
+The model is designed such that ``cell 0`` has a spike source, ``cell 1`` has
+a spike target and gap junction site, and ``cell 2`` has a gap junction site. A
+**network connection** can be fromed from ``cell 0`` to ``cell 1``; and a
```
brenthuisman, line 51: fromed -> formed

doc/concepts/recipe.rst
```diff
@@ -1,22 +1,77 @@
...
+on ``cell 2``. All this information is also registered via the recipe.
+
+The recipe is used to distribute the model across machines and is used in the simulation.
+Technical details of the recipe class are presented in the  :ref:`python <pyrecipe>` and
+:ref:`C++ <cpprecipe>` APIs.
+
+Are recipes always neccessary?
+------------------------------
+
+Yes. However, we provide a python :class:`single_cell_model <py_single_cell_model>`
```
brenthuisman, line 64: python -> Python

brenthuisman (changes requested):

doc/concepts/recipe.rst
noraabiakar: I'll make this a bit clearer. But description is not really a vague term in arbor, it's the cell itself. And our recipe method is called `get_cell_description`. I use the word description in the example to link back to the recipe terminology.

doc/concepts/recipe.rst
noraabiakar: It's not all the properties. I'll try to make it clearer below in the example.

doc/concepts/recipe.rst
```diff
@@ -1,22 +1,77 @@
...
+  * **Gap junction connections** on a cell.
+  * **Probes** on a cell.
+
+To better illustrate the content of a recipe, let's consider the following model of
+three cells:
+
+-  | ``Cell 0``: Is a single soma, with ``HH`` dynamics. In the middle of the soma, a
+     spike detector is attached, it generates a spiking event when the voltage goes
+     above 10 mV. In the same spot on the soma, a current clamp is also attached, with
+     the intention of triggering some spikes. This is the **description** of the cell.
```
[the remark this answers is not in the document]
noraabiakar: I'll make it clearer.

doc/concepts/recipe.rst
```diff
@@ -1,22 +1,77 @@
...
+   | ``Cell 0`` is a complex cell with dynamics, which should be modeled as a
+     :ref:`cable cell<model_cable_cell>`. This is the **kind** of the cell.
+   | It's quite expensive to build complex cells, so we don't want to do this too often.
```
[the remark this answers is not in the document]
noraabiakar: Well it's both, but I don't want to include that here, as it is mentioned below.

doc/concepts/recipe.rst
```diff
@@ -1,22 +1,77 @@
...
+     But when the simulation is first set up, it needs to now how cells interact with
+     one another in order to partition cells across nodes in an ideal manner. This is
```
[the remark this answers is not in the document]
noraabiakar: I'll use your suggestion

doc/concepts/recipe.rst
brenthuisman: The term 'description' is a bit overloaded. Also, there is no matching `set_description`, right? Is it (with the new `cable_cell`) everything passed in the constructor?

doc/concepts/recipe.rst
```diff
@@ -8,29 +8,33 @@ building phase to provide information about cells in the model, such as:
 
   * The **number of cells** in the model.
   * The **kind** of each cell.
-  * The **description** of each cell, e.g. with morphology, synapses, detectors, stimuli.
+  * The **description** of each cell, e.g. with morphology, dynamics, synapses, detectors,
+    stimuli etc.
```
brenthuisman, line 12: The etc. throws me off again. Are the following bullets included for example? I think not but I'm not sure.

noraabiakar: A bit more context about the description: the cell description is basically the cell itself, which includes everything I listed plus other things I didn't: the initial voltage, the temperature, the initial ion concentrations etc. I think the confusion is around the `description` term. This is directly related to the `get_cell_description()` method of the recipe, which returns a cell object (in most cases, a `cable_cell`). Take for example the [branch_cell](https://github.com/arbor-sim/arbor/blob/master/example/ring/branch_cell.hpp#L56) function from our repo. It returns a full "description" of a cell, which is actually a `cable_cell` object. This function is called from [recipe::get_cell_description()](https://github.com/arbor-sim/arbor/blob/master/example/ring/ring.cpp#L80). 

Is this still vague? Should the line be: 
`The cells themselves, referred to as the cell description of each cell, e.g. with morphology, dynamics, synapses, detectors, stimuli etc.`

As for the etc, I can remove it, but I don't see the source of the confusion. I just mean to convey that there are things that can be included in the description of the cell other than morphology, dynamics, synapses and detectors.

brenthuisman: If the "cell description", as we use the term in Arbor, is not described here in full, then it isn't anywhere as far as I'm able to establish (which may be hindered by the fact 'description' is used in many contexts in the docs). I think this is the right place to do so.

`get_cell_description()` is documented in basically the same way and so does not offer further clarification.

noraabiakar: I am adding a `Cell descriptions` section in `concepts/cells` right underneath `Cell kinds`, as I think they are related to some extent. And I'll link to it from this page. Do you mind if this is in a separate PR? What changes do you suggest for this PR?

brenthuisman: OK, that's fine too. A link would be nice! Then go with 'etc.' for this PR.

brenthuisman (approved):

Issue #1211, opened 2020-11-03
SWC files with empty lines not parsed correctly


noraabiakar:
`parse_swc` stops parsing SWC files after an empty line is encountered. 
For example, the following file generates an empty vector of `swc_record`:
```
# comment

1 1     0.0     0.0     0.0     6.30785 -1
2 1     6.30785 0.0     0.0     0.5      1
```
And only the first sample is parsed in the following file.
```
# comment
1 1     0.0     0.0     0.0     6.30785 -1

2 1     6.30785 0.0     0.0     0.5      1
```

No warning is generated in both of these cases. We should probably allow empty lines in the description.

halfflat:
Possibly wrongly, it was coded that way on purpose: I didn't think empty lines were permitted, and so they served as an end-of-SWC marker when reading from a stream.

If there are valid SWC files out there with blank lines, then this has to charge :)

noraabiakar:
We have a `ball_and_stick.swc` in `test/unit/swc` with an empty line, which is how I noticed this issue. But I couldn't find any `swc` files on modeldb with empty lines. 

We can keep it as is and fix our file. I usually add an empty line after comments so I'd like to allow empty lines, but I can be convinced to change my ways. 

I interpret this line `Subsequent non-empty lines each represent a single neuron sample point with seven data items.
` from [here](http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html) to mean they are allowed, but it's kind of vague.

halfflat:
I guess there are three options, for reading from a stream:
1. Status quo (and we fix our `ball_and_stick.swc` file).
2. We skip blank lines before the first record, but after that, a blank line (or other line that does not parse as a record) counts as end-of-data.
3. We just skip all blank lines.

Option 2 is easy enough to add, and catches the current problem.

An advantage to allowing terminating blank lines is in the parsing of SWC data in a stream that may contain other data; without the blank line for termination, the attempt to parse a line as a record may well consume that line, making it unavailable to further processing. I personally like the idea that we can contextually read an 'SWC file' embedded in a stream with potentially other information. It would allow, for example, multiple morphologies to be read from a file without fuss, e.g.:
```
# Morphology: foo
1 1 0 0 0 10 -1
…

# Morphology: bar
1 1 3 4 5 10 -1
…
```
I believe with option 3, we'd be unable to successfully capture the comment/metadata associated with the second SWC morphology.

[...]

PR #1218, opened 2020-11-05
Check trailing data in SWC 

noraabiakar:
* Remove blank line from `ball_and_stick.swc` 
* Mention that blank lines in SWC files indicate end of data
* Check for trailing data after parsing in the python interface

Addresses #1211

### DIFF
--- a/python/morphology.cpp
+++ b/python/morphology.cpp
@@ -142,7 +142,11 @@ void register_morphology(pybind11::module& m) {
                 throw pyarb_error(util::pprintf("can't open file '{}'", fname));
             }
             try {
-                return arborio::load_swc_arbor(arborio::parse_swc(fid));
+                auto data = arborio::load_swc_arbor(arborio::parse_swc(fid));
+                if (fid.rdbuf()->in_avail()) {
+                    throw pyarb_error(util::pprintf("Trailing data found at end of file '{}'", fname));
+                }
+                return data;
             }
             catch (arborio::swc_error& e) {
                 // Try to produce helpful error messages for SWC parsing errors.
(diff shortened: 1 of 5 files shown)


python/morphology.cpp
```diff
@@ -142,7 +142,11 @@ void register_morphology(pybind11::module& m) {
                 throw pyarb_error(util::pprintf("can't open file '{}'", fname));
             }
             try {
-                return arborio::load_swc_arbor(arborio::parse_swc(fid));
+                auto data = arborio::parse_swc(fid);
+                if (fid.rdbuf()->in_avail()) {
```
halfflat, line 146: Should trailing white space/blank lines be permitted?
If so, could have e.g.
```
auto data = arborio::parse_swc(fid);
if (!(fid >> std::ws).eof()) { throw …; }
```

Either way, the test and throw across the different load functions could be encapsulated into a single function to avoid duplication of the error message and logic.

halfflat (commented):
If we're blessing the use of a blank line as an SWC terminator, we should also take care of the stream error state in `parse_swc`; currently, once a record can't be read, the stream is in 'fail' state, and this is not cleared.  Should `swc1 = parse_swc(fs); fs >> std::ws; swc2 = parse_swc(fs);` 'just work'?

It's open if we should fail after reading a non-blank line that is not parsed successfully as an SWC record. 

Again, I think there are two options:
1. `parse_swc` checks if the line is empty before parsing a record, using `istream::peek()`; if empty, it stops further parsing and leave the stream in a 'good' state.
2. `parse_swc` keeps going until reading a record fails, and then just clears failbit if any data was read at all, say. I'm less happy with approach though, because it means some data in the stream might have been silently discarded.

What do you think?

python/morphology.cpp
noraabiakar: Okay here's what I think we can do: 
1. I'll encapsulate the checking and error message as you suggested.
2. `parse_swc` checks if the line is empty before parsing a record, using `istream::peek()`; if empty, it stops further parsing and leave the stream in a 'good' state. With this, `swc1 = parse_swc(fs); fs >> std::ws; swc2 = parse_swc(fs);` should work providing there are no other errors. 
3. If we read a non-blank line that is not parsed successfully as an SWC record, the fail bit is set on the stream: an error is raised in the python wrapper, we can't continue reading from the stream

halfflat (approved):
Looks good! Thanks for making the changes.

PR #1220, opened 2020-11-06
Fix ODR violation from geneated mechanisms.

### LINKED ISSUE #1219: Incorrect parameter names in Allen catalogue mechanisms from Python
Consider:
```
import arbor as A

for p in A.allen_catalogue()["Ca_HVA"].parameters:
    print(p)
```

Given the Allen catalogue mechanism with name `Ca_HVA` defined by `mechanisms/allen/Ca_HVA.mod`, the expected output is:
```
gbar
```

Actual output is:
```
gCa_HVAbar
```

The problem is an ODR violation: there is a mechanism by the name `Ca_HVA`  in both the included Allen and BBP catalogues; the generated code is not differentiated by namespace, and so we have two different definitions for the same info function `mechanism_Ca_HVA_info()`.

halfflat:
* Put generated mechanisms for each built-in catalogue in their own namespace.

Fixes #1219.


noraabiakar (approved):
lgtm!

PR #1222, opened 2020-11-09
Fix some static analysis warnings

noraabiakar:
Addresses some of the warnings from static analysis tools in #1084 

* fix some instances where exceptions were constructed, but not thrown
* fix use-after-move in a unit test.


bcumming (approved):
Looks good to me.

Issue #1160, opened 2020-09-24
Probe diversity in Python lib.


halfflat:
The Python interface currently supports two hard-coded probe types, namely cable cell membrane voltage and total ionic current density.

Task is to make the Python recipe and single-cell recipe support a variety of probe types (and corresponding variety in probe address arguments), ignoring for now the problem of representing probe metadata provided by the C++ API.

The most straightforward approach is probably to adopt the typey style of specification offered by the C++ API, rather than string dispatch: `'voltage'` or `'membrane_voltage`' is unambiguous now, but would become problematic when we support e.g. specialized single-compartment cells in addition to cable cells.

PR #1225, opened 2020-11-09
Expose diverse probes in Python API.

halfflat:
Major changes in Python library and documentation:

* Add global state (immutable after initialization) for the Python module that
  manages the mapping between probe metadata types and the method by which the
  corresponding sample data can be collected and presented as Python objects.
  This is necessary for decoupling the implementation of the simulation wrapper
  from that of the various cable cell probe types.

* Wrap each of the C++ cable cell probe types with a Python function that
  returns a `probe_info` object.

* Add code for converting and accumulating C++ probe metadata and sample
  data, registered in the module global state against the metadata type info.

* Make the `arbor.simulator` object responsible for recording all spike and
  sample data, with corresponding new methods.

* Use NumPy arrays and structured datatypes for returning spike and
  sample data.

* Place Python schedule proxies under an abstract interface so that consumers
  of the proxies can be made generic over schedule types.

* Add unit tests for the `arbor.simulator` class and its new functionality,
  including distributed spike collection.

* Rework all Python API documentation concerning probes, sampling, and spike
  recording; add information on newly exposed cable cell probe addresses.

* Add new python example `single_cell_recipe.py` which is a generic recipe
  version of `single_cell_model.py`.

* Adjust other code in the wrapper and examples to accomodate these changes.

Minor changes in Python library:

* Add equality tests for `arbor.location` and `arbor.cable` Python classes.

* Use the `py::` namespace shorthand more often in the Python wrapping sources.

* Add an implicit conversion for a pair of numbers to `arbor.cell_member`, so
  that e.g. probe ids can be specified simply as `(gid, index)`.

* Add an implicit conversion from a tuple to `arbor.mpoint` so that a
  morphological point can be specified simply as `(x, y, z, radius)`.

* Modify the interface to `arbor.regular_schedule` so that the overloads
  correspond more closely to the C++ API. In particular,
  `arbor.regular_schedule(dt)` now describes a schedule with times 0, dt, 2*dt,
  etc.

Minor changes in C++ library:

* Change test in FVM lowered cell implementation and exception message for
  `bad_source_description`: assert number of detectors is exactly the number
  of sources promised by the recipe.

* Add comment describing `probe_metadata` object in `sampling.hpp` which
  asserts that the type and value of probe-specific metadata completely
  determines the correct interpretation of sampled data provided to a callback.

Implements #1160.


thorstenhater:
Hi,

a short comment on the motivation would be helpful for me.

T

python/test/unit/test_simulator.py
```diff
@@ -0,0 +1,258 @@
...
+all tests for the simulator wrapper
+"""
+
+# Test recipe cc2 comprises two cable cells and some probes.
+
+class cc2_recipe(A.recipe):
+    def __init__(self):
+        A.recipe.__init__(self)
+        st = A.segment_tree()
+        foo = A.mpoint((10, 0, 0, 10))
```
noraabiakar, line 28: `foo` not used anywhere?

noraabiakar (changes requested):
This is great!
I have a few questions and suggestions. My main problem is with the probe/sampling docs. I don't think they're very clear. They probably should be extended, if not in this PR, then soon after.

doc/python/cable_cell.rst
```diff
@@ -93,4 +93,147 @@ Cable cells
 
     properties of an ionic species.
 
+.. _pycableprobes:
+
+Cable cell probes
+-----------------
+
+Cable cell probe addresses are definied analagously to their counterparts in
```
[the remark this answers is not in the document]
halfflat: Nice catch! (And on the other ytpos, too!)

doc/python/cable_cell.rst
```diff
@@ -93,4 +93,147 @@ Cable cells
...
+one, or more probes, one per site.
+
+Each of the functions described below generates an opaque :class:`probe`
+object for use in the recipe :py:func:`get_probes` method.
+
+Membrane voltage
+   .. py:function:: cable_probe_membrane_voltage(locations)
+
+   Cell membrane potential (mV) at the sites specified by the location
+   expression string ``locations``. This value is spatially interpolated.
```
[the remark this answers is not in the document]
halfflat: It will work with labels; the expressions are evaluated in the back end with the corresponding cell description 'in hand', whose `mprovider` object is used to resolve locset expressions.

doc/python/cable_cell.rst
```diff
@@ -93,4 +93,147 @@ Cable cells
...
+
+   Metadata: the explicit :class:`location` of the sample site.
+
+   .. py:function:: cable_probe_membrane_voltage_cell()
+
+   Cell membrane potential (mV) associated with each cable in each CV of
```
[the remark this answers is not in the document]
halfflat: 'cable' here is meant to indicate a contiguous sub-component of a branch, that is, something representable by an `mcable` object. Would 'unbranched component' make more sense?

The thing I am trying to explain is that if a CV is just a subset of a branch, then there will be one entry corresponding to that subset; if it intersects with multiple branches, there will be an entry for each of those intersections.

doc/python/cable_cell.rst
```diff
@@ -93,4 +93,147 @@ Cable cells
...
+   the cell discretization.
+
+   Metadata: the list of corresponding :class:`cable` objects.
```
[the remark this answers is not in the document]
halfflat: On the second point first, there more detailed explanation about how to capture sampled data is in the `simulation` documentation. If that bit is clear, I can add a link from here to there. There's also information there about what probe metadata is meant to be.

Could you take a look at that bit and see if it helps sufficiently?

python/cable_probes.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    m.def("cable_probe_membrane_voltage", &cable_probe_membrane_voltage,
+        "Probe specification for cable cell membrane voltage interpolated at points in a location set.",
+        "where"_a);
+
+    m.def("cable_probe_membrane_voltage_cell", &cable_probe_membrane_voltage_cell,
+        "Probe specification for cable cell membrane voltage associated with each cable in each CV.");
+
+    m.def("cable_probe_axial_current", &cable_probe_axial_current,
+        "Probe specification for cable cell axial current at points in a location set.");
```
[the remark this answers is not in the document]
halfflat: This is what I get for not putting in a unit test for every probe type. I will do that.

python/identifiers.cpp
```diff
@@ -36,7 +44,9 @@ void register_identifiers(pybind11::module& m) {
         .def("__str__", [](arb::cell_member_type m) {return pprintf("<arbor.cell_member: gid {}, index {}>", m.gid, m.index);})
         .def("__repr__",[](arb::cell_member_type m) {return pprintf("<arbor.cell_member: gid {}, index {}>", m.gid, m.index);});
 
-    pybind11::enum_<arb::cell_kind>(m, "cell_kind",
+    py::implicitly_convertible<py::tuple, arb::cell_member_type>();
```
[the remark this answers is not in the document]
halfflat: It's so we can write `(0, 1)` in arguments to functions that take a `cell_member`, rather than having to write `cell_member((0, 1))`. It cuts out a fair bit of verbosity.

python/test/unit/test_simulator.py
halfflat: Arg, stale testing code! Will remove.

halfflat:
> a short comment on the motivation would be helpful for me.

I totally forgot to link to the corresponding issue (#1160). I've added that to the description above.

doc/python/simulation.rst
```diff
@@ -183,96 +242,120 @@ Definitions
...
-    .. function:: sampler()
+       Recording is set up with the method :py:func:`simulation.sample`
+       as described above. It returns a handle that is used to retrieve the recorded data after
+       simulation.
 
-        Initialize the sample recorder.
+    3. Retrieve recorded data.
 
-    .. function:: samples(probe_id)
+       The method :py:func:`simulation.samples` method takes a handle and returns the recorded data as a list,
```
halfflat, line 296: One 'method' here is probably enough.

python/simulation.cpp
```diff
@@ -35,18 +187,35 @@ void register_simulation(pybind11::module& m) {
...
             "Set the binning policy for event delivery, and the binning time interval if applicable [ms].",
             "policy"_a, "bin_interval"_a)
-        .def("__str__",  [](const arb::simulation&){ return "<arbor.simulation>"; })
-        .def("__repr__", [](const arb::simulation&){ return "<arbor.simulation>"; });
+        .def("record", &py_simulation::record,
+            "Disable or enable local or global spike recording.")
+        .def("spikes", &py_simulation::spikes,
+            "Retrieve recorded spikes as numpy array.")
+        .def("probe_metadata", &py_simulation::get_probe_metadata,
+            "Retrieve metadata associated with given probe id.")
```
halfflat, line 205: For consistency, add `probe_id` as a named argument.

python/test/unit/test_simulator.py
```diff
@@ -0,0 +1,258 @@
...
+        self.assertEqual(2, len(r2))
+        s20, meta20 = r2[0]
+        s21, meta21 = r2[1]
+
+        # Probe id (0, 2) has probes over the two locations that correspond to probes (0, 0) and (0, 1).
+
+        # (order is not guaranteed to line up though)
+        if meta20==meta0:
+            self.assertEqual(meta1, meta21)
+            self.assertTrue((s1[:,1]==s21[:,1]).all())
```
halfflat, line 198: Should also check that `s0` and `s20` are equal here.

doc/python/cable_cell.rst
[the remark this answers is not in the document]
halfflat: I added a comment in the docs which I hope clarifies this.

doc/python/cable_cell.rst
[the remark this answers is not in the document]
noraabiakar: I think 'unbranched component' would be generate less confusion. But, I think the current description is probably better. 
Maybe we can link `cable` to https://arbor.readthedocs.io/en/latest/concepts/labels.html?highlight=cable#cables? And then, when we have a section describing what a CV is, we can link it too. 
I'll leave it up to you :)

python/identifiers.cpp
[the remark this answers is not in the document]
noraabiakar: Okay, thanks :)

doc/python/cable_cell.rst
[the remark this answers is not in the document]
noraabiakar: I thought the python 'simulation' docs were actually really helpful, and my confusion was more related to the cpp docs and the concept docs. It would be nice to have something that describes the full probing/sampling procedure and links to the correct locations in the docs. 
I will think a bit more about this and discuss offline. For the scope of this PR, the newly added changes are really good. So I think we can resolve this issue.

arbor/CMakeLists.txt
```diff
@@ -137,7 +137,7 @@ if(ARB_WITH_CUDA_CLANG)
     target_link_libraries(arbor PUBLIC  ${CUDA_LIBRARIES})
 endif()
 
-set_target_properties(arbor PROPERTIES CUDA_RESOLVE_DEVICE_SYMBOLS ON)
+set_target_properties(arbor PROPERTIES CUDA_RESOLVE_DEVICE_SYMBOLS ON POSITION_INDEPENDENT_CODE ON)
```
noraabiakar, line 140: There's already this line in the main `CMakeLists.txt`: https://github.com/arbor-sim/arbor/blob/master/CMakeLists.txt#L244
```
set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
So I think this may not be needed?

halfflat: Good point!
Any ideas about the reported relocation errors we're seeing in the Travis logs?

noraabiakar: I was able to successfully build pyarb on my system (ubuntu 20.04, gcc 9.3) with the same cmake flags that cause the error on travis. 
I can try again with gcc 8.4 as soon as it's done building. 
Could it be the order of the flag?

bcumming:
bors try

python/example/single_cell_recipe.py
```diff
@@ -0,0 +1,82 @@
...
+        print('{:3.3f}'.format(t))
+else:
+    print('no spikes')
+
+# (8) Plot the recorded voltages over time.
+
+print("Plotting results ...")
+seaborn.set_theme() # Apply some styling to the plot
+df = pandas.DataFrame({'t/ms': data[:, 0], 'U/mV': data[:, 1]})
+seaborn.relplot(data=df, kind="line", x="t/ms", y="U/mV").savefig('single_cell_recipe_result.svg')
```
bcumming, line 78: Don't forget the `ci=None` argument to `seaborn.relplot`:
https://github.com/arbor-sim/arbor/blob/master/python/example/single_cell_multi_branch.py#L112

python/identifiers.cpp
```diff
@@ -10,23 +10,31 @@
...
                 return arb::cell_member_type{gid, idx};
             }),
             "gid"_a, "index"_a,
-            "Construct a cell member with arguments:\n"
+            "Construct a cell member identifier with arguments:\n"
+            "  gid:     The global identifier of the cell.\n"
+            "  index:   The cell-local index of the item.\n")
+        .def(py::init([](py::tuple t) {
+                if (py::len(t)!=2) throw std::runtime_error("tuple length != 4");
+                return arb::cell_member_type{t[0].cast<arb::cell_gid_type>(), t[1].cast<arb::cell_lid_type>()};
```
bcumming, line 35: What happens with these casts if the user passes a tuple with non-convertible types, e.g.:
```python
m = arbor.cell_member(("hello", 23.2))
```
You can force pybind11 to do the casting and checking for you by making the arguments to be `std::tuple<arb::cell_gid_type, arb::cell_lid_type>` instead of `pybind11::typle`.

python/identifiers.cpp
```diff
@@ -47,14 +57,14 @@ void register_identifiers(pybind11::module& m) {
...
-    pybind11::enum_<arb::backend_kind>(m, "backend",
+    py::enum_<arb::backend_kind>(m, "backend",
         "Enumeration used to indicate which hardware backend to execute a cell group on.")
         .value("gpu", arb::backend_kind::gpu,
             "Use GPU backend.")
         .value("multicore", arb::backend_kind::multicore,
             "Use multicore backend.");
 
-    pybind11::enum_<arb::binning_kind>(m, "binning",
+    py::enum_<arb::binning_kind>(m, "binning",
```
bcumming, line 67: I deliberately chose `pybind11` as the namespace. By doing this we are now inconsistent (you are probably going to show me one spot where I used `py` instead of `pybind11` now.

python/morphology.cpp
```diff
@@ -188,6 +199,8 @@ void register_morphology(pybind11::module& m) {
                 "The relative position of the proximal end of the cable on its branch ∈ [0,1].")
         .def_readonly("dist", &arb::mcable::dist_pos,
                 "The relative position of the distal end of the cable on its branch ∈ [0,1].")
+        .def("__eq__",
```
bcumming, line 202: If you are overloading an operator that is defined directly in C++ (as is the case here):
```
.def(pybind11::self == pybind11::self)
```
See: https://github.com/pybind/pybind11/blob/master/tests/test_operator_overloading.cpp
You'll need to include `pybind11/operators.hpp`

python/pyarb.hpp
```diff
@@ -0,0 +1,84 @@
+#pragma once
```
bcumming, line 1: How about putting this content in `python/cable_probes.hpp`, because this is really specific to probes/samplers.

python/simulation.cpp
```diff
@@ -6,24 +8,174 @@
...
+#include "pyarb.hpp"
 #include "recipe.hpp"
+#include "schedule.hpp"
+
+namespace py = pybind11;
 
 namespace pyarb {
 
-void register_simulation(pybind11::module& m) {
+enum class py_spike_recording {
```
bcumming, line 19: A comment documenting what this is for would be helpful.

python/simulation.cpp
```diff
@@ -6,24 +8,174 @@
...
+    off, local, all
+};
+
+// Wraps an arb::simulation object and in addition manages a set of
+// sampler callbacks for retrieving probe data.
+
+class py_simulation {
```
bcumming, line 26: why use the `py_` prefix here, when it is declared in the `pyarb` namespace?
The `_shim` postfix has been used for such wrappers to date.

python/test/unit_distributed/test_simulator.py
```diff
@@ -0,0 +1,134 @@
...
+
+def suite():
+    # specify class and test functions in tuple (here: all tests starting with 'test' from class Contexts
+    suite = unittest.makeSuite(Simulator, ('test'))
+    return suite
+
+def run():
+    v = options.parse_arguments().verbosity
+
+    print('hello?')
```
bcumming, line 112: Should this, and the `'wat'` message below be removed?

python/pyarb.cpp
```diff
@@ -17,21 +23,27 @@ void register_mechanisms(pybind11::module& m);
...
 void register_mpi(pybind11::module& m);
 #endif
-}
+
+} // namespace pyarb
 
 PYBIND11_MODULE(_arbor, m) {
+    pyarb::pyarb_global_ptr global_ptr(new pyarb::pyarb_global);
+
+    PYBIND11_NUMPY_DTYPE(arb::cell_member_type, gid, index);
```
bcumming, line 40: Can we have a quick comment explaining what is going on here?

bcumming (changes requested):
Nice work. There are a few requested changes.

doc/cpp/cable_cell.rst
```diff
@@ -377,6 +377,8 @@ Overriding properties locally
    the morphology.
 
 
+.. _cable-cell-probes:
```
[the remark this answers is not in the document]
halfflat: Would this be the content that's been put under `simulation`? I'm not averse to this, but maybe we could make the concept docs a different task, as it's not exactly in scope for this PR, and would delay merging.

python/identifiers.cpp
halfflat: Yeah, there were multiple cases of using `py::` instead of `pybind11::`, so I went with the shorter for consistency and brevity.

python/identifiers.cpp
halfflat: If we do that with the current version of the PR, we get:
```
RuntimeError: Unable to cast Python instance to C++ type (compile in debug mode for details)
```

Is that sufficient?

python/morphology.cpp
halfflat: The code that's there does work though — is there a reason to use the `operators.hpp` way instead, other than being a bit shorter?

python/pyarb.cpp
halfflat: Sure

python/pyarb.hpp
halfflat: It's probes and samplers, but not cable probes specifically.

Part of the motivation for this design is to keep the simulator code uncoupled from the cable probe definitions and wrappers. This way, the cable probe wrapper code is responsible for registering its metadata handlers, and the simulator code is responsible for using those handlers to process samples.

python/simulation.cpp
halfflat: Usage is inconsistent. We have `py_recipe`  and `py_recipe_shim`, for example.

python/test/unit_distributed/test_simulator.py
halfflat: They certainly should!

python/simulation.cpp
halfflat: I did think the code in `py_simulation.record()` was pretty self-explanatory.

python/example/single_cell_recipe.py
halfflat: Will add.

halfflat:
Addressed some of the PR comments; rebased off of current master.

python/simulation.cpp
halfflat: Renamed in 14e1d9e

python/simulation.cpp
bcumming: I meant, a quick comment that explains what the `py_spike_recording` enum is to be used for?

python/morphology.cpp
bcumming: It's just shorter. I pointed it out, because I don't think that it is in the docs, so you might not have been aware of it.
It is a bit clearer, and is consistent with changes I recently made in another PR that implemented operator overloading for `cv_policy`.

doc/cpp/cable_cell.rst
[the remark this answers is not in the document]
bcumming: I don't know what you are referring to by "under simulation".

The issue here is that documentation has been created under Python API, but nothing added to concepts. The cpp and python docs might be simplified if a lot of the common part of that documentation is moved to concepts. So I think that it is actually in the scope of this PR.

If it is in a separate PR, then it should be done ASAP, otherwise it will be forgotten.

python/identifiers.cpp
bcumming: Sufficient. I was hoping that it would just work! I always have to test.

bcumming (approved):

doc/cpp/cable_cell.rst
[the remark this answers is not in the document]
halfflat: I mean, I added some text in the documentation of the `simulation` class that explained the probes and sampling. Is this suitable material for moving into Concepts?

noraabiakar (approved):

PR #1227, opened 2020-11-11
Docs: Cell kind and description 

noraabiakar:
* Modify `cell kind` concept docs.
* Add `cell description` concept docs.
* Reorder the docs so that the recipe, cell kind and description, interconnectivity, hardware management, domain decomposition, and simulation are consecutive. Then start the more detailed cable cell docs.

### DIFF
--- a/doc/concepts/cell.rst
+++ b/doc/concepts/cell.rst
@@ -63,8 +63,8 @@ Cell kind
 .. _model_cable_cell:
 1. **Cable Cells**
 
-   Cable cells are morphologically-detailed cells represented as branching linear 1D segments. They can be coupled
-   to other cell types via the following mechanisms:
+   Cable cells are morphologically-detailed cells. They can be coupled to other cells via the following
+   mechanisms:
 
    1. Spike exchange over a :ref:`connection <modelconnections>` with fixed latency.
       Cable cells can *receive* spikes from any kind of cell, and can be a *source* of spikes
@@ -116,12 +116,12 @@ morphologies and user-defined mechanisms. See the cable cell's :ref:`dedicated p
 Cell description
 ----------------
 
-The `description` of a cell details everything needed to build a cell. This degree of detail needed
-differs according to the cell kind.
+The `description` of a cell is referred to in the :ref:`recipe <modelrecipe>`, and elsewhere in the docs.
+It details everything needed to build a cell. The degree of detail differs according to the cell kind.
 
 1. **Cable Cells**
 
-   The description of a cable cell includes the following:
+   The description of a cable cell can include all the following:
 
      * :ref:`Morphology <co_morphology>`: the shape of the cell.
      * Discretisation: how to split the morphology into discrete components for the simulation.
@@ -131,15 +131,15 @@ differs according to the cell kind.
      * Initial temperature.
      * Initial ion internal and external concentrations.
      * Initial ion reversal potential.
-     * :ref:`Density mechanisms <mechanisms-density>`: commonly used to describe ion-channel dynamics accross
+     * :ref:`Density mechanisms <mechanisms-density>`: commonly used to describe ion-channel dynamics across
        :ref:`regions <labels-region>` of the cell.
      * :ref:`Ion reversal potential mechanisms <mechanisms-revpot>`: used to control the reversal potentials of
-       ions accross :ref:`regions <labels-region>` of the cell.
+       ions across :ref:`regions <labels-region>` of the cell.
      * :ref:`Point mechanisms <mechanisms-point>`: commonly used to describe synapses on specific
        :ref:`locations <labels-locset>` of the cell; typically act as targets in the recipe.
      * Stimuli: such as current clamps; placed on specific :ref:`locations <labels-locset>` on the cell.
      * Threshold detectors: used to generate spiking events on specific :ref:`locations <labels-locset>` on the
-       cell, when the voltage increases above a certain threshold; typically act as sources in the recipe.
+       cell when the voltage increases above a certain threshold; typically act as sources in the recipe.
      * Gap junction sites: used to electrically couple the cell to another gap junction site on another cell;
        placed on specific :ref:`locations <labels-locset>` on the cell.
 
@@ -149,7 +149,8 @@ differs according to the cell kind.
 
 2. **LIF Cells**
 
-   The description of a LIF cell is used to control the leaky integrate-and-fire dynamics.
+   The description of a LIF cell is used to control the leaky integrate-and-fire dynamics:
+
       * Resting potential.
       * Reset potential.
       * Initial value of membrane potential.
@@ -158,7 +159,7 @@ differs according to the cell kind.
       * Firing threshold.
       * Refractory period.
 
-   The morphology of a lif cell is automatically modeled as a single compartment; each cell has a single built-in
+   The morphology of a LIF cell is automatically modeled as a single compartment; each cell has a single built-in
    source and target which do not need to be explicitly added in the cell description.
 
 3. **Spiking cells**
--- a/doc/concepts/cell.rst
+++ b/doc/concepts/cell.rst
@@ -20,11 +20,11 @@ Cells interact with each other via spike exchange and gap junctions.
                                                                 For example, the 7th synapse on cell 42.
     ========================  ================================  ===========================================================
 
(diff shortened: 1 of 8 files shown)


bcumming:
One observation: you have changed the naming of references, e.g.
```
.. _model_cable_cell:
```
becomes
```
.. _modelcablecell:
```

I started using the following pattern:
```
.. _cablecell-paint:
```
where `cablecell` is the name of the text file, followed by a hyphen.

We should settle on a consistent way to name references.

noraabiakar:
@bcumming I agree, the reference naming varies a lot across the docs. This was my attempt at making them match with our existing references such as `modelrecipe`, `modelconnections` and `modelgapjunctions`.
I don't mind using `filename-subtitle` convention. I was planning on having a separate PR to fix all references eventually.

doc/concepts/index.rst
```diff
@@ -21,25 +21,41 @@ The python front-end further abstracts away some of these steps for single cell
...
    cell
-   cable_cell
-   morphology
-   labels
-   mechanisms
    interconnectivity
    hardware
    domdec
    simulation
+
```
brenthuisman, line 53: Have you tested this breakup of the TOC? Won't that create two separate concept sections, 'From recipe to simulation:' and 'The cable cell:'?

doc/concepts/simulation.rst
```diff
@@ -2,28 +2,33 @@
...
-The recipe describes the model, the domain decomposition describes how the cells in the model are assigned to hardware resources and the context is used to execute the simulation.
+Optionally, experimental inputs that can change between model runs, such as external spike trains, can be injected.
 
 Simulation execution and interaction
 ------------------------------------
 
 Simulations provide an interface for executing and interacting with the model:
 
-    * The simulation is executed/ *run* by advancing the model state from the current simulation time to another with maximum time step size.
+    * The simulation is executed/ *run* by advancing the model state from the current simulation time to another
```
brenthuisman, line 26: The `executed/ *run*` bit does not make much sense to me, is it correct grammatically?

doc/concepts/simulation.rst
```diff
@@ -2,28 +2,33 @@
...
-    * A context used to execute the simulation.
+    * A :ref:`recipe <modelrecipe>` that describes the cells and connections in the model.
+    * A :ref:`domain decomposition <modeldomdec>` that describes the distribution of the
+      model over the local and distributed :ref:`hardware resources <modelhardware>`.
+    * An :ref:`execution context <modelcontext>` used to execute the simulation.
 
-The workflow to build a simulation is to first generate a domain decomposition that describes the distribution of the model over the local and distributed hardware resources (see :ref:`modeldomdec`), then build the simulation from the recipe, the domain decomposition and the execution context. Optionally experimental inputs  that can change between model runs, such as external spike trains, can be injected.
-
-The recipe describes the model, the domain decomposition describes how the cells in the model are assigned to hardware resources and the context is used to execute the simulation.
+Optionally, experimental inputs that can change between model runs, such as external spike trains, can be injected.
```
brenthuisman, line 19: I can't find a definition for 'injecting', should we link to it? Or to how one might do this?

doc/concepts/cell.rst
```diff
@@ -20,90 +20,148 @@ Cells interact with each other via spike exchange and gap junctions.
...
+    **Benchmark cell**        Proxy cell used for benchmarking (developer use only).
     ========================  ===========================================================
 
-.. _model_cable_cell:
+.. _modelcablecell:
 1. **Cable Cells**
 
-   Cable cells are morphologically-detailed cells represented as branching linear 1D segments. They can be coupled
-   to other cell types via the following mechanisms:
+   Cable cells are morphologically-detailed cells. They can be coupled to other cells via the following
```
brenthuisman, line 63: Is mentioning somewhere 'branched linear 1D segments' not a good idea?

brenthuisman (changes requested):

doc/concepts/cell.rst
noraabiakar: It's mentioned in the cell description below. I thought it fit in better there. But we can mention it here too, if you think it's a good idea.

doc/concepts/index.rst
noraabiakar: That'll create 2 labeled sections on the main `Arbor concepts` page, but wont change anything on the toolbar on the lhs of the screen.

doc/concepts/simulation.rst
noraabiakar: I think this is not too important of a concept to be explained in the concepts page. It is part of the simulation [cpp API](https://arbor.readthedocs.io/en/latest/cpp/simulation.html#_CPPv4N3arb10simulation13inject_eventsERK10pse_vector), but it doesn't look like it's exposed in python yet.

noraabiakar: If you think it's too ambiguous I can modify this or mention that it's only exposed in the C++ API (with a link)

doc/concepts/simulation.rst
noraabiakar: So I think the intention is to signify that the execution of a simulation is referred to as running the simulation in Arbor. The past participle of run is still run, so it's grammatically correct.

doc/concepts/index.rst
brenthuisman: The index on the lhs is created from the TOCs in `doc/*/index.rst`. See the TOC in `doc/index.rst`. I think there may be a hide directive, but you should test this because it gets a bit finicky.

doc/concepts/simulation.rst
brenthuisman: Does the space between `executed/ *run*` disappear when rendered?

doc/concepts/simulation.rst
brenthuisman: It's Arbor-specific jargon, so some idea of what it means or where to look for details would be nice. It's covered in the C++ API page, but an problem in the Python API page is revealed: a binning class is defined and the binning policy method of `simulation` is annotated, but any mention of event injection is missing. The Python reference ( :smiley: ) reveals that quite a bit is exposed though! `arbor.event_generator` exists.

The C++ API page does not _explain_ events and injection either by the way. So maybe this is the right place?

doc/concepts/cell.rst
brenthuisman: I see it now, OK!

doc/concepts/index.rst
noraabiakar: I have tested it. And the index on the lhs is still the same.

doc/concepts/simulation.rst
noraabiakar: No, I'll remove the space.

doc/concepts/simulation.rst
noraabiakar: I'm just going to remove the line. It doesn't really add much to the description here.

noraabiakar: We could go into detail about this specific concept here, but it'll only add confusion as it is not an essential feature in arbor, that's probably also why it hasn't been exposed in python yet!

brenthuisman: So it is already exposed, at least in part (I don't know how much of a part).

I would like to keep some trace of this, because if it is implemented it should be documented and it should be here, eventually. What's a good way of keeping track of things like this? We've collected some todos in the Slack, I've collected some in an Issue before, Sphinx has the `..todo::` directive, we could file separate Issues for each thing we identify. They're all options as far as I'm concerned, but maybe we should choose one way of tracking TODOs. What has your preference?

noraabiakar: I don't think spike injection needs to appear in the concept docs at all. Not every part of the API is a concept after all. Spike injection is documented in the C++ API and when it is exposed to the python API it will appear there.
As for keeping track of todos, I would rather not have them in the docs unless they are important concepts/parts of the API that have been implemented but not yet documented. We can have a separate issue for each thing we identify, or one issue with a group of small fixes.

brenthuisman: OK, not here. Keeping track: I made [this](https://mcnest.slack.com/archives/C019BVBKS5V/p1604484691003200) post in the slack, is that enough or shall I make an Issue out of it?

brenthuisman (approved):

Issue #868, opened 2019-09-17
Document make targets better


heplesser:
As far as I can see, different `make` targets are not well documented at present. The `tests` target is only implicitly mentioned here

https://arbor.readthedocs.io/en/latest/install.html#building-and-installing-arbor

while I have not found the `examples` target mentioned anywhere in the documentation. I wonder if there are further targets, which are all not built by `make all`.

PR #1228, opened 2020-11-12
Add readme to examples.

thorstenhater:
Add a readme for the examples, aimed at beginners, who might want to learn Arbor's C++ interface by looking at the examples.

### DIFF
--- a/example/readme.md
+++ b/example/readme.md
@@ -1,0 +1,18 @@
+# C++ Examples
+
+A series of mini-apps to demonstrate specific features.
+Please refer to the respective sub-directories for more
+information.
+
+All examples can be built by setting up the standard
+CMake build process, e.g.
+
+    # go to arbor's source directory   
+    cd arbor
+    # set up a build dir
+    mkdir build
+    cd build
+    # configure (example options for an MPI-enabled build)
+    cmake .. -DARB_VECTORIZE=ON -DARB_WITH_MPI=ON -DCMAKE_CXX_COMPILER=`which g++`
+    make examples
+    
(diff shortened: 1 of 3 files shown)


noraabiakar:
I think it would be good to also mention how to build the c++ examples in the docs, maybe before [this](https://arbor.readthedocs.io/en/latest/install/build_install.html#quick-start-examples) section? 
Also the dryrun readme is wrong, can we also fix that at the same time?

thorstenhater:
Good points, done.

brenthuisman:
Does this PR fix #868?

thorstenhater:
By accident, yes ;)

example/readme.md
```diff
@@ -0,0 +1,18 @@
...
+
+All examples can be built by setting up the standard
+CMake build process, e.g.
+
+    # go to arbor's source directory   
+    cd arbor
+    # set up a build dir
+    mkdir build
+    cd build
+    # configure (example options for an MPI-enabled build)
```
noraabiakar, line 15: Can we mention that it's also vectorization-enabled here? And below, instructions how to run, with a note saying to check the readme of each example?

noraabiakar (commented):
Looks good! thanks for also fixing the dryrun readme. I have just one small suggestion.

example/readme.md
thorstenhater: Certainly I can add SIMD here. But the individual READMEs are referenced above this example.

thorstenhater: That said, I made it a bit more clear ... hopefully.

noraabiakar (approved):

PR #1231, opened 2020-11-12
Move inline overrides of public methods out of header files

noraabiakar:
Building `pyarb` fails at the linking stage on certain platforms. This appears to be related to the GCC visibility of methods overridden in header files, and is fixed by moving the implementation to the corresponding cpp files. 

Addresses the Travis CI failure of #1225

### DIFF
--- a/arbor/include/arbor/communication/mpi_error.hpp
+++ b/arbor/include/arbor/communication/mpi_error.hpp
@@ -82,18 +82,9 @@ class mpi_error_category_impl;
 const mpi_error_category_impl& mpi_error_category();
 
 class mpi_error_category_impl: public std::error_category {
-    const char* name() const noexcept override { return "MPI"; }
-    std::string message(int ev) const override {
-        char err[MPI_MAX_ERROR_STRING];
-        int r;
-        MPI_Error_string(ev, err, &r);
-        return err;
-    }
-    std::error_condition default_error_condition(int ev) const noexcept override {
-        int eclass;
-        MPI_Error_class(ev, &eclass);
-        return std::error_condition(eclass, mpi_error_category());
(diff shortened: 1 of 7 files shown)


bcumming (approved):
Looks good, and good catch!

I will wait for Travis to pass, then merge ;)

halfflat:
Just for keeping a record: if we run into this (or a related visibility-triggered problem) in the future, we can instruct the linker to regard `libarbor.a` symbols as hidden by adding `-Wl,--exclude-libs,ALL` to the `pyarb` target link options.

arbor/include/arbor/communication/mpi_error.hpp
```diff
@@ -82,18 +82,9 @@ class mpi_error_category_impl;
...
-        return err;
-    }
-    std::error_condition default_error_condition(int ev) const noexcept override {
-        int eclass;
-        MPI_Error_class(ev, &eclass);
-        return std::error_condition(eclass, mpi_error_category());
-    }
+    const char* name() const noexcept override;
+    std::string message(int ev) const override;
+    std::error_condition default_error_condition(int ) const noexcept override;
```
halfflat, line 87: Trivial formatting thing, but there is an errant space after `int` here, and for consistency with `message` above, either both should have the parameter `ev` named or neither.

halfflat (approved):
Approved! But a minor non-semantic formatting request in comments.

arbor/include/arbor/communication/mpi_error.hpp
noraabiakar: I'll remove the extra space and the extra `ev` argument. Thanks for pointing it out!

PR #1235, opened 2020-11-13
Decor

bcumming:
Make the `cable_cell` interface read only by passing the decorations
to its constructor in a new `decor` type.

C++ library
* Remove the paint/place/set_default interface from `cable_cell`
* Create a `decor` type that:
  * stores lists of paintings and placings
  * stores a set of cable cell parameters
  * uses the `paint`, `place`, `set_default` interface that was deprecated
    from `cable_cell`.
* Create `paintable`, `placeable` and `defaultable` variants that are sum types
  over the respective types that can be painted, placed and defaulted.
* Remove the overloaded `cable_cell::paint`, `cable_cell::place`,
  `cable_cell::set_default` methods to single methods that consume the sum types.

Unit Tests
* Many small changes because many many tests use cable cell API.
* There were no `cable_cell` unit tests! Not such a big deal, since cable_cell
  is tested implicitly in so many other tests
  * but I added cable_cell tests: not much at the moment but will quickly fill
    up as we work on reading, writing, etc.                                                            

Python
* Wrap the new decor type.
* Update cable cell type to reflect read only API.
* Wrap the cv_policy types and implement the same operator overloading.


python/cells.cpp
```diff
@@ -321,6 +360,36 @@ void register_cells(pybind11::module& m) {
...
+    // arb::cv_policy wrappers
+
+    pybind11::class_<arb::cv_policy> cv_policy(m, "cv_policy",
+            "Describes the rules used to discretize (compartmentalise) a cable cell morphology.");
+    cv_policy
+        .def_property_readonly("domain",
+                               [](const arb::cv_policy& p) {return util::pprintf("{}", p.domain());},
+                               "The domain on which the policy is applied.")
+        .def(pybind11::self + pybind11::self)
+        .def(pybind11::self | pybind11::self)
```
noraabiakar, line 372: Does this just call the underlying methods in C++?

noraabiakar (changes requested):
Sorry if this review is a bit premature, I was going through the changes in the code and noticed a few minor things. I haven't reviewed the changes to the docs yet.

example/dryrun/branch_cell.hpp
```diff
@@ -112,7 +112,7 @@ arb::cable_cell branch_cell(arb::cell_gid_type gid, const cell_parameters& param
 
     cell.paint("soma"_lab, "hh");
     cell.paint("dend"_lab, "pas");
-    cell.default_parameters.axial_resistivity = 100; // [Ω·cm]
+    cell.set_default(arb::axial_resistivity{100}); // [Ω·cm]
```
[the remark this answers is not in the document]
bcumming: Sorry for that!

I have updated the unit tests and python wrapper, but not all the examples (C++ and Python).

This is a premature PR because I want feedback on the changes to the API... I promise to fix the examples.

python/cells.cpp
bcumming: Yes: it is a shorthand way to specify that the operator overloads on the C++ side can be used in Python.

arbor/cable_cell_param.cpp
```diff
@@ -1,12 +1,16 @@
+#include <iostream>
```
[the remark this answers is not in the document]
bcumming: It can and it is!

example/dryrun/branch_cell.hpp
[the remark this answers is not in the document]
bcumming: :+1:  I fixed all of the examples

test/common_cells.cpp
```diff
@@ -151,13 +152,14 @@ cable_cell soma_cell_builder::make_cell() const {
     }
 
     // Make cable_cell from sample tree and dictionary.
-    cable_cell c(tree, dict);
+    //cable_cell c(tree, dict);
```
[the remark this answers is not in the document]
bcumming: :+1:

test/unit/test_cv_geom.cpp
```diff
@@ -70,7 +70,7 @@ std::ostream& operator<<(std::ostream& out, ::arb::cv_prefer::type p) {
 TEST(cv_geom, empty) {
     using namespace common_morphology;
 
-    cable_cell empty_cell{m_empty};
+    cable_cell empty_cell{m_empty, {}, {}};
```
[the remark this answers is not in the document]
bcumming: Right. I added the convenience constructor from just a morphology after getting sick of making all these changes. Will fix.

test/unit/test_fvm_layout.cpp
```diff
@@ -30,25 +30,30 @@ using util::value_by_key;
 namespace {
     struct system {
         std::vector<soma_cell_builder> builders;
-        std::vector<cable_cell> cells;
+        std::vector<cable_cell_description> descriptions;
+
+        std::vector<arb::cable_cell> cells() const {
+            return {descriptions[0], descriptions[1]};
```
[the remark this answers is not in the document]
bcumming: There are only two cells. This was an update of what was already there for the new API.

test/unit/test_mc_cell_group.cpp
```diff
@@ -66,10 +67,11 @@ TEST(mc_cell_group, sources) {
     std::vector<cable_cell> cells;
 
     for (int i=0; i<20; ++i) {
-        cells.push_back(make_cell());
+        auto desc = make_cell();
         if (i==0 || i==3 || i==17) {
-            cells.back().place(mlocation{0, 0.3}, threshold_detector{2.3});
+            desc.decorations.place(mlocation{0, 0.3}, threshold_detector{2.3});
         }
+        cells.push_back(desc);
```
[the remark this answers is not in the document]
bcumming: In this case I think it is all the same. I changed to `emplace_back` because it is a bit clearer.

bcumming:
bors try.

doc/python/cable_cell.rst
```diff
@@ -128,34 +131,44 @@ Cable cells
...
+        :param str locations: description of the locset.
         :param mechanism: the mechanism.
         :type mechanism: :py:class:`mechanism`
 
-    .. method:: place(locations, const arb::gap_junction_site& site)
+    .. method:: place(locations, site)
+        :noindex:
 
-        Place one gap junction site at each location in ``locations``.
+        Place one gap junction site at each location at each location in ``locations``.
```
noraabiakar, line 145: double 'at each location'

doc/python/cable_cell.rst
```diff
@@ -128,34 +131,44 @@ Cable cells
...
+        :param str locations: description of the locset.
+        :param site: indicates a gap junction site..
+        :type site: :py:class:`gap_junction_site`
 
-    .. method:: place(locations, const arb::i_clamp& stim)
+    .. method:: place(locations, stim)
+        :noindex:
 
-        Add a current stimulus at each location in ``locations``.
+        Add a current stimulus at each location at each location in ``locations``.
```
noraabiakar, line 154: double 'at each location'

noraabiakar (changes requested):
I just reviewed the doc changes and pointed out a few typos, otherwise looks good!

bcumming:
bors try

test/unit/test_fvm_layout.cpp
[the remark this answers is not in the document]
noraabiakar: I just meant to say that `system` could potentially hold many descriptions, so the `cells()` method returning only the first 2 cells stops it from being used for anything other than `two_cell_system`, which is okay for the unit tests, I just found it a bit strange.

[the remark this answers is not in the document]
bcumming: Fair enough: I have added a for loop to be more general.

doc/tutorial/single_cell_model.rst
```diff
@@ -72,93 +72,103 @@ dynamics, stimulii and probes to the cell. We add two labels:
...
 ----------------------------------------------------
 
-Great, we have defined our cell! Now, let's move on to the simulation. Arbor is able to simulate
-networks with multiple individual cells; this requires a *recipe* to describe the cells,
-connections, gap junctions, etc. However, for single cell models, arbor does not require the recipe
-to be provided by the user. Arbor provides a :class:`arbor.single_cell_model<arbor.single_cell_model>`
+Once the cell description has been built, the next step is to build and run the simulation.
+Arbor can simulate networks with multiple cells using a :ref:`recipe <modelrecipe>` to describe all parts of the model:
+components including the cells, connections and gap junctions.
+For single cell models, Arbor provides a shortcut for building a model requiring the user to
```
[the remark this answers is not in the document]
bcumming: I suppose part of the problem is that it starts discussing recipes before the concept has been introduced.
I just removed all mentions of recipes, which can be introduced later, when we build a simple network.

noraabiakar (approved):
Looks good to me!

arbor/include/arbor/cable_cell.hpp
```diff
@@ -183,6 +184,7 @@ struct cable_probe_ion_ext_concentration_cell {
 // Forward declare the implementation, for PIMPL.
 struct cable_cell_impl;
 
+
```
halfflat, line 187: The typed map definitions below can now be moved into the `.cpp` file now, I think.

arbor/cable_cell.cpp
```diff
@@ -168,24 +171,32 @@ const cable_cell_region_map& cable_cell::region_assignments() const {
...
+
+void cable_cell::decorate(const decor& d) {
+    for (auto& x: d.paintings) {
+        paint(x.first, x.second);
+    }
+    for (auto& x: d.placements) {
+        place(x.first, x.second);
+    }
+    decor_.defaults = d.defaults;
 }
```
halfflat, line 200: Is it always the case that `decor_` will end up with a copy of the parameter `d`?

The current implementation with `decorate` is a bit of a mix between hidden and exposed functionality. It's not clear that the private `paint` and `place` methods are only called during initialization, nor that `decor_` remains constant over the lifetime of the cable cell. All the remaining cable cell functionality lives in the `cable_cell_impl` class.

To clarify these conditions, I'd prefer if:
* `decor_` were a const member that is initialized in the `cable_cell` constructor;
* the constructor calls a private `init` method that does the work of creating the `cable_cell_impl` object, which in turn takes the `decor` as a parameter to the constructor;
* and the `cable_cell_impl` constructor does what `decorate`, `paint`, and `place` do here.

To make the dependency and constness even more clear, one could have in the header file:
```
struct cable_cell_impl;
std::unique_ptr<cable_cell_impl, void (*)(cable_cell_impl*)> make_cable_cell_impl(const arb::morphology&, const label_dict&, const decor&);
```
and with `cable_cell` constructor:
```
cable_cell(const class morphology& m, const label_dict& l, const decor& d):
    decor_(d), impl_(make_cable_cell_impl_(m, l, d)) {}
```

This and a const `decor_` member makes it clear that `decor_` is just there to answer the `decorations()` query, and that everything else is the responsibility of the pimpl class.

arbor/cable_cell_param.cpp
```diff
@@ -68,4 +70,92 @@ cable_cell_parameter_set neuron_parameter_defaults = {
...
+                }
+                else if constexpr (std::is_same_v<ion_reversal_potential_method, T>) {
+                    defaults.reversal_potential_method[p.ion] = p.method;
+                }
+                else if constexpr (std::is_same_v<cv_policy, T>) {
+                    defaults.discretization = std::forward<cv_policy>(p);
+                }
+            },
+            what);
+}
```
halfflat, line 159: You could use `util::overload` (in `util/any_visitor.hpp`) here to avoid some of the `if constexpr (std::is_same_v<X, T>)` boilerplate (example code in `test/unit/any_visitor.cpp`).

python/cells.cpp
```diff
@@ -428,179 +502,170 @@ void register_cells(pybind11::module& m) {
...
-        .def("set_properties",
-            [](arb::cable_cell& c,
+    pybind11::class_<arb::decor> decor(m, "decor",
+            "Description of the decorations to be applied to a cable cell, that is the painted,\n"
+            "placed and defaulted properties, mecahanisms, ion species etc.");
+    decor
+        .def(pybind11::init<>())
+        .def(pybind11::init<const arb::decor&>())
+        // Set cell-wide default values for properties
+        .def("set_property",
```
halfflat, line 530: At first thought, it seems better to have the same name across the C++ and the Python interfaces, i.e. `set_property` or `set_default`. Is there a reason to keep them distinct?

halfflat (changes requested):
A couple of minor things in code comments, but the main issue is one that @schmitts observed in chat: without the `lid_range` return from `place`, how can recipes that provide cable cells with synapses (or gap junctions) against various `locset` expressions determine synapse numbering?

This problem will become even more acute when we have random seeds provided at simulation-initialization time.

It may demand a rethink of how connection targets are identified in recipes.

python/example/single_cell_multi_branch.py
```diff
@@ -1,112 +0,0 @@
-#!/usr/bin/env python3
```
noraabiakar, line 1: Why has this example been removed?

arbor/cable_cell_param.cpp
```diff
@@ -68,4 +70,92 @@ cable_cell_parameter_set neuron_parameter_defaults = {
...
+    }
+    if (axial_resistivity) {
+        D.push_back(arb::axial_resistivity{*this->axial_resistivity});
+    }
+    if (membrane_capacitance) {
+        D.push_back(arb::membrane_capacitance{*this->membrane_capacitance});
+    }
+
+    //std::unordered_map<std::string, cable_cell_ion_data> ion_data;
+    for (auto& ion: ion_data) {
```
thorstenhater, line 90: How about
```
for (const auto& [name, data]: ion_data) {
   ...
}
```
?

thorstenhater (commented):
A small suggestion, otherwise good.

arbor/include/arbor/cable_cell.hpp
```diff
@@ -349,7 +285,29 @@ class cable_cell {
...
 
+    // View the decorations on the cell.
+    const decor& decorations() const {
+        return decor_;
+    }
+
+    const cable_cell_parameter_set& default_parameters() const {
+        return decor_.defaults;
+    }
+
```
halfflat, line 296: Is it necessary to have the `decorations()` method? Following on from the comment above on `cable_cell::decorate`, `default_parameters()`, like `region_assignments()` and `location_assignments()`, could be delegated to the pimpl object, and then there would be no need to hold on to a copy of the `decor_` object at all.

arbor/cable_cell_param.cpp
bcumming: Good point. I am still getting used to using the C++17 syntactic sugar.

arbor/cable_cell.cpp
bcumming: Seems reasonable. I will have a look at a time estimate for the work, and whether it can go in a separate PR or be merged here.

python/cells.cpp
bcumming: The set_property and set_ion calls were part of the Python API for global properties and cable_cell before, and this here just reflects this.

For Python, interfaces like the following
`set_ion(ion_name, {named properties})`
`set_property([named properties])`
 were easier to define and implement than a `set_default` interface for setting both cable properties and  ion species values with named arguments. 

We could have a "typey" interface:
`set_default(arbor.ion(ion_name, [named properties])`
But such an interface for membrane potential, capacitance, etc would be overly verbose and awkward (my opinion), and a solution like the following
```Python
c.set_default(arbor.ion('ca', 'reversal_potential'=12))
c.set_default(Cm=0.1, Vm=-20)
```
is hard to implement.

arbor/include/arbor/cable_cell.hpp
bcumming: One reason to keep it around (and to keep morphology and label dictionary information around) is to make it possible to "serialize" and `cable_cell`, which is straightforward with the decor, morphology and label_dict used to instantiate it.

Whether we want to do this is another matter. You could argue that the original decor, labels and morphology used to create the cell can be retained and used for that purpose. I haven't put enough thought into possible use cases and situations that could arise.

halfflat: I was wondering about the serialization interface. For descriptions, I agree that it's the decor that should be serialized, rather than the cell. For snapshot purposes, it's open: do we represent cells, or do we represent cell groups? If the former, then I think it should be in terms of thingified quantities; if the latter, then the vectors of stuff based on the discretization.

bcumming:
Good point about the lid ranges.

I had thought about this, but it slipped my mind as I proceeded with the implementation.

I have been careful to ensure that the order in which items are placed with calls to `place` is respected by the `decor`. In practice placements are pushed onto a `std::vector`, and there is no deletion or modification of operations once something has been placed or painted. The constructor of the `cable_cell` then respects this order.

With this in mind, we could have one additional level of indirection best illustrated with some pseudo C++:

```C++  
// calls to decor::place return a "tag" (an int should do)
auto syn_id = decor.place("(terminal)", my_synapse_mech);

// lids are determined and stored on instantiation of the cell
cable_cell cell(morp, labs, decor);

// ... and can be accessed later
auto lids = cell.placed_lid_range(syn_id);

std::cout << "synapses have lids in [" << lids.first << ", " << lids.second << ")\n";
```

One might also extract this information from the current API using `cable_cell::location_assignments()`, however it would be a bit cumbersome.                                                        

This kicks the can down the road on the question about what to do when generating multiple concrete cell instances from the same cable_cell instance with different seeds.

bcumming:
Good point about the lid ranges.

I had thought about this, but it slipped my mind as I proceeded with the implementation.

I have been careful to ensure that the order in which items are placed with calls to `place` is respected by the `decor`. In practice placements are pushed onto a `std::vector`, and there is no deletion or modification of operations once something has been placed or painted. The constructor of the `cable_cell` then respects this order.

With this in mind, we could have one additional level of indirection best illustrated with some pseudo C++:

```C++  
// calls to decor::place return a "tag" (an int should do)
auto syn_id = decor.place("(terminal)", my_synapse_mech);

// lids are determined and stored on instantiation of the cell
cable_cell cell(morp, labs, decor);

// ... and can be accessed later
auto lids = cell.placed_lid_range(syn_id);

std::cout << "synapses have lids in [" << lids.first << ", " << lids.second << ")\n";
```

One might also extract this information from the current API using `cable_cell::location_assignments()`, however it would be a bit cumbersome.                                                        

This kicks the can down the road on the question about what to do when generating multiple concrete cell instances from the same cable_cell instance with different seeds.

arbor/include/arbor/cable_cell.hpp
bcumming: I was referring to descriptions, and not cell state.

I agree that the decor + morphology + labels should be the inputs for serialization (that's the whole point of this refactoring, anyway). The question is whether one should be able to "extract" those three things from the `cable_cell` for serialization, or it whether the inputs used to create the cell should be retained for that purpose.

One use case I can think of would be serializing a "complete" cell description generated from a recipe, where the inputs used to generate the cell are not available (e.g. the recipe comes from a library or a python call back).

```C++
arb::cable_cell c = std::any_cast<arb::cable_cell>(rec.get_cell_description(42));
serialize(c);
```

This seems like a reasonable thing to support, however we can't do it at the moment because `cable_cells` don't retain the `label_dict`, and if the morphology is retained, it is kept in the bottom of a provider somewhere that isn't obvious.

bcumming:
bors try

arbor/cable_cell_param.cpp
bcumming: I reverted this change, because it triggers a bug in Clang (compiler error).

doc/concepts/cable_cell.rst
```diff
@@ -139,7 +159,7 @@ morphology. Each fork point in the morphology will be the responsibility of
 a single CV, and as a special case a zero-volume CV can be used to represent
 a single fork point in isolation.
 
-.. _cable-density-mechs:
+.. _cablecell-density-mechs:
```
halfflat, line 162: The subsection on Discretisation here is out of place now, but it's also made redundant given that the longer discussion has been brought into this file from the C++ API. Can you remove these lines 150–160 and redirect any references to `cablecell-discretization` to `cablecell-cv-policies` instead?

halfflat (changes requested):
Looks good! I've asked for a small documentation edit, but otherwise 👍

halfflat (approved):
Looks good!

bcumming:
bors try.

PR #1239, opened 2020-11-13
Ci/gh actions

thorstenhater:
As we have discussed elsewhere, Github Actions will have to replace Travis CI.
So, here is a first cut at the problem:

- Build arbor libraries, examples, and unit tests.
- Run unit tests
- OS: Ubuntu 18.04 LTS, 20.04 LTS, MacOS 10.15
- Compilers GCC 8, 9, 10; Clang 8, 9, 10, Apple


Helveg (commented):
In our Slack chat you mentioned OpenGL + GLFW3, was nothing required to make this work?

thorstenhater:
> In our Slack chat you mentioned OpenGL + GLFW3, was nothing required to make this work?

That would be here: https://github.com/thorstenhater/arbor-gui/actions/runs/361594333

.github/workflows/main.yml
```diff
@@ -0,0 +1,104 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    name: "Build & Test"
+    runs-on: ${{ matrix.config.os }}
+    strategy:
+      matrix:
+        config:
```
bcumming, line 15: I wonder: do we need to test all of the Ubuntu+compiler combinations?

We could focus on one version of Ubuntu, and "all the compilers" on that Ubuntu?

bcumming (commented):
This review is just a comment, to start a discussion about where to go.

Some points that are not covered:
* Using the minimum supported version of CMake on each platform: we often get caught out if we don't test with the minimum supported version.
* Running the Python unit tests
* Testing MPI builds

The CMake, Python and MPI builds might create the need for OS-specific build and run processes (e.g. different methods to install a specific version of CMake).

Ultimately we would like GH actions to replace Travis, and the above are required to do that.

We can make a first version of GH actions that doesn't cover everything, then extend it and replace Travis when ready.

Opinions?

thorstenhater:
As said above, this is a first cut, and I agree that we should add Python, MPI, and CMake variants.
To keep things manageable, I will split these files by OS.
Question: Do we need separate runs for MPI and non-MPI, or do MPI builds and test subsume the serial case?

bcumming:
By first cut do you mean "call for comment", or "merge, then improve"?

When building with MPI, there are separate unit tests (C++ and Python) that test out the distribute/mpi back end.

On Travis we split them into separate builds, so that we can test each configuration with and without `-DARB_MPI=on`. I would be more comfortable doing that, because it makes it simpler to understand CI build fails if the problem is caused by MPI.

thorstenhater:
Update: 
1. Split by OS and compiler
2. Reduce Ubuntu to 20.04 LTS
3. Span the matrix as (min, max) supported versions for CMake, Python, C++ in stead of full range
4. Split out MPI ./. non-MPI builds

2. and 3. are motivated by the combinatorics, if we were to test all version between min and max, we would
have to run around 750 jobs.

Currently only Ubuntu is set up. I will extend this in the coming days.

thorstenhater:
I'll close this due to contamination by bad submodules.

PR #1243, opened 2020-11-16
Use FindPythonModule to warn for missing svgwrite

### LINKED ISSUE #1128: Documentation won't build without an extra python module
After checking out master, updating submodules etc. and configuring, building the documentation with
```
make html
```
gives the error:
```
Scanning dependencies of target check-sphinx
Scanning dependencies of target check-sphinx_rtd_theme
Finding Sphinx
Checking for Read the Docs Sphinx theme
Built target check-sphinx
Built target check-sphinx_rtd_theme
Scanning dependencies of target html
Generating Sphinx documentation
--- generating images ---

Configuration error:
There is a programmable error in your configuration file:

  File "/usr/lib/python3/dist-packages/sphinx/config.py", line 368, in eval_config_file
    execfile_(filename, namespace)
  File "/usr/lib/python3/dist-packages/sphinx/util/pycompat.py", line 150, in execfile_
    exec_(code, _globals)
  File "XXX/arbor/doc/conf.py", line 50, in <module>
    import make_images
  File "XXX/arbor/doc/scripts/make_images.py", line 2, in <module>
    import svgwrite
ModuleNotFoundError: No module named 'svgwrite'
```

Should doc generation run things in its own virtualenv?
And should the module be somehow bundled when we make releases, so that internet access is not required at point-of-install?

brenthuisman:
Use FindPythonModule to warn for missing svgwrite. Fixes #1128.

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -6,12 +6,11 @@ add_target_if(SPHINX_FOUND
     "Finding Sphinx"
     "Sphinx must be installed to build documentation")
 
-if (SPHINX_FOUND)
-    find_python_module(svgwrite)
-    if (NOT HAVE_SVGWRITE)
-        message(WARNING "Building documentation requires the svgwrite Python module. You can obtain it from pip like so: pip3 install svgwrite --user")
-    endif()
-endif()
+find_python_module(svgwrite)
+add_target_if(HAVE_SVGWRITE
+    ext-svgwrite
+    "Finding svgwrite"
+    "The Python module svgwrite needs to be installed to build documentation. Use pip to install svgwrite")
 
 set(html_dir "${CMAKE_CURRENT_BINARY_DIR}/html")
 set(doctree_dir "${CMAKE_CURRENT_BINARY_DIR}/_doctrees")
@@ -28,7 +27,7 @@ add_custom_target(html
         -q                          # Quiet: no output other than errors and warnings.
         ${CMAKE_CURRENT_SOURCE_DIR} # Source directory
         ${html_dir}                 # Output directory
-    DEPENDS check-sphinx ext-sphinx_rtd_theme
+    DEPENDS check-sphinx ext-sphinx_rtd_theme ext-svgwrite
     COMMENT
         "Generating Sphinx documentation")
 
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -1,8 +1,17 @@
+include(FindPythonModule) # required for find_python_module
+
 find_package(Sphinx)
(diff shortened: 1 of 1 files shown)


bcumming (commented):
The `requirements.txt` was added to ensure that ReadTheDocs could build the documentation. For developers building the docs locally, we can 

Installing python dependencies for `--user` with pip is bad manners. If we wanted to automatically ensure that the libraries are available, we can set up a virtualenv to use in the build (not so trivial to do).

bcumming:
The proposal on the ticket #1128 from @halfflat was to check for the presence of the packages at build time (when `make html` is requested), and present a user-friendly error message about the missing python package `svgwrite`, like we do for `sphinx`.

doc/CMakeLists.txt
```diff
@@ -1,18 +1,25 @@
...
 find_package(Sphinx)
 add_target_if(SPHINX_FOUND
     check-sphinx
     "Finding Sphinx"
     "Sphinx must be installed to build documentation")
 
+if (SPHINX_FOUND)
+    find_python_module(svgwrite)
+    if (NOT HAVE_SVGWRITE)
+        message(WARNING "Building documentation requires the svgwrite Python module. You can obtain it from pip like so: pip3 install svgwrite --user")
```
bcumming, line 12: I would avoid giving the explicit `--user` tag: instead just say install with pip.
Best practice is to install in a virtual env, which we shouldn't nudge users away from (let's leave it up to them to manage their packages).

We could add another `add_target_if` to define a `check-py-svgwrite` target (see `check-sphinx` above), that would then be used as a dependency of the `html` target below, like `check-sphinx`.

```cmake
find_python_module(svgwrite)
add_target_if("${SPHINX_FOUND} AND ${HAVE_SVGWRITE}"
    check-py-svgwrite
    "Finding python svgwrite"
    "The python module svgwrite needs to be installed to build documentation")
```

Something like the above could be used (you will have to test and check the syntax of the boolean expression, because I don't know enough CMake off the top of my head to be certain that it is correct).

brenthuisman: I tried that first, but then the error doesn't show up in the `cmake` step, only in the `make html`. What's the right stage for this error to show up?

I pushed your proposed version just now.

bcumming: Yes, that was the intention to have the error occur at `make html`, i.e. when building the docs.

**Rationale** building the library/examples/tests should work even if everything required to build the docs is not available. So we defer throwing the error until doc build time, but we use the configure time test to ensure that we produce a nice error message.

bcumming (approved):
lgtm!

PR #1246, opened 2020-11-17
Change use of long double to double.

### LINKED ISSUE #1245: WSL: modcc unit test fails
Running the modcc unit tests on a Win10+WSL+Ubuntu 20.04 build has 1 test failing (all other tests are OK):
```
/home/user/arbor/test/unit-modcc/test_parser.cpp:663: Failure
The difference between eval(e.get()) and test_case.second is 7.4505805969238281e-09, which exceeds 1e-10, where
eval(e.get()) evaluates to -63281459.031249993,
test_case.second evaluates to -63281459.03125, and
1e-10 evaluates to 1e-10.
```

A bug has been filed in the WSL repo: https://github.com/microsoft/WSL/issues/830. `long double` is assumed to be 80 bit, but is 64 bit in WSL. 

Since as of today WSL still installs v1 rather than v2 by default, and this is a minor bug, we can change the test to a relative error, e.g.:
`EXPECT_TRUE(near_relative(eval(e.get()), test_case.second, 1e-10));`
We can instead not assume assumptions about `long double`, and change `eval()` accordingly.

brenthuisman:
Change from `long double` to `double` for storing floating point values in modcc.

WSL, and Windows generally, treats `long double` differently than Linux, which leads to inconsistencies in literal numeric values in code generated by modcc on the two platforms.

Fixes #1245

### DIFF
--- a/test/unit-modcc/test_parser.cpp
+++ b/test/unit-modcc/test_parser.cpp
@@ -580,7 +580,7 @@ TEST(Parser, parse_conserve) {
     }
 }
 
-long double eval(Expression *e) {
+double eval(Expression *e) {
     if (auto n = e->is_number()) {
         return n->value();
     }
@@ -606,7 +606,7 @@ long double eval(Expression *e) {
             default:;
         }
     }
-    return std::numeric_limits<long double>::quiet_NaN();
+    return std::numeric_limits<double>::quiet_NaN();
 }
 
 // test parsing of expressions for correctness
@@ -656,10 +656,10 @@ TEST(Parser, parse_binop) {
         EXPECT_TRUE(check_parse(e, &Parser::parse_expression, test_case.first));
 
         // A loose tolerance of 1e-10 is required here because the eval()
-        // function uses long double for intermediate results (like constant
+        // function uses double for intermediate results (like constant
         // folding in modparser).  For expressions with transcendental
         // operations this can see relatively large divergence between the
-        // double and long double results.
+        // double and double results.
         EXPECT_NEAR(eval(e.get()), test_case.second, 1e-10);
     }
 }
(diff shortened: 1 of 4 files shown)


halfflat:
We'd have to changed the uses of `long double` to `double` in `modcc` too.

brenthuisman:
What about `long long`? I think those are OK, but please confirm :)

halfflat:
I think we can consider changing `long long` to an explicit width type as something to consider in the scope of a different PR.

test/unit-modcc/test_parser.cpp
```diff
@@ -656,10 +656,10 @@ TEST(Parser, parse_binop) {
         EXPECT_TRUE(check_parse(e, &Parser::parse_expression, test_case.first));
 
         // A loose tolerance of 1e-10 is required here because the eval()
-        // function uses long double for intermediate results (like constant
+        // function uses double for intermediate results (like constant
```
bcumming, line 659: This comment no longer makes sense.

bcumming (changes requested):
Does this fix the test on WSL? I thought that the problem was that on WSL the `long double` was being treated as a `double`. If so, using `double` will still create the incorrect result.

How about we instead update the test to test for relative error, and not absolute error? I would like to keep the explicit, more accurate, compile time `long double` if possible.

halfflat:
My issue with `long double` is that it bakes in a platform-dependent numerical interpretation in generated code, making reproducibility harder; it's not just that the specific numerical environment would need to be recreated (if this is indeed possible), but this has to be done on the platform that runs `modcc` _as well_ as on the execution platform.

It's not too stringent a requirement, I think, that we demand IEEE double precision semantics everywhere for correctness and reproducibility, but expecting `long double` behaviour to be the same everywhere is unreasonable.

thorstenhater:
Question: What was the rationale for using long double in the first place?

bcumming:
Rationale was that it was at least as accurate as double, likely more so. If we were going to do folding, might as well make it accurate. That was before I knew what I know now.

test/unit-modcc/test_parser.cpp
brenthuisman: I'll remove it.

brenthuisman:
The PR at present does make the unit-modcc test pass on Windows WSL.

bcumming (approved):

PR #1247, opened 2020-11-19
Undo hip-clang bug workaround

noraabiakar:
`pow(double x, int y)` is implemented as of ROCm-3.9 (May have been fixed earlier, don't have access to earlier versions)


halfflat (commented):
Looks good!

halfflat (approved):

bcumming (approved):
:+1:

PR #1248, opened 2020-11-20
Fix pybind11 issues.

thorstenhater:
This Pr fixes three issue with our current Python bindings

1. Compile error in simulation for implicit narrowing cast.
2. Warnings for missing virtual destructors in schedule
3. A SegFault in recipe.

On 3: Here is a minimal example

``` python
#!/usr/bin/env python3
import arbor as A

class Recipe(A.recipe):
    def __init__(self, n_cell):
        A.recipe.__init__(self)
        self.n_cell = n_cell
        self.kind = A.lif_cell()

    def num_cells(self):
        return self.n_cell

    def num_targets(self, gid):
        return 0

    def num_sources(self, gid):
        return 0

    def cell_kind(self, gid):
        return A.cell_kind.lif

    def connections_on(self, gid):
        return []

    def event_generators(self, gid):
        return []

    def get_probes(self, gid):
        return []

    def cell_description(self, gid):
        return A.lif_cell()


context = A.context(threads=1)
recipe = Recipe(n_cell=1)
dd = A.partition_load_balance(recipe, context)
sim = A.simulation(recipe, dd, context)
```

This appears in MPI-enabled python tests only, namely `test/unit_distributed/test_simulator.py`. Reason 
is that the test case hits precisely this issue of returning a temporary to C++. It can be 'fixed' by
changing 
``` python
    def cell_description(self, gid):
        self.kind = A. lif_cell()
        return self.kind
```
thus demonstrating an issue with lifetime/GC.

The issue occurs to due a lifetime problem in `cell_description` when returning a temporary
value from python to C++. Following the comment in `cells.cpp` I tried acquiring the GIL before
calling `convert_cells`, which fixes the issue.

### DIFF
--- a/python/recipe.cpp
+++ b/python/recipe.cpp
@@ -24,6 +24,7 @@ namespace pyarb {
 // The py::recipe::cell_decription returns a pybind11::object, that is
 // unwrapped and copied into a arb::util::unique_any.
 arb::util::unique_any py_recipe_shim::get_cell_description(arb::cell_gid_type gid) const {
+    pybind11::gil_scoped_acquire guard;
     return try_catch_pyexception(
                 [&](){ return convert_cell(impl_->cell_description(gid)); },
                 "Python error already thrown");
--- a/python/recipe.cpp
+++ b/python/recipe.cpp
@@ -24,7 +24,6 @@ namespace pyarb {
 // The py::recipe::cell_decription returns a pybind11::object, that is
 // unwrapped and copied into a arb::util::unique_any.
 arb::util::unique_any py_recipe_shim::get_cell_description(arb::cell_gid_type gid) const {
-    pybind11::gil_scoped_acquire guard;
     return try_catch_pyexception(
                 [&](){ return convert_cell(impl_->cell_description(gid)); },
                 "Python error already thrown");
--- a/python/schedule.hpp
+++ b/python/schedule.hpp
@@ -18,6 +18,7 @@ struct schedule_shim_base {
     schedule_shim_base() = default;
     schedule_shim_base(const schedule_shim_base&) = delete;
     schedule_shim_base& operator=(schedule_shim_base&) = delete;
+    virtual ~schedule_shim_base() {}
 
     virtual arb::schedule schedule() const = 0;
 };
@@ -36,6 +37,8 @@ struct regular_schedule_shim: schedule_shim_base {
 
     regular_schedule_shim(time_type t0, time_type delta_t, pybind11::object t1);
     explicit regular_schedule_shim(time_type delta_t);
+
+    virtual ~regular_schedule_shim();
 
     // getter and setter (in order to assert when being set)
     void set_tstart(time_type t);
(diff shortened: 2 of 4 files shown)


python/recipe.cpp
```diff
@@ -24,6 +24,7 @@ namespace pyarb {
 // The py::recipe::cell_decription returns a pybind11::object, that is
 // unwrapped and copied into a arb::util::unique_any.
 arb::util::unique_any py_recipe_shim::get_cell_description(arb::cell_gid_type gid) const {
+    pybind11::gil_scoped_acquire guard;
```
bcumming, line 27: Why does this fix the problem with temporaries?

`try_catch_pyexception` uses a global mutex that we use on the C++ side, and `convert_cell` acquires the GIL explicitly. Acquiring the GIL at this level could lead to deadlock.

bcumming (changes requested):
I would like to understand the root cause of the error that is fixed in this case by acquiring the GIL. The proposed solution might cause other issues.

The rest of the PR is fine.

python/recipe.cpp
thorstenhater: Quoting `cells.cpp` comment right before `convert_cells`
```
// Warning: requires that the GIL has been acquired before calling,
// if there is a segmentation fault in the cast or isinstance calls,
// check that the caller has the GIL.
arb::util::unique_any convert_cell(pybind11::object o) {
```

thorstenhater:
The segfault only appears when upgrade pybind11 to v2.6.1, so I will retract that part.

python/schedule.hpp
```diff
@@ -18,6 +18,7 @@ struct schedule_shim_base {
     schedule_shim_base() = default;
     schedule_shim_base(const schedule_shim_base&) = delete;
     schedule_shim_base& operator=(schedule_shim_base&) = delete;
+    virtual ~schedule_shim_base() {}
```
halfflat, line 21: I'm pretty sure this line is all we need for getting the virtual destructor right; no need to implement overloads in the derived classes.

python/simulation.cpp
```diff
@@ -107,7 +107,7 @@ class simulation_shim {
     }
 
     py::object spikes() const {
-        return py::array_t<arb::spike>({spike_record_.size()}, spike_record_.data());
+        return py::array_t<arb::spike>(py::ssize_t{(long) spike_record_.size()}, spike_record_.data());
```
halfflat, line 110: I think a simple `py::ssize_t(spike_record_.size())` will do.

thorstenhater: 👍

python/schedule.hpp
thorstenhater: 👍

halfflat (approved):
Looks good!

python/recipe.cpp
thorstenhater: Retracted.

bcumming (approved):

PR #1249, opened 2020-11-20
Ci/gh actions

thorstenhater:
First step of replacing Travis with GithubActions

Here, only a basic set of tests has been added

- Build arbor with MPI and Python
- Compile examples
- Run unit tests, incl distributed.
- Run python tests, incl distributed.
- Run python examples.
- Build documentation.

Next steps
======
- [ ] Test installation.
- [ ] Test pip install.
- [ ] Fan out matrix.
- [ ] Remove Travis.


brenthuisman:
Do we remove all Travis related stuff right away? We have badges in README and doc/index.rst, maybe other places?

.github/workflows/ubuntu-gcc.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Set up cmake
+        uses: jwlawson/actions-setup-cmake@v1.4
+        with:
+          cmake-version: ${{ matrix.cmake }}
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.py }}
+      - name: Install Python packages
+        run: python${{ matrix.py }} -mpip install numpy sphinx svgwrite sphinx-rtd-theme
```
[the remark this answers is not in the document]
bcumming: We use `numpy` in the tests (actually, @halfflat added numpy support in the Python wrapper... @halfflat, we might have to add it to Arbor's `requirements.txt`?)
The other are required to build to documentation. @thorstenhater , do we need to build the docs for the ci tests?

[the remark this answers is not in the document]
thorstenhater: I would like the docs to be built as well, yes. There's two reason:
- The documentation is worked on similar to other parts, so breaking it is a blocker
- The documentation can in principle contain tests and examples.

.github/workflows/ubuntu-clang.yml
```diff
@@ -0,0 +1,67 @@
...
+      - name: Set up cmake
+        uses: jwlawson/actions-setup-cmake@v1.4
+        with:
+          cmake-version: ${{ matrix.cmake }}
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.py }}
+      - name: Install Python packages
+        run: python${{ matrix.py }} -mpip install numpy sphinx svgwrite sphinx-rtd-theme
```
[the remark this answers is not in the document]
thorstenhater: See above.

.github/workflows/macos-clang.yml
```diff
@@ -0,0 +1,65 @@
...
+      - name: Set up cmake
+        uses: jwlawson/actions-setup-cmake@v1.4
+        with:
+          cmake-version: ${{ matrix.cmake }}
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.py }}
+      - name: Install Python packages
+        run: python${{ matrix.py }} -mpip install numpy sphinx svgwrite sphinx-rtd-theme
```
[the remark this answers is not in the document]
thorstenhater: See above.

thorstenhater:
Hi all,

please take a look at `basic.yml`, this is going to be the prototype for all
tests in the set. If good, I will fan out the remaining jobs from there as discussed.

T

.github/workflows/basic.yml
```diff
@@ -0,0 +1,100 @@
...
+        cc:    ["gcc"]
+        ver:   ["9"]
+        py:    ["3.6"]
+        cmake: ["3.12.x"]
+        mpi:   ["ON"]
+    env:
+        CC:         ${{ matrix.cc }}-${{ matrix.ver }}
+        CXX:        ${{ matrix.cxx }}-${{ matrix.ver }}
+        PY:         python${{ matrix.py }}
+        PYTHONPATH: /usr/local/lib/python${{ matrix.py }}/site-packages/
```
brenthuisman, line 26: I think we've just established that Ubuntu's system-Python doesn't look in PYTHONPATH (site-packages). Apparently `actions/setup-python@v2` does look there, or we'd be seeing failing builds. Is it a good idea to rely on this working?

.github/workflows/basic.yml
```diff
@@ -0,0 +1,100 @@
...
+        cxx:   ["g++"]
+        cc:    ["gcc"]
+        ver:   ["9"]
+        py:    ["3.6"]
+        cmake: ["3.12.x"]
+        mpi:   ["ON"]
+    env:
+        CC:         ${{ matrix.cc }}-${{ matrix.ver }}
+        CXX:        ${{ matrix.cxx }}-${{ matrix.ver }}
+        PY:         python${{ matrix.py }}
```
Helveg, line 25: This is obsolete as the `setup/python` action makes sure that `python` links to the python version you specify

.github/workflows/basic.yml
Helveg: Using `PYTHONPATH` is a major code smell. What is this used for? If there are packages missing they should be installed inside the environment set up by `setup/python` rather than being "sneak peeked" from the system install.

.github/workflows/basic.yml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Install OpenMPI
+        run: |
+           echo "Going to install ompi"
+           cd ~
+           cd openmpi-4.0.2
+           sudo make install
+           sudo ldconfig
+           cd -
+      - name: Install Python packages
+        run:  $PY -mpip install numpy sphinx svgwrite sphinx-rtd-theme mpi4py pandas seaborn
```
Helveg, line 63: `setup/python` should link `pip` to be the equivalent of `python -m pip`

.github/workflows/basic.yml
thorstenhater: Good you mention this. It is caused by arbor hardwiring `site-packages` into `CMakeLists.txt`; *but* Debian uses `dist-packages`. Subject to another PR, I guess.

.github/workflows/basic.yml
thorstenhater: Thanks, I'll try it.

.github/workflows/basic.yml
thorstenhater: Fixed.

.github/workflows/basic.yml
bcumming: For the unit tests we don't `make install` on Travis. We just set PYTHONPATH to `arbor/build/python`.
I would be happy to reproduce that behavior (and have other, separate, tests that test pip at a later date)

.github/workflows/basic.yml
```diff
@@ -0,0 +1,96 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.py }}
+      - name: OpenMPI cache
+        uses: actions/cache@v2
+        id:   cache-ompi
+        with:
+          path: ~/openmpi-4.0.2
+          key:  ${{ matrix.os }}-openmpi-4.0.2-${{ matrix.cxx }}-${{ matrix.ver }}
+      - name: Build OpenMPI
```
bcumming, line 41: Installing the openMPI packages doesn't work?
https://github.com/arbor-sim/arbor/blob/master/.travis.yml#L86

.github/workflows/basic.yml
```diff
@@ -0,0 +1,96 @@
...
+  build:
+    name: "Default test set"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os:    ["ubuntu-20.04"]
+        cxx:   ["g++"]
+        cc:    ["gcc"]
+        ver:   ["9"]
+        py:    ["3.6"]
```
bcumming, line 19: Should we bump this to 3.7?
It is time according to the numpy rules.

brenthuisman: Let's test anyway, it's nice to know when it starts failing.

.github/workflows/basic.yml
thorstenhater: No, it seems to break when swapping compilers around. I tried that.

.github/workflows/basic.yml
thorstenhater: It should really say "minimum of whatever we support" in each category.
But that's hard to scrape form the website.

.github/workflows/basic.yml
thorstenhater: Sure, no problem.

Helveg: So if I understand correctly, rather than installing the built files to their location, you point Python to the build location? Could this be added in a comment with the justification?

thorstenhater: Sure. :+1:

.github/workflows/basic.yml
```diff
@@ -22,7 +22,8 @@ jobs:
     env:
         CC:         ${{ matrix.cc }}-${{ matrix.ver }}
         CXX:        ${{ matrix.cxx }}-${{ matrix.ver }}
-        PYTHONPATH: /usr/local/lib/python${{ matrix.py }}/site-packages/
+        # We set PYTHONPATH instead of installing to avoid distribution/OS specific behaviour.
```
brenthuisman, line 25: Make it a bit more specific : instead of `make install`ing

brenthuisman (approved):

PR #1253, opened 2020-11-25
Bug fix: shared state time vector swap 

### LINKED ISSUE #1251: Swapped pointers lead to wrong direction of interpolation in threshold watcher, and wrong `dt` in mechanisms.
The `threshold_watcher` in `fvm_lowered_cell_impl` has constant pointers to the `time` and `time_to` vector data stored in the `shared_state`, used to linearly interpolate the time of a spike. These pointers remain valid when the `shared_state` performs a `std::swap` operation on the `time` and `time_to` vectors (at every integration step) causing `time` to be less than `time_to` half of the time, and the interpolation direction in the `threshold_watcher` to be incorrect. 
The errors caused by this bug are of the same order of the errors in the integration, which is likely why this hasn't been noticed before. Nevertheless, a test should be added to catch the bug.

noraabiakar:
Fixes a bug caused by swapping the `time` and `time_to` vectors in the `shared_state` and not swapping the corresponding data pointers in the `threshold_watcher` and `mechanism`.
- This fix keeps the swapping optimization, and instead, the `threshold_watcher` and `mechanism` hold pointers to the std::vectors (not the underlying arrays) which are dereferenced before use. 
- The mechanism API functions have been renamed. 
- A unit test is also added that checks the interpolation behavior of the `threshold_watcher`.

Fixes #1251

### DIFF
--- a/arbor/backends/gpu/fvm.hpp
+++ b/arbor/backends/gpu/fvm.hpp
@@ -57,9 +57,9 @@ struct backend {
     {
         return threshold_watcher(
             state.cv_to_intdom.data(),
-            state.time.data(),
-            state.time_to.data(),
             state.voltage.data(),
+            &state.time,
+            &state.time_to,
             cv,
             thresholds,
             context);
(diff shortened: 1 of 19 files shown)


noraabiakar:
bors try

thorstenhater (approved):
Looks good.

PR #1264, opened 2020-12-07
Prevent CMake from building in source-dir. Prevent git from picking up CMake generated artifacts.

brenthuisman:


.gitignore
```diff
@@ -48,7 +48,13 @@
 CMakeFiles
 CMakeCache.txt
 cmake_install.cmake
+CMakeFiles/
 /compile_commands.json
+arbor-config-version.cmake
+arbor-config.cmake
+bin/
+arbor/include/arbor/assert_macro.hpp
+ext/gbench-build-prefix/
```
halfflat, line 57: This is possibly a very personal preference, but I'd rather we don't fill up the .gitignore file with lots of special cases; ultimately, we can't predict how we're going to build things, or locally install things, etc., and can at best catch only some of the common ones.

So we have to be careful when using git add and git commit anyway. .gitignore can't save us, unless we're typically making non-source files in source directories, which we really shouldn't be doing anyway.

brenthuisman: Yeah, I'm not sure about it either. Since I saw some lines blocking Cmake stuff I decided to add mine. Should that be removed from `.gitignore` as well?

halfflat: That's my preference, but I think it's a preference that may be in opposition to @bcumming's.

bcumming: Right. I think that I am somewhere in between.

The artifacts generated by cmake during configure and build phases shouldn't be in the `.gitignore`. However, that's easy for me to say, because I think I added `build` (the path in which I do my out of tree CMake) to `.gitignore` a long time ago.

The CMake artifacts currently in .gitignore were probably added a long long time ago without any thought given to long term maintenance of `.gitignore`.

@halfflat, you would probably be in favor of removing `build` from the `.gitignore`? I am not so keen, as it would force me to do completely out of tree builds (because unlike you, I don't like having to be so careful when `git add`ing).

bcumming: To make it clear: yes, let's remove all of those CMake and build artifacts from the `.gitignore` instead of expanding them.

CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
+   message(SEND_ERROR "In-source builds are not allowed.")
+endif("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
```
bcumming, line 3: You will be happy to hear that `endif()` is enough.

CMakeLists.txt
```diff
@@ -1,3 +1,7 @@
+if("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
```
bcumming, line 1: Is this a standard approach? I couldn't find any guildelines regarding this online.

Because it isn't explicitly disallowed by CMake, I don't know if we want to arbitrarily turn the feature off.

Also, everyone has to make a complete mess of their repo at least once to truly understand two key lessons in life:
1. you don't do in tree builds
2. that CMake can't be trusted
;)

.gitignore
halfflat: Yeah, for example, I personally don't use `build/` as a build directory (I typically use `b/`) so even that entry doesn't help me :)

CMakeLists.txt
brenthuisman: I can't conceive of any possible reason to make an in-source build, but I can think of several that cause me to accidentally do so, one of which prompted this PR today.

.gitignore
thorstenhater: About the `build` thing ... I tried adding a file named `build-catalogue.py` which was stopped outright by `.gitignore` rules

brenthuisman:
Travis is failing with the "In-source builds are not allowed." error. I'm checking it out.

brenthuisman:
Can't figure out why, but the proposed change disallows builds in subdirs of source as well, which causes Travis to fail. `set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)` does not work either. I'll delete this PR.

PR #1265, opened 2020-12-08
`std::allocator<void>` is deprecated in C++17

thorstenhater:
... so it has been removed.

### DIFF
--- a/arbor/memory/allocator.hpp
+++ b/arbor/memory/allocator.hpp
@@ -184,7 +184,7 @@ public:
         return &r;
     }
 
-    pointer allocate(size_type cnt, typename std::allocator<void>::const_pointer = 0) {
+    pointer allocate(size_type cnt, const_pointer = 0) {
         if (cnt) {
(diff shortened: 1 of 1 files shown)


thorstenhater:
bors try

arbor/memory/allocator.hpp
```diff
@@ -184,7 +184,7 @@ class allocator : public Policy {
         return &r;
     }
 
-    pointer allocate(size_type cnt, typename std::allocator<void>::const_pointer = 0) {
+    pointer allocate(size_type cnt, const_pointer = 0) {
```
bcumming, line 187: Now that we live in C++17 land, this could be simplified to
```
pointer allocate(size_type cnt)
```
https://en.cppreference.com/w/cpp/memory/allocator/allocate

bcumming (changes requested):

arbor/memory/allocator.hpp
thorstenhater: Good point.

bcumming (approved):

PR #1269, opened 2020-12-09
Remove mention of the Copyright Transfer Agreement

brenthuisman:
Remove mention of the Copyright Transfer Agreement


thorstenhater (changes requested):
Looks good, but I would suggest changing 'usage right' to '(unrestricted) right of use',
which is to my knowledge the proper legal term.

thorstenhater (approved):
I assume it might still not valid as a contract, but that's fine for our usecase.
:+1:

Issue #1271, opened 2020-12-10
SegFault on upgrade of pybind11


thorstenhater:
**Describe the bug**
Arbor SegFaults when upgrading pybind11 to at least 2.6.1 and using a simulation with
a recipe that returns a temporary `cell_description` object.

I am bringing this up as a bug since we plan to allow system-level pybind11 to be used
with arbor.

**To Reproduce**
1. Upgrade the pybind11 module to 2.6.1
2. Compile with Python
3. Run the example below.

```#!/usr/bin/env python3
from python import arbor as A

class Recipe(A.recipe):
    def __init__(self, n_cell):
        A.recipe.__init__(self)
        self.n_cell = n_cell
        self.kind = A.lif_cell()

    def num_cells(self):
        return self.n_cell

    def num_targets(self, gid):
        return 0

    def num_sources(self, gid):
        return 0

    def cell_kind(self, gid):
        return A.cell_kind.lif

    def connections_on(self, gid):
        return []

    def event_generators(self, gid):
        return []

    def get_probes(self, gid):
        return []

    def cell_description(self, gid):
        return A.lif_cell()

context = A.context(threads=40)
recipe = Recipe(n_cell=1000)
dd = A.partition_load_balance(recipe, context)
sim = A.simulation(recipe, dd, context)
```

**Additional context**
See also here: https://github.com/pybind/pybind11/issues/2721

**Analysis**
- Pybind11 added an error guard to object destruction to fix other bugs
- This guard manipulates the Python interpreter state
- Thus it needs to hold the GIL
- We release it when entering the simulation constructor (to avoid deadlocks from multithreading)
- Consequence: SegFault.

**Potential fixes**
- We acquire the GIL where needed
  - currently that is in `py_recipe_shim::get_cell_description` to cover the lifetime of the return value from `impl->cell_description`, which is a python object and _might_ have a temporary lifetime.
  - Potentially, more locations might be needed, eg `probes`, `event_generators above`.
- Pybind11 fixes this as a bug.

PR #1272, opened 2020-12-10
Fix GIL related error in pybind11 interface.

thorstenhater:
Fixes an issue in the python bindings, that appeared when upgrading to pybind11 v2.6.1, 
see https://github.com/arbor-sim/arbor/issues/1271.

This is triggered by the upgrade, but is actually incorrect in any case. Here is why:

- (PY) Simulation::init releases the GIL
- (CXX) Simulation calls through to the Recipe
- (PY) A python class deriving from (CXX) Recipe returns a python object.
- This (PY) object is then destroyed in python land, which requires the GIL.

Pybind11 only guarantees acquisition of the GIL at edges C++ -> Python.
So, we need an extra GIL acquisition here. Our usecase is actually an example
here
https://pybind11.readthedocs.io/en/stable/advanced/misc.html?highlight=gil#global-interpreter-lock-gil.

Note: We use `OVERLOAD_PURE` here, which acquires the GIL at the edge Python -> C++
for the duration of the wrapped call, *but* that does not cover the destruction of the return
value, thus triggering the bug.

I tested the fix with multiple threads on my machine, but it's quite hard to test.

T

Update:
In addition -- like mentioned in #1271 -- event generators have been fixed. None of the other
methods in recipe produce this issues. Only simulation releases the GIL, no other trampolines 
are used. This should eliminate all of the issues.

### DIFF
--- a/python/recipe.cpp
+++ b/python/recipe.cpp
@@ -25,9 +25,9 @@ namespace pyarb {
 
 // Convert a cell description inside a Python object to a cell description in a
 // unique_any, as required by the recipe interface.
-// This helper is only to be called while holding the GIL We require this guard
-// across the lifetime of the description object `d`, since this function can be
-// called without holding the GIL, ie from `simulation::init`, and `d` is a
+// This helper is only to be called while holding the GIL. We require this guard
+// across the lifetime of the description object `o`, since this function can be
+// called without holding the GIL, ie from `simulation::init`, and `o` is a
 // python object that can only be destroyed while holding the GIL. The fact that
 // `cell_description` has a scoped GIL does not help with destruction as it
 // happens outside that scope. `Description` needs to be extended in Python,
@@ -93,8 +93,7 @@ static std::vector<arb::event_generator> convert_gen(std::vector<pybind11::objec
 std::vector<arb::event_generator> py_recipe_shim::event_generators(arb::cell_gid_type gid) const {
     return try_catch_pyexception([&](){
         pybind11::gil_scoped_acquire guard;
-        auto result = convert_gen(impl_->event_generators(gid), gid);
-        return result;
+        return convert_gen(impl_->event_generators(gid), gid);
     },
         "Python error already thrown");
 }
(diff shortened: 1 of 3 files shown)


python/recipe.cpp
```diff
@@ -21,22 +23,53 @@
 
 namespace pyarb {
 
+// Convert a cell description inside a Python object to a cell description in a
+// unique_any, as required by the recipe interface.
+// This helper is only to be called while holding the GIL We require this guard
```
bcumming, line 28: Full stop in `GIL. We`

python/recipe.cpp
```diff
@@ -21,22 +23,53 @@
...
+// across the lifetime of the description object `d`, since this function can be
```
bcumming, line 29: What is `d`? Do you mean `o`?

python/recipe.cpp
```diff
@@ -58,9 +91,12 @@ std::vector<arb::event_generator> convert_gen(std::vector<pybind11::object> pyge
 }
 
 std::vector<arb::event_generator> py_recipe_shim::event_generators(arb::cell_gid_type gid) const {
-    return try_catch_pyexception(
-                [&](){ return convert_gen(impl_->event_generators(gid), gid); },
-                "Python error already thrown");
+    return try_catch_pyexception([&](){
+        pybind11::gil_scoped_acquire guard;
+        auto result = convert_gen(impl_->event_generators(gid), gid);
```
bcumming, line 96: Any reason why we don't just `return convert_gen(...);`?

bcumming (changes requested):
Great work tracking down this nasty latent bug.

There are just a couple of nit-picks.

python/recipe.cpp
thorstenhater: Good point.

python/recipe.cpp
thorstenhater: :+1:

python/recipe.cpp
thorstenhater: Yes.

bcumming (approved):

PR #1273, opened 2020-12-10
Expose py_recipe::global_properties

noraabiakar:
* Rename `pyrecipe::get_probes` to `py_recipe::probes`
* Remove `global_properties_shim` : It's easy to make mistakes with the class, as it holds a `mechanism_catalogue` and a `cable_cell_global_properties` which holds a pointer to the catalogue. This would likely have caused issues with the users.
* Expose `py_recipe::global_properties`


python/single_cell_model.cpp
```diff
@@ -270,6 +270,7 @@ void register_single_cell(pybind11::module& m) {
                 return m.traces();},
             "Holds sample traces after a call to run().")
         .def_readwrite("properties", &single_cell_model::gprop, "Global properties.")
+        .def_readwrite("catalogue", &single_cell_model::cat, "Mechanism catalogue.")
```
bcumming, line 273: The global properties hold a pointer to a catalogue.
At the moment this pointer is only set to point to the `single_cell_model::cat` when `::run` is called, so there will be a mismatch with the catalogue in the properties.

Instead, can users just set the catalogue by accessing `properties`, i.e:
```Python
model.properties.catalogue = ...;
```

bcumming (changes requested):
A few small nits.

python/cells.cpp
```diff
@@ -60,6 +60,15 @@ arb::util::unique_any convert_cell(pybind11::object o) {
                       + "\" which does not describe a known Arbor cell type");
 }
 
+// Convert global properties inside a Python object to a
+// std::any, as required by the recipe interface.
+std::any convert_gprop(pybind11::object o) {
```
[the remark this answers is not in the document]
noraabiakar: Right, I'll add that fix here too.

python/single_cell_model.cpp
noraabiakar: Yes, that's true. But we don't expose the `catalogue` of the global properties to the user any more. So `model.properties.catalogue` won't work. Instead we have `model.properties.register(cat)` which will point the global properties catalogue to `cat`.

bcumming (approved):

PR #1277, opened 2020-12-11
Changed default Python library install path. Refined Python instructions.

### LINKED ISSUE #1100: Python site-packages directory unavailable in Debian (derived) distro's
The Cmake script wishes to install the Python package to `${ARB_PYTHON_PREFIX}/lib/python${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}/site-packages`, e.g. `/usr/local/lib/python3.6/site-packages/arbor` on Ubuntu 18.04. This probably works everywhere, except Debian derived distro's, see <https://wiki.debian.org/Python#Deviations_from_upstream>.

`make install` therefore exits with an error on Ubuntu and Debian. A solution could be (and perhaps should be?) to install the package as a user package. I don't know if that conflicts with anything else, hence me bringing the issue up.

I've seen the config option to set your own ARB_PYTHON_PREFIX, but might it be nicer to set it to `python3 -m site --user-site` by default rather than `/usr/local`?

brenthuisman:
* Docs: emphasised the option of having pip install a local copy of the Arbor source.
* CMake: default Python library install path to `${PYTHON_EXECUTABLE} -c "import sysconfig; print(sysconfig.get_path('platlib'))"`. Fixes #1100.
* CMake: remove ARB_PYTHON_PREFIX option, added ARB_PYTHON_LIB_PATH option. We cannot know in general where in the prefix the packages are expected by Python.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,12 +60,12 @@ option(ARB_WITH_NEUROML "build NeuroML support library" OFF)
 option(ARB_WITH_PYTHON "enable Python front end" OFF)
 
 if (ARB_WITH_PYTHON)
-    # Path in which to install the Python module.
+    # Find path in which to install the Python module.
     find_package(PythonInterp REQUIRED)
     # Ask the above found Python where it keeps its system (platform) packages.
-    execute_process (COMMAND ${PYTHON_EXECUTABLE} -c "import sysconfig; print(sysconfig.get_path('platlib'))" OUTPUT_VARIABLE ARB_PYTHON_LIB_PATH_PYTHON OUTPUT_STRIP_TRAILING_WHITESPACE)
+    execute_process (COMMAND ${PYTHON_EXECUTABLE} -c "import sysconfig; print(sysconfig.get_path('platlib'))" OUTPUT_VARIABLE ARB_PYTHON_LIB_PATH_DEFAULT OUTPUT_STRIP_TRAILING_WHITESPACE)
     # Default to installing in that path
-    set(ARB_PYTHON_LIB_PATH ${ARB_PYTHON_LIB_PATH_PYTHON} CACHE PATH "path for installing Python module for Arbor.")
+    set(ARB_PYTHON_LIB_PATH ${ARB_PYTHON_LIB_PATH_DEFAULT} CACHE PATH "path for installing Python module for Arbor.")
 endif()
 
 #----------------------------------------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,12 +59,14 @@ option(ARB_WITH_NEUROML "build NeuroML support library" OFF)
 
 option(ARB_WITH_PYTHON "enable Python front end" OFF)
 
-# Path in which to install the Python module.
-# For Python 3.8, the module would ibe installed in
-#       ${ARB_PYTHON_PREFIX}/python3.8/site-packages
-# To install Arbor eqivalently to `pip install --user` on a linux system:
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -59,12 +59,14 @@ option(ARB_WITH_NEUROML "build NeuroML support library" OFF)
...
-# To install Arbor eqivalently to `pip install --user` on a linux system:
-#       -DARB_PYTHON_PREFIX="${HOME}/.local/lib""
-set(ARB_PYTHON_PREFIX ${CMAKE_INSTALL_PREFIX} CACHE PATH "path for installing Python module for Arbor.")
+if (ARB_WITH_PYTHON)
+    # Path in which to install the Python module.
+    find_package(PythonInterp REQUIRED)
+    # Ask the above found Python where it keeps its system (platform) packages.
+    execute_process (COMMAND ${PYTHON_EXECUTABLE} -c "import sysconfig; print(sysconfig.get_path('platlib'))" OUTPUT_VARIABLE ARB_PYTHON_LIB_PATH_PYTHON OUTPUT_STRIP_TRAILING_WHITESPACE)
+    # Default to installing in that path
+    set(ARB_PYTHON_LIB_PATH ${ARB_PYTHON_LIB_PATH_PYTHON} CACHE PATH "path for installing Python module for Arbor.")
```
bcumming, line 68: How about `ARB_PYTHON_LIB_PATH_DEFAULT` is a name for the value return by `sysconfig.get_path('platlib')`?

doc/install/python.rst
```diff
@@ -46,14 +41,28 @@ to see information about the version and enabled features:
...
+^^^^^^^^^^^^^^^^^
+
+If you want to work on Arbor's code, you can get a copy of our repo and point `pip` at the local directory:
+
+.. code-block:: bash
+
+    # get your copy of the Arbor source
+    git clone https://github.com/arbor-sim/arbor.git --recursive
+    # make your changes and then instruct pip to build and install the local source
+    pip3 install ./arbor/
```
bcumming, line 60: Have you tested this?

Does it work with pip v20.3 and earlier?

The more that I think about it, the better I think it is to advertise using CMake here
* with pip it does a fresh configure and build from scratch every time, which is very time consuming (no iterative builds)
* unless you pass `--verbose` to pip, you get no feedback about the build.

I would also add a very strong (emphasize using a text box) hint that a venv should be used, otherwise `pip3 install` won't work without sudo, and we don't want to ask people to call `sudo` (too much power)

bcumming (changes requested):
Looks good. Some questions and suggestions.

CMakeLists.txt
brenthuisman: Sure!

doc/install/python.rst
brenthuisman: I tested it (naturally :) ). Since new pip installs as user package if it does not find the system package dir writable, I consider this solved/clear. Presumably they are already familiar with pip.

Wether or not a venv is used is also up to the user imho, I don't want to prescribe venv over conda environments are whatever pops up next in the landscape.

doc/install/python.rst
```diff
@@ -142,8 +151,8 @@ with MPI support would add the following to its requirements:
     arbor >= 0.3 --install-option='--gpu=cuda' \
```
thorstenhater, line 151: Not 0.4?

brenthuisman: It is an example, so it does not matter. I however am not sure if we need the section at all. Has anyone needed this instruction so far?

bcumming (approved):

PR #1278, opened 2020-12-11
modcc : report linear test error.

### LINKED ISSUE #1275: Clipping to min/max conductance results in segfault of modcc
I want to restrict the synaptic conductance to a minimum value (0) and preferable also to some maximum (here 10).

```
diff --git a/mechanisms/default/expsyn.mod b/mechanisms/default/expsyn.mod
index 44da369..d5828d2 100644
--- a/mechanisms/default/expsyn.mod
+++ b/mechanisms/default/expsyn.mod
@@ -33,6 +33,6 @@ DERIVATIVE state {
 } NET_RECEIVE(weight) {
-    g = g + weight
+    g = max(0, min(g + weight, 10))
 }
```

Doing so results in a segfault of `modcc expsyn.mod -t cpu`:

```
Program received signal SIGSEGV, Segmentation fault.
0x00005555555e4162 in constant_simplify (e=0x0) at ../../modcc/symdiff.cpp:552
552         e->accept(&csimp_visitor);
(gdb) where
#0  0x00005555555e4162 in constant_simplify (e=0x0) at ../../modcc/symdiff.cpp:552
#1  0x00005555555a18a6 in constant_simplify (e=std::unique_ptr<Expression> = {...}) at ../../modcc/symdiff.hpp:57
#2  0x00005555555e4307 in symbolic_pdiff (e=0x5555556f1f60, id="g") at ../../modcc/symdiff.cpp:565
#3  0x000055555559a603 in Module::semantic (this=0x7fffffffdbf0) at ../../modcc/module.cpp:511
#4  0x0000555555574a77 in main (argc=1, argv=0x7fffffffe348) at ../../modcc/modcc.cpp:226
```

I'm on commit ce945f2085d1c3d7b7ca03f0128695c1a30a43df.

noraabiakar:
* Fixes segfault in modcc: can be reproduced with a synapse that use `max` or `min` expressions in the `net_receive` block: we test for linearity to check whether they can be coalesced; we cannot differentiate the expression; an error is raised but not reported; seg fault. 
* Fixes ` as_c_double` printing which does not print integers as doubles as intended. 

Fixes #1275

### DIFF
--- a/modcc/printer/cexpr_emit.cpp
+++ b/modcc/printer/cexpr_emit.cpp
@@ -23,10 +23,11 @@ std::ostream& operator<<(std::ostream& out, as_c_double wrap) {
     default:
         double val;
         std::stringstream s;
+        // If wrap.value is an int print it as X.0, this is needed for std::max and std::min
         if (std::modf(wrap.value, &val) == 0) {
             s << io::classic << std::fixed << std::setprecision(1) << wrap.value;
         } else {
-            s << io::classic << std::setprecision(1) << wrap.value;
+            s << io::classic << std::setprecision(17) << wrap.value;
         }
         return out << s.rdbuf();
     }
--- a/modcc/printer/cexpr_emit.cpp
+++ b/modcc/printer/cexpr_emit.cpp
@@ -21,8 +21,14 @@ std::ostream& operator<<(std::ostream& out, as_c_double wrap) {
     case FP_ZERO:
         return out << (neg? "-0.": "0.");
     default:
-        return out <<
-            (std::stringstream{} << io::classic << std::setprecision(17) << wrap.value).rdbuf();
+        double val;
+        std::stringstream s;
+        if (std::modf(wrap.value, &val) == 0) {
+            s << io::classic << std::fixed << std::setprecision(1) << wrap.value;
+        } else {
+            s << io::classic << std::setprecision(1) << wrap.value;
+        }
+        return out << s.rdbuf();
     }
 }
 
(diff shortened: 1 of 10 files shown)


modcc/printer/cexpr_emit.cpp
```diff
@@ -21,8 +21,14 @@ std::ostream& operator<<(std::ostream& out, as_c_double wrap) {
     case FP_ZERO:
         return out << (neg? "-0.": "0.");
     default:
-        return out <<
-            (std::stringstream{} << io::classic << std::setprecision(17) << wrap.value).rdbuf();
+        double val;
+        std::stringstream s;
+        if (std::modf(wrap.value, &val) == 0) {
+            s << io::classic << std::fixed << std::setprecision(1) << wrap.value;
```
bcumming, line 27: Why `setprecision(1)`? A comment that explains what is going on would be helpful.

bcumming (changes requested):
Thanks for the quick turnaround.
What is the result of using `min+max` in the `NET_RECEIVE` block now? Will it work, but without synapse coalescing?

noraabiakar:
> Thanks for the quick turnaround.
> What is the result of using `min+max` in the `NET_RECEIVE` block now? Will it work, but without synapse coalescing?

It will work but without coalescing, because it's not a linear mechanism. I'm going to add some unit tests to double check that.

modcc/printer/cexpr_emit.cpp
noraabiakar: `std::max` and `std::min` were complaining about comparing `value_type` and `int` because the generated expression was 
`max( 1, min(g[i_]+weight,  10));`.
We have the `as_c_double` struct which should print all numbers as doubles AFAIK, but it doesn't actually do that for `int`. `std::setprecision(1)` will print `1.0` and `10.0` instead of `1` and `10` which fixes the issue.

I'll add a comment explaining this.

bcumming: Thanks.

bcumming (approved):

PR #1282, opened 2020-12-15
Expose `(segment id)` region expression.

thorstenhater:
Expose and document the `(segment id)` region expressions.

Also fixes some issue found along the way:
- `gen-labels.py` was not up to speed with `decor` #1235.
- `inputs.py` has been hand edited, so changes where lost.
- The SWC example used an invalid format.

These have been rectified and now:
- The SWC example is read from disk and inserted into the docs verbatim.
- The same file is used to generate the image.
- The missing `decor` constructor has been added.

Update
- Generating `inputs.py` on the fly when building docs. That should prevent this issue from coming up again.


doc/concepts/labels.rst
```diff
@@ -385,6 +385,18 @@ Region expressions
 
       Branches 0 and 3, selected using ``(branch 0)`` and ``(branch 3)`` respectively.
 
+.. label:: (segment segment_id:integer)
+
+    Refer to a segment by its id. Note that segment ids depend on the construction
+    order of the morphology. While arbor's morphology loaders are stable in this regard,
+    it can become an issue when building morphologies by hand.
```
bcumming, line 392: How is this an issue when building by hand?
Segments are given ids in the order that they are added to the segment tree, and this is a strong guarantee.

bcumming (changes requested):
Good fix.

There  is one thing to change.

doc/concepts/labels.rst
thorstenhater: Hand-constructed `segment_trees` might be re-factored, producing the same morphology, but mixing up the labels.
Or is this a non-issue?

doc/scripts/inputs.py
```diff
@@ -85,8 +85,8 @@
 ysoma_morph3 = representation.make_morph(tmp)
 
 tmp = [
-    [[Segment((0.0, 0.0, 1.0), (2.0, 0.0,  1.0), 1)], [Segment((-3.0, 0.0, 0.7), (0.0, 0.0, 1.0), 2)], [Segment((2.0, 0.0, 1.0), (20.0, 0.0, 1.0), 3)]],
-]
+    [[Segment((0.0, 0.0, 1.0), (2.0, 0.0, 1.0), 1), Segment((2.0, 0.0, 1.0), (20.0, 0.0, 1.0), 3)]],
```
noraabiakar, line 88: So, `inputs.py` is generated by `gen-images.py`. The proper procedure is to modify `gen-images.py`, use it to regenrate `inputs.py` and then push the changes to both files?

thorstenhater: Yes. I would vote we move to generating it on the fly soon.

thorstenhater: I decided to fix that as well.

noraabiakar: Thanks :)

doc/concepts/labels.rst
bcumming: I don't think that this is an issue.
This region expression is not intended for use by general users, because you are referring explicitly to a detail of a particular morphology, instead of a generic rule that can be applied to multiple morphologies.
It's intended use (and where it is used internally) is in software that constructs morphologies, which is expected to understand this.

E.g. the NeuroML segments do not correspond one-to-one with segments in Arbor, because a NeuroML segment can attached half way along another segment, in which case you need to split the parent segment into two Arbor segments. The NeuroML wrapper keeps track of the mapping between NeuroML segments and the Arbor segments that they are composed of, and presents them to the user using nice labels.

This could have a "hint" box that says that refering to segments directly is not best practice, and should be avoided because such rules can't be applied to generic morphologies.

bcumming (approved):

PR #1283, opened 2020-12-15
Throw exception if probes are added to non-cable cells

### LINKED ISSUE #1274: Throw exception if probes are added to non-cable cells
Currently, if a cable cell probe is specified on a non-cable cell (e.g. lif cell), the simulation ignores the probes silently, which leads to confusion when users can't later sample the cell.

Throw an exception inside cell group constructors when probes for a different kind of cell are specified in the recipe.

noraabiakar:
Fixes #1274


bcumming (approved):
lgtm!

PR #1289, opened 2020-12-16
Added some lines to help sell the recipe concept

brenthuisman:
Now we connect to a feature we see in the API and new tutorials: why a recipe overrides functions related to connections based on gid. The 'why recipes' section now has something to build on as well.

### DIFF
--- a/doc/concepts/recipe.rst
+++ b/doc/concepts/recipe.rst
@@ -17,6 +17,12 @@ building phase to provide information about cells in the model, such as:
   * **Gap junction connections** on a cell.
   * **Probes** on a cell.
 
+Recipes are structured in such a way that during the lifetime of the simulation properties
+of each cell in the network can be retrieved rapidly by their global identifier (`gid`).
+To be able to quickly look-up properties related to the connections going in and out of a
+cell (think of synapses, gap junctions, but also probes and input events) is a particularly
+important ingredient that helps make Arbor both easily distributable over many nodes and fast.
+
 To better illustrate the content of a recipe, let's consider the following network of
 three cells:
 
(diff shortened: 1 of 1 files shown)


doc/concepts/recipe.rst
```diff
@@ -17,17 +17,23 @@ building phase to provide information about cells in the model, such as:
   * **Gap junction connections** on a cell.
   * **Probes** on a cell.
 
+Recipes are structured in such a way that during the lifetime of the simulation properties
```
bcumming, line 20: This isn't strictly true.
A recipe is only needed when instantiating (constructing) a simulation: once the simulation has been constructed, the recipe can be destructed.

```
Recipes are structured to provide a consistent interface for describing each cell in the network using their global identifier (`gid`).
```

Also, the recipe provides an interface for describing connections coming "in" to the cell. The term `input events` is vague, and isn't needed.

doc/concepts/recipe.rst
```diff
@@ -17,17 +17,23 @@ building phase to provide information about cells in the model, such as:
   * **Gap junction connections** on a cell.
   * **Probes** on a cell.
 
+Recipes are structured to provide a consistent interface for describing each cell in the
+network using their global identifier (`gid`).
+To be able to quickly look-up properties related to the connections going in and out of a
```
noraabiakar, line 22: I don't want to nitpick, but I find this sentence quite difficult to read. How about: 

`This allows the simulator to be able to quickly look-up properties related to the connections going in and out of a cell (think of synapses, gap junctions, but also probes and spike inputs); which helps make Arbor fast and easily distributable over many nodes.`

noraabiakar (approved):

PR #1290, opened 2020-12-17
Fix CMake to use correct PYTHON_EXECUTABLE

brenthuisman:
Locally `find_package(PythonInterp REQUIRED)` does not overwrite `${PYTHON_EXECUTABLE}`, but in the macos builds [on Travis](https://travis-ci.org/github/arbor-sim/arbor/jobs/750174548) it seems to go wrong. Since I can't reproduce this locally, I use this PR to test.

Of interest:
* https://gitlab.kitware.com/cmake/cmake/-/issues/19492
* https://reviews.llvm.org/D64881

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -14,13 +14,6 @@ endif()
 set(PYBIND11_CPP_STANDARD -std=c++17)
 add_subdirectory(pybind11)
 
-# Find path in which to install the Python module.
-find_package(PythonInterp REQUIRED)
-# Ask the above found Python where it keeps its system (platform) packages.
-execute_process (COMMAND ${PYTHON_EXECUTABLE} -c "import sysconfig; print(sysconfig.get_path('platlib'))" OUTPUT_VARIABLE ARB_PYTHON_LIB_PATH_DEFAULT OUTPUT_STRIP_TRAILING_WHITESPACE)
-# Default to installing in that path
-set(ARB_PYTHON_LIB_PATH ${ARB_PYTHON_LIB_PATH_DEFAULT} CACHE PATH "path for installing Python module for Arbor.")
-
 set(pyarb_source
     cable_probes.cpp
     cells.cpp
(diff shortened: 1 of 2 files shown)


bcumming (changes requested):
One approach is to move this to `python/CMakeLists.txt` *after* the pybind11 subdirectory is loaded.

* Pybind11 handles finding Python, and does it's best to work around CMake's tendency to ignore/override the `PYTHON_EXECUTABLE` variable.
* The variable `ARB_PYTHON_LIB_PATH_DEFAULT` is only consumed in one place: `python/CMakeLists.txt`
* The logic for looking at the Python environment logically should go into the Python CMake path.

brenthuisman:
I'm a bit confused: .travis.yml specifies python: 3.6 but this does not seem to be used anywhere. scripts/travis/build.sh hardcodes the PY_FLAGS (PY_FLAGS="-DPYTHON_EXECUTABLE=$pypref/bin/python3.6 ...). In the last commit that was building and testing without problems, the actually used Python version seems to be 3.8 (which is what CMake finds): https://travis-ci.org/github/arbor-sim/arbor/jobs/749650003#L2967

brenthuisman:
Seems to be passing now.

python/CMakeLists.txt
```diff
@@ -14,6 +14,13 @@ endif()
 set(PYBIND11_CPP_STANDARD -std=c++17)
 add_subdirectory(pybind11)
 
+# Find path in which to install the Python module.
+find_package(PythonInterp REQUIRED)
```
bcumming, line 18: Do we need the `find_package(PythonInterp REQUIRED)`?
Can we instead ensure that we use the `PYTHON_EXECUTABLE` variable?

bcumming (changes requested):

python/CMakeLists.txt
brenthuisman: How is `PYTHON_EXECUTABLE` set if we don't use `find_package(PythonInterp REQUIRED)`? Locally I see that is it set to the `python` in the path, but I don't see where or how. I thought you had to tell CMake this.

brenthuisman:
I do not understand why this works, but it does.

bcumming (approved):

Issue #1284, opened 2020-12-15
Different CV areas using equivalent trees 


noraabiakar:
DIfferent CV areas are generated starting from 2 seemingly equivalent morphologies and discretisations (1 CV total for both morphologies)

**Morphology 1**

1. cables:
```
(segment 0 (point -6.3 0 0 6.3) (point 6.3 0 0 6.3) 1)
(segment 1 (point 6.3 0 0 0.5) (point 206.3 0 0 0.2) 3)
```
2. CV area = `938.583` (this is the correct answer = (2 * π * 6.3 * 12.6) + (2 * π * ((0.5+0.2)/2) * 200) )

**Morphology 2**
1. cables: 
```
(segment 0 (point -6.3 0 0 6.3) (point 6.3 0 0 6.3) 1)
(segment 1 (point 6.3 0 0 6.3) (point 6.3 0 0 0.5) 3)
(segment 2 (point 6.3 0 0 0.5) (point 206.3 0 0 0.2) 3)
```
2. CV area = `1062.49`
Segment 1 is a zero length cable and shouldn't contribute to the CV area (as far as I understand), which would mean that the answer is incorrect. 

***************************************
This can be replicated using the `single-cell` [example](https://github.com/arbor-sim/arbor/blob/master/example/single/single.cpp).

Morphology 1 is the default morphology.
Morphology 2 is the generated from the following SWC file: 

```
1 1  -6.3 0.0 0.0 6.3 -1
2 1   6.3 0.0 0.0 6.3  1
3 3   6.3 0.0 0.0 0.5  2
4 3 206.3 0.0 0.0 0.2  3
```

halfflat:
It turns out there is an explanation, and it might not be a bug.

The difference between the two is that the zero-length segment contributes a surface area that represents the anulus with outer radius 6.3 and inner radius 0.5. Conceptually, it's what you would be left with if you had a non-zero length segment with these radii, and took the limit as the length went to zero. With the zero-length segment in-place, the gap between the two segments on other side is 'filled-in'.

Because of the hand-wavey limit picture above, I am inclined to think it is not a bug: the two segment trees describe distinct morphologies, one where the first and last segments are disconnected, and one where they are not.

bcumming:
The zero-length segment with varying radius has non-zero surface area and zero volume?

halfflat:
Yup! In dodgy ASCII art, the difference is between:
```
   |   |
   |   |
|         |
|         |
```
and
```
[...]

PR #1291, opened 2020-12-23
Add unit test for 0-length segment area.

halfflat:
* Add unit test that checks the expected conbtribution of an
annulus to the embed_pwlin integrated area, when a zero-length segment
with differing radii exists in the segment tree.

See issue #1284; this may or may not be working as intended, but the
test at least makes the current semantics explicit.


noraabiakar (approved):
This looks fine to me!

PR #1293, opened 2021-01-04
Remove dependence of docs on the Python module

bcumming:
Revert a change introduced in #1282 that required the python module for arbor to build the documentation.

This breaks the readthedocs build, where arbor is not available.

* Remove arbor dependence from `doc/conf.py`, reverting to the original "out of core" approach
* Fix other recently introduced issues that lead to CMake errors when configuring the documentation
  when the Python target is not configured.
  * The CMake variable `PYTHON_EXECUTABLE` wasn't set, but was required to make calls to the
     Python interpreter.
  * Instead of relying on pybind11 to find Python, explicitly detect it using `find_package(Python3 ...)`
  * Make Python3 a hard requirement when building the Python wrapper
  * And a soft requirement otherwise, that will fail only when the docs are built.


brenthuisman (approved):
Looks good. The change that relied on an importable Python package was not inadvertent, but we'll sort that out in another PR.

bcumming:
Right, it wasn't inadvertent, but it did inadvertently break read the docs. I have updated the description to reflect this.

PR #1294, opened 2021-01-05
Python API: neuron_cable_propetries->neuron_cable_properties

brenthuisman:
Python API: typofix. neuron_cable_propetries->neuron_cable_properties


noraabiakar (approved):
I need to start using a spell-checker. Thanks for catching that!

PR #1295, opened 2021-01-05
Add docs tutorial page for single_cell_recipe.py example

brenthuisman:
* Add docs tutorial page for single_cell_recipe.py example
* Adapted single_cell_recipe.py slightly
* Improved labels and fixed some broken cross-linking
* Fixed a typo here and there


doc/tutorial/single_cell_recipe_result.svg
```diff
@@ -0,0 +1,691 @@
+<?xml version="1.0" encoding="utf-8" standalone="no"?>
```
noraabiakar, line 1: If it builds exactly the same model then we get the same results and we can use the same [file](https://arbor.readthedocs.io/en/latest/_images/single_cell_model_result.svg) as the previous example.

noraabiakar (commented):
As we discussed a few weeks ago, the goal of the tutorial pages has been to introduce new concepts in each page. It seems to me that [this tutorial page](https://arbor.readthedocs.io/en/latest/tutorial/single_cell_detailed_recipe.html) exposes all of the same concepts as this proposed page. If we add this tutorial example, we will have 2 examples that take a `single cell model` and recreate it in `recipe` form.

Is there a new concept which I missed?

doc/tutorial/single_cell_recipe_result.svg
brenthuisman: True, I'll change it.

brenthuisman:
It's a bit different in scope due to the model being differently scoped, this is just about the simplest `recipe` possible.

All external users I've spoken with so far, and myself, are rather intrigued by the `recipe`, and since we had the example worked out, I think it's a great addition.I did my best to not leave a single line unexplained, because as you know, what's obvious to seasoned Arbor devs may not be that to others ;)

doc/tutorial/single_cell_recipe.rst
```diff
@@ -0,0 +1,183 @@
...
+    seaborn.relplot(data=df, kind="line", x="t/ms", y="U/mV", ci=None).savefig('single_cell_recipe_result.svg')
+
+    df.to_csv('single_cell_recipe_result.csv', float_format='%g')
+
+Step **(9)** plots the measured potentials during the runtime of the simulation.
+Retrieving the sampled quantities is a little different, these have to be accessed through the simulation object: :meth:`arbor.simulation.spikes` and :meth:`arbor.simulation.samples`.
+
+We should be seeing something like this:
+
+.. figure:: single_cell_result.svg
```
noraabiakar, line 177: No image is being rendered.

noraabiakar (changes requested):

doc/tutorial/single_cell_detailed.rst
```diff
@@ -3,7 +3,7 @@
 A detailed single cell model
 ============================
 
-We can expand on the :ref:`single segment cell example <tutorialsimplecell>` to create a more
+We can expand on the :ref:`single segment cell example <tutorialsinglecell>` to create a more
```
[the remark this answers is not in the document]
brenthuisman: Whoops!

doc/tutorial/single_cell_recipe.rst
brenthuisman: Fixed.

noraabiakar:
If the available recipe example is not properly explained then maybe we can modify it so it is? Otherwise, I think the 2 recipes are actually almost identical. I will leave it up to the other reviewers to determine whether or not they agree.

brenthuisman:
First and foremost: I wouldn't worry about duplication in the tutorials. Repetition is an elementary ingredient in study, and that's why people come to have a look at the tutorials.

A few specific reasons:

* Having the two examples demonstrates you actually don't have to implement _all_ methods of the recipe base class. Copy-paste development will happen anyway, but it's nice to not just _say_ that you don't have to implement everything.
* single_cell_detailed_recipe is a bit a more complex and therefore does not give the most limited demonstration of a recipe. Showing your minimum viable implementation is I think always a good pedagogical thing. I expect many Neuron users are going to be interested in some more examples of `recipe` (so far they have!).

Once we have a larger pile of tutorials, and hopefully had some feedback from users, we can maybe prune, but I don't think we're there yet. I have learned things making this one, that I did not with single_cell_detailed_recipe, among which which default catalog `single_cell_model` uses.

python/example/single_cell_recipe.py
```diff
@@ -1,13 +1,35 @@
...
 
-# (1) Define a recipe for a single cell and set of probes upon it.
+# (1) Create a morphology with a single (cylindrical) segment of length=diameter=6 μm
+tree = arbor.segment_tree()
+tree.append(arbor.mnpos, arbor.mpoint(-3, 0, 0, 3), arbor.mpoint(3, 0, 0, 3), tag=1)
+
+# (2) Define the soma and its center
+labels = arbor.label_dict({'soma':   '(tag 1)',
+                           'center': '(location 0 0.5)'})
+
```
halfflat, line 15: Is there a good reason to use 'center' over 'centre'? Because otherwise I'd like to stick to UK/EU spellings.

python/example/single_cell_recipe.py
```diff
@@ -1,13 +1,35 @@
...
+# (3) Create cell and set properties
+decor = arbor.decor()
+decor.set_property(Vm=-40)
+decor.paint('"soma"', 'hh')
+decor.place('"center"', arbor.iclamp( 10, 2, 0.8))
+decor.place('"center"', arbor.spike_detector(-10))
+
+# (4) Create cell and the single cell model based on it
+cell = arbor.cable_cell(tree, labels, decor)
```
halfflat, line 24: Comment on line 23 mentions 'single cell model', which probably isn't correct for this example.

python/example/single_cell_recipe.py
brenthuisman: I know you do ;) But this is what's we use in all other examples, even here (you wrote this line).

halfflat:
I don't really have a problem with duplication of material across the simple examples, because as Brent said, it's helpful pedagogically, and the common code is hopefully not so great that we will need to worry too much about them getting out of sync.

python/example/single_cell_recipe.py
brenthuisman: Correct(ed)!

python/example/single_cell_recipe.py
halfflat: Maybe we should change all of them!

halfflat: (And in this example, I did use the other spelling...)

doc/tutorial/single_cell_recipe.rst
```diff
@@ -0,0 +1,181 @@
...
+            # (4.6) Override the probes method
+            return self.the_probes
+
+        def global_properties(self, kind):
+            # (4.7) Override the global_properties method
+            return self.the_props
+
+Step **(4)** describes the recipe that will reflect our single cell model.
+
+Step **(4.1)** defines the class constructor. It can take any shape you need, but it is important to call base class' constructor. In general it's a good idea to make objects that'll be return by methods that need to be overridden members of the class. With this constructor, we could easily change the cell and probes of the model, should be want to do so. Here we initialize the cell properties to match Neuron's defaults using Arbor's built-in :meth:`arbor.neuron_cable_properties` and extend with Arbor's own :meth:`arbor.default_catalogue`.
```
noraabiakar, line 95: Can we split the sentence over several lines? 
Also: 
- `that'll be return*ed*`
- Whole sentence is ambiguous: `In general it's a good idea to make objects that'll be return by methods that need to be overridden members of the class` -> ` If the overridden methods of the class need to return an object, it may be a good idea to have the returned object be a member of the class`. 
- `should *we* want to do so`

doc/tutorial/single_cell_recipe.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Step **(4.2)** defines that this model has one cell.
+
+Step **(4.3)** defines that this model has one source.
+
+Step **(4.4)** returns the :class:`arbor.cell_kind.cable`. Make sure you don't instantiate the model with a non cable cell! Since we have just one cell, we can ignore :gen:`gid`.
```
noraabiakar, line 101: returns :class:`arbor.cell_kind.cable` (minus the _the_). 
I think instead of cautioning users against using a non-cable cell, we should explain that the reason we are returning `arbor.cell_kind.cable` is because our modeled cell is of that kind.

doc/tutorial/single_cell_recipe.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Step **(4.5)** returns the cell description passed in on class initialisation. If you would have multiple cells of different kinds, make sure the right types are returned when :meth:`arbor.recipe.cell_kind` and :meth:`arbor.recipe.cell_description` are called with the same :gen:`gid`.
```
noraabiakar, line 103: Kind of ambiguous?
'If you would have multiple cells of different kinds, make sure the right types are returned when :meth:`arbor.recipe.cell_kind` and :meth:`arbor.recipe.cell_description` are called with the same :gen:`gid`.'
->
'If we were modeling  multiple cells of different kinds, we would need to make sure that the cell returned by :meth:`arbor.recipe.cell_description` has the same cell kind as returned by :meth:`arbor.recipe.cell_kind` for every  :gen:`gid`.'

doc/tutorial/single_cell_recipe.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Step **(4.6)** returns the probes passed in at class initialisation.
+
+Step **(4.7)** returns the properties that will be applied to all cells of that kind in the model.
+
+More methods can be overridden if your model requires that, see :class:`arbor.recipe` for options.
+
+Step **(5)** instantiates the recipe with the cable cell described earlier, and a single voltage probe located at "center".
```
noraabiakar, line 111: center -> centre? depends on the outcome of the discussion with Sam. There are other occurrences in this file

doc/tutorial/single_cell_recipe.rst
```diff
@@ -0,0 +1,181 @@
...
+
+The context and domain decomposition
+------------------------------------
+
+:class:`arbor.single_cell_model` does not only take care of the recipe, it also takes care of defining how the simulation will be ran. When you create and use your own recipe, you'll need to do this manually, in the form of defining a hardware context and a domain decomposition. Fortunately, the default constructors of :class:`arbor.context` and :class:`arbor.partition_load_balance` are sufficient for this model, and is what :class:`arbor.single_cell_model` does under the hood! We'll leave the details of this subject for another tutorial.
```
noraabiakar, line 116: Split sentence over several lines. 
`will be *run*`

doc/tutorial/single_cell_recipe.rst
```diff
@@ -0,0 +1,181 @@
...
+
+.. code-block:: python
+
+    # (6) Create a default execution context and a default domain decomposition.
```
noraabiakar, line 120: We refer to it here as an execution context and before as a hardware context, I would stick to execution context in both cases, because it matches the C++ API.

doc/tutorial/single_cell_recipe.rst
```diff
@@ -0,0 +1,181 @@
...
+
+    # (7) Create and run simulation and set up 10 kHz (every 0.1 ms) sampling on the probe.
+    # The probe is located on cell 0, and is the 0th probe on that cell, thus has probe_id (0, 0).
+
+    sim = arbor.simulation(recipe, domains, context)
+    sim.record(arbor.spike_recording.all)
+    handle = sim.sample((0, 0), arbor.regular_schedule(0.1))
+    sim.run(tfinal=30)
+
+Step **(7)** instantiates the simulation and sets up the probe add in step 6. In the :class:`arbor.single_cell_model` version of the model the probe was set to record at a frequency of 10 kHz, and a simulation duration of 30 ms. Note that spike recording must be switched on. For extraction of the probe traces later on, we store a handle.
```
noraabiakar, line 140: - Split over several lines.
-`sets up the probe add*ed*`
-`version of this *example*`
-ambiguous: `the probe was set to record at a frequency of 10 kHz, and a simulation duration of 30 ms` -> `the probe frequency and simulation duration are the same.`

noraabiakar (changes requested):

python/example/single_cell_recipe.py
brenthuisman: In this example, you used both spellings in fact, which was problematic ;)

doc/tutorial/single_cell_recipe.rst
brenthuisman: I thought about this (second comment). I thought that is obvious enough, but not that nothing will prevent you from constructing this recipe with non-cable cell `cell`, so hence the warning. It would be nice to query a `cell` for its `cell_kind` and set or test in the constructor.

doc/tutorial/single_cell_recipe.rst
brenthuisman: I'll leave this for another time. I definitely don't want to make this example different from all the others.

doc/tutorial/single_cell_recipe.rst
noraabiakar: It is the responsibility of the user to not mess this up. And it's not just the `cell_kind`, it's also the `num_sources` and `num_targets` that need to match the cell description. 
We can keep the comment as is, I just think the logic is reversed: we don't build a cable cell because the kind of the cell is `cable`; the kind of the cell is `cable` because we're building a cable cell.

brenthuisman: > It is the responsibility of the user to not mess this up.

I'm not happy with this line of reasoning. Python is typed (typing was invented to help users not make mistakes), so we should warn users when we deviate and suddenly request users to repeat themselves.

noraabiakar: It may be asking too much of the user, and I support having a "recipe checker" of some sort, but it should be, yet again, up to the user to check the recipe. In any case, we do warn the user, in the form of exceptions at runtime.

As to this actual comment, I still think the logic is reversed. Warning not to use a something other than a cable cell places the importance on the cell kind: it has a cable kind, therefore we create must create a cable cell, when it should be the other way around. I would put the `cell_description` right beneath `num_cells` in the recipe, and then when describing the `cell_kind` I would refer back to the `cell_description` saying that the kind needs to be `cable` because the actual cell is a cable cell.

noraabiakar: In any case, I don't have any major objections to the comment itself. The recipe concept can be complicated to describe to the users, I think making it so that `cell_description` is the main method that returns the cell, and `cell_kind`, `num_sources` and `num_targets` provide easy access to information about the cell, without actually building the cell by calling `cell_descrition`, could make the concept easier to understand, potentially.

brenthuisman: One person interested in Arbor wanted to simulate one or some cells modelled as a cable cell in a large network of LIF cells. So use cases with multiple kinds might not be rare.

I have reworded the warning.

noraabiakar (approved):

PR #1302, opened 2021-01-08
Docs: define branch, mlocation, cv default.

brenthuisman:
* doc: Used Sphinx glossary directive to create terms that can be referred to in `concepts/morphology.rst`. Added note box on difference with Neuron.
* doc: Added note box on difference with Neuron.
* doc: Parameters added to svg generation script: turn off colors, root
* doc: Fix cpp TOC duplication, match ordering in other sections.

Termification of other documentation can come later.


doc/concepts/morphology.rst
```diff
@@ -3,56 +3,117 @@
...
-A *point* is a three-dimensional location and a radius, used to mark the cross-sectional
-centre and radius of the cable.
+Let's first define the building blocks out of which we can construct a
+proper definition for a morphology:
 
-.. csv-table::
+.. glossary::
+
+  mpoint
+    A point in 3D space has three coordinates. In Arbor, we add a fourth coordinate: radius.
```
noraabiakar, line 20: I think this is missing some info that was better explained originally, namely what the `mpoint` represents the _centre_ of a cable and the radius represents the _cross-sectional_ radius of the cable.

doc/concepts/morphology.rst
```diff
@@ -3,56 +3,117 @@
...
+
+.. csv-table:: The properties of a :term:`point <mpoint>`, in the context of cable cell morphologies.
    :widths: 10, 10, 30
+   :align: center
 
    **Field**,   **Type**, **Description**
-   ``x``,       real, x coordinate of centre of cable (μm).
-   ``y``,       real, y coordinate of centre of cable (μm).
-   ``z``,       real, z coordinate of centre of cable (μm).
+   ``x``,       real, x coordinate of center of cable (μm).
```
noraabiakar, line 27: I don't think we should be changing these. Aren't we going with the British English spelling?

doc/concepts/morphology.rst
```diff
@@ -3,56 +3,117 @@
...
+.. csv-table:: The properties of a :term:`segment`.
    :widths: 10, 10, 30
+   :align: center
 
-   **Field**,   **Type**, **Description**
+   **Field**,      **Type**,                           **Description**
    ``prox``,       :py:class:`point <arbor.mpoint>`,   the center and radius of the proximal end.
    ``dist``,       :py:class:`point <arbor.mpoint>`,   the center and radius of the distal end.
-   ``tag``,        integer, tag meta-data.
+   ``tag``,        integer (:term:`tag`),              "differentiate soma, axon, dendrite, etc."
```
noraabiakar, line 47: I think the description of the tag is quite limiting. Maybe we can say: 
"tag meta-data, can be used to classify segments of the same kind (ex: soma, dendrite, but also arbitrary use-defined groups)"

doc/concepts/morphology.rst
```diff
@@ -3,56 +3,117 @@
...
+.. figure:: ../gen-images/term_segments.svg
+  :width: 300
+  :align: center
+
+  Four segments arranged head to toe.
+
+.. glossary::
+
+  branch
+    A branch is an unbranched sequence of :term:`segments <segment>`.
```
noraabiakar, line 58: It's the longest possible sequence of unbranched `segments`. Also the image does not accurately describe what a branch is, because it implies that any group of unbranched segments can be a branch.

doc/concepts/morphology.rst
```diff
@@ -3,56 +3,117 @@
...
+.. figure:: ../gen-images/term_branch.svg
+  :width: 300
+  :align: center
+
+  A branch corresponding to the previous segments.
+
+.. glossary::
+
+  mlocation
+    A location is not a point in 3D space, but a point in the cable cell morphology's
```
noraabiakar, line 69: Do we need to say that it is not a point in 3D space?

doc/concepts/morphology.rst
```diff
@@ -3,56 +3,117 @@
...
+    coordinate system. It is defined by a specific branch, and a fraction that
```
noraabiakar, line 70: a fraction that?

doc/concepts/morphology.rst
```diff
@@ -3,56 +3,117 @@
...
-to SWC `structure identifiers <http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html>`_.
+.. figure:: ../gen-images/term_cable.svg
+  :width: 300
+  :align: center
+
+  A cable shown in black is defined as a section :term:`located <mlocation>` between 0.2 and 0.8 of the length of the previous branch.
+
+.. note::
+
+  NEURON uses different nomenclature for segments and branches. What's called a *section* in NEURON corresponds to an Arbor :term:`segment`, and sometimes an Arbor :term:`branch`. The number of *segments* (`nseg`) that NEURON uses for control over discretisation corresponds to an Arbor :term:`control volume`.
```
noraabiakar, line 93: This is not actually true. I described earlier what Neuron segments and sections refer to.
Neuron section ~= Arbor branch 
Neuron segment ~= Arbor CV

doc/concepts/morphology.rst
```diff
@@ -144,11 +206,24 @@ uses 4 segments to model the soma.
...
-and their associated tree structure.
-Every segment tree can be used to generate a unique morphology, which derives and enumerates
-*branches* from the segments.
-The branches of a morphology are unbranched cables, composed of one or more segments, where:
+.. glossary::
+
+  morphology
+    Morphologies in Arbor are modelled as a set of one dimensional :term:`cables <cable>` of variable radius,
+    joined together to form a tree. Only :ref:`cable cells <modelcablecell>` support custom
+    morphologies in Arbor. Morphologies can be create by :ref:`loading a file<morph-formats>` with a cell description,
```
noraabiakar, line 214: created*

arbor/include/arbor/cv_policy.hpp
```diff
@@ -182,7 +182,7 @@ struct cv_policy_every_segment: cv_policy_base {
 };
 
 inline cv_policy default_cv_policy() {
-    return cv_policy_fixed_per_branch(1);
+    return cv_policy_every_segment();
```
halfflat, line 185: This is a big change to the default behaviour, and will have us diverge from NEURON's default. Is there some discussion supporting this?

doc/concepts/morphology.rst
```diff
@@ -144,11 +206,24 @@ uses 4 segments to model the soma.
...
+    or by manually constructing one from a :term:`segment tree`.
+
+A segment tree and a morphology can both describe the exact same cable cell geometry,
```
noraabiakar, line 217: I find this sentence misleading: "A segment tree and a morphology can both describe the exact same cable cell geometry". 
Segment trees by themselves can not describe the cell geometry. They have to be used to create morphologies. As such I don;t think we can compare the two. Maybe what you mean to say is: "A segment tree differs from a morphology mainly in it's coordinate system ... "

doc/concepts/morphology.rst
```diff
@@ -183,6 +258,8 @@ which is illustrated along with its branches below.
   :width: 800
   :align: center
 
+  Left, the same 10 segment cable cell seen before. On the right, the associated morphology and branches.
+  Note that the proximal point of the soma is always the start (and possibly end) of a branch.
```
noraabiakar, line 262: I don't understand what this sentence means. The proximal point of a soma does not have to be the start of a branch, unless you mean the soma branch itself?

noraabiakar (changes requested):
Some of the Neuron definitions are not completely accurate (understandable as the neuron docs are not easy to read)

arbor/include/arbor/cv_policy.hpp
brenthuisman: I'm probably wrong, so I'll undo this. Yesterday @KirHarms discovered some deviation with Neuron, and we though this change would match Neuron better. Meanwhile understanding of the difference in definitions between Neuron and Arbor has improved, and new results support that.

doc/concepts/cable_cell.rst
```diff
@@ -335,16 +335,24 @@ See :ref:`modelgapjunctions`.
...
-{*x*: *h* ≤ *x* and ¬∃ *y* ∈ *B* s.t *h* < *y* < *x*}, where < and ≤ refer to the
-geometrical partial order of locations on the morphology. A fork point is
-owned by a CV if and only if all of its corresponding representative locations
-are in the CV.
+.. glossary::
 
-The set of boundary points used by the simulator is determined by a *CV policy*.
+  control volume
+    For the purpose of simulation, Arbor discretises cable cell :term:`morphologies <morphology>`
+    into control volumes, or CVs. Discretising happens through a :term:`CV policy`.
```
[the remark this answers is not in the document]
halfflat: That's almost certainly my fault. From my reading of the EU style guides, it probably should be 'discretise' and 'discretisation'. Though it almost hurts to type the latter :)

doc/concepts/cable_cell.rst
```diff
@@ -335,16 +335,24 @@ See :ref:`modelgapjunctions`.
...
+    The CVs are uniquely determined by a set of *B* :term:`mlocation` boundary points. For each non-terminal
+    point *h* in *B*, there is a CV comprising the points {*x*: *h* ≤ *x* and ¬∃ *y* ∈ *B* s.t *h* < *y* < *x*},
+    where < and ≤ refer to the geometrical partial order of locations on the morphology. A fork
+    point is owned by a CV if and only if all of its corresponding representative locations are
+    in the CV.
+
+.. Note::
+    In NEURON discretisation is controlled through splitting a NEURON section into a number of NEURON segments (`nseg`, by default to `1`). Since NEURON sections don't exactly map to Arbor :term:`segments <segment>`, there is not always a way to match the precise discretisation. Therefore results may be slightly different for otherwise identical simulations.
```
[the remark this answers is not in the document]
brenthuisman: I'll say that NRN segment is similar to an Arbor branch.

Can you elaborate on the last paragraph of your comment? Would be nice to include!

doc/concepts/cable_cell.rst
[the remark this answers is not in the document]
brenthuisman: I can set my spellchecker to en_GB or en_US, not both ;)

arbor/include/arbor/cv_policy.hpp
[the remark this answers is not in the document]
brenthuisman: See above!

doc/concepts/morphology.rst
brenthuisman: So, none of this was actually consistent, and both were used in this page and elsewhere. Let's have the discussion of which dict to use separately. In the meantime, having 1 spelling is better than 2.

doc/concepts/morphology.rst
brenthuisman: Added!

doc/concepts/morphology.rst
brenthuisman: :+1:

doc/concepts/morphology.rst
brenthuisman: 1. Can't they?
2. I'll make it "A branch is the longest possible unbranched sequence of segments" unless you are telling me branched segments exist!

doc/concepts/morphology.rst
brenthuisman: Yes. It helps enormously when defining things to say how they are the same or different from other things you're defining.

doc/concepts/morphology.rst
brenthuisman: Whoops!

doc/concepts/morphology.rst
brenthuisman: Thanks, I've updated this.

doc/concepts/cable_cell.rst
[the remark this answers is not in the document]
noraabiakar: I guess it's a question of consistency vs EU style compliance?

doc/concepts/morphology.rst
brenthuisman: 1. Why can't a segment tree describe a cell geometry?
2. You can instantiate a morphology from just a segment tree.
3. The difference are two-fold: the coordinate system and the fact that angles between branches are not defined in morphologies, or, equivalently, the branching points have no specific position in space, other than in relation to a specific distance from other branching points through any branch that links them. I've added that, it is more complete.

doc/concepts/morphology.rst
brenthuisman: Indeed. Note that there is no such thing as a soma branch: the segments of the dendrite are here included in the 'first branch' up until the first branching point. The axon here is making the proximal point of the soma _also_ the distal point of another branch. What would make it more clear?

doc/concepts/cable_cell.rst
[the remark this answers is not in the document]
noraabiakar: Neuron's **section** is similar to Arbor's branch.

I'm not very familiar with Neuron's discretization process, but I'm fairly certain we made no effort to accurately reproduce it in Arbor. If I had to guess I would say that a neuron branch (nrn: section) is made of `nseg` equally sized CVs (nrn: segments). Prior to the introduction of CV policies, we made special considerations at fork points, and the CVs at the edges of a branch were half the size of the CVs in the middle of a branch, so the CVs in Arbor and Neuron just didn't match up. This has changed quite a bit with the introduction of CV policies, which have only made the differences more pronounced: we can use several cv policies on the same cell; we can describe exactly how to split a cell into arbitrarily sized CVs; etc (@halfflat can say more about that since he designed and wrote the code!). 

In any case, the discretization not matching exactly between Arbor and Neuron is probably not the biggest source of discrepancy in results. Arbor uses truncated cones to represent branches and segments, Neuron uses their average radius to create cylinders; I've observed that that has more of an effect on the results than the discretization matching up exactly. If the issue is indeed the discretization, then by using a finer discretization in both Arbor and Neuron, both simulators should converge to the same results.

doc/concepts/morphology.rst
noraabiakar: 1. Nope they cant.
2. I worded that incorrectly ^^. I meant what you wrote.

doc/concepts/cable_cell.rst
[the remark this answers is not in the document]
brenthuisman: Sorry, I mistyped (curse all those similar quite subtly differently defined terms): I meant to write section!

Thanks for the explanation. I've added the cylinder vs cone. @halfflat if you care to elaborate on how Arbor discretises around branching points, please do :)

doc/concepts/morphology.rst
brenthuisman: 1. Then what do you propose, because to me this seems quite clear and correct. Maybe you mean that the segments should be joined?
2. OK!

doc/concepts/morphology.rst
noraabiakar: 1. Because you can't build a cable cell from a segment tree, you need to use a morphology. 
The morphology **has to** describe the same geometry as the segment tree it was built from. So saying that a segment tree and a morphology *can* describe the same geometry sounds strange: either the morphology was built from that segment tree and therefore they **have to** describe the same thing; or the morphology was built from a different segment tree in which case it makes more sense to compare the underlying segment trees or the resulting morphologies.

doc/concepts/morphology.rst
noraabiakar: In theory, there is nothing stopping the soma from being in the middle of a branch. I think you mean the `root` instead of the `proximal point of the soma.`

doc/concepts/morphology.rst
noraabiakar: I mis-worded a lot of my first comment, let me try again: 

> A branch is the longest possible unbranched sequence of segments. Also the image does not accurately describe what a branch is, because it implies that any unbranched group of segments can be a branch.

What I mean to say is that the wording and image do not accurately represent the "longest possible" part of the definition. It implies that any group of segments can be a branch as long as there is no fork in the midst. But actually, a branch needs to have a forking point (or root/terminal) at either end to be a branch. Because the image is not part of a bigger branching structure, it doesn't really illustrate that concept.

doc/concepts/cable_cell.rst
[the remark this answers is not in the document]
halfflat: If there's any doubt about which spelling to use, I think it best we follow the EU style guidelines, because then we'll be consistent across code, documentation, and reports. My personal failing in this is a tendency to use -ize endings — which is fine by the OED, but not the EU guidelines.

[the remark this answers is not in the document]
noraabiakar: I don't mind using `ise`, I'm just concerned about all our `discretize` functions not matching the documentation. It's a small issue though, so I don't mind leaving this as is.

doc/concepts/morphology.rst
brenthuisman: Right!

doc/concepts/cable_cell.rst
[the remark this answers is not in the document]
halfflat: We can definitely leave it for now, but I'll make a ticket to remind me to revisit this later.

doc/concepts/morphology.rst
brenthuisman: Actually, you can instantiate a `cable_cell` with either a `morphology` and a `segment_tree`, therefore they must (be able to) describe the same cell geometry, right?

Of course a morphology created from a segment tree will describe the same thing. Maybe the word 'can' is the source of conflict, I use it because not every segment tree and morphology describe the same geometry. I'll add a second sentence stating this.

noraabiakar: > Actually, you can instantiate a cable_cell with either a morphology and a segment_tree, therefore they must (be able to) describe the same cell geometry, right?

True, but the `segment_tree` is implicitly used to construct a morphology. 

Yes, I think the "can" is the issue here.

doc/concepts/cable_cell.rst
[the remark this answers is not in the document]
brenthuisman: Good point @noraabiakar, I prefer consistency over a particular spelling. This is also my beef with en_GB, nearly all software uses en_US spellings in APIs and documentation. Let's gather these points in the issue made by @halfflat.

doc/concepts/morphology.rst
brenthuisman: OK, if I understand you correctly, the absence of branching points makes it unclear. I thought of that, but then I could no longer have a sequence of segment>branch>cable with the same geometry without first discussing yet another concept (branching point). This sequence I want to keep. And strictly speaking, there is nothing wrong: this is the correct branch for that set of segments.

If the reader makes it to the morphology section, the morph-label-morph-fig shows what you mean. I think that is enough: we can't explain/show every detail at once.

noraabiakar: I have no problem with not modifying the image, as long as the wording is clear.

brenthuisman: The figure caption does not make any claims other than that it is the branch which corresponds to the previously shown segments, which is correct. I fail to see what's not clear.

noraabiakar: Looks fine!I hadn't checked the changes.

doc/concepts/morphology.rst
```diff
@@ -3,94 +3,164 @@
...
+  tag
+    A tag is an integer label on every segment, which can be used to define disjoint regions on cells.
+    The meaning of tag values are not fixed in Arbor, however we typically use tag values that correspond
+    to SWC `structure identifiers <http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html>`_.
+
+With these definitions, we can create proper definitions for *morphology* and *segment tree*.
+
+.. note::
+
+  NEURON uses different nomenclature for segments and branches. What's called a *section* in NEURON corresponds to an Arbor :term:`branch`. Arbor uses truncated cones to represent branches and segments, Neuron uses a weighted average radius to create cylinders (see "trapezoidal integration" in the `Neuron documentation <https://www.neuron.yale.edu/neuron/static/py_doc/modelspec/programmatic/topology/geometry.html>`_.). The number of *segments* (`nseg`) that NEURON uses for control over discretisation corresponds to an Arbor :term:`control volume`.
```
noraabiakar, line 102: Can this be split across several lines? Also, I would prefer:
"what's called a *section* in NEURON most closely corresponds to a :term:`branch` in Arbor" 
and 
"The *segments* that NEURON uses for control over discretisation ( by assigning `nseg` segments per section) most closely correspond to :term:`control volumes <control volume>` in Arbor."
It's probably better not to claim that they correspond exactly.

doc/concepts/morphology.rst
```diff
@@ -3,94 +3,164 @@
...
+  segment tree
+    A segment tree describes a morphology as a set of :term:`segments <segment>` and their connections,
+    designed to support both the diverse descriptions of cell morphologies (e.g. SWC, NeuroLicida, NeuroML),
+    and tools that iteratively construct cell morphologies (e.g. L-system generators, interactive cell-builders).
+
+Segment trees comprise a sequence of segments starting from
+at lease one :term:`root` segment, together with a parent-child adjacency relationship
+where a child segment is distal to its parent. Branches in the tree occur where a segment
+has more than one child. Furthermore, a segment can not have more than one parent.
+In this manner, neuron morphologies are modelled as a :term:`tree`, where cables that
```
noraabiakar, line 120: `tree` isn't in the glossary.

noraabiakar (changes requested):
A few more minor comments. (Also the branch has conflicts)

brenthuisman:
> A few more minor comments. (Also the branch has conflicts)

All addressed and conflicts resolved.

doc/concepts/morphology.rst
```diff
@@ -109,6 +125,7 @@ Segment trees
 .. glossary::
 
   segment tree
+  segment_tree
```
noraabiakar, line 128: Why a second segment_tree?

brenthuisman: I resolved the `m` prefix situation by adding a second term in every definition that has a typname that deviates. In the case of the segment tree, there is no `m` prefix, but the space is an `_`.

noraabiakar (approved):

PR #1306, opened 2021-01-11
More pythonic membership on mechanism_catalogue.

thorstenhater:
Introduce two minor changes to the Python API
to handle `mechanism_catalogue`s idiomatically.

Instead of 

``` Python
import arbor as A

cat = A.default_catalogue()

if cat.has('hh'):
  print("Found HH.")

for mech in cat.keys():
   print("*", mech)
```
we can now write
``` Python
import arbor as A

cat = A.default_catalogue()

if 'hh' in cat:
  print("Found HH.")

for mech in cat:
   print("*", mech)
```
which is closer to the expectations of Python users.

### DIFF
--- a/doc/python/mechanisms.rst
+++ b/doc/python/mechanisms.rst
@@ -246,9 +246,18 @@ Mechanism catalogues
     2. A further hierarchy of *derived* mechanisms, that allow specialization of
        global parameters, ion bindings, and implementations.
 
-    .. py:method:: has(name)
+    .. py:method:: __contains__(name)
 
         Test if mechanism with *name* is in the catalogue.
+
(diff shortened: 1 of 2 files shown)


brenthuisman:
Looks good! Documentation needs a few small updates, like: https://arbor.readthedocs.io/en/latest/python/mechanisms.html#catalogue

thorstenhater:
> Looks good! Documentation needs a few small updates, like: https://arbor.readthedocs.io/en/latest/python/mechanisms.html#catalogue

These are included?! Take a look at the first changed file.

brenthuisman (approved):
Yes!

PR #1308, opened 2021-01-11
add numpy as runtime req

brenthuisman:


thorstenhater:
Looks good. That requirement is already codified in `setup.py`.

thorstenhater (approved):

brenthuisman:
Yes, forgot to mention it here too.

PR #1310, opened 2021-01-13
Docs reorder cable cell

brenthuisman:
* Sync Cable Cell page names and ordering between sections (concept, python, cpp). Except cpp cable cell page. Is still one big page.
* Move CV definition across categories to consistent location.
* Move decor to separate page (cpp currently missing decor documentation) because it's a top-level concept that should be in the TOC like the other ingredients of the cable cell.
* Slight rework of Concept intro and Cable Cell pages.
* The Cable Cell page is now basically an intro, and lay out how the other pages/concepts relate to each other.

Preview: https://arbor-rtd-dev.readthedocs.io/en/latest/concepts/index.html


doc/concepts/cable_cell.rst
```diff
@@ -9,15 +9,39 @@ channels, synapses, gap junction sites, stimuli and spike detectors.
...
 Cable cells are constructed from three components:
 
-* :ref:`Morphology <morph>`: a decription of the geometry and branching structure of the cell shape.
-* :ref:`Label dictionary <labels>`: a set of rules that refer to regions and locations on the cell.
+* :ref:`Morphology <morph>`: a description of the geometry and branching structure of the cell shape.
+* :ref:`Label dictionary <labels>`: a set of definitions and a :abbr:`DSL (domain specific language)` that refer to regions and locations on the cell morphology.
 * :ref:`Decor <cablecell-decoration>`: a description of the dynamics on the cell, placed according to the named rules in the dictionary.
 
-When a cable cell is constructued the following steps are performed using the inputs:
+The decor can make reference to particular mechanisms and mechanism catalogues. More on mechanisms at :ref:`mechanisms`.
```
noraabiakar, line 16: Maybe this can be shortened and added to the bullet point of the decor? 
:ref:`Decor <cablecell-decoration>`: a description of the dynamics on the cell, placed according to the named rules in the dictionary. It can reference :ref:`mechanisms from mechanism catalogues. <mechanisms>

doc/concepts/cable_cell.rst
```diff
@@ -9,15 +9,39 @@ channels, synapses, gap junction sites, stimuli and spike detectors.
...
 
-1. Concrete regions and locsets are generated for the morphology for each labeled region and locset in the dictionary
+
+.. graphviz::
+  :align: center
+  :caption: The three main ingredients of the cable cell: morphology, labels, decor. A defined morphology enables referring to parts of it through the label DSL, references which can be stored as named labels. A decor places (or paints) properties such as mechanisms onto precise places (labels) on the morphology. Straight lines represent inputs, dashed lines references.
```
noraabiakar, line 21: The sentence is a bit unclear. Maybe something like this? 
`The three main ingredients of the cable cell: morphology, label dictionary, decor. Regions and locations of the morphology can be identified using a DSL (maybe link to the DSL?). The definitions can be stored with a label in the label dictionary. The regions and locsets can then be referred to using those labels when painting and placing dynamics and properties using a decor. Straight lines represent inputs, dashed lines references.`

doc/concepts/index.rst
```diff
@@ -17,11 +15,43 @@ To be able to simulate a model, three basic steps need to be considered:
 2. Define the computational resources available to execute the model;
 3. Initiate and execute a simulation of the recipe on the chosen hardware resources.
 
-The python front-end further abstracts away some of these steps for single cell models, where users only need to
+
+.. graphviz::
```
noraabiakar, line 19: I'm unsure about this graph:
1. It shows some of the methods of a recipe, but not all.
2. It doesn't show any info about what goes into a context or a domain decomposition. Why is the recipe special? 
3. The nodes of the graph are identical, which indicates that their contents are similar types. But the nodes that make up a recipe aren't objects, they are method definitions. Maybe they can have a different color to indicate that?

doc/cpp/cable_cell.rst
```diff
@@ -8,6 +8,8 @@ Cable cells
    has changed significantly; some of the documentation below is
    out of date.
 
+   The C++ cable cell documentation should have the same structure as the Python cable cell documentation.
```
noraabiakar, line 11: Why does this need to go in the docs?

doc/python/decor.rst
```diff
@@ -0,0 +1,402 @@
...
+        :param index: the unique index of the placement.
+        :type index: int
+        :rtype: tuple(int, int)
+
+
+.. py:class:: ion
+
+    properties of an ionic species.
+
+.. _pycablecell-probes:
```
noraabiakar, line 252: This is relevant for both cpp and python `decor.rst`. The probes are not part of the decor. They should either have their own page or stay in the cable_cell page; moving them to the `decor` page is likely to be confusing.

noraabiakar (changes requested):
I think this restructuring is a good idea, but I don't think the probe docs belong under the decor page. 
Having a concepts/decor page is a good start, once this is merged we should really consider refactoring this page.

doc/concepts/index.rst
brenthuisman: 1. You are right. I'm seeking a compromise between showing everything, and then ending up with a figure that's no longer helpful for getting an overview. I also didn't want to add ellipsis nodes (seemed like more clutter). The caption 'Outline' was meant to convey this is not a complete reference.
2. The recipe is special because it contains the elements of what most neuroscientists expect to see in an overview over a morphologically detailed simulator: where are the cells and the network? The recipe has so far been the structure they are most interested in/confused by when I speak them, so I think it's a good idea to tell them up-front that that's where the biology goes; what they usually describe as a model.
    'context' should be 'execution context', which might make it clearer what it is about. However, 'domain decomposition' is perhaps a bit too cryptic. I want to both match what we actually call these concepts later on though to not add to the new-nomenclature overload. What would you think is a good solution? Maybe we could circle 'recipe' and it's ingredients and annotate it with 'neuroscience' or so, and 'execution context+'domain decomp.' with 'computational' or so?
3. Cell descriptions such as `cable_cell` are objects too no? The 'cell kind' and 'number of cells' not, you're right. The description below could have a word or two on cell desc+kinds. 'number of cells' is self-descriptive enough I'd say.

doc/cpp/cable_cell.rst
brenthuisman: Since there was already a comment about being out of date, I thought a hint as to what needs to happen wouldn't hurt. Can be removed if you prefer.

doc/python/decor.rst
brenthuisman: Actually, I think they should have their own page, what do you think?

doc/concepts/index.rst
brenthuisman: Removed for now.

doc/python/decor.rst
```diff
@@ -0,0 +1,246 @@
...
+        :rtype: int
+
+    .. method:: discretization(policy)
+
+        Set the cv_policy used to discretise the cell into control volumes for simulation.
+
+        :param policy: The cv_policy.
+        :type policy: :py:class:`cv_policy`
+
+.. py:class:: cable_cell
```
noraabiakar, line 188: Shouldn't the `cable_cell` class API be in `python/cable_cell.rst`? I think there's some duplication going on.

noraabiakar (changes requested):
Looks like there's some duplication of the Python `ceble_cell` API:
https://arbor-rtd-dev.readthedocs.io/en/latest/python/cable_cell.html#arbor.cable_cell and 
https://arbor-rtd-dev.readthedocs.io/en/latest/python/decor.html#arbor.cable_cell

doc/python/decor.rst
brenthuisman: And resolved!

noraabiakar (approved):
Looks good to me!

PR #1312, opened 2021-01-15
arborio: read decor and cable cell parameters from JSON

### LINKED ISSUE #1332: JSON Format
Proposal for a JSON format representing a set of inputs that can be easily parsed by Arbor and NEURON to generate equivalent single-cell models. We propose 2 JSON formats: one for describing the defaults of the system, to be applied at the model level and including most of the parameters in a `cable_cell_parameter_set`; and another for describing the decorations of a cell and including a subset of the `decor` description. 

We can choose to stop after implementing this issue. It's enough to ensure basic interoperability between Arbor and NEURON. Or we can extend the JSON format to cover more Arbor-exclusive extensions which are detailed in issue #1333.

`defaults.json`:

```
{
  "version" : 1,
  "type" : "default-parameters",
  "data":
  {
    "axial-resistivity": 35.4,
    "init-membrane-potential": -60.0,
    "temperature-K": 6.3,
    "membrane-capacitance": 0.01,
    "ions": {
      "ca": {
        "init-int-concentration": 5e-05,
        "init-ext-concentration": 2.0,
        "init-reversal-potential": 132.4579341637009,
        "reversal-potential-method": "constant"
      },
      "k": {
        "init-int-concentration": 54.4,
        "init-ext-concentration": 2.5,
        "init-reversal-potential": -77.0,
        "reversal-potential-method": "nernst"
      },
      "na": {
        "init-int-concentration": 10.0,
        "init-ext-concentration": 140.0,
        "init-reversal-potential": 50.0,
        "reversal-potential-method": "nernst"
      }
    }
  }
}
```
This JSON file would be parsed into a `cable_cell_parameter_set`. It can represent everything in the class except the discretization policy. `version` and `type` fields are mandatory.

`decor.json`:

```
{
  "version" : 1,
  "type" : "decor",
  "data" :
  {
    "global":
     {
        "axial-resistivity": 35.4,
        "membrane-capacitance": 0.01,
        "ions": {
          "ca": { "init-reversal-potential": 132.4579341637009 }
        }
     },
    "local":
    [
      {
        "region": "dend",
        "membrane-capacitance": 0.02,
        "ions": { "k":  {"init-reversal-potential": -85} }
      },
      {
        "region": "soma",
        "membrane-capacitance": 0.02
      }
    ],
    "mechanisms":
    [
      {
        "region": "dend",
        "mechanism": "pas",
        "parameters": {"e": -75, "g": 3e-5}
      },
      {
        "region": "soma",
        "mechanism": "CaDynamics_E2",
        "parameters": {"gamma": 0.000609, "decay": 210.485284}
      }
    ]
  }
}
```

This JSON file would be parsed into a `decor`. It can represent all `defaultables`, except the discretization policy; all the `paintables`; and none of the `placeables`. It only accepts labels of regions which have to be defined in a `label_dictionary`

noraabiakar:
* Adds to arborio the functionality to read and write the `cable_cell_parameter_set` and `decor` from and to JSON files (similar to the ones seen [here](https://github.com/bcumming/cnrn-arb-cell/tree/master/common))
* Exposes the new functionality, and relevant classes to the python interface. 
* Adds a new python example.
* Adds unit tests.
* Adds documentation.

Closes #1332.


doc/formats/default_parameters.rst
```diff
@@ -0,0 +1,84 @@
...
+   axial resistivity                         Ra                         Ω·cm
+   membrane capacitance                      cm                         μf⋅cm⁻²
+   initial internal concentration (per ion)  internal-concentration     mM
+   initial external concentration (per ion)  external-concentration     mM
+   initial reversal potential (per ion)      reversal-potential         mV
+   reversal potential method (per ion)       method                     --
+   CV policy                                 --                         --
+   ========================================  =========================  =========
+
+Of all possible default parameters, all but the CV policy can be specified using the
```
thorstenhater, line 28: Is there a reason for the exclusion of the discretisation policy (I know that NRN does not support it, but that's
no reason for us, right?)

doc/formats/morphology.rst
```diff
@@ -0,0 +1,142 @@
...
+~~~
+
+Arbor supports reading morphologies described using the
+`SWC <http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html>`_ file format.
+
+SWC files may contain comments, which are stored as metadata. And a blank line anywhere in the file is
+interpreted as end of data.
+
+The description of the morphology is encoded as a list of samples with an id,
+an `x,y,z` location in space, a radius, a tag and a parent id. Arbor parses these samples, performs some checks,
```
thorstenhater, line 18: Please add 'separated by whitespace'. At least I keep forgetting.

doc/formats/morphology.rst
```diff
@@ -0,0 +1,142 @@
...
+   :linenos:
+
+Samples 1 and 2 will form the soma; samples 1 and 3 will form the axon, connected to the soma at the proximal end;
+samples 2 and 4 will form the dendrite, connected to the soma at the distal end. The morphology will look something
+like this:
+
+.. figure:: ../gen-images/swc_morph.svg
+   :width: 400
+   :align: center
+
```
thorstenhater, line 61: Maybe it's worth adding a note here saying something to the effect of this
- morphologies are abstract things, *not* 3d physical objects ie neither the given positions nor relative locations have any effect on the simulation
- we _render_ them like this for demonstration.

doc/formats/morphology.rst
```diff
@@ -0,0 +1,142 @@
...
+This interpretation also expects that samples have the same tag as their parent samples, with the exception of samples
+that have the soma sample as a parent. In this case, when a sample's parent is the soma, no *segment* is created
+between the 2 samples; instead there is a gap in the morphology (represented electrically as a zero-resistance wire).
+Samples with the soma as a parent start new segments, that connect to the distal end of the soma if they are dendrites,
+or to the proximal end of the soma if they are axons or apical dendrites. Only axons, dendrites and apical dendrites
+(tags 2, 3 and 4 respectively) are allowed in this interpretation, in addition to the spherical soma.
+
+Finally the Allen institute interpretation of SWC files centres the morphology around the soma at the origin (0, 0, 0)
+and all samples are translated in space towards the origin.
+
```
thorstenhater, line 78: Would the difference in interpretation actually show up in a way visible in the renderings? If so, maybe add a three way comparison.

python/example/single_cell_json.py
```diff
@@ -0,0 +1,75 @@
...
+swc_file      = sys.argv[3]
+
+# Load inputs from file
+defaults = arbor.load_default_parameters(defaults_file)
+decor    = arbor.load_decor(decor_file)
+morph    = arbor.load_swc_arbor(swc_file)
+
+# Test round trip of default parameter descriptions and decor.
+arbor.store_default_parameters(defaults, "defaults_out.json")
+arbor.store_decor(decor, "decor_out.json")
```
thorstenhater, line 21: This has no effect, as nothing actually is tested. Suggestion: drop this.

arborio/jsonio.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+jsonio_decor_mech_set_error::jsonio_decor_mech_set_error(const std::string& reg, const std::string& mech, const std::string& err):
+    jsonio_error("Decor: Error painting mechanism \"" + mech + "\" on region \"" + reg + "\": " + err)
+{}
+
+jsonio_json_parse_error::jsonio_json_parse_error(const std::string err):
+    jsonio_error("Error parsing JSON : " + err)
+{}
+
+arb::cable_cell_parameter_set load_cable_cell_parameter_set(nlohmann::json& params_json) {
```
thorstenhater, line 61: I am wondering if this here

https://github.com/nlohmann/json#basic-usage

is not a better option?

thorstenhater (changes requested):
Hi,

I like this, but here are some things that might be worth discussion (offline, potentially)
- I would like to see a way to ser/de `label_dict`s in this way.
- Potentially, a (Python) method to consume a bundle of files (delivered by tar or directory) to instantly instantiate a cell.
- I am not really happy with some names proposed in the format (as expressed on the format proposal), eg `rL` should be `axial-resisitvity`. Reasons include that this removes ambiguity for the human creating these files and is consistent with the other parameter namings. That NEURON does it another way is not an argument for me (nor that our own Python API spells it differently).

T

doc/formats/default_parameters.rst
noraabiakar: We can do it. Each `cv_policy` expects a different set of parameters, so we'd have to decide how to represent that, which shouldn't be too challenging, I just limited the scope of this PR of the agreed upon json format, we can always add `cv_policy` support later.

doc/formats/morphology.rst
noraabiakar: Sure. BTW, this whole file was copied with no changes. I'll add a separate note for other reviewers.

doc/formats/morphology.rst
noraabiakar: Some of it would, some wouldn't. The main difference is in what can/can't be represented.
We can add some form of visual representation of the difference, but I'd prefer to keep it out of this PR. As I mentioned, this file is just copied from elsewhere in the code, so none of this file is new.

python/example/single_cell_json.py
noraabiakar: Maybe I will add the lines to a different example to illustrate the functionality.

arborio/jsonio.cpp
noraabiakar: I'll check it out, thanks!

arborio/jsonio.cpp
```diff
@@ -0,0 +1,317 @@
...
+#include "arborio/jsonio.hpp"
+#include "json_helpers.hpp"
+
+namespace arb {
+void to_json(nlohmann::json& j, const cable_cell_parameter_set& params) {
+    if (auto tempK = params.temperature_K) j["celsius"] = tempK.value() - 273.15;
+    if (auto Vm = params.init_membrane_potential) j["Vm"] = Vm.value();
+    if (auto Ra = params.axial_resistivity) j["Ra"] = Ra.value();
+    if (auto cm = params.membrane_capacitance) j["cm"] = cm.value();
+    for (auto ion: params.ion_data) {
```
thorstenhater, line 18: Since I cannot advertise this feature enough ;)

```
for  (const auto&[name, data]: params.ion_data) {
```

In any case, this should be a `const &` to avoid a copy.

arborio/jsonio.cpp
```diff
@@ -0,0 +1,317 @@
...
+    // Local
+    if (auto locals_json = find_and_remove_json<std::vector<nlohmann::json>>("local", j_copy)) {
+        for (auto l: locals_json.value()) {
+            auto region = find_and_remove_json<std::string>("region", l);
+            if (!region) {
+                throw arborio::jsonio_decor_local_missing_region();
+            }
+            std::string reg = region.value();
+
+            // if the region expression does not start with an open parenthesis,
```
thorstenhater, line 147: We should add documentation here and in the format proposal clarifying that 
regions are in ASCII,  _names_ must start with a letter (or number?), and that 
expressions of type `(...)` will be turned into regions via the s-exp format. Also, 
what happens if the first character is `"`?

Sorry for not bringing that up earlier.

doc/formats/morphology.rst
thorstenhater: Sure thing, that was mainly my curiosity, but let's keep it in mind for another pass on the docs.

thorstenhater (changes requested):
A few comments.

arborio/jsonio.cpp
noraabiakar: I will make the change.

arborio/jsonio.cpp
noraabiakar: I attempted to do that in the [docs](https://github.com/arbor-sim/arbor/pull/1312/files?file-filters%5B%5D=.rst#diff-70cb3593cb83f3a29cd7d8710d8e01cccfb180a4cc99edbc49623db1fc19d64cR70). But I can add a comment here too. 

I am considering all region strings that don't start with an open parenthesis to be labels, which should therefore be surrounded by quotations. that should take care of all string labels that don't start with a parenthesis. So, if a user chooses to name their region `(my-region`, that's not going to work (the s-expression parser with raise an exception). I wanted to discuss this with @bcumming, I'd rather we force the double quotation in the json file than add them implicitly here, but that'll make the json file less "standard".

thorstenhater: That's my issue here, too, I do not think here's a good, general solution.
Maybe this could help:

- The importer checks whether a region is quoted or is an expression.
- If neither an error is raised.
- The exporter adds quotes when needed. That take care of the round trip.
- Document the need for quotes in the format spec.

Alternatively, the format could include
```
...
"region": {"kind": "s-exp", "value": "(tag 2)"}
...
"region": {"kind": "label", "value": "axon"}
...
```

noraabiakar: You're saying that labels should be quoted, so in the JSON file the user would write:
```
"region" : "\"soma\""
```
right? 
I don't have an issue with that except it's not very standard. Originally this JSON format was proposed as something that can be read by Arbor and Neuron. We'd be forcing an arbor-specific format in this case by requiring double quotes. 

The exporter already only prints out s-expressions. So never:
```
"region" : "\"soma\""
```
Always: 
```
"region" : "(region \"soma\")"
```
Which is potentially a problem. 

I'd like to discuss the specifics of this format with the rest of the team. All these questions are why I was uncertain about the format (and whether this PR would ultimately get merged) in the first place.

thorstenhater: Absolutely, I just had doubts while seeing it in code. To add a third version for consideration:

It has been a long standing point whether to allow region literals over labels in paint/place.
Maybe it is in general better to not do it and rely on the label_dict? We could add a section
to the format specification that deals with labels and allows non-portable things.

doc/formats/default_parameters.rst
bcumming: Agreed.
@noraabiakar, can you provide a link to the ticket where you discuss extensions to what we have here (e.g. how to described discretisations).

doc/python/cable_cell.rst
```diff
@@ -249,6 +248,153 @@ Cable cells
 
     properties of an ionic species.
 
+.. _pycablecell-props:
```
bcumming, line 251: Thanks for adding these docs!

python/param_io.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <arborio/jsonio.hpp>
+
+#include "error.hpp"
+#include "strprintf.hpp"
+
+namespace pyarb {
+void register_param_loader(pybind11::module& m) {
+    m.def("load_default_parameters",
+          [](std::string fname) {
```
bcumming, line 18: Instead of implementing these as lambdas, they could be implemented above as stand alone functions (with return types clearly defined).

python/param_io.cpp
```diff
@@ -0,0 +1,114 @@
...
+                  arb::check_global_properties(G);
+              }
+              catch (std::exception& e) {
+                  throw pyarb_error("Error loading default parameters from \"" + fname + "\": default parameter check failed : " + std::string(e.what()));
+              }
+              return params;
+          },
+          "Load default model parameters from file.");
+
+    m.def("load_decor",
```
bcumming, line 44: All of these methods are either reading or writing JSON?

If so, then at least the name should reflect that.

bcumming: It is a pity that the Python JSON representation and our C++ nlohmann are not the same: then we could pass JSON objects instead of file names. We could then just construct the objects from json representations.

bcumming (changes requested):
Nice work.

I have one large-ish suggestion that we could debate.

We could consider adding a bit more structure the high-level json data structures. Specifically, a set of global parameters would have a field
`"type": "global-parameters"`
A decor:
`"type": "decor"`
A label dictionary (when we add it)
`"type": "label-dict"`

When all of the high level descriptions have "types", you can then write a generic reader that will first look at the type, then parse it appropriately. This has the benefits of doing things like
```python
decor = arbor.io.json_load('decor.json')
labels = arbor.io.json_load('labels.json')

arbor.io.json_write('decor.json', my_decor)
```

doc/formats/default_parameters.rst
noraabiakar: It's issue #1322. I'll add it to the PR description as well.

arborio/include/arborio/jsonio.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+namespace arborio {
+
+struct jsonio_error: public arb::arbor_exception {
+    jsonio_error(const std::string& msg);
+};
+
+// Error parsing JSON
+struct jsonio_json_parse_error: jsonio_error {
+    explicit jsonio_json_parse_error(const std::string err);
```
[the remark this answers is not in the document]
noraabiakar: I hadn't noticed that. I'll make the changes. Thanks!

python/param_io.cpp
noraabiakar: I'll make the change.

python/param_io.cpp
noraabiakar: I'll fix the names. 
As for the second point, do you mean doing something like: 
```
import json
with open(sys.argv[1]) as json_file:
    arbor.load_json_decor(json.load(json_file))
```
I didn't actually try it, but it seems a bit verbose.

noraabiakar:
I think adding the `type` field is a good idea. I'll also add a `version` field as Sam suggested.

python/param_io.cpp
bcumming: It is a bit more verbose than just passing in a file name.
But it is more "typey", because you don't need the `json` in the function name.
`load_decor` would be enough, and could take multiple different types.

It won't work as is, though, because we would have to teach our parser to understand Pythons JSON representation.

arborio/jsonio.cpp
noraabiakar: The solution I went with for this PR may not be the best. At the moment, if the json file contains: 
```
"region" : "soma"
```
The code will check the first character of the string "soma"; notice that it's not a `(`; assume it's a label; and add quotes around it so it can be parsed correctly by arbor.

If the json file contains instead: 
```
"region" : "(my-region"
```
The code will assume that's an s-expression instead of a label and parser will fail. 

As (I think) you both suggest, maybe the best thing to do is just force the double quotes as we do everywhere else in the library. So: `"region" : "soma"` would have to be replaced by `"region" : "\"soma\""`. A neuron parser will need to remove the quotes, but that should be fine for the sake of compatibility.

python/param_io.cpp
noraabiakar: I see, I agree that that would be nice to have.

arborio/jsonio.cpp
noraabiakar: Actually my solution is too special case-y for multiple reasons. I think double-quoting is the best solution now that I've taken a closer look at our s-expression parser.

noraabiakar:
Issue #1332 is addressed in bcumming/cnrn-arb-cell#4
The implementation will not be part of the main arbor repository. Instead we will focus our efforts on designing an Arbor-specific format #1336

PR #1313, opened 2021-01-15
Tutorials: ring network

brenthuisman:
* Tutorial ring network added. A few typos fixed, crosslinks fixed.
* MPI tutorial added.
* Improved SVG generation, added separate label scaling, increased default label size.
* Bugfix: didnt generate cable region in gen-labels.py used in `concepts/morphology.rst`.
* Termified Interconnectivity page.
* Termified probe nomenclature page. Will be moved in [this PR](https://github.com/arbor-sim/arbor/issues/1353) from python/simulation to concepts/probes_samplers.


doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,337 @@
...
+3. Define the **execution context** of the model: a description of the underlying system
+   on which the simulation will run.
+4. Define the **meter manager** that will record execution time and memory consumption.
+5. Define the **domain decomposition** of the network: how the cells are distributed on
+   the different ranks of the system.
+6. Define the **simulation**.
+7. **Run** the simulation.
+8. Collect and visualize the **results**.
+
+The cell
```
bcumming, line 33: Could we simplify this by using a cell model developed in an earlier tutorial? We can change the existing ring example, or create a new one, and we don't need to make a fully stand alone page in the docs for each example.

Then you could say something like
```
The ring model that we will develop will have 4 cells based on the single cell model developed earlier <link>.
We wrap this in a function that returns the cell.
```
Then you could have a code sample that shows the function that generates the cell. That could later
be called in the recipe?

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,337 @@
...
+
+   # Attach a spike detector with threshold of -10 mV.
+   decor.place('"root"', arbor.spike_detector(-10))
+
+   cell = arbor.cable_cell(tree, labels, decor)
+
+The recipe
+**********
+
+Our objective is creating a network of :ref:`connected <interconnectivity>` cells, so we need to create a :class:`arbor.recipe` accommodating multiple cells. Besides multiple cells, the recipe is also where the connections between cells is defined.
```
bcumming, line 103: To create a model with multiple connected cells, we need to use a recipe that describes the model.
The recipe is where the different cells and the connections between them are defined.

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,337 @@
...
+
+Before we go there, let's first create a function that returns the above cell. This tutorial's objective is to demonstrate creating the network after all. Simply wrap the above code in a function definition, and let's add the imports while we're at it:
+
+.. code-block:: python
+
+   import arbor
+   import pandas, seaborn
+   from math import sqrt
+
+   def make_cable_cell(gid):
```
bcumming, line 113: See my comment above.

Maybe the `import pandas, seaborn` is a distraction at this stage, where they are not used.

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,337 @@
...
+      def probes(self, gid):
+         return [arbor.cable_probe_membrane_voltage('"root"')]
+
+      def global_properties(self, kind):
+         return self.props
+
+The execution context
+*********************
+
+The execution context contains all system-specific information needed by the simulation: it contains the
```
bcumming, line 181: An execution context describes the hardware resources on which the simulation will run. Then provide a link to the appropriate documentation.

It also does not hold GPU attributes.
It `It contains the thread pool used to parallelise work on the local CPU, and optionally describes GPU and MPI resources to use`.

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,337 @@
...
+capabilities of the system and the way it was configured. Most Arbor objects can
+be printed out to get obtain some information about its configuration.
+
+.. code-block:: python
+
+   # (10) Set up the hardware context
+   context = arbor.context(threads=12, gpu_id=None)
+   print(context)
+
+The meter manager
```
bcumming, line 200: Maybe the meter manager is a distraction? If this is the first time a user is learning about recipes, then it might be best to not introduce any distractions from that core mission.

We can include discussion about meters to a later example, that takes this one, and extends it with performance measurement. Don't be afraid to modify the original example.

Originally, there was a "tutorial", that was separate from the examples, and was going to be extended step by step guide users through the process of going from a single cell to a network model running with MPI.

And I think the approach of having a clear path, that starts simple, and builds up that way with layer on layer is the way to go.

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,337 @@
...
+   # (11) Set up and start the meter manager
+   meters = arbor.meter_manager()
+   meters.start(context)
+
+   # (12) Instantiate recipe
+   ncells = 4
+   recipe = ring_recipe(ncells)
+   meters.checkpoint('recipe-create', context)
+
+The domain decomposition
```
bcumming, line 223: The domain decomposition is not required in the Python interface, and I fear that it is a bit of a distraction here.

How about the following idea: don't discuss hardware contexts or domain decomposition in this example. Mention that the next example/tutorial will cover the topic of GPUs, MPI, threads and distributing work over them.

Then that example can take our simple ring model, and add these features, once the reader has comfortably set up their first recipe example?

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,273 @@
+.. _tutorialsinglecellallen:
```
bcumming, line 1: This isn't a good candidate for a tutorial. Maybe it can go in "examples", not "tutorials", with a simple comment at the top of the file that says what the example shows.

The Allen SDK is pretty messy, and this was more a proof of concept. The topics of how to load morphologies and decorate them are better covered without that noise.

python/example/single_cell_allen.py
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env python3
```
bcumming, line 1: This is a good example, but not a good tutorial.

Maybe we have to go back to the structure of a `tutorial` path, which has the tutorial code that leads the user through the process from a single cell to a full model.

Once a user has learnt the interface from the tutorial, or by reading the concepts docs, they can read an example like this that shows the messy process of loading inputs from something like Allen SDK.

bcumming (changes requested):
Good work on the docs.

My feedback is that I think we can split the example code in Python into two paths:
* `tutorial` which has the python scripts that are used to demonstrate a gentle step by step approach from a single cell model to a distributed network.
* `example` which has additional examples, that illustrate how to use the API to do things like load Allen SDK models. These don't need a dedicated page with a full walk through in the docs.

doc/tutorial/network_ring.rst
brenthuisman: I've thought of having a small library where for instance this and the Allen fit.json loader could be parked.

In this case, we're creating the first network in the tutorial section, so a simple cell seems a good choice.

doc/tutorial/network_ring.rst
brenthuisman: This is mostly copied from `single_cell_detailed_recipe.rst:L287` so I've made the same modification there.

doc/tutorial/network_ring.rst
brenthuisman: I did remove the MPI stuff from this example already. I did not find the meter a distraction. I agree that we should keep things orthogonal-ish, but showing a hint of a feature that will be fully described in a future example/tutorial if it is easy enough to understand is OK with me. The link to the class doc is there (not sure if it works already, but if not it's coming up!). The Arbor meter manager is simple to use, and checkpointing to record runtime and memory consumption is something anyone will understand I think.

Are you saying that I should split this out now? So, have a tutorial/example with just the meter added to `network_ring.rst`?

doc/tutorial/network_ring.rst
brenthuisman: I can break it out, together with the manager. I actually started on the MPI version of this network, so I'll introduce it there.

doc/tutorial/single_cell_allen.rst
brenthuisman: On the subject of tutorials vs examples, I want to have every example as a tutorial for now. It's what people will see first, and running an Allen model is a close to 'real world' use that we have right now.

The messy part is in parsing the fit.json, which is actually not really that messy (it unfortunately does not include everything, but all that it doesn't I made it a point to describe).

Having this json -> decor converter (eventually) live somewhere else might be a good idea?

python/example/single_cell_allen.py
brenthuisman: See above.

brenthuisman:
The Allen demo was removed. The network tutorial was rewritten to be 100% complete and to not expand on concepts already introduced in earlier tutorials. Some more termification and minor updates across concepts pages.

brenthuisman:
Since it wasn't much more in terms of concepts, I added an MPI demo that distributes the network_ring tutorial over the default MPI communicator, and explains how to handle the different execution.

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,261 @@
...
+
+The two endpoints are of type :class:`arbor.cell_member`, and can be initialized with a ``(gid,index)`` tuple.
+The cells have one synapse (step **3**), so the target endpoint has the 0th index. The cell has one
+spike generator (step **4**), so its source index is also 0.
+
+Lastly, we must inform the recipe how many connections we have on each cell (``gid``).
+:func:`arbor.cable_cell.num_targets` and :func:`arbor.cable_cell.num_sources` must be set to 1: each cell has one
+source and one target endpoint.
+
+Step **(9)** attaches an :py:class:`arbor.event_generator` on the 0th target (synapse) on the 0th cell; it is
```
noraabiakar, line 113: instead of `; it is connected to the ..`, I would say `this means it is connected to the ...`

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,261 @@
...
+default domain decomposition. You can print the objects to see what defaults they produce on your system.
+
+Step **(13)** sets all spike generators to record using the :py:class:`arbor.spike_recording.all` policy.
+This means the timestamps of the generated events will be kept in memory. Be default, these are discarded.
+
+In addition to having the timestamps of spikes, we want to extract the voltage as a function of time.
+
+Step **(14)** sets the probes (step **10**) to measure at a certain schedule. This is sometimes described as attaching a :term:`sampler` to a :term:`probe`. :py:func:`arbor.simulation.sample` expects a :term:`probe id` and the desired schedule (here: a recording frequency of 10 kHz). Note that the probe id is a separate index from those of :term:`connection` endpoints; probe ids correspond to the index of the list produced by :py:func:`arbor.recipe.probes` on cell ``gid``.
+
+:py:func:`arbor.simulation.sample` returns a handle to the :term`samples <sample>` that will be recorded. We store these handles for later use.
```
noraabiakar, line 195: typo: `:term:`

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,261 @@
...
+   handles = [sim.sample((gid, 0), arbor.regular_schedule(0.1)) for gid in range(ncells)]
+
+   # (15) Run simulation
+   sim.run(100)
+   print('Simulation finished')
+
+The results
+***********
+
+Step **(16)**: we can print the timestamps of the spikes:
```
noraabiakar, line 219: Minor nitpick. For the previous steps, you used `step x does/creates/starts` etc. Can we continue the same patter for steps 16 and 17?

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,261 @@
...
+
+.. code-block:: python
+
+   # Print spike times
+   print('spikes:')
+   for sp in sim.spikes():
+      print(' ', sp)
+
+Step **(17)**: Let's make a plot of the sampling data.
+``sim.samples()`` takes a ``handle`` of the probe we wish to examine. It returns a list
```
noraabiakar, line 229: Isn't `sim.samples` in the api? Can't we link it?

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,261 @@
...
+   df_list = []
+   for gid in range(ncells):
+      samples, meta = sim.samples(handles[gid])[0]
+      df_list.append(pandas.DataFrame({'t/ms': samples[:, 0], 'U/mV': samples[:, 1], 'Cell': f"cell {gid}"}))
+
+   df = pandas.concat(df_list)
+   seaborn.relplot(data=df, kind="line", x="t/ms", y="U/mV",hue="Cell",ci=None).savefig('network_ring_result.svg')
+
+
+Since we have create ``ncells`` cells, we have ``ncells`` traces. We should be seeing phase shifted traces, as the action potential propagated through the network.
```
noraabiakar, line 249: `created`

doc/tutorial/mpi.rst
```diff
@@ -0,0 +1,115 @@
...
+   df_list = []
+   for result in results:
+      df_list.append(pandas.read_csv(result))
+
+   df = pandas.concat(df_list)
+   seaborn.relplot(data=df, kind="line", x="t/ms", y="U/mV",hue="Cell",ci=None).savefig('mpi_result.png')
+
+We plot the results using pandas and seaborn (it's a bit crowded!):
+
+.. figure:: mpi_result.png
```
noraabiakar, line 107: Why can't this be `svg` like the rest of the tutorials?

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,261 @@
+.. _tutorialnetworkring:
+
+A ring network
+==============
+
+In this example, a small *network* of cells, arranged in a ring, will be created and the simulation distributed over multiple threads or GPUs if available. Data on execution time and memory consumption will be recorded.
```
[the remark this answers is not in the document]
noraabiakar: But the data isn't actually being recorded, so this whole line can be removed.

doc/tutorial/mpi.rst
```diff
@@ -0,0 +1,115 @@
...
+
+   # (11) Instantiate recipe
+   ncells = 500
+   recipe = ring_recipe(ncells)
+
+The hardware context
+********************
+
+Step **(12)** uses the Arbor-built-in :py:class:`MPI communicator <arbor.mpi_comm>`, which identical to the
+``COMM_WORLD`` communicator you'll know if you are familiar with MPI. The :py:class:`arbor.context` takes it as an
```
noraabiakar, line 31: `MPI_COMM_WORLD`

doc/tutorial/mpi.rst
```diff
@@ -0,0 +1,115 @@
...
+argument to the ``mpi`` parameters. Note that you can also pass in communicators created with ``mpi4py``.
```
noraabiakar, line 32: `to` not needed

doc/tutorial/mpi.rst
```diff
@@ -0,0 +1,115 @@
...
+
+.. code-block:: python
+
+   # (16) Run simulation
+   sim.run(ncells*5)
+   print('Simulation finished')
+
+An important change in the execution is how the script is run. Whereas normally you run the Python script by passing it as an argument to the ``python`` command, you need to use ``srun`` or ``mpirun`` (depending on your MPI distribution) to execute a number of jobs in parallel. You can still execute the script using ``python``, but then MPI will not execute on more than one node.
+
+From the commandline, we can run the script using ``mpirun`` or ``srun`` and specify the number of ranks (``NRANKS``) or nodes. Arbor will spread the cells evenly over the ranks, so with ``NRANKS`` set to 5, we'd be spreading the 500 cells over 5 nodes, calculating 100 cells each.
```
noraabiakar, line 56: `simulating` instead of `calculating`?

doc/tutorial/mpi.rst
```diff
@@ -0,0 +1,115 @@
...
+.. code-block::
+
+   srun -n NRANKS python network_ring_mpi.py
+   mpirun -n NRANKS python network_ring_mpi.py
+
+The results
+***********
+
+Before we execute the simulation, we have to understand how Arbor distributes the computational load over the ranks
+and executes the rest of the script. You can visualize this as a fork from the moment :py:func:`arbor.simulation.run`
```
noraabiakar, line 67: I don't know about the fork once `run` is reached. The whole script is executed in parallel across all the machines. Even the partitioning happens in parallel. So I think this is misleading.

doc/tutorial/mpi.rst
```diff
@@ -0,0 +1,115 @@
...
+is reached. Here the computational load will be distributed and the subsequent execution of the script be different.
+This is relevant as to the collection of results: these are not gathered for you. Remember that in step **(14)** we
+store the handles to the probes; these referred to particular ``gid``s. The ``gid``s are now distributed, so on one
+node, the script will not find the cell referred to by the handle and therefore return an empty list (no results were found).
+
+In step **(18)** we check if the list has a nonzero length, so we can collect the results generated on this particular node. Since Arbor does not collect the results automatically, we have to write the results to disk and analyse them later. We generate a random string to use in the filename where the result is stored.
```
noraabiakar, line 73: Why do we need a random string? It's more useful to use the rank id, no?

noraabiakar (changes requested):
Looks good. My comments are mostly about typos and some small suggestions.

doc/tutorial/mpi.rst
brenthuisman: Old habits!

doc/tutorial/mpi.rst
brenthuisman: The flow of the script forks at this point. From here I can expect execution to be different between nodes, and handles to be useless etc. How would you describe this, if not fork?

doc/tutorial/mpi.rst
```diff
@@ -0,0 +1,115 @@
+.. _tutorialmpi:
+
+MPI
```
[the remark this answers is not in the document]
brenthuisman: I'll make it `Distributed ring network (MPI)` to make it easier to scan.

doc/scripts/make_images.py
```diff
@@ -18,8 +18,13 @@ def translate_all(points, f, xshift):
 
 # Draw one or more morphologies, side by side.
 # Each morphology can be drawn as segments or branches.
-def morph_image(morphs, methods, filename, drawnumbers=True, colors=True, sc=20):
+def morph_image(morphs, methods, filename, **kwargs):
     assert(len(morphs)==len(methods))
+
+    lab_sc = kwargs.get('lab_sc',2)
```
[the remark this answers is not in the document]
brenthuisman: Nope, I can't see some of them at all. This makes the letters more close in size to the text as rendered by Sphinx and viewed in the browser.

doc/tutorial/network_ring.rst
```diff
@@ -0,0 +1,261 @@
...
+The recipe
+**********
+
+To create a model with multiple connected cells, we need to use a :py:class:`recipe <arbor.recipe>`.
+The recipe is where the different cells and the :ref:`connections <interconnectivity>` between them are defined.
+
+Step **(5)** shows a class definition for a recipe with multiple cells. Instantiating the class requires the desired
+number of cells as input. Compared to the :ref:`simple cell recipe <tutorialsinglecellrecipe>`, the main differences
+are connecting the cells, returning a configurable number of cells **(6)** and returning a new cell per ``gid`` **(7)**
+(``make_cable_cell()`` returns the cell above).
```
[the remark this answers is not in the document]
brenthuisman: I removed this on Bens request (see above). If you insist, I can add it back.

doc/tutorial/mpi.rst
brenthuisman: I takes ages to generate due to the amount of data, and is rather large. I chose png to save space and time.

doc/tutorial/network_ring.rst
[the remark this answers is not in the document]
noraabiakar: No it's okay

doc/tutorial/mpi.rst
noraabiakar: We can run it for less time or sample at larger intervals. The sample rate is 10 samples per ms, we can go down to 1.
Another suggestion is to generate the graph by running the same network for a shorter period of time, and display that one. This graph is so cramped that no one will be able to see what's happening anyway.

doc/tutorial/mpi.rst
noraabiakar: I think it's misleading. There's no fork, every rank is running an independent process. When we launch the simulation with `mpirun` on let's say 10 ranks, the script is being run on each of these 10 ranks. When we reach `partition_load_balance` that's when the work is being distributed according to the `rank_id`. So rank 0 takes the 0'th chunk of cell, rank 1 takes the 1st chunk and so on. By the time we reach `sim.run` each rank knows which cells it is supposed to be simulating. 
The problem arises when when we get to assigning `handles`. We are generating handles for all the cells on all the cells on each rank. But not all of them are available, that's why we get empty handles later on. I'll address this issue and how to solve it on slack.

noraabiakar (approved):

PR #1318, opened 2021-01-20
Extend range descriptions (both <,> and FROM/TO) 

thorstenhater:
- Allow negative and real numbers  in `<..., ...>` and `FROM ... TO ...`
- Add and extend tests.

### DIFF
--- a/modcc/blocks.hpp
+++ b/modcc/blocks.hpp
@@ -23,44 +23,42 @@ struct IonDep {
         return writes_variable(name) || reads_variable(name);
     };
     bool uses_current() const {
-        return has_variable("i"+name);
+        return has_variable("i" + name);
     };
     bool uses_rev_potential() const {
-        return has_variable("e"+name);
+        return has_variable("e" + name);
     };
     bool uses_concentration_int() const {
-        return has_variable(name+"i");
+        return has_variable(name + "i");
     };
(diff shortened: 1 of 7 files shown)


noraabiakar (changes requested):
I like the actual changes to modcc and the unit tests. However, I think there are too many whitespace and code-style changes in the code, which are unrelated to the purpose of this PR. 
Generally speaking we try not to make too many of these changes in lines that aren't being modified in some other way as it clutters the history of the files and the PR. If you really want to see those changes in the code, can they be split out into another PR? (Note that we're more likely to use `a+b` in the code than `a + b` and making this sort of changes actually go against convention.)

bcumming (approved):
The content look good.

I second that adding many small subjective formatting changes unrelated to the makes the PR hard to read.

thorstenhater:
Hi,

as we agreed upon a few weeks back, I am using the clang-format style that @noraabiakar shared. These changes
are a consequence of that.

bcumming:
OK, that explains the changes.
Can we apply the changes only to parts of the code that are touched, or have separate PRs that do just formatting.
Mixing the two into one PR makes it difficult to review the actual, functional, changes.

noraabiakar (approved):

Issue #1224, opened 2020-11-09
Contribution guidelines


brenthuisman:
This issue has been updated.

- [ ] Enumerate and write down processes.
    - [x] How to add documentation?
    - [ ] How to add a test
    - [ ] How to add a model/simulation (to examples)
- [x] Issues template.
- [x] Issues beginner tag. 
  - [x] We should start using the tag and use in backlog review.
- [x] PR template.
- [ ] Remaining Github 'community health' items:
    - [ ] Code of Conduct. Do we need this or do we wait for a major incident?
    - [x] Content Reports: I can't find a clear description of what this is, I'm guessing a way to flag contributions to us/Github as unwanted.
- Good to know: Nest calls an overview of all of the above their [developer space](https://nest.github.io/nest-simulator/index). Maybe it's a good thing to have something like this to point at. #1437 is a way to have something similar.

A list of resources when considering the codifying contributing to Arbor (the above list can be amended):

[1] https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/setting-up-your-project-for-healthy-contributions
[2] https://github.com/arbor-sim/arbor/community
[3] https://github.com/stevemao/github-issue-templates/
[4] https://nest.github.io/nest-simulator/development_workflow
[5] https://raw.githubusercontent.com/microsoft/vscode/master/.github/pull_request_template.md

brenthuisman:
Update: I've been through the motions of the CII best practices badge, see [1]. Relevant for this discussion is the following 'fail':

* The project SHOULD respond to a majority (>50%) of enhancement requests in the last 2-12 months (inclusive).

We don't have an explicit time frame for handling Issues / Enhancement requests. We don't collects stats, but also don't have many Issues filed by external users yet. Personally I have noticed PR review can fall a little by the wayside. Maybe some explicit desired timeline would be helpful to give external users a good impression. Also, Backlog Pruning...

[1] https://bestpractices.coreinfrastructure.org/en/projects/4476

thorstenhater:
I think we went as far as we can for now!

PR #1319, opened 2021-01-22
Add AEP and PR template with ref to contrib guidelines and license

brenthuisman:
* Issue templates reduced in size
* Add Arbor Enhancement Proposal template (sets labels automatically mainy)
* Addressed part of #1224


bcumming (commented):
What is the purpose of an `Arbor Enhancement Proposal`?

Is it intended for internal use (i.e. the Arbor dev team would use them to define work before starting) or for external use?

How does it differ from a Feature request?

brenthuisman:
The difference would be that it's a more explicit invitation to flesh something out. A feature request might be no more than that, a AEP (like PEP/RFC) would imply there's a worked out and detailed proposal, and input from more than one person is given.

It's perhaps not a sharp divide. I hope it would encourage discussion be a bit more public (no need to use Slack, just talk in the AEP!), public as in both for non Arbor-devs and other Arbor-dev not necesarily involved in the development of the feature, but still interested in following along and understanding compromises/decisions.

bcumming:
Maybe this could all be handled with the one issue type, i.e. just have a `feature or enhancement request`? They are a little ambiguous, and the discussion under both types of issue will probably be very similar (and a feature request would probably turn into an AEP in many cases).

brenthuisman:
There was some enthusiasm in the channel for this, hence this PR. Maybe we should insist then that a discussion about a feature request takes places in an Issue?

noraabiakar:
Can we postpone this issue until we have a better idea of what our feature discussion/approval workflow is going to look like?

brenthuisman:
I think last retrospective we agreed to make formal proposals by way of Issue for larger plans. How we handle them (==Issues w/ this label) can be perhaps described in the contribution docs, once ironed out. This PR (the AEP part) only adds the label if you choose the Issue template.

thorstenhater:
bors try

bcumming:
bors try

brenthuisman:
bors try

brenthuisman:
bors try

brenthuisman:
bors try

haampie:
bors try-

haampie:
bors try

bcumming (approved):

PR #1325, opened 2021-01-22
Respect CMAKE_INSTALL_PREFIX in python lib path

### LINKED ISSUE #1324: CMake install ignores CMAKE_INSTALL_PREFIX for python components
Consider:
```
% mkdir build; cd build
% cmake .. -DCMAKE_INSTALL_PREFIX=/tmp/foo -DARB_WITH_PYTHON=ON
% make install
```

Current behaviour:
```
[…]
-- Installing: /usr/lib/python3.7/site-packages/arbor
CMake Error at python/cmake_install.cmake:54 (file):
  file INSTALL cannot make directory
  "/usr/lib/python3.7/site-packages/arbor": No such file or directory
Call Stack (most recent call first):
  cmake_install.cmake:92 (include)


make: *** [Makefile:74: install] Error 1
```

Note that the attempted path is not under `/tmp/foo`.

Expected behaviour:
```
[…]
-- Installing: /tmp/foo/lib/python3.7/site-packages/arbor
-- Installing: /tmp/foo/lib/python3.7/site-packages/arbor/VERSION
-- Installing: /tmp/foo/lib/python3.7/site-packages/arbor/__init__.py
-- Installing: /tmp/foo/lib/python3.7/site-packages/arbor/_arbor.cpython-37m-x86_64-linux-gnu.so
```

When installing in `/usr/local`, or in a stow directory, or in a staging install directory, or in fact in pretty much every circumstance where one wants to set `CMAKE_INSTALL_PREFIX`, all the installed files should be installed under that prefix unless there is a specific direction not to (e.g. via `ARB_PYTHON_LIB_PATH`).

halfflat:
* Set `base` and `platbase` to the CMAKE_INSTALL_PREFIX, if defined, when asking Python for the default python lib install path.

Fixes #1324.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -81,7 +81,15 @@ file(COPY "${PROJECT_SOURCE_DIR}/VERSION" DESTINATION "${python_mod_path}")
 # Set the installation path
 
 # Ask Python where it keeps its system (platform) packages.
-execute_process (COMMAND ${PYTHON_EXECUTABLE} -c "import sysconfig; print(sysconfig.get_path('platlib'))" OUTPUT_VARIABLE ARB_PYTHON_LIB_PATH_DEFAULT OUTPUT_STRIP_TRAILING_WHITESPACE)
+
+file(WRITE "${CMAKE_BINARY_DIR}/install-prefix" "${CMAKE_INSTALL_PREFIX}")
+execute_process(
+    COMMAND ${PYTHON_EXECUTABLE} -c
+    "import sys,sysconfig;pfx=sys.stdin.read();print(sysconfig.get_path('platlib',vars={} if pfx=='' else {'base':pfx,'platbase':pfx}))"
+    INPUT_FILE "${CMAKE_BINARY_DIR}/install-prefix"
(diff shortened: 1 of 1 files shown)


halfflat:
I notice that if `CMAKE_INSTALL_PREFIX` is not set, it defaults to `/usr/local`. If one _were_ to install under `/usr/local`, then the correct spot for the python libs would still likely be somewhere under `/usr/local/lib`. 

With this patch then as it stands, the system site-packages path will basically never get picked, unless `CMAKE_INSTALL_PREFIX` is set to the python install `base` directory. The only circumstance where this might be an issue is in packaging, but there people are going to have to futz with the paths anyway, so I'm personally not concerned.

bcumming (approved):
A good improvement to the current approach.

During discussion of this PR during the weekly meeting we discovered that neither this approach, nor the one that it replaces, install in the correct path on Debian-based distro (i.e. Ubuntu) when installing the default path chosen by python when not in a virtual environment (it should, apparently go to something like `/usr/local/lib/python3.x/dist-packages`, and instead Python will return `/usr/local/lib/python3.x/site-packages`.

The solution is... don't try to solve that mess. Instead we should update the installation guide for that small subset of users who try to `sudo make install` are warned.

PR #1327, opened 2021-01-25
Move neuroml handling to arbor-io and wrap for pyarb

### LINKED ISSUE #1234: Move arbornml into arborio.
Like it says on the tin. Cuts down on the number of libraries we're building and installing, and it's a natural fit,
We can still keep NeuroML support a configure time option, dependent upon libxml2.

### LINKED ISSUE #1256: Add Python wrapper for NeuroML.
We need to expose the NeuroML reader to Python users.

noraabiakar:
* Move contents of `arbornml` to to `arborio` 
* `arborio` will be built with NeuroML2 morphology support if `ARB_CMAKE_NEUROML` is set. 
* Wrap `arborio::neuroml` and `arborio::morphology_data` for pyarb.
* Add documentation.
* add `neuroml` flag to `setup.py`

Fixes #1256, and fixes #1234


setup.py
```diff
@@ -54,6 +55,7 @@ class install_command(install):
                         'none, cuda, cuda-clang, hip'),
         ('vec',   None, 'enable vectorization'),
         ('arch=', None, 'cpu architecture, e.g. haswell, skylake, armv8.2-a+sve, znver2 (default native).'),
+        ('neuroml', None, 'enable parsing neuroml morphologies in Arbor (regiures libxml)')
```
bcumming, line 58: small typo: **requires** libxml.

bcumming, line 58: small typo: **requires** libxml.

bcumming (changes requested):
Looks good.

One small fix.

setup.py
noraabiakar: Wouldn't call that small! Thanks for pointing it out.

bcumming (approved):

PR #1329, opened 2021-01-27
replace sphinx rtd theme submodule with copy

schmitts:


bcumming (commented):
There is a lot of content here, particularly regarding the number of files, and the 8 MB size of the contents of the path (which is nearly all fonts), which adds another 33% to the size of our repository.

Can we reduce the number of files and fonts required?

brenthuisman:
@schmitts What's the reason to have the theme source in the repo?

Maybe we should take the opportunity to change the theme. In search of a visually and resource-wise lighter theme, I found [Furo](https://pradyunsg.me/furo/quickstart/) the best candidate. I couldn't find any that satisfies @halfflat 's wish of no-js.

halfflat:
IIRC the motivation was to have a modifiable copy so we didn't have to resort to CSS hacks to get proper formatting of C++ API methods etc.

That said, if we're going to do this, I'd prefer not to have the whole of the theme included in the Arbor repo, especially the fonts. Some alternative options, perhaps:
* We fork the theme in another repo, and submodule the fork in.
* We just copy in what we need from the theme, being careful to abide by the Apache License. 

Either way though, we should update the top level ATTRIBUTIONS file to suit.

bcumming:
It was my impression that one reason for not having the docs as a submodule was to make it easier to install the docs from the source distribution (that would be packaged up in the MANIFEST file for pip installation)

The Furo theme is smaller and clean, however I understand that it requires node-js to build the docs. Which is:
1. Additional dependencies.
2. A sign that the world has gone mad.

Are there no simple themes that are actually simple?

halfflat:
Re: no-JavaScript desideratum: I'd still be happy with graceful degradation without it :)

schmitts:
Yes, the reason is to get rid of the last remaining mandatory git submodule. I'm strongly against to include anything node-js.

brenthuisman:
Can we rely on pipping furo? We do so for svgwrite as well.

(I don't understand why, the theme is distributed all built, but I just can't get Sphinx to pick up on it outside of site-packages)

schmitts:
We can also use the package for the rtd theme. But the argument was that we want the possibility to customize the theme, i.e. use a copy or point to a specific version.

brenthuisman:
OK, I wasn't aware of that argument. Currently the modifications we do make are quite minimal, and not really necessary in my opinion. I'd be absolutely willing to let go of customization for now and rely on pip for the theme.

bcumming (changes requested):
My review has been requested, and in light of the discussion on the PR, we can reject it in its current form.

Options are:
* find a minimal RTD theme that does not have python or (heaven-forbid) JS dependencies, its own fonts, or other overheads, and include that in the repo.
* keep using the rtd theme as a sub module, and package the rtd code in future release tar balls so that they are available to tools like spack. That way we could have our own fork of rtd theme, to modify as we see fit.

schmitts:
I second @bcumming's remarks.

brenthuisman:
There is another option:

* Do not package the theme. Rely on pip to get it.

PR #1330, opened 2021-01-28
Set correct minimum GCC version.

thorstenhater:
Sets the minimum GCC version we test against to the correct value of eight.


brenthuisman (approved):

brenthuisman:
The docs say `8.4.0`. I don't see minor version number in the yaml file, so not sure if that's an important distinction.

https://arbor.readthedocs.io/en/latest/install/build_install.html

PR #1331, opened 2021-01-28
Clarify sample collection in distributed contexts.

brenthuisman:


doc/python/simulation.rst
```diff
@@ -135,6 +135,11 @@ over the local and distributed hardware resources (see :ref:`pydomdec`). Then, t
 
         Retrieve a list of sample data associated with the given ``handle``.
         There will be one entry in the list per probe associated with the probe id used when the sampling was set up.
+        E.g. if a probe was placed on a locset describing three positions, the returned list will have three elements.
```
bcumming, line 138: Starting a sentence with E.g. looks odd.

**For example**, if a probe... the returned list will **contain** three elements.

doc/python/simulation.rst
```diff
@@ -135,6 +135,11 @@ over the local and distributed hardware resources (see :ref:`pydomdec`). Then, t
...
+
+        An empty list will be returned if no (local) output was found. This is expected for execution contexts with
```
bcumming, line 140: Good point. How about making it clear that the "local" reference is in the context of MPI?

```
An empty list will be returned if no output was recorded for the cell. For simulations that are distributed using MPI, handles associated with non-local cells will return an empty list, and it is not an error to query the results for a cell on a different rank, however it is the responsibility of the caller to gather results over the ranks.
```

bcumming (changes requested):

bcumming (approved):

Issue #1340, opened 2021-02-03
Arbor fails to build with Python support disabled


halfflat:
Unit test for dynamically loaded catalogue is built even when Arbor is built with PIC disabled (PIC is enabled globally if PYTHON support is configured). With the current implementation of loadable catalogues, the catalogue must link against Arbor, and we get errors at build time such as:
```
FAILED: lib/dummy-catalogue.so
: && /usr/bin/c++ -fPIC -O3 -DNDEBUG   -shared -Wl,-soname,dummy-catalogue.so -o lib/dummy-catalogue.so test/unit/CMakeFiles/dummy-catalogue.dir/dummy_catalogue.cpp.o test/unit/CMakeFiles/dummy-catalogue.dir/generated/dummy/dummy_cpu.cpp.o  lib/libarbor.a -ldl -lstdc++fs -lpthread && :
/usr/bin/ld: lib/libarbor.a(mechanism.cpp.o): relocation R_X86_64_PC32 against symbol `_ZTVSt9bad_alloc@@GLIBCXX_3.4' can not be used when making a shared object; recompile with -fPIC
```

This in particular breaks nsuite/dev as things stand.

Proposed fix: disable building the dummy catalogue and omit the corresponding unit case if Arbor is not being built as a shared library.

Longer term suggestion: we build a static lib, and a shared lib in addition if building Python, and when the mechanism ABI is in place, we will not have a link dependency of catalogues on the Arbor library.

PR #1341, opened 2021-02-04
Do not build/use dynamic catalogue tests if not built with -fPIC

thorstenhater:
Fixes a regression mentioned in #1340 .


halfflat (approved):
Looks good!

Issue #1337, opened 2021-02-02
Expand GH CI


thorstenhater:
As travis.org will be shutting down soon, we will need to expand our CI at Github.
Currently, all tests are run, documentation is build, but only on Ubuntu 20.04 LTS.

We  are commited to support the following OSs (although MacOS is not really fixed)
- Ubuntu LTS 
  - 18.04
  - 20.04
- MacOS
  - 10.15
  - 11.0

And the following compilers on Linux
- GCC 8, 9, 10
- Clang 8, 9, 10, 11

plus the default Apple Clang on Mac.

CMake 3.12+

Python 3.6.9+

Running all combinations is obviously infeasible; we have discussed the following compromise
before
- Ubuntu LTS 
  - 18.04
    - GCC 8 and 10
    - Clang 8 and 11
  - 20.04
    - GCC 8 and 10
    - Clang 8 and 11
- MacOS
  - 10.15
    - AClang default
  - 11.0
    - AClang default
    
This results in 10 runs _before_ factoring in CMake and Python. If we were to follow the same procedure
([min, max]) for both, the total would come in at 40 runs. For my taste that is still too much.

My suggestion is to pare this further down to
- Ubuntu LTS 
  - 18.04 (collecting the minimum versions)
    - GCC 8 
    - Clang 8 
    - Python 3.6.9
    - CMake 3.12
  - 20.04 (collecting the current versions)
    - GCC 10
    - Clang 11
    - Python 3.9.1
    - CMake 3.19.4   
- MacOS
  - 10.15 (collecting the minimum versions)
    - AClang default
    - Python 3.6.9
    - CMake 3.12    
  - 11.0 (collecting the current versions)
    - AClang default
    - Python 3.9.1
    - CMake 3.19.4 
 
for a total of 6 runs. Note that the versions need to be adjusted when new supported releases appear (max) and
when we bump the minima (min).

bcumming:
The proposed configurations strike a good balance between comprehensiveness and a combinatorial explosion of configurations.

Testing bother versions of MacOS is important, because the differences between the two are significant.

PR #1342, opened 2021-02-04
Expand Build Matrix on GH CI

thorstenhater:
Implements #1337 minus the realities of GitHub Runners:

- GCC max version is 10, not 11
- MacOS 11 is not officially supported yet


bcumming (approved):

bcumming:
Add more GitHub actions CI configurations
* min and max versions of ubuntu, clang and gcc for linux
* MacOS tests with min and max clang versions.

Fixes #1337

PR #1345, opened 2021-02-04
Example for a plastic synapse

schmitts:


python/example/single_cell_stdp.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python3
```
noraabiakar, line 1: Can we add the example to github actions? https://github.com/arbor-sim/arbor/blob/master/.github/workflows/basic.yml

noraabiakar (changes requested):
Looks good, thanks! Just a small request.

python/example/single_cell_stdp.py
schmitts: done

python/example/single_cell_stdp.py
```diff
@@ -0,0 +1,124 @@
...
+
+    reg_sched = arbor.regular_schedule(0.1)
+    handle_mem = sim.sample((0, 0), reg_sched)
+    handle_g = sim.sample((0, 1), reg_sched)
+    handle_apost = sim.sample((0, 2), reg_sched)
+    handle_apre = sim.sample((0, 3), reg_sched)
+    handle_weight_plastic = sim.sample((0, 4), reg_sched)
+
+    sim.run(tfinal=600)
+    mem, meta = sim.samples(handle_mem)[0]
```
noraabiakar, line 98: I think this line can be removed?

noraabiakar (approved):
Looks good! There's just one line in the example that I think can be removed, I didn't notice it before.

python/example/single_cell_stdp.py
schmitts: ack

PR #1350, opened 2021-02-05
Use JFrog as a registry and fix build errors

noraabiakar:
* Make Gitlab CI push images to CSCS' JFrog registry reg.giuv.cscs.ch so that sarus can pull images on Daint again. 
  This URL is probably temporary until JFrog officially goes into production later this month.
  Unfortunately JFrog is currently behind the firewall.
* Upgrade g++ to g++8 on the Docker image.
* Only require libxml when compiling with NeuroML support.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -16,7 +16,7 @@ stages:
   variables:
     GIT_SUBMODULE_STRATEGY: recursive
   script:
-    - docker login -u $CI_REGISTRY_USER -p $CI_REGISTRY_PASSWORD $CI_REGISTRY
+    - docker login -u $CSCS_REGISTRY_USER -p $CSCS_REGISTRY_PASSWORD $CSCS_REGISTRY
     - docker build -f $BUILD_DOCKERFILE --network=host --cache-from $BUILD_IMAGE --build-arg BUILDKIT_INLINE_CACHE=1 -t $BUILD_IMAGE .
     - docker push $BUILD_IMAGE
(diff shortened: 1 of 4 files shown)


noraabiakar:
bors try

haampie:
Oh, can you maybe cherry-pick the first commit from #1351, then I'll close it and trust you that you know how to build arbor :p

noraabiakar:
Sure, and I'll add a fix for building arbor :)

noraabiakar:
bors try

noraabiakar:
bors try

haampie:
That's not a very helpful failure :D

noraabiakar:
@haampie thanks!

haampie:
The codecov situation with ubuntu 18.04 and GCC is really dire :/ apparently GCC 8 changed their coverage format and lcov didn't support it until much later, but that version can't be installed with apt. I thought it would be nice to use [grcov](https://github.com/mozilla/grcov) because it is _much_ faster and easier than lcov, but they don't support GCC 8 either. Too bad...

bors try

noraabiakar:
Would this be fixed if we use Ubuntu 20.04?

haampie:
Yeah, I thought that was the easiest solution, but it looks like NVIDIA doesn't officially support it for CUDA 10.2 (which is what Daint is on): https://hub.docker.com/r/nvidia/cuda/tags?page=1&ordering=last_updated&name=10.2-devel-ubuntu

noraabiakar:
That's too bad.

haampie:
Yeah, the other thing is that ubuntu 20 ships with GCC 9 by default, and you'd have to configure it to use GCC 8 anyways (if that is what you want to test)

bcumming (approved):
Thanks @haampie and @noraabiakar !

PR #1352, opened 2021-02-05
modcc: fix bug in simd cprinter

noraabiakar:
Caused by a typo. The correct version is being used in the non-simd printer.


schmitts:
The bug was not found by any test? Could you add one?

noraabiakar:
Unfortunately, our test coverage of modcc is quite lacking. I'll see if I can add something to catch this.

bcumming:
@noraabiakar is right: we don't have thorough tests in place for the modcc printers, and adding a test for this case would require quite a lot of infrastructure. Given that we don't plan to implement something "better" than NMODL in the future, that is some technical debt I don't think we want to spend time addressing.

bcumming (approved):
Oops! LGTM!

PR #1354, opened 2021-02-05
Refactor `algorithms.hpp`

### LINKED ISSUE #539: Refactor `algorithms.hpp` into `util`.
* Split general purpose routines in `algorithms.hpp` from cell-tree routines; more former into `util/`.
* Move cell-tree routines into own header, check names.
* Update tests to suit.

Part of #98 work.

noraabiakar:
* Refactor the functions from `algorithms.hpp`:
  - `mean`: moved to `profile/meter_manager.cpp`
  - `make_index`: moved to `util/index.hpp`
  - `is_strictly_monotonic_increasing`: moved to `/backends/gpu/forest.hpp`
  - `has_contiguous_compartments`:  moved to `/backends/gpu/forest.hpp`.
  - `branches`: moved to `/backends/gpu/forest.hpp`.
  - `expand_branches`: moved to `/backends/gpu/forest.hpp`.
  - `tree_reduce`: moved to `/backends/gpu/forest.hpp`.
  - `is_minimal_degree`: moved to `tree.hpp`.
  - `child_count`: moved to `tree.hpp`.
  - `in_interval`: not used, deleted.
  - `is_strictly_monotonic_decreasing`: not used, deleted.
  - `all_positive`:  not used, deleted.
  - `all_negative`: not used, deleted.
  - `find_branch`: not used, deleted.
  - `is_unique`: not used, deleted.
  - `binary_find`: not used, deleted.
* Remove `make_parent_index` from `tree.hpp`: not used. 
* Add unit tests to `test_forest.cpp`, `test_tree.cpp` and `test_index.cpp`

 Fixes #539


arbor/profile/meter_manager.cpp
```diff
@@ -18,6 +17,13 @@ namespace profile {
 using timer_type = timer<>;
 using util::strprintf;
 
+template <typename C>
+typename util::sequence_traits<C>::value_type
+mean(C const& c)
+{
+    return util::sum(c)/std::size(c);
+}
+
```
halfflat, line 26: There was always a latent issue in `mean` where the mean of sequence over an integral type would give an integral result. In this context, in `meter_manager.cpp`, it's always assigned to a double value, and so I think it's worthwhile making it a bit more correct and a bit less general, with e.g. `template <typename C> double mean(const C& c) { return util::sum(c)/(double)std::size(c); }`

arbor/communication/communicator.cpp
```diff
@@ -104,7 +104,7 @@ communicator::communicator(const recipe& rec,
     // The loop above gave the information required to construct in place
     // the connections as partitioned by the domain of their source gid.
     connections_.resize(n_cons);
-    connection_part_ = algorithms::make_index(src_counts);
+    connection_part_ = util::make_index(src_counts);
```
halfflat, line 107: `util::make_partition` is a possible alternative here (from `util/partition.hpp`).

halfflat (changes requested):
Glad to see this cleaned up!

There's one comment about the mean function (it was previously simultaneously too general and too specific), and I'm wondering if we can replace the uses of `make_index` with `make_partition`. I've never really been happy with the name `make_index`, to be honest; what do you think about a name such as `left_scan` or `scanl`?

noraabiakar:
I've replaced all instances of `util::make_index` with `util::make_partition`, which I think are equivalent. 
In `tree.cpp`, I've removed the `memory::copy` call because I don't think it's necessary. As far as I can tell none of the vectors used live on the GPU.

noraabiakar:
bors try

noraabiakar:
bors try

noraabiakar:
bors try

bcumming:
bors try

halfflat (approved):
All good! Sorry for the delay in approval.

PR #1356, opened 2021-02-08
Ebrains mirror update

brenthuisman:
Mirror only master branch and tags.


bcumming (approved):
I agree with limiting it to just tags and master. We can add other branches as needed.

Has this been tested in any way?

brenthuisman:
Yes: https://github.com/brenthuisman/arbor/runs/1854266993

PR #1358, opened 2021-02-08
Add missing comma in setup.py

llandsmeer:
I'm sorry for creating an entire pull request for such a small change, but this line breaks my build


noraabiakar (approved):
Thanks for catching that!

bcumming (approved):
Thanks!

Did that even work?

bcumming:
It might be time to add an extra CI test that
* clones the repository to `${arbor_path}`
* sets up a virtual env with numpy
* runs `pip install ${arbor_path}`
That would catch issues like this

llandsmeer:
I didn't test the --sysdeps flag (this was just my first checkout of the repository), but enabling it leads to `cmake` receiving `-DARB_USE_BUNDLED_LIBS=off` at least (didn't check further).

PR #1360, opened 2021-02-10
Modcc: comments & seg fault

noraabiakar:
- Fixes bug with `COMMENT`/`ENDCOMMENT` comments in nmodl, where `location.line` was not updated within the comment block.
- Fixes segmentation fault where `api_state->semantic(symbols_);` was being called without checking whether `api_state` is null.


noraabiakar:
FYI: most of the changes in `module.cpp` are whitespace changes.

thorstenhater (approved):
Looks good.

PR #1361, opened 2021-02-10
Modcc: parse the unit of `FUNCTION`

noraabiakar:
Parse and ignore the units of a FUNCTION in nmodl. 
For example, parse `(mV)` in the following:
```
FUCNTION foo(v) (mV) {
  foo = v + 5;
}
```


thorstenhater (changes requested):
Looks good, but I think it could benefit from a unit test. Also, I remember adding something similar a while back, this here is used to indicate the return value's unit (as opposed to that of the parameters), right?

thorstenhater (approved):
LGTM

PR #1363, opened 2021-02-10
Gitpod: update cmake 

noraabiakar:


modcc/parser.cpp
```diff
@@ -973,6 +973,14 @@ symbol_ptr Parser::parse_function() {
     auto p = parse_prototype();
     if (p == nullptr) return nullptr;
 
+    // Functions may have a unit attached
```
brenthuisman, line 976: Is this related to the CMake update?

noraabiakar: Sorry I accidentally pushed the contents of another branch. I've removed it now.

brenthuisman (approved):

PR #1368, opened 2021-02-12
Shell script to create a tar ball

schmitts:
The script can be used to create a tar ball containing all sources without relying on further downloads from, e..g, git submodules.
A self contained package of an Arbor release can be better with package managers like Spack compared to a multi stage retrieval of sources.


thorstenhater:
Looks good, but could you add a short description along the lines of what this does and why?

schmitts:
> Looks good, but could you add a short description along the lines of what this does and why?

I've updated the description.

thorstenhater (approved):
Looks good.

PR #1370, opened 2021-02-12
Fixes for codecov

haampie:
- remove repeated CMAKE_BUILD_TYPE
- bump lcov to 1.15
- use new --exclude flag to skip generated sources

With gcc 8 as a compiler + a new version of lcov (#1350) codecov started tracking generated source files and coverage dropped to < 60%. Not sure if this is a bug in lcov or not. I've added these new flags for lcov: `--exclude "*/ext/*" --exclude "*/test/*" --exclude "*/mechanisms/*"` to fix it.


haampie:
bors try

bcumming (approved):
Thanks @haampie !

PR #1372, opened 2021-02-12
Make `pas/e` GLOBAL instead of RANGE

### LINKED ISSUE #1052: Examine the built-in `pas` mechanism.
This came up again in discussions about #1051 ...

While we still have the mechanism-based approach to passive currents, we should look at the built-in `pas` mechanism.

* The default reversal potential is different from NEURON's, but it probably ought not be.
* The reversal potential is represented by a `RANGE` parameter, despite it not being a conductance multiplier. For correctness, we should probably change this to a `GLOBAL` mechanism parameter, so that we do not engage in incorrect linear averaging.

As a global parameter, the reversal potential would be customized via the mechanism catalogue derivation process (which can be done implicitly by using a mechanism name such as `pas/e=61`).

thorstenhater:
- Make reversal potential `e` a global parameter of `pas` to avoid inconsistency 
  (reversal potential is represented by a RANGE parameter, despite it not being a conductance multiplier see #1052)
- Adjust default value of `e` to -70mV (as in Neuron)
- Streamline docs,  which use `passive`, `el`, and `gl` instead of `passive`, `e`, and `g`.

Closes #1052

### DIFF
--- a/doc/concepts/decor.rst
+++ b/doc/concepts/decor.rst
@@ -139,23 +139,23 @@ combination of global parameter values.
 
 Take for example a mechanism passive leaky dynamics:
 
-* Name: ``"passive"``.
-* Global variable: reversal potential ``"el"``.
+* Name: ``"passive"`` (the built-in mechanism is called just ``"pas"``).
+* Global variable: reversal potential ``"e"``.
 * Range variable: conductance ``"g"``.
 
 .. code-block:: Python
 
-    # Create pas mechanism with default parameter values (set in NMODL file).
+    # Create passive mechanism with default parameter values (set in NMODL file).
     m1 = arbor.mechanism('passive')
 
-    # Create default mechanism with custom conductance (range)
+    # Create mechanism with custom conductance (range)
     m2 = arbor.mechanism('passive', {'g': 0.1})
 
-    # Create a new pas mechanism with that changes reversal potential (global)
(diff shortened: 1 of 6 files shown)


doc/concepts/decor.rst
```diff
@@ -139,23 +139,23 @@ combination of global parameter values.
...
-* Global variable: reversal potential ``"el"``.
+* Name: ``"passive"`` (the built-in mechanism is called just ``"pas"``).
+* Global variable: reversal potential ``"e"``.
 * Range variable: conductance ``"g"``.
 
 .. code-block:: Python
 
-    # Create pas mechanism with default parameter values (set in NMODL file).
+    # Create passive mechanism with default parameter values (set in NMODL file).
     m1 = arbor.mechanism('passive')
```
halfflat, line 149: Should the example code also refer to `'pas'` instead of `'passive'`?

halfflat (changes requested):
I'm glad to see this resolved! Just one query re: example code in decor docs.

halfflat (approved):

bcumming:
Can you please update the title of the PR and the description to describe what was done?

That way it is possible to understand the PR from looking at the output of `git log` , without going to the GitHub repo.

thorstenhater:
Sure, see above.

PR #1373, opened 2021-02-12
Modcc: solve statement

### LINKED ISSUE #1384: modcc: adding statements before the `SOLVE` statement in the `BREAKPOINT` block
When we perform function inlining, we have to create some local variables and initialize them to zero. These statements go on the top of the statements list. This causes a compiler bug if there is a function call in the `BREAKPOINT` block of an nmodl file: 
```
BREAKPOINT {
SOLVE state method cnexp
i = foo(g) * (v -e)
}
FUNCTION foo(x) {
foo = x + 5
}
```
After foo is inlined, the BREAKPOINT block looks as follows (pseudo-code)
```
local t_
t_ = 0
solve(state, cnexp)
t_ = g + 5
i_ = t_ * (v - e) 
```
Right now, we only allow local declarations before a `SOLVE` statement. The compiler will complain about the assignment `t_ = 0`. We should either allow initializing local variables before a `SOLVE` statement, or remove the requirement that the `SOLVE` statement be first in a `BREAKPOINT` block.

If we remove the requirement that `SOLVE` has to be first, then all code before that statement is part of the current update, and only the `SOLVE` statement will be part of the state update. For example: 
```
BREAKPOINT {
g = 9
SOLVE state method cnexp
i = foo(g) * (v -e)
}
```
If the `SOLVE` statement reads `g`, the state update will not see that it was written to `9`. Only the current update will. This behavior is consistent with NEURON (tested with `nrnivmodl`).

noraabiakar:
- Allow SOLVE statement anywhere in the BREAKPOINT block.
- Add unit test.

Fixes #1384.


halfflat:
I'm concerned about this change, because non-SOLVE statements will be executed as part of the `nrn_current` phase, that is preceded by the `nrn_state` call. Any changes in state or assigned variables that precede the SOLVE statement, will in fact have an effect after the SOLVE is performed.

noraabiakar:
Yes, I was also concerned about that. However, I *think* that is the expected behavior. I looked at the code generated by `nrnivmodl` and it lumps any code before the `SOLVE` statement with the code of `nrn_current`. I asked @bcumming and he also believes that is the expected behavior. 
I should have mentioned in the PR description the reason why I am making this change: when we perform function inlining, we place some statements at the top of the `BREAKPOINT` block. It used to be just local declarations, which were skipped; but later, that caused compiler warnings for uninitialized variables in the SIMD backend, so we started initializing the local variables. It's becoming more and more difficult to enforce that the `SOLVE` statement is first in the `BREAKPOINT` block or to allow only special statements to go before it.

bcumming:
I think that it is the expected behavior that all statements except `SOLVE` are part of the current update, even those before the `SOLVE` statements.

modcc/module.cpp
```diff
@@ -189,6 +189,24 @@ bool Module::semantic() {
...
+    // in the block.
+    if (has_symbol("breakpoint", symbolKind::procedure)) {
+        bool found_non_solve = false;
+        auto breakpoint = symbols_["breakpoint"]->is_procedure();
+        for (const auto& s: breakpoint->body()->statements()) {
+            if(!s->is_solve_statement()) {
+                found_non_solve = true;
+                continue;
+            }
+            if (found_non_solve) {
```
bcumming, line 203: should this be `else if (found_non_solve)`, because it is only an error if a SOLVE statement is encountered after the first non-solve statement.

bcumming (changes requested):
Looks good: just one potential issue with the check for non-solve statements before a solve.

modcc/module.cpp
noraabiakar: Doesn't the `continue` take care of that? 
But I think removing the `continue` and using `else` is clearer, so I'll make the change.

bcumming: Ah, yes, you are right. But I think that the logic is a bit confusing relying on the `continue`.

bcumming (approved):

PR #1383, opened 2021-02-16
Remove transmogrify layer from s-expression parser

bcumming:
Remove transmogrify layer from s-expression parser.

The transmogrify layer isn't needed if asc files won't be parsed as s-expressions.
It complicates and slows down parsing of s-expressions.


noraabiakar (approved):
Looks good to me. If we decide to treat `\r` as whitespace as @halfflat suggested, I can do the same in the modcc lexer later.

PR #1386, opened 2021-02-16
Add neuroML python example

noraabiakar:
* Fix missing docs.
* Add methods to import label dictionaries. 
* Add neuroml example (should be identical to `single_cell_swc.py`)


brenthuisman (approved):
Looks good.

Duplication can be addressed in a future PR.

brenthuisman (approved):
Didn't know you could do this!

PR #1387, opened 2021-02-17
Hush catalogue build output

thorstenhater:
This produces less output when building catalogues, it used to spam the CMake configure log.
The old behaviour can be toggled back on by the advanced flag `ARB_CAT_VERBOSE`. I also
added some fixes build scripts.
- [CMake] Correctly pick up the compiler used for arbor.
- [CMake] Set the path to scripts properly.
- [build-cat] Produce less output per default
- [build-cat] Add `quiet` flag for even less output


bcumming (approved):
Looks good: thanks!

PR #1388, opened 2021-02-17
Update `set_ion` method of `cable_global_properties` to allow the addition of new ions

noraabiakar:


python/cells.cpp
```diff
@@ -384,6 +384,18 @@ void register_cells(pybind11::module& m) {
...
+                int valence, double int_con, double ext_con, double rev_pot)
+             {
+               props.add_ion(ion, valence, int_con, ext_con, rev_pot);
+             },
+             pybind11::arg_v("ion", "name of the ion species."),
+             pybind11::arg_v("valence", "valence of the ion species."),
+             pybind11::arg_v("int_con", "initial internal concentration [mM]."),
+             pybind11::arg_v("ext_con", "initial external concentration [mM]."),
+             pybind11::arg_v("rev_pot", "reversal potential [mV]."),
+             "Add a new ion species named 'ion' to the global default propoerties.")
```
bcumming, line 398: Typo `propoerties`

bcumming: Why the `add_ion` method when there is a `set_ion` method?
Should we pick one method, and stick to it?

bcumming (changes requested):

python/cells.cpp
noraabiakar: Oh, I should fix where I copied that from too. 
And I can rename it to `set_ion`.

bcumming (approved):

PR #1390, opened 2021-02-18
tiny themefix, tutorialfix

brenthuisman:


doc/tutorial/mpi.rst
```diff
@@ -64,8 +64,8 @@ cells over 5 nodes, simulating 100 cells each.
 
 .. code-block::
 
-   srun -n NRANKS python network_ring_mpi.py
-   mpirun -n NRANKS python network_ring_mpi.py
+   srun -n NRANKS python mpi.py
```
bcumming, line 67: Should this be `${NRANKS}`, to make it bashy, and cut and paste freindly?

bcumming (changes requested):

doc/tutorial/mpi.rst
thorstenhater, line 67: _also_ ``srun`` is specific to SLURM, not MPI, which might cause quite some confusion on a laptop w/o a resource manager.

thorstenhater (changes requested):

doc/tutorial/mpi.rst
brenthuisman: I'm not expecting that variable is set, so I'd say it's good that the line will fail if left unedited. The context is quite clear about this, I think.

brenthuisman: Aight.

bcumming: Source code examples should be valid code. You could just say `mpirun -n 5 python mpi.py`.

brenthuisman: done

thorstenhater (approved):

PR #1392, opened 2021-02-19
Use test programs to enable ``std::filesystem``

thorstenhater:
Use CMake `check_cxx_source_compiles` and `check_cxx_source_runs` to detect how `std::filesystem`
is enabled. The probes are executed first, if none succeeds, we use link-only to cover the possibility that we
are crosscompiling.


thorstenhater:
bors try

thorstenhater:
bors try

thorstenhater:
bors try

cmake/CompilerOptions.cmake
```diff
@@ -18,20 +21,60 @@ if(${ARBDEV_COLOR})
...
-endif()
-# TODO Remove when upgrading Clang
-if(CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
-  if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
-    target_link_libraries(arbor-compiler-compat INTERFACE stdc++fs)
+
+cmake_push_check_state()
+# Check how to use std::filesystem
+string(CONFIGURE [[
+  #include <cstdlib>
```
halfflat, line 29: For `printf` this should be `#include <cstdio>`, and `std::printf` below.

cmake/CompilerOptions.cmake
```diff
@@ -18,20 +21,60 @@ if(${ARBDEV_COLOR})
...
+  #include <filesystem>
+  int main() {
+    auto cwd = std::filesystem::current_path();
+    printf("%s %d", cwd.c_str(), std::filesystem::exists(cwd));
+  }
+  ]] arb_cxx_fs_test @ONLY)
+
+#Test whether we can run probes
+set(STD_FS_LIB "")
+set(CMAKE_REQUIRED_FLAGS -std=c++17 ${CMAKE_REQUIRED_FLAGS})
```
halfflat, line 39: Should this be adding to an existing value for `CMAKE_REQUIRED_FLAGS`? I don't think it is set by anything in CMake, so it would have to be something the user provided…

halfflat (commented):
Looks good, but two minor nits in the comments.

cmake/CompilerOptions.cmake
thorstenhater: True.

cmake/CompilerOptions.cmake
thorstenhater: Yes, this variable is used by other feature tests. The reason why I can simply manipulate this 
setting are the push/pop macros around this block. In all honesty, the CMAKE_REQUIRED_LIBS
should also be saved and appended to in turn. Will add that tomorrow.
_Not_ using push_state/pop_state led to weird errors elsewhere in the CMake wilderness.

halfflat: Answered in private chat, thanks @thorstenhater! It's good practice because a script such as this may be called in the context of a Find module invoked from the generated CMakeLists.txt from an enclosing try_compile. Wow.

halfflat: Sorry for the out-of-timeline comment; didn't see your reply until just now. (I don't know if it was GH, or me, to be honest.)

thorstenhater:
bors try

halfflat (approved):
Great! Hopefully this will resolve all the fslib issues!

PR #1397, opened 2021-02-23
Remove all mention of `initial_ion_data`

### LINKED ISSUE #1396: Remove `initial_ion_data`
The `initial_ion_data` struct was previously used to paint/set_default ion properties on a cell. It predates the `init_int_concentration`, `init_ext_concentration` and `init_reversal_potential` structs, which can now also be used for the same purpose. 
It can be useful to paint/set_default ion properties in bulk, but I think it's not necessary and we can remove this struct. It's not used in many places in the repository.

noraabiakar:
Closes #1396


halfflat (approved):
Nice clean up!

PR #1399, opened 2021-02-24
Binary Python wheels and sdist generated by Github Action

### LINKED ISSUE #1326: Have Python wheel
Having a Python wheel would speed up installation and remove the requirement for users to have a compilation pipeline set up.

Currently, if you pip install arbor, you will see output like:
```
Building wheels for collected packages: arbor
  Building wheel for arbor (setup.py) ... done
  Created wheel for arbor: filename=arbor-0.5-cp38-cp38-linux_x86_64.whl size=1322418 sha256=83fc9fe7d03f7afc73f69210bb9ea4373b08ee5d4c66180b9cfbb52c1aad35ac
  Stored in directory: /p/home/jusers/huisman1/jusuf/.cache/pip/wheels/09/18/68/ceac2887642e9c2e49a20630e3e5a94572bec0a994058e82dc
Successfully built arbor
```
If you take that wheel, it's perfectly installable in other envs/machines right now (just tested), as long as you match the platform. Turning that into manylinux (most recent version is 2014, see [PEP 599](https://www.python.org/dev/peps/pep-0599/)) seems like it shouldn't be so difficult (famous last words, I know).

brenthuisman:
## Changes

* Auto-generate binary wheels, triggered by git tags 'v*'.
  * Wheels are generated for Python versions 3.6-3.9, for Linux using the [PyPA manylinux-2014 image](https://github.com/pypa/manylinux) and MacOS using macos-latest (10.15 at time of writing).
  * The Action generates `sdist` for any other platforms.
  * The cibuildwheel action automatically prepares the wheels using [auditwheel](https://cibuildwheel.readthedocs.io/en/stable/options/#repair-wheel-command), which means bundling of external deps (e.g. libxml2 for NeuroML).
  * The output is all that's needed for publication of a new version to (Test)PyPI.
* `setup.py` defaults `neuroml` to on.
* A small change to `CMakeLists.txt` was required to build the wheels: based on this [Pybind hint](https://pybind11.readthedocs.io/en/stable/compiling.html#findpython-mode) `Development.Module`is now required, and not `Development` when CMake searches for Python. I confirm this change is needed to make Arbor build in the manylinux docker images, but I do not understand what the change implies, other than that we now don't link to `libpython`.
* (Unrelated to the above) Only run the `ebrains.yml` action on commits to master.

## Not changed

* Vectorization is off
  * Arbor's vectorization requires AVX (on x86-64), but that might be too much of a demand on all users (Intel CPUs >2010, AMD CPUs >2013 come with AVX), so I've left it `off` right now.
  * Reference for archs: https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html
    * Building Arbor with vectorization on but an arch without AVX support fails.

## Example

Example output (scroll down): https://github.com/brenthuisman/arbor/actions/runs/614579564
The artifacts can be download from the Github Action run, and you'll be just a `twine` away from uploading a new version.  

Fixes #1326

### DIFF
--- a/.github/workflows/ciwheel.yml
+++ b/.github/workflows/ciwheel.yml
@@ -100,7 +100,7 @@ jobs:
         with:
           output-dir: dist
         env:
-          CIBW_BEFORE_BUILD: pip install numpy setuptools
+          CIBW_BEFORE_BUILD: python -m pip install numpy setuptools
           CIBW_BUILD: "cp3?-macosx_x86_64"
           CIBW_SKIP: "cp35-*"
           CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
@@ -111,13 +111,13 @@ jobs:
         with:
           output-dir: dist
         env:
-          CIBW_BEFORE_BUILD: pip install numpy setuptools
+          CIBW_BEFORE_BUILD: python -m pip install numpy setuptools
           CIBW_BUILD: "cp3?-macosx_x86_64"
           CIBW_SKIP: "cp35-*"
 
-        run: |
-          pip install cibuildwheel
-          cibuildwheel --output-dir dist
+      - uses: actions/upload-artifact@v2
+        with:
+          path: dist/*.whl
 
       - name: Publish distribution 📦 to Test PyPI
         if: github.ref == 'refs/heads/master'
--- a/.github/workflows/ciwheel.yml
+++ b/.github/workflows/ciwheel.yml
@@ -1,4 +1,4 @@
-name: Arbor
+name: Arbor in Wheels
 
 on:
   push:
@@ -8,8 +8,8 @@ on:
 
 jobs:
   build_wheels:
-    name: Build wheels on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
+    name: Build wheels on ${{ matrix.config.os }}
+    runs-on: ${{ matrix.config.os }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ciwheel.yml
+++ b/.github/workflows/ciwheel.yml
@@ -61,8 +61,6 @@ jobs:
           $CC --version
           $CXX --version
           python --version
-          mpic++ --show
-          mpicc --show
           echo $PYTHONPATH
 
       ##cibuildwheel setup
--- a/.github/workflows/ciwheel.yml
+++ b/.github/workflows/ciwheel.yml
@@ -52,32 +52,6 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.config.py }}
-      - name: OpenMPI cache
-        uses: actions/cache@v2
-        id:   cache-ompi
-        with:
-          path: ~/openmpi-4.0.2
-          key:  ${{ matrix.config.os }}-openmpi-4.0.2-${{ matrix.config.cxx }}
-      - name: Build OpenMPI
-        if: ${{ steps.cache-ompi.outputs.cache-hit != 'true' }}
-        run: |
-           echo cache-hit='${{ steps.cache-ompi.outputs.cache-hit }}'
-           cd ~
-           wget https://download.open-mpi.org/release/open-mpi/v4.0/openmpi-4.0.2.tar.gz
-           tar -xvf ./openmpi-4.0.2.tar.gz
-           cd openmpi-4.0.2
-           ./configure --disable-mpi-fortran
-           make -j4
-      - name: Install OpenMPI
-        run: |
-           echo "Going to install ompi"
-           cd ~
-           cd openmpi-4.0.2
-           sudo make install
-           cd -
-      - name: Update shared library cache
-        if: ${{ startsWith(matrix.config.os, 'ubuntu') }}
-        run: sudo ldconfig
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
--- a/.github/workflows/ciwheel.yml
+++ b/.github/workflows/ciwheel.yml
@@ -24,23 +24,20 @@ jobs:
             mpi:   "ON",
             simd:  "OFF"
           }
-        - {
-            name:  "MacOS Max",
-            os:    "macos-10.15", # TODO: 11.0 is still private preview, fix later.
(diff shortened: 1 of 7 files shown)


Helveg:
Lovin the commit messages, CI fidgeting, yay. On the topic of the `arbor` vs `testarbor` I'd say use the `dev` version tag. PyPI's semver allows you to give a version like `0.5dev123` for exactly this purpose.

brenthuisman:
I think we're already doing that (see `VERSION`). However, @bcumming has been doing all PyPI testing and releasing manually, so this represents a change of procedure. Also, so far `VERSION` is updated by hand.

CMakeLists.txt
```diff
@@ -246,7 +246,11 @@ endif()
 if(ARB_WITH_PYTHON)
     cmake_dependent_option(ARB_USE_BUNDLED_PYBIND11 "Use bundled pybind11" ON "ARB_WITH_PYTHON;ARB_USE_BUNDLED_LIBS" OFF)
 
-    find_package(Python3 ${arb_py_version} COMPONENTS Interpreter Development REQUIRED)
+    if(DEFINED ENV{CIBUILDWHEEL})
+        find_package(Python3 ${arb_py_version} COMPONENTS Interpreter Development.Module REQUIRED)
```
thorstenhater, line 250: This conditional seems weird, I would expect that we can always load either Development.Module, Development, or both and have success?

doc/install/python.rst
```diff
@@ -73,7 +73,7 @@ be used to configure the installation:
 * ``--mpi``: Enable MPI support (requires MPI library).
 * ``--gpu``: Enable GPU support for NVIDIA GPUs with nvcc using ``cuda``, or with clang using ``cuda-clang`` (both require cudaruntime).
   Enable GPU support for AMD GPUs with hipcc using ``hip``. By default set to ``none``, which disables gpu support.
-* ``--vec``: Enable vectorization. This might require choosing an appropriate architecture using ``--arch``.
+* ``--vec``: Enable vectorization. This might require choosing an appropriate architecture using ``--arch``. Note that on x86-64 platforms compilation will fail if you enable vectorization, but the CPU or ``--arch`` does not support any form of AVX.
```
thorstenhater, line 76: ... but defaults to the detected 'native', see below.

setup.py
```diff
@@ -163,8 +213,12 @@ def run(self):
     zip_safe=False,
     ext_modules=[cmake_extension('arbor')],
     cmdclass={
-        'build_ext': cmake_build,
-        'install':   install_command,
+        'build_ext':   cmake_build,
+        'install':     install_command,
+        'bdist_wheel': bdist_wheel_command,
+    } if WHEEL_INSTALLED else {
```
thorstenhater, line 219: Maybe store `cmdclass` in a variable set in the first if/else above?

.github/workflows/ciwheel.yml
```diff
@@ -0,0 +1,91 @@
...
+        uses: joerick/cibuildwheel@v1.9.0
+        with:
+          output-dir: dist
+        env:
+          MACOSX_DEPLOYMENT_TARGET: "10.15" #needed to undo some CIBW settings
+          CIBW_BEFORE_BUILD: python -m pip install numpy setuptools
+          CIBW_BUILD: "cp3?-macosx_x86_64"
+          CIBW_SKIP: "cp35-*"
+          CIBW_ARCHS_MACOS: x86_64 universal2
+          CIBW_BUILD_VERBOSITY: 3
```
thorstenhater, line 47: That's left from debugging?

.github/workflows/ciwheel.yml
```diff
@@ -0,0 +1,91 @@
...
+          # CIBW_TEST_COMMAND: TODO
+
+      - name: Build wheels macos
+        if: ${{ startsWith(matrix.os, 'macos') }}
+        uses: joerick/cibuildwheel@v1.9.0
+        with:
+          output-dir: dist
+        env:
+          MACOSX_DEPLOYMENT_TARGET: "10.15" #needed to undo some CIBW settings
+          CIBW_BEFORE_BUILD: python -m pip install numpy setuptools
```
thorstenhater, line 43: This (and CIBW_SKIP) can be set once on the job level env

thorstenhater: In fact, it's probably cleaner to use not the cartesian product matrix, but instead the config variant to set these variables (see basic.yml for how this looks like). The macos specifics can be set always (or, if we are worried about pollution using `run ::set ENV=VAR`)

thorstenhater (changes requested):

CMakeLists.txt
brenthuisman: Unfortunately, local builds with `Development.Module` fail, and in-container builds fail with `Development` (this seems to be on purpose). The only piece of documentation I could find is what I linked, but I don't understand it, only that it works.

setup.py
brenthuisman: OK

.github/workflows/ciwheel.yml
brenthuisman: Yep

.github/workflows/ciwheel.yml
brenthuisman: The Action runs this step in a docker image in the case of Linux (the manylinux2014 image) that does not inherit those things. It's clearer that this step is 'special' by keeping it contained like this.

Come to think of it, I probably don't need to setup Python.

doc/install/python.rst
brenthuisman: I know, but if you specify an arch that hasn't AVX, build will fail. That's a gotcha that I though worth mentioning.

setup.py
brenthuisman: Turns out this creates a problem: https://github.com/brenthuisman/arbor/runs/2031776731?check_suite_focus=true

.github/workflows/ciwheel.yml
thorstenhater: Seems weird that the same keyword has different scope, but well.

thorstenhater (approved):
Looks good, issues have been clarified.

Issue #820, opened 2019-07-15
Fix up our GPU back-end FVM/cable cell tests


halfflat:
They're a bit of a mess!

Ideally, any test currently that checks state or time evolution for cable cells should be able to run with both the cpu and gpu back-ends, and the built unit test should run them on both, if we're GPU-enabled and there is a GPU present.

halfflat:
The alternative is to make two unit test runs: the runner sets some state or flag that determines the cpu or gpu execution context.  Each specific test then runs with that context, and anything that is truly GPU-specific checks the context first.

noraabiakar:
Our unit tests don't always go through the `execution_context`, there are specific parts of the backend code that are (or should be) explicitly tested on the GPU/CPU, for example: matrix assemble/solve and  mechanism instantiation. Are those tests in a good state, or is one of the backends more covered than the other? Will it be enough (to your knowledge) to make sure that all *existing* unit tests that require an `execution_context` run twice with and without GPU support (when possible) or are there some hidden assumptions made in the code about the backend which will require refactoring/extending?

PR #1401, opened 2021-02-25
Bug/multiplicity

thorstenhater:
Fix a bug in GPU mechanism init that would have overwritten ion indices if in-place multiplication was enabled.
Also refactor the surrounding code to avoid similar bugs in the future.

### DIFF
--- a/arbor/backends/gpu/mechanism.cpp
+++ b/arbor/backends/gpu/mechanism.cpp
@@ -152,17 +152,13 @@ void mechanism::instantiate(unsigned id,
 
     auto base_ptr = indices_.data();
 
-    auto append_chunk = [&](const auto& input, auto& base_ptr, auto& output) {
+    auto append_chunk = [&](const auto& input, auto& output) {
         memory::copy(make_const_view(input), device_view(base_ptr, width_));
         output = base_ptr;
         base_ptr += width_padded_;
     };
 
-    append_chunk(pos_data.cv, base_ptr, pp->node_index);
-
-    // memory::copy(make_const_view(pos_data.cv), device_view(base_ptr, width_));
-    // pp->node_index_ = base_ptr;
-    // base_ptr += width_padded_;
+    append_chunk(pos_data.cv, pp->node_index);
 
     auto ion_index_tbl = ion_index_table();
     arb_assert(num_ions_==ion_index_tbl.size());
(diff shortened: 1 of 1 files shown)


thorstenhater:
bors try

thorstenhater:
bors try

noraabiakar (commented):
Looks good! Can we add a unit test for the bug?

arbor/backends/gpu/mechanism.cpp
```diff
@@ -170,15 +173,15 @@ void mechanism::instantiate(unsigned id,
...
-        auto index_start = indices_.data()+(i+1)*width_padded_;
-        ion_index_ptr = index_start;
-        memory::copy(make_const_view(mech_ion_index), device_view(index_start, width_));
+        memory::copy(make_const_view(mech_ion_index), device_view(base_ptr, width_));
+        *ion_ptr = base_ptr;
+        base_ptr += width_padded_;
     }
 
     if (mult_in_place_) {
-        memory::copy(make_const_view(pos_data.multiplicity), device_view(indices_.data() + width_padded_, width_));
```
noraabiakar, line 180: The bug is here: should have been `device_view(indices_.data() + (num_ions_+1)*width_padded_` like below.

thorstenhater: Yes. That's fixed by stepping `base_ptr` everytime a chunk is appended.

thorstenhater:
I think it would make sense to have a test here, but I also think we should postpone it to another PR:
- Currently we have no mechanism unit tests #1402 
- Our GPU tests are very sparse #820
As this is a GPU issue I would suggest waiting until #820 is resolved and consider this part of #1402.
That said, we should make #820 a priority.

arbor/backends/gpu/mechanism.cpp
noraabiakar: I think your refactoring makes things easier to reason about. It just took me a second to find the bug that's why I made the comment :)

noraabiakar (approved):

arbor/backends/gpu/mechanism.cpp
```diff
@@ -170,15 +173,15 @@ void mechanism::instantiate(unsigned id,
...
-        pp->multiplicity_ = indices_.data() + (num_ions_ + 1)*width_padded_;
+        memory::copy(make_const_view(pos_data.multiplicity), device_view(base_ptr, width_));
+        pp->multiplicity_ = base_ptr;
+        base_ptr += width_padded_; // Theoretically redundant, but for consistency
```
noraabiakar, line 184: minor comment: this line is not needed right?

thorstenhater: Yes, therefore the comment. It's more consistent and also makes it safer to tack on another chunk.

arbor/backends/gpu/mechanism.cpp
thorstenhater: 👍

thorstenhater:
bors try

thorstenhater:
bors try

thorstenhater:
bors try

arbor/backends/gpu/mechanism.cpp
thorstenhater: I did.

noraabiakar (approved):
lgtm!

PR #1404, opened 2021-02-25
Port of Brunel example

### LINKED ISSUE #1393: Python version of Brunel example.
There are currently no examples of LIF cells and networks in Python, which would also provide examples of using recipes with a variety of cell kinds.

A Python version of the C++ Brunel example would provide such an example and also exercise an undertested part of the Python interface.

brenthuisman:
* Python version of brunel.cpp
* Consistency in units: both C++ and Python APIs use unit `kHz` for all frequencies.
  * C++ `poisson_schedule` expected `kHz`, other schedules already expect `Hz`.
  * Docs updated.
* Consistency in function signature: both C++ and Pythons `poisson_schedule` no longer defaults frequency.
  * Python unit tests adapted.
* Tiny fix in brunel.cpp (ignored flag) and some cosmetic updates.
* A few minor doc additions related to networks and interconnectivity.
* Add `brunel.py` to `basic.yml` tests

Fixes #1393 #1420


python/example/brunel.py
```diff
@@ -0,0 +1,161 @@
...
+    parser.add_argument('-l', '--lambda', dest='poiss_lambda', type=float, default=1, help='Expected number of spikes from a single poisson cell (kHz)')
+    parser.add_argument('-t', '--tfinal', dest='tfinal', type=float, default=100, help='Length of the simulation period (ms)')
+    parser.add_argument('-s', '--dt', dest='dt', type=float, default=1, help='Simulation time step (ms)')
+    parser.add_argument('-G', '--group-size', dest='group_size', type=int, default=10, help='Number of cells per cell group')
+    parser.add_argument('-S', '--seed', dest='seed', type=int, default=42, help='Seed for poisson spike generators')
+    parser.add_argument('-f', '--write-spikes', dest='spike_file_output', action='store_true', help='Save spikes to file')
+    parser.add_argument('-z', '--profile-rank-zero', dest='profile_only_zero', action='store_true', help='Only output profile information for rank 0')
+    parser.add_argument('-v', '--verbose', dest='verbose', action='store_true', help='Print more verbose information to stdout')
```
halfflat, line 119: These two options `--profile-rank-zero` and `--verbose` don't actually do anything in this example.

python/example/brunel.py
```diff
@@ -0,0 +1,161 @@
...
+    meters.checkpoint('simulation-run', context)
+
+    # Print profiling information
+    print(f'{arbor.meter_report(meters, context)}')
+
+    # Print spike times
+    print(f"{len(sim.spikes())} spikes generated.")
+
+    if opt.spike_file_output:
+        with open('spikes.txt', 'w') as the_file:
```
halfflat, line 159: The C++ example writes to `spikes.gdf` instead. Ideally, neither this example nor the C++ one would write to a hardcoded path, but instead use a path supplied as an argument to `-f`.

halfflat (changes requested):
Some minor requests in comments on the code.
But a big problem is that the output is non-deterministic, at least when I tried it out, with it reporting 6996, 6997, or 6998 spikes.

python/schedule.cpp
```diff
@@ -236,15 +241,19 @@ void register_schedules(py::module& m) {
...
     poisson_schedule
         .def(py::init<time_type, time_type, std::mt19937_64::result_type>(),
-            "tstart"_a = 0., "freq"_a = 10., "seed"_a = 0,
+            "tstart"_a = 0., "freq"_a, "seed"_a = 0,
             "Construct a Poisson schedule with arguments:\n"
             "  tstart: The delivery time of the first event in the sequence [ms], 0 by default.\n"
-            "  freq:   The expected frequency [Hz], 10 by default.\n"
+            "  freq:   The expected frequency [kHz], 10 by default.\n"
             "  seed:   The seed for the random number generator, 0 by default.")
+        .def(py::init<time_type>(),
```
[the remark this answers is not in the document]
brenthuisman: Let's, because it should be done consistently.

.github/workflows/basic.yml
```diff
@@ -165,6 +165,7 @@ jobs:
             python python/example/single_cell_model.py
             python python/example/single_cell_recipe.py
             python python/example/single_cell_stdp.py
+            python python/example/brunel.py -n 400 -m 100 -e 20 -p 0.1 -w 1.2 -d 1 -g 0.5 -l 5 -t 100 -s 1 -G 50 -S 123
```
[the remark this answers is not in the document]
brenthuisman: We don't check the outputs, so it does not really matter what the params are, but this way we test if they're all settable. brunel.cpp had a typo in its args (capital vs lowercase `s`) which might have been spotted this way.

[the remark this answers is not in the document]
halfflat: Fair enough! My concern was that we should be testing something similar to what a random user might try running — having all the arguments explicitly set looks a bit intimidating :)

python/schedule.cpp
```diff
@@ -148,6 +148,12 @@ poisson_schedule_shim::poisson_schedule_shim(
     seed = s;
 }
 
+poisson_schedule_shim::poisson_schedule_shim(arb::time_type f) {
+    set_tstart(0.);
+    set_freq(f);
+    seed = 0;
+}
+
```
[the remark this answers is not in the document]
halfflat: Turns out the shim classes are a bit of a messy solution. What you have is great :)

brenthuisman (commented):
For posterity: why not have py_init with a lambda function for constructor specializations:

I'm running into the limits of my C++ foo:

        .def(py::init<>(
            [](arb::time_type f){
                return poisson_schedule_shim(0.,f,0);}),
            "freq"_a,
            "Construct a Poisson schedule, starting from t = 0, default seed, with:\n"
            "  freq:   The expected frequency [kHz], 10 by default.\n")

generates a compilation error:

    /home/brent/sync/julich/code/arbor/python/pybind11/include/pybind11/detail/init.h:150:59: fout: static assertion failed: pybind11::init() return-by-value factory function requires a movable class
      150 |     static_assert(std::is_move_constructible<Cpp<Class>>::value,
          |                                                           ^~~~~
    /home/brent/sync/julich/code/arbor/python/pybind11/include/pybind11/detail/init.h:155:27: fout: use of deleted function 'pyarb::poisson_schedule_shim::poisson_schedule_shim(pyarb::poisson_schedule_shim&&)'
      155 |         v_h.value_ptr() = new Cpp<Class>(std::move(result));
          |                           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    In file included from /home/brent/sync/julich/code/arbor/python/schedule.cpp:8:
    /home/brent/sync/julich/code/arbor/python/schedule.hpp:78:8: note: 'pyarb::poisson_schedule_shim::poisson_schedule_shim(pyarb::poisson_schedule_shim&&)' is implicitly deleted because the default definition would be ill-formed:
       78 | struct poisson_schedule_shim: schedule_shim_base {
          |        ^~~~~~~~~~~~~~~~~~~~~
    /home/brent/sync/julich/code/arbor/python/schedule.hpp:78:8: fout: use of deleted function 'pyarb::schedule_shim_base::schedule_shim_base(const pyarb::schedule_shim_base&)'
    /home/brent/sync/julich/code/arbor/python/schedule.hpp:19:5: note: declared here
       19 |     schedule_shim_base(const schedule_shim_base&) = delete;
          |     ^~~~~~~~~~~~~~~~~~

event_generator_shim and spike_source_cell are simple structs, might that be why this works for them?

python/example/brunel.py
brenthuisman: Correct. This is inherited from `brunel.cpp` ;)

python/example/brunel.py
```diff
@@ -0,0 +1,161 @@
...
+    parser.add_argument('-S', '--seed', dest='seed', type=int, default=42, help='Seed for poisson spike generators')
+    parser.add_argument('-f', '--write-spikes', dest='spike_file_output', action='store_true', help='Save spikes to file')
+    parser.add_argument('-z', '--profile-rank-zero', dest='profile_only_zero', action='store_true', help='Only output profile information for rank 0')
+    parser.add_argument('-v', '--verbose', dest='verbose', action='store_true', help='Print more verbose information to stdout')
+
+    for k,v in vars(parser.parse_args()).items():
+        setattr(opt, k, v)
+
+    context = arbor.context()
+    print(context)
```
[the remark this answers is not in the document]
brenthuisman: I find it quite concise and clear! Maybe a comment could solve this (print some concise info about how Arbor is running)?

Alternatively, we might improve the way `__str__` prints the info (a full title per field).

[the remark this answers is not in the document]
halfflat: The `__str__` representations definitely need a look at, but I think this falls out of the scope of this PR. (It's not just the context, of course). In the interim, we can't really rely on the printing of any particular arbor object to be especially useful, and I think it's better to take charge of the presentation of that information in user-side code, at least until we clean up the Python lib's `__repr__` and `__str__` calls.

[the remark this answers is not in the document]
brenthuisman: I've made an issue to track updating `__str__`, #1433 . But I'm not sure what you'd like to see here, the particular objects printed here do show their parameters, which is all I can imagine that is relevant.

python/schedule.hpp
```diff
@@ -78,9 +78,9 @@ struct explicit_schedule_shim: schedule_shim_base {
...
     using rng_type = std::mt19937_64;
 
-    arb::time_type tstart = arb::terminal_time;
-    arb::time_type freq = 10.;  // kHz
-    rng_type::result_type seed = 0;
+    arb::time_type tstart; // ms
+    arb::time_type freq; // kHz
+    rng_type::result_type seed;
 
     poisson_schedule_shim() = default;
```
halfflat, line 85: Without in-line defaults, this constructor will do whatevs, and initialize the object in an undetermined fashion. Remove this ctor too?

brenthuisman: yes!

python/example/brunel.py
```diff
@@ -0,0 +1,159 @@
...
+class brunel_recipe (arbor.recipe):
+    def __init__(self, nexc, ninh, next, in_degree_prop, weight, delay, rel_inh_strength, poiss_lambda, seed = 42):
+
+        arbor.recipe.__init__(self)
+        # self.props = arbor.neuron_cable_properties()
+        # self.cat = arbor.default_catalogue()
+        # self.props.register(self.cat)
+
+        # Make sure that in_degree_prop in the interval (0, 1]
+        if 0.0>=in_degree_prop>1.0:
```
halfflat, line 41: This test is always false (and indeed, something like `-p 1.1` gives us an infinite loop.)

python/example/brunel.py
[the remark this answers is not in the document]
halfflat: Fine by me!

python/example/brunel.py
```diff
@@ -0,0 +1,159 @@
...
+    # Print profiling information
+    print(f'{arbor.meter_report(meters, context)}')
+
+    # Print spike times
+    print(f"{len(sim.spikes())} spikes generated.")
+
+    if opt.spike_file_output:
+        with open(opt.spike_file_output, 'w') as the_file:
+            for sp in sim.spikes():
+                the_file.write('{:3.3f}\n'.format(sp[-1]))
```
halfflat, line 159: In the C++ example, the spiking cell gid is also printed. It's a small change, and would bring the python example more in line with the C++ one.

python/example/brunel.py
```diff
@@ -0,0 +1,159 @@
...
+
+Call with parameters, for example:
+./brunel.py -n 400 -m 100 -e 20 -p 0.1 -w 1.2 -d 1 -g 0.5 -l 5 -t 100 -s 1 -G 50 -S 123 -f spikes.txt
+
+'''
+
+# Samples m unique values in interval [start, end) - gid.
+# We exclude gid because we don't want self-loops.
+def sample_subset(gid, start, end, m):
+    gen = numpy.random.default_rng(gid+42)
```
halfflat, line 24: The only issue here is that this interface is new in numpy 1.17, and the latest version available in e.g. Debian stable is 1.16.2. The old interface is still preserved under `numpy.random.RandomState`, and we could get 1.16 compatibility with e.g.
```
gen = numpy.random.RandomState(gid+42)
…
val = gen.randint(low=start, high=end)
```

That said though, `Generator.integers` will pick numbers from [low, high) unless `endpoint=True`, and it is by default false, so line 27 should be, IIUC, `val = gen.integers(low=start, high=end)` in the new `numpy.random` interface.

halfflat (changes requested):
Looks good, but there's one bug (that I spotted at least!) and a numpy compatibility issue.

python/example/brunel.py
brenthuisman: Whoops!

python/example/brunel.py
```diff
@@ -33,12 +33,9 @@ class brunel_recipe (arbor.recipe):
...
 
         arbor.recipe.__init__(self)
-        # self.props = arbor.neuron_cable_properties()
-        # self.cat = arbor.default_catalogue()
-        # self.props.register(self.cat)
 
         # Make sure that in_degree_prop in the interval (0, 1]
-        if 0.0>=in_degree_prop>1.0:
+        if not 0.0<=in_degree_prop<1.0:
             print("The proportion of incoming connections should be in the interval (0, 1].")
```
halfflat, line 39: This should be `not 0.0<in_degree_prop<=1.0`, for agreement with the error message, and for consistency with the C++ version.

brenthuisman: :+1:

halfflat (approved):
All good!

PR #1407, opened 2021-02-26
Move current clamps to shared state.

### LINKED ISSUE #1262: Account for stimulus currents in total transmembrane current probes.
Currently cable cell per-CV total transmembrane currents represent _all_ non intracellular fluxes, and this includes stimulus currents. These should _not_ be included, however, for the purposes of computing LFP.

I propose one of:
* Remove stimulus currents explicitly from total current reporting, by tracking their contribution in a sparse per-CV representation and doing the sum in the cable cell group implementation, or
* Track the contribution, but report it via a different probe query, or
* Have two versions of the per-CV total current probe, one with and one without stimulus currents.

The first option is definitely the easiest from the point of view of providing the functionality required for LFP calculation.

halfflat:
* Change implementation of current clamp from a builtin mechanism to  functionality provided by the back-end shared state classes.
* Extend clamp functionality to cover more use cases: it now supports DC  and AC stimuli governed by a piecewise-linear envelope.
* Update Python iclamp wrapper to suit; add unit test.
* Account for total applied stimulus currents separately, and remove their contribution to the reported transmembrane current and   ionic (i.e. non-capacitive) transmembrane current.
* Add new probe specifically for tracking total current stimulus in a cell.
* Update docs, and add an entry for current clamps to the concepts doc.

Fixes #1262.


doc/concepts/decor.rst
```diff
@@ -287,6 +287,27 @@ See :ref:`modelgapjunctions`.
 4. Stimuli
 ~~~~~~~~~~
 
+A current stimulus is described by a *frequency* in Hertz and an amplitude *envelope*.
```
thorstenhater, line 290: I think it is unclear what the frequency actually _does_ here. Could you explain?
Maybe even drawing the resulting waveforms.

doc/concepts/decor.rst
```diff
@@ -287,6 +287,27 @@ See :ref:`modelgapjunctions`.
...
+
+Stimulus objects in the C++ and Python interfaces have simple constructors for describing
+constant stimuli and constant amplitude stimuli restricted to a fixed time interval.
+
+.. code-block:: Python
+
+    # Constant stimulus, amplitude 10 nA.
+    decor.place('(root)', arbor.iclamp(10))
+
+    # Constant ampltidude 10 nA stimulus at 20 Hz.
```
thorstenhater, line 305: Typo: amplitude.

arbor/mc_cell_group.cpp
```diff
@@ -241,6 +241,51 @@ void run_samples(
...
+    sample_records.clear();
+
+    auto& tmp = std::get<std::vector<double>>(scratch);
+    tmp.clear();
+    tmp.reserve(n_interp_per_sample*n_sample);
+
+    for (sample_size_type j = 0; j<n_sample; ++j) {
+        auto offset = j*n_raw_per_sample+sc.begin_offset;
+        for (sample_size_type i = 0; i<n_interp_per_sample; ++i) {
+            tmp.push_back(raw_samples[offset+i]*p.coef[0][i]+raw_samples[offset+n_interp_per_sample+i]*p.coef[1][i]);
```
thorstenhater, line 270: Maybe untangle this a bit by putting `offset + i` into variable, it's quite dense.

test/unit/test_probe.cpp
```diff
@@ -1044,16 +1052,16 @@ void run_total_current_probe_test(const context& ctx) {
     // at the fork points, and once with a non-trivial CV centred on the fork
     // point.
 
-    trace_data<std::vector<double>, mcable_list> traces[2];
+    trace_data<std::vector<double>, mcable_list> traces[2];
```
thorstenhater, line 1055: Whitespace?

arbor/backends/gpu/shared_state.cpp
```diff
@@ -86,6 +87,81 @@ void ion_state::reset() {
...
+    // Initialize ppack pointers.
+    ppack_.accu_index = accu_index_.data();
+    ppack_.accu_to_cv = accu_to_cv_.data();
+    ppack_.frequency = frequency_.data();
+    ppack_.envl_amplitudes = envl_amplitudes_.data();
+    ppack_.envl_times = envl_times_.data();
+    ppack_.envl_divs = envl_divs_.data();
+    ppack_.accu_stim = accu_stim_.data();
+    ppack_.envl_index = envl_index_.data();
+    ppack_.time = nullptr; // These fields must be set in add_current() before queing kernel.
```
thorstenhater, line 140: 1. Typo: queuing?
2. Why?
3. 'These' implies more than one, which exactly?

arbor/builtin_mechanisms.cpp
```diff
@@ -9,21 +9,9 @@
 
 namespace arb {
 
-template <typename B>
-concrete_mech_ptr<B> make_builtin_stimulus();
-
 mechanism_catalogue build_builtin_mechanisms() {
```
thorstenhater, line 12: Are we sure we cannot delete this completely?

arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -892,10 +924,44 @@ void resolve_probe(const cable_probe_total_current_cell& p, probe_resolution_dat
...
+                r.metadata.push_back(cable);
+            }
+        }
+    }
+
     r.shrink_to_fit();
     R.result.push_back(std::move(r));
 }
 
+
```
thorstenhater, line 964: Whitespace?

thorstenhater (commented):
Some minor comments and:

Why only sine waves? I can imagine interest in at least square waves, but possibly other forms (sine + phase, sawtooth, triangle, ...)

doc/concepts/decor.rst
halfflat: I'm not going to draw a waveform :) But I can use terms like AC and DC to make this clearer.

arbor/backends/gpu/shared_state.cpp
halfflat: 1. Definitely ytpo.
2. Because otherwise they will be null? For time, it's because it is being flipped around with time_to every timestep, but for the others, it's just a division of responsibility: the shared state code knows what it is doing with the shared vectors it is creating (especially `time`), so let's leave it to that code to do the requisite setting.
3. This line and the following, but I can make that more clear.

arbor/builtin_mechanisms.cpp
halfflat: Well, we could! I just didn't want to be hasty. Different PR perhaps?

halfflat:
Re: other waveforms: we can add them, but I want to have a use case in hand first, as we do have for sine waves.

The envelope can be used to describe any finite stimulus current (up to a certain degree of fidelity), but it may be useful to support periodic envelopes, so that freq=0 and a periodic envelope can describe any waveform you want. I didn't add this, because I wanted to keep things more simple, but it would not be that disruptive an extension. (The most annoying part would be the documentation.)

doc/concepts/decor.rst
```diff
@@ -287,6 +287,37 @@ See :ref:`modelgapjunctions`.
 4. Stimuli
 ~~~~~~~~~~
 
+A current stimilus is a DC or sinusoidal current of fixed frequemcy with a time-varying amplitude
+governed by a piecewise-linear envelope.
+
+The stimilus is described by two parameters: a frequency in Hertz, where a value of zero denotes DC;
```
thorstenhater, line 293: Typo: stimulus

halfflat: I am particularly impressed by how I managed to mistype it twice.

halfflat:
Rebased onto current master, rehomed doc additions.

thorstenhater (approved):
Looks good.

noraabiakar:
bors try

PR #1411, opened 2021-03-02
Extend available version information.

### LINKED ISSUE #900: Better VERSION defines
The `ARB_VERSION` macro defined in `version.hpp` is just a char string, which is fine for reporting, but terrible for compile-time testing.

We should also have macros that allow the testing of specific version numbers, e.g. `ARB_VERSION_MAJOR`, `ARB_VERSION_MINOR`, `ARB_VERSION_PATCHLEVEL`, and an additional macro `ARB_VERSION_DEV` which is e.g. defined only for `-dev` versions of Arbor.

Names of course are up for debate,

halfflat:
* Provide defines and variables for version string components: major version; minor version; patch level version; development suffix.
* Change commit date format to strict ISO.
* Add define/variable for Arbor build configuration.
* Add usage/help info to `git-source-id` script.
* Add define/variable for a full build identification string, incorporating source info, version, build configuration, feature options.
* Add documentation for the version and build defines and variables.
* Extend version unit tests to suit.

Fixes #900.

### DIFF
--- a/arbor/include/CMakeLists.txt
+++ b/arbor/include/CMakeLists.txt
@@ -50,10 +50,12 @@ if(ARB_VECTORIZE)
     list(APPEND arb_features VECTORIZE)
 endif()
 
+string(TOUPPER "${CMAKE_BUILD_TYPE}" arb_config_str)
+
 add_custom_command(
     OUTPUT version.hpp-test
     DEPENDS _always_rebuild
-    COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/git-source-id ${FULL_VERSION_STRING} ${ARB_ARCH} ${arb_features} > version.hpp-test
+    COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/git-source-id ${FULL_VERSION_STRING} ${ARB_ARCH} ${arb_config_str} ${arb_features} > version.hpp-test
 )
 
 set(version_hpp_path arbor/version.hpp)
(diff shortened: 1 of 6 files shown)


brenthuisman:
Two questions:
* I see a patch number: how do we increase that?
* For autosubmission of wheels to TestPyPI (coming in a followup PR to #1399) a unique version for every push is required. The biggest problem is how we generate that string, so we may as well discuss this now: is the increase of the minor number automatable with a Github Action? A nice solution to that problem would be to bump patch on every commit against master.

halfflat:
Everything is currently derived entirely from the contents of the `VERSION` file. For unique tags for TestPyPI wheels can we append arbitrary strings? One option would be to add a hash of the contents of e.g. `ARB_FULL_BUILD_ID`.

brenthuisman:
I _think_ it can be arbitrary. At first I thought of something like 0.5.dev-commithash. The problem is that this isn't guaranteed to increase, so pipping the 'latest' `arbor` package probably isn't the latest. You'd have to construct the version yourself and request that version from `pip`. Every hash will have this problem.

halfflat:
The TestPyPI wheel naming problem could probably be tackled separately, but the newly recorded info in `version.hpp` might be a way to accommodate those needs.

halfflat:
Rebased against current master.

noraabiakar (approved):
Looks good!

PR #1413, opened 2021-03-03
Clarify Python installation from CMake

### LINKED ISSUE #1403: Python installation through CMake improvements
Three ways to improve our Python installation when the user uses CMake:

* [ ] `python setup.py egg_info` generates an `arbor.egg-info` dir, which should be copied along with `arbor` when `make install` copies to `site-packages`. This will register the package to pip, and allows pip to manage it.
* [x] We should NOT be overriding the install destination (obtained with `sysconfig.get_path('platlib')`) when a CMAKE_INSTALL_PREFIX is provided. This ignores pip/conda environments, and is thus unexpected behavior for Python users. Destination for C++ artifacts is entirely separate for Python folks, but still may be needed (modcc).
* [ ] ARB_USE_BUNDLED_LIBS should default to ON. The default should be to go with our vetted and provided versions of dependencies.
  * Related: we don't check nonbundled deps such as `pybind` for versions. Should we? I'd like to avoid users [successfully compiling broken packages](https://github.com/arbor-sim/arbor/issues/1398).

bcumming:
Always install Python module inside `CMAKE_INSTALL_PREFIX`, unless the user explicitly sets the `ARB_PYTHON_LIB_PATH` CMake variable.

Update documentation on installing and building the Python packing using CMake to have a detailed example of how to install the Python module inside a virtual env.

Fixes #1403

### DIFF
--- a/doc/install/python.rst
+++ b/doc/install/python.rst
@@ -136,7 +136,7 @@ below demonstrate this for both pip and ``setup.py``.
     :ref:`Python configuration <install-python>` section of the
     :ref:`installation guide <in_build_install>`.
     CMake is recommended for developers, integration with package managers such as
-    Spack and EasyBuild, and users who require fine grained control over compilation
+    Spack and EasyBuild, and expert users who require fine grained control over compilation
     and installation.
 
 Dependencies
--- a/doc/install/python.rst
+++ b/doc/install/python.rst
@@ -136,8 +136,7 @@ below demonstrate this for both pip and ``setup.py``.
     :ref:`Python configuration <install-python>` section of the
     :ref:`installation guide <in_build_install>`.
     CMake is recommended for developers, integration with package managers such as
-    Spack and EasyBuild, and expert users who require fine grained control over compilation
-    and installation.
+    Spack and EasyBuild, and fine-grained control over compilation and installation.
 
 Dependencies
 ^^^^^^^^^^^^
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -81,12 +81,7 @@ file(COPY "${PROJECT_SOURCE_DIR}/VERSION" DESTINATION "${python_mod_path}")
 # Set the installation path
 
 # Ask Python where it keeps its system (platform) packages.
-
-if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-    file(WRITE "${CMAKE_BINARY_DIR}/install-prefix" "")
-else()
-    file(WRITE "${CMAKE_BINARY_DIR}/install-prefix" "${CMAKE_INSTALL_PREFIX}")
-endif()
+file(WRITE "${CMAKE_BINARY_DIR}/install-prefix" "${CMAKE_INSTALL_PREFIX}")
 
 execute_process(
     COMMAND ${PYTHON_EXECUTABLE} -c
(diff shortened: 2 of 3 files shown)


doc/install/python.rst
```diff
@@ -136,7 +136,7 @@ below demonstrate this for both pip and ``setup.py``.
     :ref:`Python configuration <install-python>` section of the
     :ref:`installation guide <in_build_install>`.
     CMake is recommended for developers, integration with package managers such as
-    Spack and EasyBuild, and users who require fine grained control over compilation
+    Spack and EasyBuild, and expert users who require fine grained control over compilation
```
brenthuisman, line 139: I'd like to be explicit: let's mention users who expect to use Arbor's C++ API.

python/CMakeLists.txt
```diff
@@ -81,12 +81,7 @@ file(COPY "${PROJECT_SOURCE_DIR}/VERSION" DESTINATION "${python_mod_path}")
...
-
-if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
-    file(WRITE "${CMAKE_BINARY_DIR}/install-prefix" "")
-else()
-    file(WRITE "${CMAKE_BINARY_DIR}/install-prefix" "${CMAKE_INSTALL_PREFIX}")
-endif()
+file(WRITE "${CMAKE_BINARY_DIR}/install-prefix" "${CMAKE_INSTALL_PREFIX}")
 
 execute_process(
     COMMAND ${PYTHON_EXECUTABLE} -c
```
brenthuisman, line 87: This still sets the python package destination by default to what sysconfig says, not `${CMAKE_INSTALL_PREFIX}/lib/python3.9/site-packages`

brenthuisman (changes requested):

doc/install/python.rst
bcumming: This is the installation guide for the Python wrapper, not the C++ installation guide.
So, this is targeting users who want fine-grained control over compilation options for their Python installation.

python/CMakeLists.txt
bcumming: Actually, it sets it to `${CMAKE_INSTALL_PREFIX}/lib/python3.9/site-packages`.
The contents of `${CMAKE_BINARY_DIR}/install-prefix` are set to `CMAKE_INSTALL_PREFIX`.
That is passed into the Python script, which will append the `/lib/python3.x/site-packages`.
I just double checked, and it works.
Try yourself by hard-coding a non empty value for prefix:
```
 > python -c "import sys,sysconfig;pfx='/home/brent/fancy-neuro-env';print(sysconfig.get_path('platlib',vars={} if pfx=='' else {'base':pfx,'platbase':pfx}))"
/home/brent/fancy-neuro-env/lib/python3.9/site-packages
```

doc/install/python.rst
brenthuisman: Yes, but this like is telling them to look to the CMake build is they're 'expert users'. It's clearer if we say 'C++ users'.

python/CMakeLists.txt
brenthuisman: Right!

doc/install/python.rst
bcumming: Oops, sorry for being a bit slow to reply. My point is that it isn't specifically for C++ users: it is for all users who want fine grained control

It is for users of Python, C++ and applications that use Arbor, that want fine grained control over the installation.

It is also for HPC users and admins who use spack, etc.

brenthuisman: Then let's leave the word out altogether: 'users who require fine grained control' is clear enough.

bcumming: Done!

brenthuisman (approved):

PR #1415, opened 2021-03-05
Initialize contexts to the number of available threads on the current system by default

brenthuisman:


halfflat:
I don't think this is the way we should do it, based on how we've organized the code so far.
* The `arborenv` library has the routines already defined for thread and env var queries: `arbenv::get_env_num_threads()` and `arbenv::thread_concurrency()`.
* We have issue #988 up already, which hopes to solve the problem this issue addresses, but still within the context of using `arborenv` functionality.

Apart from the existing division of responsibilities between the libraries, one of the motivations for making the default context only ever a single thread was 'consistency-by-default': while we should make it easy for users to use the available computing resources, it is also worthwhile to have behaviour of the core library be, as much as possible. independent of the environment.

thorstenhater:
Also, that would give the illusion that more threads equal more performance. This is likely not true in the general case.

halfflat:
I think having different default behaviour in Python and C++ should be avoided. Can we follow the outline in #988, together with a Python interface to that?

brenthuisman:
`arbenv::thread_concurrency()` seems to be returning `1` on my system, while `std::thread::hardware_concurrency()` returned the expected number. Maybe I used it wrong?

On Python I'd definitely expect use-all-local-threads be default, but my preference is indeed also no difference between the APIs

halfflat:
Regarding `thread_concurrency` incorrectly giving the value one: this is part of the problem. There's no solid way of determining what the best thread count to use is. Even if we could work around the various platform differences, there is still the issue that the user has more information than we do: is something else running, and occupying four of the cores? we can't tell.

If `arborenv::threasd_concurrency()` is giving a bad value, we should definitely address that. But it will always be a matter of better or worse, rather than correct or incorrect.

bcumming:
The default is set to 1 for consistency, so that the default value is predictable (same as C++, not dependent on env variables, etc).

Instead, we can wrap `arbor-env`, and provide its interface for querying the environment.

@brenthuisman, if you compile the C++ examples and run ring, how many threads does it used:
```
> make ring
> ./bin/ring

gpu:      no
threads:  4
mpi:      no
ranks:    1
```

brenthuisman:
I ran `brunel.cpp` as-is; indeed used the expected number of threads. Neither of these commits use >1 thread in a Python test:
https://github.com/arbor-sim/arbor/pull/1415/commits/bcecd66e995e133008f247a7bc0f597378a48d79
https://github.com/arbor-sim/arbor/pull/1415/commits/0a84226cac9e90c088a32c2d1579a2f6ed2f6ebe

brenthuisman:
Solution will come through #988

PR #1416, opened 2021-03-05
Fix Issues yaml, now all templates should show up

brenthuisman:
Turns out some templates are hidden due to malformed yaml.


halfflat (approved):
Looks fine!

Issue #1233, opened 2020-11-12
Morphology writers


halfflat:
Add support for writing morphology representations in arborio:

1. <s>`save_swc_XXX` counterparts to `load_swc_XXX` functions, which will roundtrip morphologies provided they abide by the constraints of the corresponding representation.</s>

2. Writing morphologies (and maybe regions, too, via segmentGroups) to NeuroML. Handling regions, at least where they do not align exactly with segments, will require subdividing segments, presumably linearly.

3. (New) Specify and implement an S-expr representation for segment trees, and make the corresponding reader and writer. While regions etc. already have a perfectly good representation, should we also have a representation of their concrete realisations, so that they can be more easily shared across other sims?

bcumming:
Very good idea!
However, in the past you and I have decided that `write_swc` should not be implemented in Arbor
1. We can't represent our morphology representations using swc (except for a handful of corner cases)
2. We really don't want to have solve the "reverse interpretation" problem of converting the different soma representations into swc samples.
2. Users should not be encouraged to use SWC.

halfflat:
The discouraging people from using SWC at all I think is the strongest argument here :)

thorstenhater:
Hey,

ACC was decided to be our internal format of choice and we do not consider ourselves a neuroscience I/O library.
Closed until considered interesting.

PR #1423, opened 2021-03-10
Arbor Cable-Cell Format (ACC)

### LINKED ISSUE #1336: Cable_cell file formats
## Goal:
File format for representing a `label_dictionary`, `decor`, `morphology`, `cable_cell_parameter_set`, `cable_cell` and perhaps `cv_policy`.

### Requirements:
1) Round-trippable.
2) Easy to produce/consume by machines (little code to implement the reader/writer).
3) Can refer to modules within the same file or from another file.
4) If possible, easily human readable/edittable.

### Options: 
(Note: the featured examples are suggestions meant to illustrate what the files would look like, the actual formats need more work.) 
1) JSON-based format:
   * Examples:
     Common fields:
            - `version`: Format version.
            - `type`: One of "decor", "label-dictionary", "morphology", "default-parameters", "cv-policy", "cable-cell".
            - `modules`: Modules referenced in the `data` description.
            - `data`: JSON with format corresponding to the specified `type`.

       1. default-parameters file:
       ```
       {
         "version" : 1,
         "type" : "default-parameters",
         "data":
         {
           "axial-resistivity": 35.4,
           "init-membrane-potential": -60.0,
           "temperature-K": 6.3,
           "membrane-capacitance": 0.01,
           "ions": {
             "ca": {
               "init-int-concentration": 5e-05,
               "init-ext-concentration": 2.0,
               "init-reversal-potential": 132.4579341637009,
               "reversal-potential-method": "constant"
             },
             "k": {
               "init-int-concentration": 54.4,
               "init-ext-concentration": 2.5,
               "init-reversal-potential": -77.0,
               "reversal-potential-method": "nernst"
             },
             "na": {
               "init-int-concentration": 10.0,
               "init-ext-concentration": 140.0,
               "init-reversal-potential": 50.0,
               "reversal-potential-method": "nernst"
             }
           }
         }
       }
       ```
       2. cable-cell file:
       ```
       {
         "version" : 1, 
         "type": "cable-cell",
         "modules":
         [
           {"name": "decor_0", "path" : "/path/to/the/file.json"}, 
           {"name": "label_dict_1" , "path" : "/path/to/the/file.json"}, 
           {"name": "mopho_5", , "path" : "/path/to/the/file.json"} 
         
         ]
         "data": 
         {
           "decor": "decor_0",
           "label-dict": "label_dict_1",
           "morphology": "morpho_5"
         }
       }
       ```
   * **Pros & Cons**:
       (+) JSON is widely used.
       (+) There are good and lightweight C++ and python libraries for reading and writing JSON.
       (-) We would most likely need to use s-expressions to describe regions/locsets/morphology. These would have to be added as strings in the JSON file, and seperate parsers need to be implemented for that purpose. 

2) S-expr based format
   * Examples: 
       1. label_dictionary: 
       ```
        (label-dict 
          (locset-def "term" (terminal))
          (locset-def "stim-site" (location 0 0.5))
          (region-def "soma" (tag 1))
          (region-def "axon" (tag 2))
          (region-def "dend" (tag 3)))
       ```
     
        2. decor: 
       ```
       (decorations 
         (place (terminal) (threshold-detector -10.000000))
         (paint (region "axon") (mechanism "hh" (("bar" 1.2) ("foo" 12.0))))
         (default (ion-reversal-potential "na" 50.000000)))
       ```
     
       3. morphology: 
       ```
       (morphology 
         (branch 0 -1 
           (segment 0 
             (point 0.000000 0.000000 0.000000 10.000000)
             (point 0.000000 0.000000 10.000000 10.000000)
             1))
         (branch 1 0 
           (segment 1 
             (point 0.000000 0.000000 10.000000 10.000000)
             (point 0.000000 0.000000 10.000000 1.000000)
             3)
           (segment 2 
             (point 0.000000 0.000000 10.000000 1.000000)
             (point 0.000000 0.000000 100.000000 1.000000)
             3))
         (branch 2 0 
           (segment 3 
             (point 0.000000 0.000000 10.000000 10.000000)
             (point 0.000000 0.000000 0.000000 2.000000)
             2)
           (segment 4 
             (point 0.000000 0.000000 0.000000 2.000000)
             (point 0.000000 0.000000 -20.000000 2.000000)
             2)))
       ```
   * Questions: 
      How to refer to expressions in the same file or across files?
          - We can name decors/label_dictionaries/morphologies. 
          - A file containing s-expressions could then be thought of as a dictionary containing named decors/label_dictionaries/morphologies etc. 
          - We can have functions to read the s-expression associated with a label from a given file.
           
   * **Pros & Cons**:
       (+) We already use s-expressions to form regions/locsets.
       (+) S-expressions are easy to read and write, they're compact and readable.
       (-) Referencing objects within a file or across files is not straight forward. 
       (-) Not as popular as XML/JSON. 

3) XML based format:
   * Examples: 
       1. morphology: 
       ```
       <morphology id="m2">
         <segment id="0" tag="1">
           <proximal x="1" y="-2" z="3.5" diameter="8"/>
           <distal x="3" y="-3.5" z="4" diameter="8.5"/>
         </segment>
         <segment id="1" tag="3">
           <parent segment="0"/>
           <distal x="4.5" y="-5" z="5" diameter="0.5"/>
         </segment>
       </morphology>
       ```
       
       2. label_dictionary: 
       ```
       <label-dict id="dict0">
         <region-def id="soma">
           <tagged id="1"/>
         </region-def>
         <locset-def id="root">
           <terminal>
         </locset-def>
       </label-dict>
       ```
       
    * **Pros & Cons**
        (+) Well established format.
        (+) Has existing standards for refering to external modules (e.g. XInclude and XPath).
        (+) Already used for NeuroML, so there's a precedent. 
        (-) Almost certainly needs an external library for reading/writing.
        (-) Quite verbose.

noraabiakar:
* Implements parsers and writers in `arborio` for the s-expression based Arbor Cable-Cell Format defined in #1336.
* Moves `s_expr.hpp` to the public interface.
* Adds python wrappers to support the features in pyarb.  
* Adds unit tests.
* Adds docs.

Fixes #1336
Partially addresses #1233


arbor/include/arbor/cable_cell_param.hpp
```diff
@@ -46,55 +46,87 @@ struct i_clamp {
     i_clamp(value_type delay, value_type duration, value_type amplitude):
         delay(delay), duration(duration), amplitude(amplitude)
     {}
+
+    friend std::ostream& operator<<(std::ostream& o, const i_clamp& p) {
+        return o << "(current-clamp " << p.delay << ' ' << p.duration << ' ' << p.amplitude << ')';
```
thorstenhater, line 51: Seeing these makes me wish for `fmt` or `std::format`. Maybe these can be made less verbose by using `util::pprintf`?

On the other hand, did you consider to have a method `s_expr to_expr(const i_clamp&)` (etc) and delegating the actual printing part to the `s_expr` functionality? I believe @halfflat mentioned something similar.

arborio/include/arborio/cableio.hpp
```diff
@@ -0,0 +1,39 @@
...
+    cable_cell_variant component;
+};
+
+std::ostream& write_component(std::ostream&, const cable_cell_component&);
+std::ostream& write_component(std::ostream&, const arb::decor& x, const meta_data& m = {});
+std::ostream& write_component(std::ostream&, const arb::label_dict& x, const meta_data& m = {});
+std::ostream& write_component(std::ostream&, const arb::morphology& x, const meta_data& m = {});
+std::ostream& write_component(std::ostream&, const arb::cable_cell& x, const meta_data& m = {});
+
+parse_hopefully<cable_cell_component> parse_component(const std::string&);
```
thorstenhater, line 36: I have a suggestion on the readers here. Maybe it's wise to allow to pass an expected version or accept a range of compatible versions. The reasoning is that users might have older files they wish to read in. Storing should only happen in the latest format, however.

arborio/cableio.cpp
```diff
@@ -0,0 +1,799 @@
...
+    }
+    // A branch should have either 0 or >1 children.
+    auto it = std::find_if(branch_children.begin(), branch_children.end(), [](unsigned i){return i==1;});
+    if (it != branch_children.end()) {
+        throw cableio_morphology_error(it - branch_children.begin());
+    }
+
+    // Append segments according to id order.
+    std::sort(segs.begin(), segs.end(), [](const auto& lhs, const auto& rhs){return lhs.first.id < rhs.first.id;});
+    for (const auto& spair: segs) {
```
thorstenhater, line 284: You know what I am going to suggest here, right? ;)

`const auto& [seg, s_pid]`

arborio/cableio.cpp
```diff
@@ -0,0 +1,799 @@
...
+        auto seg = spair.first;
+        auto s_pid = spair.second;
+        tree.append(s_pid, seg.prox, seg.dist, seg.tag);
+    }
+    return morphology(tree);
+}
+
+// Define cable-cell maker
+// Accepts the morphology, decor and label_dict arguments in any order as a vector
+cable_cell make_cablecell(const std::vector<std::variant<morphology, label_dict, decor>>& args) {
```
thorstenhater, line 294: Our typography suggests an `_` in `cablecell`.

doc/fileformat/cable_cell.rst
```diff
@@ -0,0 +1,410 @@
...
+
+   This describes an *initial membrane potential* object with value ``val`` (unit [mV]).
+
+.. label:: (axial-resistivity val:real)
+
+   This describes an *axial resistivity* object with value ``val`` (unit [Ω·cm]).
+
+.. label:: (temperature-kelvin val:real)
+
+   This describes a *temperature* object with value ``val`` (unit [kelvin]).
```
thorstenhater, line 103: Tiny nitpick: The rest of the unit is written as SI symbols, Kelvin should be just K for consistency.

thorstenhater (changes requested):
A few nitpicks, rest looks good.

arbor/include/arbor/cable_cell_param.hpp
noraabiakar: These methods aren't really needed in the implementation, so they can be removed. I kept them in because I thought they could be useful. I can use util::pprintf or just remove them completely. 
We do have `mksexp` functions in the `arborio` lib, but I didn't want s-expressions to bleed too much into the core arbor library (especially public headers) because I'd rather we move the `s_expr` implementation completely into `arborio` eventually. 
@halfflat's suggetsion would be useful here, but I think it'll take some more work, which I will do in another PR after we discuss the details a bit more.

arborio/include/arborio/cableio.hpp
noraabiakar: The string argument is expected to contain meta-data containing the version. At the moment only version 1 is supported, and anything else will return an error. Is this what you meant, or was it something else?

arborio/cableio.cpp
noraabiakar: Right, still not used to that. I'll make the change, thanks!

arborio/cableio.cpp
noraabiakar: Okay.

doc/fileformat/cable_cell.rst
noraabiakar: Okay.

arbor/include/arbor/cable_cell_param.hpp
thorstenhater: I think it's worth keeping them. Our python bindings could make use of them to implement `__str__`/`__repr__`.

thorstenhater: Agreed on the rest.

arborio/include/arborio/cableio.hpp
thorstenhater: Not exactly. Assume someone writes out a cable-cell on version 1 of the format, then upgrades to a version of arbor using version 2. Now they try reading the old file. What are we doing then. Here's the options I see

1. Flat out refuse to load and throw `unsupported_version_error`.
2. Like 1. but offer also a script to convert version 1 -> 2.
3. Support more than one version and automatically select the loader.
4. Like 3., but the load function takes a version, which is our expected version.

I think it's worth to lay the foundation for this now.

noraabiakar: Right now, we do option (1). We could do one of the other options in the future once we support more versions. 
Personally, I like either option (2) or (3), but I haven't thought of it too deeply.

thorstenhater: Ok, that potentially entails changing the public API, if going with 4. We can leave it is for this, as long we consider this be acceptable.

thorstenhater (approved):
👍

doc/python/format.rst
```diff
@@ -0,0 +1,80 @@
...
+
+   .. py:attribute:: meta_data meta
+
+      Stores meta-data pertaining to the description of a cable cell component.
+
+   .. py:attribute:: component
+
+      Stores one of :class:`decor`, :class:`label_dict`, :class:`morphology` or :class:`cable_cell`.
+
+Reading arbor-components
```
brenthuisman, line 33: I think that you combine reading and writing here, as opposed to the C++ section. if so, the heading should reflect that.

doc/fileformat/cable_cell.rst
```diff
@@ -0,0 +1,429 @@
+.. _formatcablecell:
+
+Cable cell format
```
brenthuisman, line 3: This sounds a bit too non specific. I suppose it's one of those hard things, but I think we need one (The Arbor Cable Cell Format). Or, are we so ambitious (maybe we are!) that this in indeed _The_ Cable Cell Format?

doc/python/cable_cell.rst
```diff
@@ -3,6 +3,16 @@
 Cable cells
 ===========
 
+.. toctree::
```
brenthuisman, line 6: Good catch, I though I'd already done that.

doc/fileformat/cable_cell.rst
```diff
@@ -0,0 +1,429 @@
...
+
+.. label:: (ion-reversal-potential-method ion:string method:mechanism)
+
+   This creates a *reversal potential method* (able to modify the reversal potential) of ion ``ion`` from
+   mechanism ``method``.
+   For example:
+
+   .. code:: lisp
+
+      (ion-reversal-potential-method "ca" (mechanism "nersnt/ca"))
```
brenthuisman, line 143: a typo in nernst

doc/fileformat/cable_cell.rst
```diff
@@ -0,0 +1,429 @@
...
+
+.. label:: (current-clamp (envelope-pulse delay:real duration:real amplitude:real) freq:real)
+
+   This creates a *current clamp*. If the frequency ``freq`` (unit [Hz]) is zero, the current is a square
```
brenthuisman, line 147: Probably not related to this PR as such, and I hadn't looked at the clamps yet, but elsewhere I believe the unit for frequencies are [kHz], which in turn matches [ms] better than [s]. Would it be possible to use [kHz] here?

doc/fileformat/cable_cell.rst
```diff
@@ -0,0 +1,429 @@
...
+
+.. label:: (threshold-detector val:real).
+
+   This describes a *threshold-detector* object with value ``val`` (unit [mV]]).
+
+.. label:: (gap-junction-site)
+
+   This describes a *gap-junction-site*.
+
+*Paintable* and *placeable* properties and dynamics are placed on regions and locsets respectively.
```
brenthuisman, line 177: More precisely: region/locset expressions.

doc/fileformat/cable_cell.rst
```diff
@@ -0,0 +1,429 @@
...
+        (paint (region "soma") (membrane-potential -50.000000))
+        (paint (all) (mechanism "pas"))
+        (paint (tag 4) (mechanism "Ih" ("gbar" 0.001)))
+        (place (locset "root") (mechanism "expsyn"))
+        (place (terminal) (gap-junction-site)))
+
+Morphology
+----------
+
+The morphology of a cable cell can be described in terms of points, tagged segments and branches.
```
brenthuisman, line 240: The difference between segment tree and morphology is a bit subtle, and gets more confusing if we points out the segments in a morphology. Does the format in fact decribe a morphology or segment_tree?

doc/fileformat/cable_cell.rst
```diff
@@ -0,0 +1,429 @@
...
+expressions.
+
+.. label:: (cable-cell morph:morphology dec:decor dict:label-dict)
+
+   The arguments of the cable-cell can be in any order, as long as all 3 components are listed.
+   For example:
+
+   .. code:: lisp
+
+      (cable-cell
```
brenthuisman, line 315: The indentation does not matter, but is it outputted this way? Also: the placement of linebreaks are a matter of choice/optional. Since few users will be familiar with lispy formats, it might be nice to state these properties explicitly.

doc/python/format.rst
```diff
@@ -0,0 +1,80 @@
+.. _pycablecellformat:
```
brenthuisman, line 1: Currently, (links to) formats live in `doc/python/morphology.rst`. Maybe that should link here instead, and the other formats listed here as well.

Same for the Concepts section.

brenthuisman (changes requested):
I went through the documentation so a few comments there.

doc/fileformat/cable_cell.rst
noraabiakar: You're right, we should be using kHz. Sam is going to fix this and the underlying class units in a future PR.

doc/fileformat/cable_cell.rst
noraabiakar: They're placed on the locsets/regions generated from the expressions, not the expressions themselves. Is it unclear in the text?

doc/fileformat/cable_cell.rst
noraabiakar: Technically, we build a segment tree and create a morphology from it. I know there's a subtle difference in the `morphology` and `segment_tree` classes, but we create morphologies from segment trees all the time in the docs and the examples; so I don't think using `morphology` here will cause confusion, in fact I think `segment-tree` would be more confusing.

doc/python/format.rst
noraabiakar: The formats that live in `doc/python/morphology.rst` are morphology specific (SWC and neuroML) so it makes sense for them to live there. Of course we could place them under this section, I'd rather not do this in this PR though, as it is already quite big.

doc/fileformat/cable_cell.rst
noraabiakar: The output respects the indentation. I'll add a note.

doc/fileformat/cable_cell.rst
brenthuisman: OK! Will that follow this PR quickly? If not, we'll need to bump the version quite soon for a reason we already foresee, which is a tiny bit unfortunate :) If we're sure of that other PR, then can we do it 'dirty' and divide/multiply outputs/inputs right now, in order to keep the format unchanged?

doc/fileformat/cable_cell.rst
noraabiakar: I kept it vague because I didn't know what to call it. I'll bring up the naming issue on slack.

doc/fileformat/cable_cell.rst
brenthuisman: I see the sentence is correct, however, when using the terms as such, I'm inclined to think of them by the names I have given them in the label dict (I have region Region1 and locset Locset1). Which in turn might lead me to believe I can only address them with those labels. But, I can use expressions here, so I do not in fact need to explicitly label them.

"placed on regions and locsets respectively, generated from a given expression." ?

doc/fileformat/cable_cell.rst
brenthuisman: :+1:

doc/fileformat/cable_cell.rst
noraabiakar: I'm make the change

doc/python/format.rst
brenthuisman: Sure, but I'm not sure too much nesting is helping figure out how supported file formats work. A separate `format.rst` shows up in the TOC, and thus will be the first place to look, and the first place to list the formats. It's actually not that much text that would move.

doc/fileformat/cable_cell.rst
noraabiakar: It should follow quite soon. We'll have to make another change to the format (adding a `phase` parameter to the current-clamp) anyway once that future PR is in. I don't think this will be a huge issue, I don't know if we'll need to bump the version for example.

doc/fileformat/cable_cell.rst
noraabiakar: We can remove it from the concepts, maybe.

doc/python/format.rst
noraabiakar: I agree. I'll make sure to move them right after this PR is merged, in a separate PR, just to keep the scope of this one limited. Does that sound good?

brenthuisman: :+1:

doc/fileformat/cable_cell.rst
brenthuisman: :+1:

doc/fileformat/cable_cell.rst
brenthuisman: :+1:

doc/fileformat/cable_cell.rst
```diff
@@ -0,0 +1,440 @@
...
+
+We define an s-expression format for describing :ref:`cable cells <cablecell>`.
+Cable cells are constructed from three components: a :ref:`label dictionary <labels>`,
+a :ref:`decoration object <cablecell-decoration>` and a :ref:`morphology <morph>`.
+The cable cell *format* is constructed in the same way.
+
+.. Note::
+
+   Lines breaks and indentation in the s-expressions presented below have been added
+   for clarity. They are not a requirement of the format and will be ignored by the
```
thorstenhater, line 14: It's probably more precise to say 'treated as whitespace', but I unsure whether this matters, eg
```LISP
(loc
  ation 0 0)
```
will be parsed differently than
```LISP
(location 0 0)
```

doc/fileformat/cable_cell.rst
```diff
@@ -0,0 +1,440 @@
...
+   point mechanism,                    ✓,             --,            --
+   current clamp,                      ✓,             --,            --
+   threshold detector,                 ✓,             --,            --
+   gap junction site,                  ✓,             --,            --
+
+The various properties and dynamics of the decor are described as follows:
+
+.. label:: (membrane-potential val:real)
+
+   This describes an *initial membrane potential* object with value ``val`` (unit [mV]).
```
thorstenhater, line 101: Physicists nitpick: `[X]` means 'whatever the units of the quantity `X` are'. So, `[L] = m`, but `[m]` does not make sense in this context. Just drop the `[]`, saying 'unit' is enough. I know this is appearing in other place of Arbor.

doc/cpp/cable_cell_format.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Reading and writing of the arbor-component description format is delegated to the ``arborio``
+library and the responsible classes and functions are present in the ``arborio`` namespace.
+
+The arbor-components and meta-data
+----------------------------------
+
+.. cpp:type:: cable_cell_variant = std::variant<arb::morphology, arb::label_dict, arb::decor, arb::cable_cell>
+
+.. cpp:type:: template <typname T> parse_hopefully = arb::util::expected<T, cableio_parse_error>
```
thorstenhater, line 21: Typo: typename

python/cable_cell_io.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <arbor/cable_cell.hpp>
+
+#include <arborio/cableio.hpp>
+
+#include "error.hpp"
+#include "strprintf.hpp"
+
+namespace pyarb {
+
+arborio::cable_cell_component load_component(const std::string& fname) {
```
thorstenhater, line 16: A final suggestion here: I would like to see the option to load from a File-like object, either as an 'overload' (aka switch-on-type) or as the only interface.

This the boilerplate to read from file.
```Python
with open('fn.acc') as fd:
  comp = arb.load_component(fd)
```
 But this pattern allows to do this
```Python
text = get_acc_from_http('https://arb-db.acc/foo')
buff = StringIO(text)
comp = arb.load_component(buff)
```

thorstenhater (changes requested):
A few nitpicks and an Python API suggestion. Rest is good as before.

brenthuisman (approved):

doc/fileformat/cable_cell.rst
noraabiakar: Yeah, the first will fail. I mean indentations and white space between symbols. I can say 'treated as whitespace', will that be enough?

thorstenhater: Yes, as I said 'Treated as whitespace' is more precise, but whether the reader cares (or sees the equivalence directly) I do not know.

noraabiakar: "Extra line breaks and indentations in the s-expressions presented below have been added for clarity. They are not a requirement of the format and will be treated as whitespace."?

thorstenhater: 👍

python/cable_cell_io.cpp
noraabiakar: This affects most of the intefaces of arborio. Can we open a separate issue for it?

thorstenhater: 👍 
And we should get to it quickly, it's a commonly used pattern in Python (yaml/json/... libs).

doc/cpp/cable_cell_format.rst
```diff
@@ -0,0 +1,88 @@
...
+
+   ``arborio::cableio_parse_error`` is derived from ``arb::arbor_exception`` which in turn is derived
+   from ``std::runtime_error``. It contains information about the ``line`` and ``position`` of an encountered
+   error in a document.
+
+   ``arb::util::expected`` contains either an object of type ``T`` or an error object.
+
+.. cpp:class:: meta_data
+
+   .. cpp:member:: int version
```
thorstenhater, line 31: Shouldn't this be capable of holding '0.1-dev'?

thorstenhater (changes requested):
Looks good, but docs got a bit out of sync.

doc/cpp/cable_cell_format.rst
noraabiakar: Yes, thanks :)

thorstenhater: 👍

Issue #1414, opened 2021-03-04
Discussion: Automate Release / Deployment


thorstenhater:
# Goals

- Reduce workload
- Enable more frequent releases to make it easier to focus on tasks and prioritize
- Simpler review for partners and 'stakeholders'
- *Non-goal*: completely automated releases

# Release Notes

- Where do we store them?
   - Proposal: CHANGELOG file in tree.
    - What do we record? Options
        1.  Closed Issues since last release/tag. Filtered by release manager.
        2. List of commits since last release/tag. Filtered by release manager.
            ```bash
            git log $(git describe --tags --abbrev=0)..HEAD --oneline
            ```

# Release planning

- within month of release, release planning session for the *next* release
- create an aspirational Release Note wiki page like this.

# Things to Automate
- [ ] CI
  - [x] Run the C++ examples
  - [ ] on `-rc` tag: auto-upload Python wheels to TestPyPI
  - [ ] on `-rc` tag: install from wheels and run tests and examples
  - [ ] on `-rc` tag : install from pip via source with all (possible!) flags and run tests and examples
- [ ] Add a script to
   1. Bump VERSION (should be patch-level + `-rc` by default)
   2. git tag from VERSION file
   3. push tagged branch to read-the-docs.io
   4. push to PyPI
   5. update spack package
- Github release
  - e.g. https://github.com/ncipollo/release-action

# Other

- [ ] Update [Release Process](https://github.com/arbor-sim/arbor/wiki/Release-Process) page.
   - [ ] which steps are automated and link to the script/thing that automates them.
   - [X] add spack package to process.
- [ ] Investigate and document what pip does when asked for a flag that's not present in the wheel.
- [ ] Investigate how `MANIFEST` and `setup.cfg` are updated.

schmitts:
* release tarball incl. checkout out submodules (cf. https://github.com/arbor-sim/arbor/blob/master/scripts/create_tarball.sh)
* update spack package (cf. https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/arbor/package.py)

bcumming:
Can we please have a description of what is meant by "automate release" ;)

thorstenhater:
This is a placeholder at the moment from our ongoing discussion.

brenthuisman:
Be patient, my young padawans :) @thorstenhater and I will round the message out ASAP.

halfflat:
Should a change log be regarded as release notes, or do we keep these separate?
[...]

PR #1432, opened 2021-03-13
Run C++ examples in CI

thorstenhater:
- Add C++ examples to default CI.
- Makes the release process easier, see #1414 
- ... and the CI more comprehensive.


brenthuisman (approved):
Great! One question: running examples as a test relies on them having a nonzero exit code in case of problems. Is there any way this may not work as expected?

thorstenhater:
Well, yes, but hard to catch in any case. Assume that `example` contains this
```C++
int main() {
   try { ... } catch { return 0; }
}
```
Not catching this exception results in a non-zero return value.

brenthuisman (approved):

PR #1438, opened 2021-03-18
Check validity of event_generator targets in recipe

### LINKED ISSUE #1435: Segmentation fault: placing an event_generator on a non-existing target
Adding an `event_generator` on a target that doesn't exist will cause a segmentation fault. 
The target of an `event_generator` is not directly accessible because is not necessarily unique for all events in an event generator (see `explicit_generator`). But we should still be able to add a check at setup time: 
Add a method to `event_generator` that returns all of the targets of an `event_generator`, then iterate over the targets, and check their validity.

noraabiakar:
- Add a method to `event_generator` that returns all of its targets. 
- When the simulation is constructed, check the targets of each event generator, and throw an exception if an the `gid` or `lid` is invalid.
- Add unit test.

Fixes #1435

### DIFF
--- a/arbor/arbexcept.cpp
+++ b/arbor/arbexcept.cpp
@@ -38,7 +38,8 @@ bad_connection_source_gid::bad_connection_source_gid(cell_gid_type gid, cell_gid
 {}
 
 bad_connection_source_lid::bad_connection_source_lid(cell_gid_type gid, cell_lid_type src_lid, cell_size_type num_sources):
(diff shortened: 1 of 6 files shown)


halfflat (changes requested):
Looks good! The only thing missing is an update to the comments at the top of `event_generator.hpp` that covers the new `targets()` method requirement.

halfflat (approved):

PR #1440, opened 2021-03-18
fix bug attaching branches to the soma in ASCII file descriptions

bcumming:
Fix bug attaching branches to the soma in ASCII file descriptions.

Branches in an ASCII description attached to the soma were incorrectly creating a segment
from the center of the soma to the first point in the branch.


thorstenhater (approved):
👍

thorstenhater (approved):

PR #1443, opened 2021-03-18
Amend segment cables for zero length branches.

### LINKED ISSUE #1442: Segment extents on zero length branches
On one hand, segments (geometrical) associated with a branch should cover the branch (topological). But if the branch has zero length (as defined geometrically) — all of its segments are defined by colacated points — the cables corresponding to those segments will all be of the form (_branch id_, 0, 0).

Consider segments _s_<sub>1</sub> … _s_<sub>_k_</sub> comprising a branch _b_, with _s_<sub>_i_+1</sub> having parent segment _s_<sub>_i_</sub>. Let  _s_<sub>_i_</sub> have morphological representation as a cable (_b_, _p_<sub>_i_</sub>, _d_<sub>_i_</sub>). Then for consistency, we should have:
* _p_<sub>_i_</sub> ≤ _d_<sub>_i_</sub> for all _i_ = 1 … k,
* _d_<sub>_i_</sub> ≤ _p_<sub>_i+1_</sub> for all _i_ = 1 … k-1,
* ∪<sub>_i_ </sub> [_p_<sub>_i_</sub>,  _d_<sub>_i_</sub>] = [0, 1], that is, these cables should cover the branch.

I propose that in the zero length situation the segment cables should be (_b_, 0, 0) for all but the most distal segment, which should be (_b_, 0, 1). Other partitions of the interval would work too, but this is simple to implement, explain, and test.

halfflat:
* Ensure cables corresponding to segments cover even zero length branches, by setting the segment cables to (b, 0, 0) for all but the last segment in a zero-length branch b, and setting that last segment cable to (b, 0, 1).

Fixes #1442

### DIFF
--- a/arbor/morph/embed_pwlin.cpp
+++ b/arbor/morph/embed_pwlin.cpp
@@ -249,11 +249,18 @@ embed_pwlin::embed_pwlin(const arb::morphology& m) {
         if (branch_length!=0) {
             for (auto& d: seg_pos) {
                 d /= branch_length;
+                all_segment_ends_.push_back({bid, d});
             }
         }
-
-        for (auto d: seg_pos) {
-            all_segment_ends_.push_back({bid, d});
+        else {
+            // In zero length branch, set all segment ends to be 0,
+            // except for last, which is 1. This ensures that the
+            // union of the cables corresponding to a branch cover
+            // the branch.
+            seg_pos.back() = 1;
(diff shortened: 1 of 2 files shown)


halfflat:
bors try

noraabiakar (approved):
lgtm!

PR #1458, opened 2021-03-25
Fix bug in `single_cell_model::probe`, fix probe frequency in examples

### LINKED ISSUE #1457: Empty plots are being generated in python examples
All the python examples using the `single_cell_model` are generating empty plots, specifically: 
* single_cell_detailed.py
* single_cell_model.py
* single_cell_nml.py
* single_cell_swc.py

noraabiakar:
* Fix the regular schedule generated in `single_cell_model::run`
* Adjust frequency of probes in python examples. 

Fixes #1457

### DIFF
--- a/python/example/single_cell_nml.py
+++ b/python/example/single_cell_nml.py
@@ -80,9 +80,9 @@ print(cell.locations('"axon_end"'))
 m = arbor.single_cell_model(cell)
 
 # Attach voltage probes that sample at 50 kHz.
-m.probe('voltage', where='"root"',  frequency=50000)
-m.probe('voltage', where='"stim_site"',  frequency=50000)
-m.probe('voltage', where='"axon_end"', frequency=50000)
+m.probe('voltage', where='"root"',  frequency=50)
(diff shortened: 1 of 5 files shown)


python/single_cell_model.cpp
```diff
@@ -194,7 +194,7 @@ class single_cell_model {
 
             traces_.push_back({"voltage", p.site, {}, {}});
 
-            auto sched = arb::regular_schedule(p.frequency);
+            auto sched = arb::regular_schedule(1.0/p.frequency);
```
brenthuisman, line 197: Wow, can't believe it was not noticed before!

brenthuisman (approved):

python/example/single_cell_nml.py
```diff
@@ -80,9 +80,9 @@
 m = arbor.single_cell_model(cell)
 
 # Attach voltage probes that sample at 50 kHz.
```
schmitts, line 82: 50 Hz vs. 50 kHz?

PR #1464, opened 2021-03-25
Add optional spherical root for NeuroML.

### LINKED ISSUE #1439: Zero-length NeuroML segments are not interpreted as spheres.
Zero length segments in NeuroML morphology descriptions should be interpreted as spheres, or for Arbor at least, the equivalent cylinder. Currently this appears not to be done.

halfflat:
* Adds a `neuroml_options` namespace with enum for use with `neuroml::morphology(...)` and `neuroml::cell_morphology(...)`.
* Add support for a 'spherical' root segment with option `neuroml_options;:allow_spherical_root`; a zero-length root segment with identical proximal and distal radius will then be converted into an area-equivalent cylinder in the resultant morphology.
* Amend `place_pwlin::all_at` so that locations on a zero-length segment with different radii at the endpoint will give an mpoint for either end.
* Add `allow_spherical_root` boolean optional parameter to Python NeuroML morphology methods.
* Update docs; C++ unit tests.

Fixes #1439.

### DIFF
--- a/arbor/morph/place_pwlin.cpp
+++ b/arbor/morph/place_pwlin.cpp
@@ -62,7 +62,16 @@ std::vector<mpoint> place_pwlin::all_at(mlocation loc) const {
     double pos = is_degenerate(pw_index)? 0: loc.pos;
 
     for (auto [bounds, index]: util::make_range(pw_index.equal_range(pos))) {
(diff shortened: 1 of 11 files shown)


noraabiakar (approved):
Looks good!

PR #1466, opened 2021-03-26
Add spack package file to the repository

### LINKED ISSUE #1465: maintain spack config in master
Add the spack `package.py` file for Arbor to the repository for two reasons:

* keep the package information up to date, so that we have a working up-to-date package description on hand to push to the main spack repository when we tag a release
* to test the spack config as part of CI.

https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/arbor/package.py

bcumming:
Copied the spack `package.py` from the main spack repository into the arbor repo.

Some small updates were made:
- Correctly provide a url for the GitHub tagged release
- Specify minimum pybind11 version 2.6

Fixes #1465

### DIFF
--- a/spack/package.py
+++ b/spack/package.py
@@ -13,7 +13,7 @@ class Arbor(CMakePackage, CudaPackage):
     homepage = 'https://arbor-sim.github.io/'
     git      = 'https://github.com/arbor-sim/arbor.git'
     url      = 'https://github.com/arbor-sim/arbor/releases/download/v0.5/arbor-v0.5-full.tar.gz'
-    maintainers = ['bcumming', 'halfflat']
+    maintainers = ['bcumming', 'brenthuisman', 'haampie', 'halfflat']
 
     version('master', branch='master', submodules=True)
     version('0.5', 'd0c8a4c7f97565d7c30493c66249be794d1dc424de266fc79cecbbf0e313df59', url='https://github.com/arbor-sim/arbor/releases/download/v0.5/arbor-v0.5-full.tar.gz')
@@ -27,7 +27,7 @@ class Arbor(CMakePackage, CudaPackage):
 
     # https://arbor.readthedocs.io/en/latest/install/build_install.html?highlight=requirements#compilers
     conflicts('%gcc@:8.3')
-    conflicts('%clang@:7.0')
+    conflicts('%clang@:7')
     # Cray compiler v9.2 and later is Clang-based.
     conflicts('%cce@:9.1')
     conflicts('%intel')
--- a/spack/package.py
+++ b/spack/package.py
@@ -10,9 +10,9 @@ class Arbor(CMakePackage, CudaPackage):
     """Arbor is a high-performance library for computational neuroscience
     simulations."""
 
-    homepage = 'https://github.com/arbor-sim/arbor/'
+    homepage = 'https://arbor-sim.github.io/'
     git      = 'https://github.com/arbor-sim/arbor.git'
-    url      = 'https://arbor.readthedocs.io/en/latest/'
+    url      = 'https://github.com/arbor-sim/arbor/releases/download/v0.5/arbor-v0.5-full.tar.gz'
     maintainers = ['bcumming', 'halfflat']
 
     version('master', branch='master', submodules=True)
@@ -27,7 +27,7 @@ class Arbor(CMakePackage, CudaPackage):
 
     # https://arbor.readthedocs.io/en/latest/install/build_install.html?highlight=requirements#compilers
     conflicts('%gcc@:8.3')
-    conflicts('%clang@:8.0')
+    conflicts('%clang@:7.0')
     # Cray compiler v9.2 and later is Clang-based.
     conflicts('%cce@:9.1')
     conflicts('%intel')
--- a/spack/package.py
+++ b/spack/package.py
@@ -1,0 +1,76 @@
+# Copyright 2013-2021 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack import *
+
+
+class Arbor(CMakePackage, CudaPackage):
+    """Arbor is a high-performance library for computational neuroscience
+    simulations."""
+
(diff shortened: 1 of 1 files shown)


spack/package.py
```diff
@@ -0,0 +1,76 @@
...
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack import *
+
+
+class Arbor(CMakePackage, CudaPackage):
+    """Arbor is a high-performance library for computational neuroscience
+    simulations."""
+
+    homepage = 'https://github.com/arbor-sim/arbor/'
```
brenthuisman, line 13: Shall we start using arbor-sim.github.io?

spack/package.py
```diff
@@ -0,0 +1,76 @@
...
+    git      = 'https://github.com/arbor-sim/arbor.git'
+    url      = 'https://arbor.readthedocs.io/en/latest/'
+    maintainers = ['bcumming', 'halfflat']
+
+    version('master', branch='master', submodules=True)
+    version('0.5', 'd0c8a4c7f97565d7c30493c66249be794d1dc424de266fc79cecbbf0e313df59', url='https://github.com/arbor-sim/arbor/releases/download/v0.5/arbor-v0.5-full.tar.gz')
```
brenthuisman, line 19: I find another commithash for the v0.5 tag:
[2158d2c61c0608363c66b9665ca4785e8c290b37](https://github.com/arbor-sim/arbor/releases/tag/v0.5)

spack/package.py
```diff
@@ -0,0 +1,76 @@
...
+    variant('assertions', default=False, description='Enable arb_assert() assertions in code.')
+    variant('doc', default=False, description='Build documentation.')
+    variant('mpi', default=False, description='Enable MPI support')
+    variant('neuroml', default=True, description='Build NeuroML support library.')
+    variant('python', default=True, description='Enable Python frontend support')
+    variant('vectorize', default=False, description='Enable vectorization of computational kernels')
+
+    # https://arbor.readthedocs.io/en/latest/install/build_install.html?highlight=requirements#compilers
+    conflicts('%gcc@:8.3')
+    conflicts('%clang@:8.0')
```
brenthuisman, line 30: We list clang 8 as min version. Why would it conflict? Because of its stdlib?

spack/package.py
```diff
@@ -0,0 +1,76 @@
...
+    depends_on('cmake@3.12:', type='build')
+
+    # misc dependencies
+    depends_on('nlohmann-json')
+    depends_on('cuda@10:', when='+cuda')
+    depends_on('libxml2', when='+neuroml')
+
+    # mpi
+    depends_on('mpi', when='+mpi')
+    depends_on('py-mpi4py', when='+mpi+python', type=('build', 'run'))
```
brenthuisman, line 44: mpi4py is not a compile type dep, and only a runtime dep if the user wants to configure their own communicator.

brenthuisman (commented):
For some extra compilation speed, how about telling Spack to use Ninja?
https://spack.readthedocs.io/en/latest/build_systems/cmakepackage.html#generators

spack/package.py
haampie: It's not the commit hash, it's the tarball hash

spack/package.py
bcumming: I just copied over our current configuration.
It is true that `:8.0` means `<=8`, so it should be `:7`.

The spack syntax of using only open intervals for specifying version ranges is awkward.

spack/package.py
bcumming: This is something we discussed when writing the spack package.
Because we have spack handling the dependencies, it is easy to ask it to configure mpi4py if installing arbor with mpi and python.

spack/package.py
bcumming: I double checked, and that is the correct sha256 hash for the tar ball.

spack/package.py
bcumming: I will fix this.

spack/package.py
bcumming: We can.
Though, I have to make some changes to that site (e.g. stop it from opening up links in new tabs).

bcumming:
I don't have a strong opinion about using ninja in spack. I think that by default the vast majority of the spack dependencies will be built using gmake, so it might be safer to use the same tool. One less dependency, and all that.

spack/package.py
haampie: There's an open PR for this, but not there yet: https://github.com/spack/spack/pull/20258

spack/package.py
```diff
@@ -27,7 +27,7 @@ class Arbor(CMakePackage, CudaPackage):
 
     # https://arbor.readthedocs.io/en/latest/install/build_install.html?highlight=requirements#compilers
     conflicts('%gcc@:8.3')
-    conflicts('%clang@:8.0')
+    conflicts('%clang@:7.0')
```
brenthuisman, line 30: `7.0` will probably allow clang 7.1+ to be used. Would just `7` block and clang with major version 7 to be blocked?

spack/package.py
```diff
@@ -10,9 +10,9 @@ class Arbor(CMakePackage, CudaPackage):
     """Arbor is a high-performance library for computational neuroscience
     simulations."""
 
-    homepage = 'https://github.com/arbor-sim/arbor/'
+    homepage = 'https://arbor-sim.github.io/'
     git      = 'https://github.com/arbor-sim/arbor.git'
-    url      = 'https://arbor.readthedocs.io/en/latest/'
+    url      = 'https://github.com/arbor-sim/arbor/releases/download/v0.5/arbor-v0.5-full.tar.gz'
     maintainers = ['bcumming', 'halfflat']
```
brenthuisman, line 16: The A-team?

bcumming: This was based on a draft version that was written 18 months ago, so please don't be offended!
I will add you and @haampie, because he is a real "Spack A teamer"!

brenthuisman: I actually meant 'The A-team', literally. As in Arbor team ;)

Didn't @schmitts (co-)write this file?

spack/package.py
bcumming: Right, I thought I typed 7. Fixed!

brenthuisman (approved):
lgtm

PR #1467, opened 2021-03-29
Remove `gid` attribute of target of a connection, and local gap-junction site of a gap-junction connection

### LINKED ISSUE #511: Remove target gid from events
If post-synaptic events are placed directly into per-cell queues, i.e. into a queue with all events to be delivered to a unique target gid, there is no need to use the gid in event.

Removing the gid, and using just the lid, would reduce memory and  bandwidth overheads.

noraabiakar:
* Change the type of the target of a `connection` from `cell_member_type` to `cell_lid_type`. The target cell's gid is always known in both the recipe and simulation.
* Change the type of the local site of a `gap_junction_connection` from `cell_member_type` to `cell_lid_type`, and make it such that the first argument of the connection is the peer site, and the second the local site. The local cell's gid is always known in both the recipe and simulation.
* Change the type of the target of an `event_generator` from `cell_member_type` to `cell_lid_type`. The target cell's gid is always known in both the recipe and simulation.
* New `cell_spike_events` and `cse_vector` for `simulation::inject_events`.
* Simplify recipe sanity checks and remove associated exception types. 
* Fix unit tests, examples, docs

Fixes #511.

### DIFF
--- a/doc/cpp/interconnectivity.rst
+++ b/doc/cpp/interconnectivity.rst
@@ -9,18 +9,16 @@ Interconnectivity
     post-synaptic destination. The source is typically a threshold detector on
     a cell or a spike source. The destination is a synapse on the post-synaptic cell.
 
-    .. cpp:type:: cell_connection_endpoint = cell_member_type
+    A :class:`cell_connection` is associated with the destination cell of the connection
+    and its gid.
 
-        Connection end-points are represented by pairs
-        (cell index, source/target index on cell).
+    .. cpp:member:: cell_member_type source
 
-    .. cpp:member:: cell_connection_endpoint source
+        Source end point, represented by the pair (cell gid, source index on the cell)
 
-        Source end point.
+    .. cpp:member:: cell_lid_type dest
 
-    .. cpp:member:: cell_connection_endpoint dest
-
-        Destination end point.
+        Destination target index on the cell, target cell's gid is implicitly known.
 
     .. cpp:member:: float weight
 
@@ -37,15 +35,17 @@ Interconnectivity
 .. cpp:class:: gap_junction_connection
 
     Describes a gap junction between two gap junction sites.
-    Gap junction sites are represented by :cpp:type:cell_member_type.
 
-    .. cpp:member:: cell_member_type local
-
-        gap junction site: one half of the gap junction connection.
+    A :class:`gap_junction_connection` is associated with the local cell of the connection and
+    its gid.
 
     .. cpp:member:: cell_member_type peer
 
-        gap junction site: other half of the gap junction connection.
+        Peer gap junction site, represented by the pair (cell gid, gap junction site index on the cell)
+
+    .. cpp:member:: cell_lid_type local
+
+        Local gap junction site index on the cell, the gid of the local site's cell is implicitly known.
 
     .. cpp:member:: float ggap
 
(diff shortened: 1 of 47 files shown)


doc/cpp/interconnectivity.rst
```diff
@@ -37,15 +35,17 @@ Interconnectivity
 .. cpp:class:: gap_junction_connection
 
     Describes a gap junction between two gap junction sites.
-    Gap junction sites are represented by :cpp:type:cell_member_type.
-
-    .. cpp:member:: cell_member_type local
 
-        gap junction site: one half of the gap junction connection.
+    A :class:`gap_junction_connection` is associated with the local cell of the connection and
+    its gid.
```
halfflat, line 40: I'm not sure what this sentence means — would it be possible to rephrase it?

doc/cpp/recipe.rst
```diff
@@ -74,18 +74,17 @@ Recipe
     .. cpp:function:: virtual std::vector<cell_connection> connections_on(cell_gid_type gid) const
 
         Returns a list of all the **incoming** connections for `gid` .
-        Each connection ``con`` should have post-synaptic target ``con.dest.gid`` that matches
-        the argument :cpp:any:`gid`, and a valid synapse id ``con.dest.index`` on `gid`.
+        Each connection ``con`` should have a valid synapse id ``con.dest`` on the pos-synaptic target `gid`,
```
halfflat, line 77: 'pos-synaptic' → 'post-synaptic'

doc/python/interconnectivity.rst
```diff
@@ -10,6 +10,8 @@ Interconnectivity
     Describes a connection between two cells, defined by source and destination end points (that is pre-synaptic and post-synaptic respectively),
     a connection weight and a delay time.
 
+    A :class:`connection` is associated with the destination cell of the connection and its gid.
+
```
halfflat, line 14: Also this sentence, analogously to the one in the gap junction documentation, is not clear…

test/unit/test_merge_events.cpp
```diff
@@ -234,7 +234,7 @@ TEST(merge_events, tourney_poisson)
 
     std::vector<event_generator> generators;
     for (auto i=0u; i<ngen; ++i) {
-        cell_member_type tgt{0, i};
+        cell_lid_type tgt =i;
```
halfflat, line 237: (Trivial formatting) `tgt =i` → `tgt = i`

halfflat (changes requested):
Excellent clean up! Some trivial edits requested in docs/formatting.

doc/cpp/interconnectivity.rst
noraabiakar: This was my attempt at clarifying why we don't need the `gid` of the `local` cell anymore. Connections used to be "complete" by themselves, now they're only complete in the context of the recipe, when they are bound to a cell. I can remove it, or I can replace it with:
> Describes a gap junction between two gap junction sites. The `local` site does not include the gid of a cell, this is because a `gap_junction_connection` is bound to the local cell which means that the `gid` is implicitly known.

What do you think?

noraabiakar: Also I should probably clarify that to create a bidirectional gap junction, 2 gap junction connections are needed. I'll add that to the concepts/cpp/python pages

halfflat: I think the revised statement is very clear, thanks!

halfflat (approved):

PR #1472, opened 2021-04-01
install v8 of clang, gcc

brenthuisman:
The Github Actions Ubuntu image updates of March 29, 2021 removed gcc and clang versions 8. Since we use those in our tests, this installs them [as per instruction](https://github.com/actions/virtual-environments/issues/2950).


halfflat (approved):
lgtm!

PR #1474, opened 2021-04-01
Add phase parameter to current clamps.

### LINKED ISSUE #1454: Add phase to sinusoidal stimulus current specification.
From a chat with @espenhgn, it looks as though the currently implemented current stimulus is sufficient, with the possible exception of adding an additional phase parameter for the sinusoidal component.

Task is to add this parameter, update tests, docs, etc.

Edit: we're also consistently using kHz for frequencies in other parts of the API now, so we should do so here too. This can be rolled into the same piece of work.

halfflat:
* `i_clamp` constructors take new optional phase argument.
* Pulse or box envelope clamps are now provided by the static method `i_clamp::box`, taking onset, duration, and amplitude parameters.
* Bump pybind11 version so we can have keyword-only arguments in the Python API.
* Make frequency and phase keyword-only arguments to Python `iclamp`.
* Change freuqnecy units for clamps from Hz to kHz for consistency with other interfaces.

Fixes #1454.

### DIFF
--- a/doc/concepts/decor.rst
+++ b/doc/concepts/decor.rst
@@ -290,9 +290,10 @@ See :ref:`modelgapjunctions`.
 A current stimulus is a DC or sinusoidal current of fixed frequency with a time-varying amplitude
 governed by a piecewise-linear envelope.
 
-The stimulus is described by two parameters: a frequency in Hertz, where a value of zero denotes DC;
-and a sequence of points (*t*\ :sub:`i`\ , *a*\ :sub:`i`\ ) describing the envelope, where the times
-*t*\ :sub:`i` are in milliseconds and the amplitudes *a*\ :sub:`i` are in nanoamperes.
+The stimulus is described by theww parameters:
+a sequence of points (*t*\ :sub:`i`\ , *a*\ :sub:`i`\ ) describing the envelope, where the times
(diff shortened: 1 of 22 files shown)


halfflat:
bors try

halfflat:
bors try

halfflat:
bors try

doc/concepts/decor.rst
```diff
@@ -290,9 +290,10 @@ See :ref:`modelgapjunctions`.
 A current stimulus is a DC or sinusoidal current of fixed frequency with a time-varying amplitude
 governed by a piecewise-linear envelope.
 
-The stimulus is described by two parameters: a frequency in Hertz, where a value of zero denotes DC;
-and a sequence of points (*t*\ :sub:`i`\ , *a*\ :sub:`i`\ ) describing the envelope, where the times
-*t*\ :sub:`i` are in milliseconds and the amplitudes *a*\ :sub:`i` are in nanoamperes.
+The stimulus is described by theww parameters:
```
noraabiakar, line 293: Typo 'three'

noraabiakar (changes requested):
Looks good!
I found one typo, and there's also the docs for ACC that weren't modified: https://arbor.readthedocs.io/en/latest/fileformat/cable_cell.html#label-current-clamp-envelope-pulse-delay-real-duration-real-amplitude-real-freq-real

noraabiakar (approved):

PR #1475, opened 2021-04-02
Update NEURON-style swc reader

### LINKED ISSUE #1445: Follow neuromorpho guidelines from soma representations
There is a set of rules for reconstructing soma morphologies from swc and asc files here: http://neuromorpho.org/SomaFormat.html, that describes how to handle somas described with cylinders, countours and single points.

I think that Neuron attempts to follow these rules, however in the past there were issues in at least with how this was performed in practice for SWC files.

Possible action items:

* fix asc file reader to orient the soma along the y axis, not z axis
* look at making the Allen SWC reader follow these conventions.
* revisit the NEURON SWC interpretation, because NEURON's behavior was fixed recently (though I still don't think that the intention of the loader is documented).

bcumming:
* update and simplify the `load_swc_neuron` reader to match the simplified and more correct interpretation in NEURON 8
* update the documentation on the interpretation, and add details about how it still differs from the neuromorpho descriptions. 
* update unit tests to reflect new interpretation
* add a unit test that reproduces a complex case from the NEURON test suite.
* remove the `load_swc_allen` C++ and Python calls
* add links to NeuroMorpho site in the documention

Fixes #1445.

### DIFF
--- a/arborio/swcio.cpp
+++ b/arborio/swcio.cpp
@@ -226,6 +226,10 @@ arb::morphology load_swc_neuron(const swc_data& data) {
         }
 
         // No soma: fall back to Arbor interpretation.
+        // Check for tags unsupported by NEURON beforehand.
+        if (auto it=std::find_if(R.begin(), R.end(), [](const auto& r) {return r.tag<1 || r.tag>4;}); it!=R.end()) {
+            throw swc_unsupported_tag(R[std::distance(R.begin(), it)].id);
+        }
         return load_swc_arbor(data);
     }
 
@@ -239,7 +243,10 @@ arb::morphology load_swc_neuron(const swc_data& data) {
         record_index[r.id] = i;
         old_record_index[i] = r.id;
         r.id = i;
-        r.parent_id = record_index.at(r.parent_id);
+        if (!record_index.count(r.parent_id)) {
+            throw swc_no_such_parent(r.parent_id);
+        }
+        r.parent_id = record_index[r.parent_id];
     }
 
     // Calculate meta-data
--- a/arborio/swcio.cpp
+++ b/arborio/swcio.cpp
@@ -1,8 +1,8 @@
 #include <cmath>
 #include <ios>
-#include <iostream>
 #include <limits>
 #include <numeric>
+#include <iostream>
 #include <set>
 #include <string>
 #include <sstream>
@@ -11,6 +11,8 @@
 #include <vector>
 
 #include <arbor/morph/segment_tree.hpp>
+
+#include "arbor/morph/primitives.hpp"
 
 #include <arborio/swcio.hpp>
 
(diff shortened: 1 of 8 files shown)


doc/python/morphology.rst
```diff
@@ -332,69 +332,30 @@ Cable cell morphology
...
+        as a sphere. SWC files with single point somas are common, for example
         `SONATA <https://github.com/AllenInstitute/sonata/blob/master/docs/SONATA_DEVELOPER_GUIDE.md#representing-biophysical-neuron-morphologies>`_
         model descriptions.
 
-        Such representations are unfortunate because simulation tools like Arbor and NEURON require
-        the use of cylinders or fustrums to describe morphologies, and it is not possible to
-        infer how branches attached to the soma should be connected.
+        Such representations are challenging to consistently interpret in different
+        simulation tools because they require heuristics and, often undocumented, rules
+        for how to interpret the connectin of axons and dendrites to the soma.
```
noraabiakar, line 341: 'connection'

doc/python/morphology.rst
```diff
@@ -332,69 +332,30 @@ Cable cell morphology
...
 
-        The :func:`load_swc_allen` and :func:`load_swc_neuron` functions provide support for interpreting
-        such SWC files.
+        The :func:`load_swc_neuron` function provide support for interpreting
```
noraabiakar, line 343: 'provides'

doc/python/morphology.rst
```diff
@@ -332,69 +332,30 @@ Cable cell morphology
...
-        * A single-sample soma at is constructed as a cylinder with length=diameter.
-        * If a non-soma sample is to have a soma sample as its parent, it must have the most distal sample of the soma
-          as the parent.
-        * Every non-soma sample that has a soma sample as its parent, attaches to the created soma cylinder at its midpoint.
-        * If a non-soma sample has a soma sample as its parent, no segment is created between the sample and its parent,
-          instead that sample is the proximal point of a new segment, and there is a gap in the morphology (represented
-          electrically as a zero-resistance wire)
-        * To create a segment with a certain tag, that is to be attached to the soma, we need at least 2 samples with that
-          tag.
+    Loads the :class:`morphology` from an SWC file according to to how NEURON's ``Import3D``
```
noraabiakar, line 352: 'to to how' -> 'to'

python/morphology.cpp
```diff
@@ -322,23 +285,8 @@ void register_morphology(py::module& m) {
...
-        "* Every non-soma sample that has a soma sample as its parent, attaches to the\n"
-        "  created soma cylinder at its midpoint.\n"
-        "* If a non-soma sample has a soma sample as its parent, no segment is created\n"
-        "  between the sample and its parent, instead that sample is the proximal point of\n"
-        "  a new segment, and there is a gap in the morphology (represented electrically as a\n"
-        "  zero-resistance wire)\n"
-        "* To create a segment with a certain tag, that is to be attached to the soma,\n"
-        "  we need at least 2 samples with that tag."
-        );
+        "See the documentation https://arbor.readthedocs.io/en/latest/fileformat/swc.htmla\n"
```
noraabiakar, line 288: '.htmla' -> '.html'

arborio/swcio.cpp
```diff
@@ -241,272 +206,121 @@ arb::morphology load_swc_arbor(const swc_data& data) {
...
+    std::vector<int> old_record_index(n_samples);
 
-        prev_tag = r.tag;
-        prev_id = r.id;
+    for (std::size_t i=0; i<n_samples; ++i) {
+        auto& r = records[i];
+        record_index[r.id] = i;
+        old_record_index[i] = r.id;
+        r.id = i;
+        r.parent_id = record_index.at(r.parent_id);
```
noraabiakar, line 242: Should we provide a better exception if the parent id doesn't exist?

arborio/swcio.cpp
```diff
@@ -241,272 +206,121 @@ arb::morphology load_swc_arbor(const swc_data& data) {
...
+            // soma sample.
+            throw swc_mismatched_tags(it->id);
         }
 
-        // Find record index of the parent
-        auto parent_iter = record_index.find(r.parent_id);
-
-        if (parent_iter == record_index.end() || records[parent_iter->second].id == r.id) throw swc_no_such_parent{r.id};
+        // No soma: fall back to Arbor interpretation.
+        return load_swc_arbor(data);
```
noraabiakar, line 229: This would mean that tags other than 1, 2, 3 and 4 would be allowed in this case. It's not a huge problem but it would be against the specs, right?

noraabiakar (changes requested):
looks good! I have a couple of questions, and I've highlighted a few typos.

arborio/swcio.cpp
```diff
@@ -1,8 +1,8 @@
 #include <cmath>
 #include <ios>
-#include <iostream>
 #include <limits>
 #include <numeric>
+#include <iostream>
```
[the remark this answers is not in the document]
bcumming: It's required for the `std::ostream` and `std::istream` definitions, instead of including both `<ostream>` and `<istream>`

arborio/swcio.cpp
bcumming: Fixed to throw `swc_no_such_parent`.

arborio/swcio.cpp
halfflat: This was possibly at my insistence. As a general rule, I think it is better not to impose constraints that are not necessary for a consistent and unambiguous interpretation, unless we're trying to validate against the conventions — interpretation and validation are different activities, so to speak. If we writing though, it would be a different story.

noraabiakar: Okay, but then it's still different behavior depending on whether or not a soma is present in the description.

bcumming: Indeed. I have added a tag check before the call to `load_swc_arbor`.

bcumming: @halfflat, for the NEURON interpretation we explicitly only allow tags = {1,2,3,4}, because of the special way that NEURON interprets these. These get bundled into soma, axon, dendrite and apical "section lists" by NEURON.

noraabiakar (approved):

arborio/swcio.cpp
halfflat: But, doesn't the reader (our reader) return just a morphology where these 'section lists' correspond to tagged regions?

PR #1476, opened 2021-04-06
Add code of conduct

brenthuisman:
Based on internal discussion and discussion with the governance coordinator of HBP and EBRAINS (Florent Gaillard).

Addresses #1224


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,27 @@
...
+* Please avoid using overtly provoking aliases or nicknames that might detract from a friendly, safe and welcoming environment for all.
+* Please be kind and courteous. There’s no need to be mean or rude.
+* Respect that people have differences of opinion and that every design or implementation choice carries a trade-off and numerous costs. There is seldom a right answer.
+* Please keep unstructured critique to a minimum. If you have solid ideas you want to experiment with, make a fork and see how it works.
+
+## Concerns
+
+If you believe one of these standards has been violated, you can either file an issue on an appropriate repository, reach out to the [Arbor project by email](mailto:arbor-sim@fz-juelich.de) or confidentially contact someone outside of the Arbor team, see below. Keep in mind that most mistakes are due to ignorance rather than malice.
+
+Arbor is an Ebrains/HBP project. Its policies and recourse options in case of concern are therefore available to to you. In addition, the European Commission recognizes the ALLEA Code of Conduct as a model for all research it funds, which includes Ebrains/HBP/Arbor.
```
noraabiakar, line 13: double to: available to to you

doc/contrib/index.rst
```diff
@@ -108,6 +108,28 @@ We're always happy to hear about Arbor being used! Have you used Arbor to create
...
+* Please avoid using overtly provoking aliases or nicknames that might detract from a friendly, safe and welcoming environment for all.
+* Please be kind and courteous. There’s no need to be mean or rude.
+* Respect that people have differences of opinion and that every design or implementation choice carries a trade-off and numerous costs. There is seldom a right answer.
+* Please keep unstructured critique to a minimum. If you have solid ideas you want to experiment with, make a fork and see how it works.
+
+If you believe one of these standards has been violated, you can either file an issue on an appropriate repository, reach out to the `Arbor project by email <mailto:arbor-sim@fz-juelich.de>`_ or confidentially contact someone outside of the Arbor team, see below. Keep in mind that most mistakes are due to ignorance rather than malice.
+
+Arbor is an Ebrains/HBP project. Its policies and recourse options in case of concern are therefore available to to you. In addition, the European Commission recognizes the ALLEA Code of Conduct as a model for all research it funds, which includes Ebrains/HBP/Arbor.
```
noraabiakar, line 126: same double to typo

noraabiakar (changes requested):
Looks good. Just found 2 small typos.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,27 @@
+## The Arbor Code of Conduct
+
+* We are committed to providing a friendly, safe and welcoming environment for all, regardless of level of experience, gender identity and expression, sexual orientation, disability, personal appearance, body size, race, ethnicity, age, religion, nationality, or any other characteristic.
+* Please avoid using overtly provoking aliases or nicknames that might detract from a friendly, safe and welcoming environment for all.
+* Please be kind and courteous. There’s no need to be mean or rude.
+* Respect that people have differences of opinion and that every design or implementation choice carries a trade-off and numerous costs. There is seldom a right answer.
```
halfflat, line 6: Just in terms of content: I wouldn't say 'There is seldom a right answer.' It's true of course that there are almost always trade-offs, but I don't want the document to convey the impression that this means that most changes cannot be evaluated in an objective way, even in principle.

halfflat (commented):
Apart from the specific comment on the document itself, I've only two concerns:
* It's not clear who reads mail sent to arbor-sim@fz-juelich.de — to avoid potential or perceived conflicts of interest, someone with a complaint against an Arbor developer may well choose to bypass this and go through one of the EBRAINS mechanisms, where in fact it might be something that could be dealt with fairly within the team.
* We don't mention what we would do, or the range of responses available, in the instance of a raised issue. We don't have to put this in the document, but making it clear, both to ourselves and to the community at large, I believe would help indicate that this is something we take seriously.

CODE_OF_CONDUCT.md
brenthuisman: "There may not be a right answer." or leave it out altogether?

thorstenhater: "There may not always be a right answer."

thorstenhater:
1. On the address: Absolutely yes, there should be a decicated address, and it should be very clear _who_ reads it. Who should be more than one person, for obvious reasons.
2. These are the rules we strive to live by and want to build our community upon. In the end, we can only exclude bad elements from that community, either temporarily or permanently. A concrete catalogue or something like 'first infraction that, second this' would feel wrong to me at least.

brenthuisman:
@halfflat "The Arbor Team" reads the email sent to that address, so I don't think there is any confusion: it will not be read by a 100% impartial party, but it does allow you to reach out formally (and we don't have a phone number or postal address). Then, if the person specifically wishes to avoid the team and speak to an impartial outsider directly, the appropriate ways to reach out to Ebrains are listed.

In the absence of anyone having actually raised any such issue with the Arbor team, I think it would be speculating to list anything specific in terms of handling and response. It would vary wildly by the specific concern raised and who is approached (Ebrains will set off a different chain of actions than contacting us). As is, the basic house rules are established, the path to raising concerns is described and the choice of who to reach out to is left up to the person. That to me sounds clear and fair.

CODE_OF_CONDUCT.md
halfflat: I'd just leave it out.

halfflat:
Regarding responses, it doesn't have to be a letter-of-law-and-punishment decree, but rather just what steps we would take when an email turns up, a commitment to protecting private information, that sort of thing. Both the referenced external codes have something for this — Rust in terms of its moderation guidelines, which doesn't apply directly, and Julia under its description of community stewards (direct link: https://julialang.org/community/stewards/).

brenthuisman:
Committing to handling private info with care I'll add, of course.

We could copy the complaint procedure from Julia, it sounds fair. The minor problem is that we don't and for the time being can't have a more or less independent team of stewards, on account of manpower. The obvious criticism would be that the Arbor team is investigating perhaps one of its own, which is not good. We could replace the procedure under point 3 by referring to the Ebrains ombudsman ourselves.

brenthuisman:
We have agreed that we will name names for Arbor contactpersons, and sort out how we can forward our arbor-sim address to all of us.

noraabiakar (approved):
lgtm!

brenthuisman:
On more thing: the email address. I hope to get an answer today on that domain. Do we push this with our old address in the meantime (I provisionally updated it :) ).

PR #1477, opened 2021-04-06
Expose `assign` from `sve.hpp`

noraabiakar:
Move the `assign` function out of the `detail` namespace.

### DIFF
--- a/arbor/include/arbor/simd/sve.hpp
+++ b/arbor/include/arbor/simd/sve.hpp
@@ -894,7 +894,7 @@ struct simd_cast_impl {
 };
 }  // namespace detail
 
-template <typename T, typename Other>
+template <typename T, typename Other, typename = std::enable_if_t<detail::is_sve<T>::value>>
 void assign(T& a, const Other& b) {
     a = detail::simd_cast_impl<T>::cast(b);
 }
--- a/arbor/include/arbor/simd/sve.hpp
+++ b/arbor/include/arbor/simd/sve.hpp
@@ -800,10 +800,12 @@ void compound_indexed_add(
     }
 }
 
+[[maybe_unused]]
 static int width(const svfloat64_t& v) {
     return svlen_f64(v);
 };
 
+[[maybe_unused]]
 static int width(const svint64_t& v) {
     return svlen_s64(v);
 };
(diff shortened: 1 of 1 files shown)


arbor/include/arbor/simd/sve.hpp
```diff
@@ -890,13 +892,13 @@ struct simd_cast_impl {
         return r;
     }
 };
+}  // namespace detail
 
 template <typename T, typename Other>
 void assign(T& a, const Other& b) {
     a = detail::simd_cast_impl<T>::cast(b);
 }
```
halfflat, line 900: My only concern is that this will match any `T`, and not just types which are SVE types. That nothing is breaking in testing shows that this isn't a problem in the current code, but I'm a bit worried that this might cause us grief later on.

noraabiakar: The other `assign` is a stronger match for the other simd types. In what situation could this cause an issue?

noraabiakar: I guess it's not guaranteed it's a stronger match. Should I specialize on the underlying types? 
```
template <typename Other>
void assign(svint64_t& a, const Other& b) {
    a = detail::simd_cast_impl<svint64_t>::cast(b);
}

template <typename Other>
void assign(svfloat64_t& a, const Other& b) {
    a = detail::simd_cast_impl<svfloat64_t>::cast(b);
}

template <typename Other>
void assign(svbool_t& a, const Other& b) {
    a = detail::simd_cast_impl<svbool_t>::cast(b);
}

```

halfflat: Could it perhaps be guarded with `is_sve<T>`?

noraabiakar: I forgot we added that. I'll make the change.

halfflat: I think the `enable_if_t` will need to be incorporated in a defaulted parameter to the function, so that we don't collide with another set of overloads. For example, if we want to have an interface for XYZ as well as SVE (I am thinking XYZ might be e.g. SVE2):
```
// SVE overload
template <typename T>
void assign(T& a, const Other& b, typename std::enable_if_t<is_sve<T>::value>* = 0) {
    …;
}

// XYZ overload
template <typename T>
void assign(T& a, const Other& b, typename std::enable_if_t<is_xyz<T>::value>* = 0) {
    …;
}
```

noraabiakar: Okay, I can implement it this way. Out of curiosity, is the current implementation still not safe?

halfflat: It would break in the circumstance that we have two different sorts of type that we have to handle like the SVE sorts, because the two implementations would be templated function overloads with the same signature.

halfflat (approved):

PR #1478, opened 2021-04-06
Update tinyopt to v1.0.

halfflat:
The new version of tinyopt fixes at least one bug, removes the tinyopt/smolopt distinction, and combines everything into a single header.


halfflat:
bors try

noraabiakar (approved):
Looks good!

PR #1479, opened 2021-04-08
Make proper aggregates for C++20

thorstenhater:
This removes the deletion of default constructors in arbor's `cable_cell_parameter`'s strong typedefs. When compiling with a C++20 compiler (as done in the GUI) this idiom prevents aggregate initialisation like this
```
arb::membrane_capacitance cm{42.0}; // OK in 17, fails in 20
```
Instead, values are set to `NAN`.

These are the only occurences of this idiom.
See here for the change in C++ http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1008r1.pdf

### DIFF
--- a/arbor/include/arbor/cable_cell_param.hpp
+++ b/arbor/include/arbor/cable_cell_param.hpp
@@ -89,41 +89,34 @@ struct gap_junction_site {};
 // cell-wide default:
 
 struct init_membrane_potential {
-    init_membrane_potential() = delete;
-    double value; // [mV]
+    double value = NAN; // [mV]
 };
 
 struct temperature_K {
-    temperature_K() = delete;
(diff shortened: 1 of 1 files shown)


brenthuisman:
Is a signalling NaN a better default?

thorstenhater:
Arbor throws an exception a bit deeper in the stack if these values are invalid; copy `fvm_layout.cpp`. And we use NAN everywhere else.

brenthuisman (approved):
lgtm

thorstenhater:
bors try

thorstenhater:
bors try

halfflat (approved):
While I shake my puny fist at the C++ committee, I approve of the changes.

PR #1483, opened 2021-04-12
Docs markup improvements

brenthuisman:
* Adapt CSS for the HTML5 transition in docutils 0.17, see https://github.com/readthedocs/sphinx_rtd_theme/issues/1115
   * e.g. `<div class=figure>` was changed to `<figure>`; the css is updated to match these sorts of changes.
* Fix some markup errors
* Add meta-tag to enable indexation control at Google (current indexation appears at least a year old).

### DIFF
--- a/doc/concepts/morphology.rst
+++ b/doc/concepts/morphology.rst
@@ -607,7 +607,7 @@ because it has two children: the dendrites attached to its distal end.
 Supported file formats
 ----------------------
 
-See the main menu under the "file formats" caption, or have a look in the ``doc/fileformat`` directory.
+See the supported :ref:`file formats <format-overview>`.
 
 .. _morph-cv-policies:
 
--- a/doc/concepts/morphology.rst
+++ b/doc/concepts/morphology.rst
@@ -607,17 +607,7 @@ because it has two children: the dendrites attached to its distal end.
 Supported file formats
 ----------------------
 
-Arbor supports morphologies described using the SWC file format and the NeuroML file format.
-
-SWC
-~~~
-
-See :ref:`formatswc`
-
-NeuroML
-~~~~~~~
-
-See :ref:`formatneuroml`
+See the main menu under the "file formats" caption, or have a look in the ``doc/fileformat`` directory.
 
 .. _morph-cv-policies:
 
--- a/doc/concepts/recipe.rst
+++ b/doc/concepts/recipe.rst
@@ -6,16 +6,16 @@ Recipes
 An Arbor *recipe* is a description of a model. The recipe is queried during the model
 building phase to provide information about individual cells in the model, such as:
 
-  * The **number of cells** in the model.
-  * The **kind** of each cell.
-  * The **description** of each cell, e.g. with morphology, dynamics, synapses, detectors,
-    stimuli etc.
-  * The number of **spike targets** on each cell.
-  * The number of **spike sources** on each cell.
-  * The number of **gap junction sites** on each cell.
-  * Incoming **network connections** from other cells terminating on a cell.
-  * **Gap junction connections** on each cell.
-  * **Probes** on each cell.
+* The **number of cells** in the model.
+* The **kind** of each cell.
+* The **description** of each cell, e.g. with morphology, dynamics, synapses, detectors,
+  stimuli etc.
+* The number of **spike targets** on each cell.
+* The number of **spike sources** on each cell.
+* The number of **gap junction sites** on each cell.
+* Incoming **network connections** from other cells terminating on a cell.
+* **Gap junction connections** on each cell.
+* **Probes** on each cell.
 
 Recipes are structured to provide a consistent interface for describing each cell in the
 network using their global identifier (`gid`).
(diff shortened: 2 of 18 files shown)


doc/concepts/morphology.rst
```diff
@@ -607,17 +607,7 @@ because it has two children: the dendrites attached to its distal end.
...
-SWC
-~~~
-
-See :ref:`formatswc`
-
-NeuroML
-~~~~~~~
-
-See :ref:`formatneuroml`
+See the main menu under the "file formats" caption, or have a look in the ``doc/fileformat`` directory.
```
noraabiakar, line 610: Why not just keep the links? It's easier. They don't need their own headers though.

doc/concepts/recipe.rst
```diff
@@ -26,31 +26,31 @@ which helps make Arbor fast and easily distributable over many nodes.
...
--  | ``Cell 1``: Is a soma and a single dendrite, with ``passive`` dynamics everywhere.
-     It has a single synapse at the end of the dendrite and a gap junction site in the
-     middle of the soma. This is the **description** of the cell.
-     It's also a cable cell, which is its **cell kind**. It has 0 **spike sources**, 1
-     **spike target** (the synapse) and 1 **gap junction site**.
--  | ``Cell 2``: Is a soma and a single dendrite, with ``passive`` dynamics everywhere.
-     It has a gap junction site in the middle of the soma. This is the **description**
-     of the cell. It's also a cable cell, which is its **cell kind**. It has 0
-     **spike sources**, 0 **spike targets** and 1 **gap junction site**.
+- ``Cell 0``: Is a single soma, with ``hh`` (Hodgkin-huxley) dynamics. In the middle
```
noraabiakar, line 29: Does this look right when the html is generated?

doc/fileformat/cable_cell.rst
```diff
@@ -1,7 +1,7 @@
 .. _formatcablecell:
 
-Arbor Cable Cell format (ACC)
-=============================
+Arbor Cable Cell
```
noraabiakar, line 3: We should at least mention ACC somewhere. It's a useful acronym.

noraabiakar (changes requested):
I don't know if I'm qualified to review the changes to the CSS files. 
As for the doc fixes, why are we changing bullet point indentation? It changes the rendering and not necessarily for the best

doc/concepts/morphology.rst
brenthuisman: I wanted to avoid having to keep this up to date (because it had already drifted not). Maybe a link to a `doc/fileformat/index` page?

doc/concepts/recipe.rst
brenthuisman: In fact, the surplus indentation was correctly so interpreted as a [quote block](https://docutils.sourceforge.io/docs/user/rst/quickref.html#block-quotes). For some reason the old theme did not render it as such, but this one does. I have removed indentation for this reason from various places and I have verified they render correctly.

doc/fileformat/cable_cell.rst
brenthuisman: I removed it to have consistent-ish file format headings. Shall we mention the extensions in all format headings?

brenthuisman:
Regarding the CSS changes: these are in response to changes in docutils (you can verify it by running the version with the change: https://github.com/readthedocs/sphinx_rtd_theme/issues/1115). I don't think you need to, I already have ;)

Bullet indentation: this was in fact incorrect, see comment above.

doc/fileformat/cable_cell.rst
noraabiakar: Not necessarily. I would just like it mentioned somewhere, maybe in the body of the text?

brenthuisman: Sure! I can add it for all formats, maybe even have a matrix on the fileformat index.

noraabiakar:
Can you explain what the CSS changes are supposed to do? 
I tested this branch locally, and the generated html files didn't look right.

brenthuisman:
I mentioned that in the message above you: depending on which docutils you use, output will be not the same. You probably have 0.16 or lower.These changes are compatible with the newer docutils,  used on RTD. I could require that docutils version if that helps.

noraabiakar (changes requested):

doc/fileformat/index.rst
```diff
@@ -0,0 +1,22 @@
+.. _format-overview:
```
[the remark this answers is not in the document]
brenthuisman: Not sure! This page it redundant from a TOC perspective (the formats are already listed in there), but it convenient to link to from other places in the docs when a list of supported formats needs to be referenced.

doc/index.rst
```diff
@@ -162,3 +162,6 @@ A full list of our software attributions can be found `here <https://github.com/
    contrib/doc
    contrib/example
    contrib/test
+
+.. meta::
+   :google-site-verification: KbkW8d9MLsBFZz8Ry0tfcQRkHsgxzkECCahcyRSjWDo
```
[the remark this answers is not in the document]
brenthuisman: The arborsim@gmail.com account.

doc/fileformat/index.rst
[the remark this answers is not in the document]
noraabiakar: Okay, I just thought you'd want the table to be accessible. I'll approve and leave it up to you :)

noraabiakar (approved):

doc/fileformat/index.rst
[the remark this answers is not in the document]
brenthuisman: I link to it from at least one place, I think on the morphology page)

PR #1492, opened 2021-04-16
Verify checksums for downloads

haampie:
Considering https://about.codecov.io/security-update/ it might not hurt to actually verify checksums of donwloads...

### DIFF
--- a/ci/codecov/build.Dockerfile
+++ b/ci/codecov/build.Dockerfile
@@ -20,26 +20,36 @@ RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
     update-alternatives --config gcc && \
     rm -rf /var/lib/apt/lists/*
 
+RUN cd /usr/local/bin && \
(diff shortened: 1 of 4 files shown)


haampie:
bors try

haampie:
bors try-

haampie:
bors try

bcumming:
About CodeCov: we are not actively using it, and they took 14 days (14!) between finding out that they had been hacked and notifying users.
So I would also be happy to simplify our CI/CD by removing it.
Feedback?

thorstenhater:
Is there a reason not use it locally (during CI)? The value added by the Webservice seems small.

haampie:
Hm yeah, dropping it is fine too. It should also be much easier to add codecov in github actions btw.

> Is there a reason not use it locally

What does this mean?

bcumming (approved):
Looks good!

Once this is merged, we can then make another PR to remove the code coverage report from our Daint CI, because ... well, CodeCov.

PR #1495, opened 2021-04-16
default arch='armv8-a' for arm64 platforms (M1 macs)

### LINKED ISSUE #1494: `error: the clang compiler does not support '-march=native'` on M1 macs
**Describe the bug**
A clear and concise description of what the bug is, steps to reproduce, expected behavior, etc.

For native compilation on arm64/M1 equipped machines (without Rosetta2 virtualization) the fallback `--arch=native` result in an error:
```
% which pip3
/usr/bin/pip3
% which python
/usr/bin/python
% which clang 
/usr/bin/clang
% clang --version
Apple clang version 12.0.0 (clang-1200.0.32.29)
Target: arm64-apple-darwin20.3.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
% pip3 install git+https://github.com/arbor-sim/arbor --user
...
  [ 34%] Built target build_catalogue_default_mods
  Scanning dependencies of target build_all_mods
  [ 34%] Built target build_all_mods
  Scanning dependencies of target arbor
  [ 34%] Building CXX object arbor/CMakeFiles/arbor.dir/arbexcept.cpp.o
  [ 36%] Building CXX object arbor/CMakeFiles/arbor.dir/assert.cpp.o
  clang: clang: error: the clang compiler does not support '-march=native'
  error: the clang compiler does not support '-march=native'
  gmake[2]: *** [arbor/CMakeFiles/arbor.dir/build.make:95: arbor/CMakeFiles/arbor.dir/assert.cpp.o] Error 1
  gmake[2]: *** Waiting for unfinished jobs....
...
```

In order to install it the install option `--arch=armv8`  needs to be specified which appears to be not so well documented (there is a mention of `armv8.2-a+sve` in `setup.py` however):
``` 
pip3 install -v --install-option='--arch=armv8' ...
```

Would it be helpful to incorporate a quick fix to fall back to `arch='armv8'` instead of `'native'` if say `platform.machine() == 'arm64'`? 

**Additional context**
Add any other context about the problem here (e.g. OS, Arbor version, etc.).

- Arbor: 9da761d6b7cc486b9370958c1be244d7761f4211
- MacBook Pro, M1
- macOS 11.2.3
-

espenhgn:
Please make sure your PR follows our [contribution guidelines](https://github.com/arbor-sim/arbor/tree/master/doc/contrib) and agree to the terms outlined in the [PR procedure](https://github.com/arbor-sim/arbor/tree/master/doc/contrib/pr.rst).

Closes #1494

### DIFF
--- a/cmake/CompilerOptions.cmake
+++ b/cmake/CompilerOptions.cmake
@@ -172,7 +172,9 @@ function(set_arch_target optvar arch)
 
         # Use -mcpu for all supported targets _except_ for x86, where it should be -march.
 
-        if(target_model MATCHES "x86|i[3456]86" OR target_model MATCHES "amd64" OR target_model MATCHES "aarch64")
+        if("${target}" MATCHES "aarch64-apple-darwin")
+            set(arch_opt "-mcpu=${arch}")
+        elseif(target_model MATCHES "x86|i[3456]86" OR target_model MATCHES "amd64" OR target_model MATCHES "aarch64")
             set(arch_opt "-march=${arch}")
         else()
             set(arch_opt "-mcpu=${arch}")
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@ import os
 import sys
 import setuptools
 import pathlib
+from platform import machine
 from setuptools import Extension
 from setuptools.command.build_ext import build_ext
 from setuptools.command.install import install
@@ -41,7 +42,7 @@ user_options_ = [
         ('gpu=',  None, 'enable nvidia cuda support (requires cudaruntime and nvcc) or amd hip support. Supported values: '
                         'none, cuda, cuda-clang, hip'),
         ('vec',   None, 'enable vectorization'),
-        ('arch=', None, 'cpu architecture, e.g. haswell, skylake, armv8.2-a+sve, znver2 (default native).'),
+        ('arch=', None, 'cpu architecture, e.g. haswell, skylake, armv8-a, armv8.2-a+sve, znver2 (default native).'),
         ('neuroml', None, 'enable parsing neuroml morphologies in Arbor (requires libxml)'),
         ('sysdeps', None, 'don\'t use bundled 3rd party C++ dependencies (pybind11 and json). This flag forces use of dependencies installed on the system.')
     ]
@@ -91,7 +92,8 @@ class install_command(install):
         #   vec  : generate SIMD vectorized kernels for CPU micro-architecture (boolean).
         opt['vec']  = self.vec is not None
         #   arch : target CPU micro-architecture (string).
-        opt['arch'] = "native" if self.arch is None else self.arch
+        opt['arch'] = ("armv8-a" if machine() == 'arm64' else "native"
+                       ) if self.arch is None else self.arch
         #   neuroml : compile with neuroml support for morphologies.
         opt['neuroml'] = self.neuroml is not None
         #   bundled : use bundled/git-submoduled 3rd party libraries.
@@ -126,7 +128,8 @@ if WHEEL_INSTALLED:
             #   vec  : generate SIMD vectorized kernels for CPU micro-architecture (boolean).
             opt['vec']  = self.vec is not None
             #   arch : target CPU micro-architecture (string).
-            opt['arch'] = "native" if self.arch is None else self.arch
+            opt['arch'] = ("armv8-a" if machine() == 'arm64' else "native"
+                           ) if self.arch is None else self.arch
             #   neuroml : compile with neuroml support for morphologies.
             opt['neuroml'] = self.neuroml is not None
             #   bundled : use bundled/git-submoduled 3rd party libraries.
@@ -237,4 +240,3 @@ setuptools.setup(
         'Bug Reports': 'https://github.com/arbor-sim/arbor/issues',
     },
 )
-
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,6 @@ import os
 import sys
 import setuptools
 import pathlib
-from platform import machine
 from setuptools import Extension
 from setuptools.command.build_ext import build_ext
 from setuptools.command.install import install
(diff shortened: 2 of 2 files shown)


thorstenhater (commented):
Hi Espen,

thanks for trying to fix this. We do not currently have access to a M1 machine. However,
I think it might be helpful to address this at a deeper layer, in CMake (which is used by 
setup.py). Could you try if patching `cmake/CompilerOptions.cmake` helps. Specifically
line 175
```cmake
       if(target_model MATCHES "x86|i[3456]86" OR target_model MATCHES "amd64" OR target_model MATCHES "aarch64") # ADD OR target_model MATCHES "amd64" here
            set(arch_opt "-march=${arch}")
        else()
            set(arch_opt "-mcpu=${arch}")
        endif()
```

espenhgn:
Hi @thorstenhater, 
I checked a bit and on my M1 mac where `message("${target})"` added to `make/CompilerOptions.cmake` returns `aarch64-apple-darwin20.0.0`. Cmake should use `arch_opt "-mcpu=${arch}"` then. 

I now reverted the changes to setup.py. The changes I made to `cmake/CompilerOptions.cmake` should hopefully not affect other `aarch64` builds, but perhaps there are neater ways to do this. I'm not that familiar with cmake.

cmake/CompilerOptions.cmake
```diff
@@ -172,7 +172,9 @@ function(set_arch_target optvar arch)
 
         # Use -mcpu for all supported targets _except_ for x86, where it should be -march.
 
-        if(target_model MATCHES "x86|i[3456]86" OR target_model MATCHES "amd64" OR target_model MATCHES "aarch64")
+        if("${target}" MATCHES "aarch64-apple-darwin")
+            set(arch_opt "-mcpu=${arch}")
```
thorstenhater, line 176: Interesting. For `setup.py`, it was setting `arch=armv8-a`, on CMake's level it's `cpu=aarch64`.
If it this produces correct code for you, we will gladly merge it. If you want to double-check that
vectorisation works, I tried this on my machine
```bash
objdump -d lib/libarbor.a | rg 'kamt.*rates.*:' -A 100
```
and looked for AVX instructions. 
There's another diagnostic you can enable, hidden in the `test/unit/CMakeLists.txt`
```cmake
if(${CMAKE_POSITION_INDEPENDENT_CODE})
  make_catalogue(
    NAME dummy
    SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/dummy"
    OUTPUT "CAT_DUMMY_SOURCES"
    MECHS dummy
    PREFIX "${PROJECT_SOURCE_DIR}/mechanisms"
    STANDALONE ON
    VERBOSE ON)  # Set this to ON
 ```
 Your variant will be slightly different, but only the `VERBOSE` setting is relevant.
 Running `make tests` produced this for me
 ```bash
 [0/1] Re-running CMake...
-- PYTHON_EXECUTABLE: /usr/local/Frameworks/Python.framework/Versions/3.9/bin/python3.9
Catalogue name:       dummy
Catalogue mechanisms: dummy
Catalogue sources:    /Users/hater/src/arbor/test/unit/dummy
Catalogue output:     /Users/hater/src/arbor/build/test/unit/generated/dummy
Build as standalone:  ON
Arbor arch:           -march=native
Arbor cxx compiler:
Script prefix:        /Users/hater/src/arbor/mechanisms
Current cxx compiler: /usr/local/opt/llvm/bin/clang++
```
where you can check the `arch` setting.

espenhgn: Hi, 

Yes, this PR will install locally without issues as `pip install -v .` on my machine (miniforge conda environment), without specifying `arch` (didn't check builds w. MPI yet). I can run some examples from `python/example` with the build without issues.

As for vectorization,
```
objdump -d lib/libarbor.a | grep 'kamt.*rates.*:' -A 100
```
produces a bunch of output like:
```
00000000000001d0 __ZN3arb17default_catalogue25kernel_mechanism_cpu_kamt6tratesEPNS0_22mechanism_cpu_kamt_pp_Eidd:
     1d0: ed 33 bb 6d                  	stp	d13, d12, [sp, #-80]!
     1d4: eb 2b 01 6d                  	stp	d11, d10, [sp, #16]
     1d8: e9 23 02 6d                  	stp	d9, d8, [sp, #32]
     1dc: f4 4f 03 a9                  	stp	x20, x19, [sp, #48]
     1e0: fd 7b 04 a9                  	stp	x29, x30, [sp, #64]
     1e4: fd 03 01 91                  	add	x29, sp, #64
     1e8: 08 1c a0 4e                  	mov.16b	v8, v0
     1ec: f3 03 00 aa                  	mov	x19, x0
     1f0: 08 b4 40 f9                  	ldr	x8, [x0, #360]
     1f4: 34 7c 7d 93                  	sbfiz	x20, x1, #3, #32
     1f8: 00 69 74 fc                  	ldr	d0, [x8, x20]
     1fc: 02 10 77 1e                  	fmov	d2, #-24.00000000
     200: 21 28 62 1e                  	fadd	d1, d1, d2
     204: 08 00 00 90                  	adrp	x8, #0
     208: 02 01 40 fd                  	ldr	d2, [x8]
     20c: 21 08 62 1e                  	fmul	d1, d1, d2
     210: 00 00 00 94                  	bl	#0 <__ZN3arb17default_catalogue25kernel_mechanism_cpu_kamt6tratesEPNS0_22mechanism_cpu_kamt_pp_Eidd+0x40>
     214: 09 1c a0 4e                  	mov.16b	v9, v0
     218: 60 7e 40 fd                  	ldr	d0, [x19, #248
...
```
Don't know what to look for to be frank. I see no pattern matching "AVX". 

This is what I see during build with the VERBOSE parameter ON
```
...
-- PYTHON_EXECUTABLE: /Users/espenhagen/miniforge3/envs/arbor/bin/python3.9
....
Catalogue name:       dummy
  Catalogue mechanisms: dummy
  Catalogue sources:    /private/var/folders/3r/gg20mbx93rq7b7g2q_4wgl6c0000gn/T/pip-req-build-pqzib_ei/test/unit/dummy
  Catalogue output:     /private/var/folders/3r/gg20mbx93rq7b7g2q_4wgl6c0000gn/T/pip-req-build-pqzib_ei/build/temp.macosx-11.0-arm64-3.9/test/unit/generated/dummy
  Arbor source tree:    /private/var/folders/3r/gg20mbx93rq7b7g2q_4wgl6c0000gn/T/pip-req-build-pqzib_ei
  Build as standalone:  ON
  Arbor arch:           -mcpu=native
  Arbor cxx compiler:
  Current cxx compiler: /Users/espenhagen/miniforge3/envs/arbor/bin/arm64-apple-darwin20.0.0-clang++
...
```
which looks correct w. `Arbor arch:           -mcpu=native`

thorstenhater (commented):
Looks good, and I'll approve later today. In the meantime I'll see if I can test this PR on another ARM machine.

espenhgn:
> Looks good, and I'll approve later today. In the meantime I'll see if I can test this PR on another ARM machine.

Sounds good!

thorstenhater (approved):

PR #1501, opened 2021-04-23
Add Leak and Address Sanitizers.

thorstenhater:


thorstenhater:
bors try

.github/workflows/sanitize.yml
```diff
@@ -0,0 +1,132 @@
...
+        simd:      ["ON", "OFF"]
+        sanitizer: ["address", "leak", "undefined", "memory", "thread"]
+    env:
+        CC:         ${{ matrix.cc }}
+        CXX:        ${{ matrix.cxx }}
+        # We set PYTHONPATH instead of installing arbor to avoid distribution/OS specific behaviour.
+        PYTHONPATH: ${{ github.workspace }}/build/python
+        # This is a workaround for the unfortunate interaction of MacOS and OpenMPI 4
+        # See https://github.com/open-mpi/ompi/issues/6518
+        OMPI_MCA_btl: "self,tcp"
```
halfflat, line 30: I know this has been copied from the other workflow, and perhaps leaving it in is best to avoid future surprises, but we shouldn't need this for the Ubuntu image.

thorstenhater: Yes, that can go. I do not expect us going for more MacOS.

test/unit/test_recipe.cpp
```diff
@@ -46,18 +46,23 @@ namespace {
             return cell_kind::cable;
         }
         std::vector<gap_junction_connection> gap_junctions_on(cell_gid_type gid) const override {
+            if (gid >= gap_junctions_.size()) return {};
```
halfflat, line 49: Here and below, if this is to keep the sanitizer happy, I'd go with `gap_junctions_.at(gid)` instead. Then at least in a non-sanitized environment, we'd get a useful exception if the implicit contract is broken.

modcc/printer/cexpr_emit.hpp
```diff
@@ -60,9 +60,9 @@ class SimdExprEmitter: public CExprEmitter {
 
 protected:
     static std::unordered_set<std::string> mask_names_;
-    bool processing_true_;
-    bool is_indirect_;
-    std::string current_mask_, current_mask_bar_, input_mask_;
+    bool processing_true_ = false;
+    bool is_indirect_ = false;
+    std::string current_mask_ = "", current_mask_bar_ = "", input_mask_ = "";
```
halfflat, line 65: Strings do not need to be explicitly initialized. Was this a warning?

arbor/util/piecewise.hpp
```diff
@@ -484,11 +484,13 @@ inline pw_elements<void> zip(const pw_elements<void>& a, const pw_elements<void>
...
     double b_right = b.interval(bi).second;
 
-    while (left<rmin) {
+    for(;;) {
         double right = std::min(a_right, b_right);
         right = std::min(right, rmin);
 
         z.push_back(left, right);
+        if (right==rmin) break;
+
```
halfflat, line 493: I suspect there might be some further dodginess in the zip routines, when it comes to the case of zero-length elements at the end of the common support. What was the error that sanitizing was reporting?

halfflat (changes requested):
Some comments above/below. Most important request is the change in `test_recipe.cpp` — I've found the corresponding problem in `communicator.cpp` and will make a separate fix PR.

arbor/util/piecewise.hpp
thorstenhater: If I remember correctly, one level further down, the `second` of a tuple of iterators is dereferenced.

modcc/printer/cexpr_emit.hpp
thorstenhater: Actually, just `processing_true_` _needs_ to be initialised to avoid an uninitialised read. However, I added the rest for consistency.

test/unit/test_recipe.cpp
thorstenhater: The test assumes a `arb::bad_source/_target` to be thrown; this change would break that assumption, right?

modcc/printer/cexpr_emit.hpp
halfflat: `bool` members really should be initialized, I agree. I think we can leave the strings alone though, because we can be sure they won't ever be a problem.

test/unit/test_recipe.cpp
halfflat: Yes: a bad source error _should_ be thrown, but isn't because of a bug in communication.cpp:
```
diff --git a/arbor/communication/communicator.cpp b/arbor/communication/communicator.cpp
index b93638f8..16b08fb5 100644
--- a/arbor/communication/communicator.cpp
+++ b/arbor/communication/communicator.cpp
@@ -80,11 +80,10 @@ communicator::communicator(const recipe& rec,
     for (const auto& cell: gid_infos) {
         auto num_targets = rec.num_targets(cell.gid);
         for (auto c: cell.conns) {
-            auto num_sources = rec.num_sources(c.source.gid);
             if (c.source.gid >= num_total_cells) {
                 throw arb::bad_connection_source_gid(cell.gid, c.source.gid, num_total_cells);
             }
-            if (c.source.index >= num_sources) {
+            if (auto num_sources = rec.num_sources(c.source.gid); c.source.index >= num_sources) {
                 throw arb::bad_connection_source_lid(cell.gid, c.source.index, num_sources);
             }
             if (c.dest >= num_targets) {
```

modcc/printer/cexpr_emit.hpp
thorstenhater: As a matter of preference I do initialise all members, but yes, unneeded and I can remove it.

halfflat (approved):
Looks good, and it was nice to catch those other issues!

PR #1502, opened 2021-04-23
Expose cv_policy_explicit in Python and Python docs.

brenthuisman:


doc/python/morphology.rst
```diff
@@ -522,6 +537,9 @@ constitute part of the CV boundary point set.
 
 .. _pyneuroml:
 
+NeuroML
```
noraabiakar, line 540: There's also an SWC section here somewhere. If this doesn't have a header, that probably doesn't either.

brenthuisman: It was all a bit messed up, I fixed the ordering and added headings to the file formats.

bcumming: I checked over the file, and the new ordering (input file formats at the bottom of the file) is more logical.

bcumming (approved):
looks good!

PR #1503, opened 2021-04-27
spike sorting on python side

bcumming:
Ensure that spikes output by the standard Python spike recorder are sorted lexigraphically by time, then gid, then lid.

Also remove deprecated spike handling code from the Python wrapper while we are at it.


halfflat (commented):
Looks great!
Should the PR update docs, and add a unit test for the sorting?

python/test/unit/test_spikes.py
```diff
@@ -0,0 +1,95 @@
...
+        # run simulation in 5 steps, forcing 5 epochs
+        sim.run(1, 0.01)
+        sim.run(2, 0.01)
+        sim.run(3, 0.01)
+        sim.run(4, 0.01)
+        sim.run(5, 0.01)
+
+        spikes = sim.spikes().tolist()
+        times = [t for s, t in spikes]
+        gids  = [s[0] for s, t in spikes]
```
halfflat, line 80: Just for clarity, I propose changing the above to:
```
spikes = sim.spikes()
times = spikes["time"].tolist()
gids = spikes["source"]["gid"].tolist()
```

halfflat (changes requested):
Looks good! I've requested a small change in the unit test to make clearer what is going on with the spikes array. We're not testing for sorting on source index, but I don't the hassle is worth it.

python/simulation.cpp
```diff
@@ -87,7 +87,14 @@ class simulation_shim {
 
     void record(spike_recording policy) {
         auto spike_recorder = [this](const std::vector<arb::spike>& spikes) {
+            auto old_size = spike_record_.size();
+            // Append the new spikes to the end of the spike record.
             spike_record_.insert(spike_record_.end(), spikes.begin(), spikes.end());
+            // Sort the newly appended spikes.
+            std::sort(spike_record_.begin()+old_size, spike_record_.end(),
+                    [](const auto& lhs, const auto& rhs) {
+                        return std::tie(lhs.time, lhs.source.gid, lhs.source.index)<std::tie(rhs.time, rhs.source.gid, rhs.source.index);
```
schmitts, line 96: What about implementing `<=>` for `arb::spike`?

halfflat: I'd prefer not to, but only because the ordering is not natural: we could have just as easily sorted first on GID, for example.

bcumming:
bors try

halfflat (approved):
Looks good!

python/test/unit/test_spikes.py
bcumming: Done

PR #1504, opened 2021-04-28
Labels instead of indices for placeable item identfication.

### LINKED ISSUE #1394: Discussion: synapse identification.
As raised in the comments on PR #1235, it is hard to identify synapses associated with a decor placement given by a locset expression:
* The number of created synapses depends upon the thingification of the locset expression. If this number is not exactly one for all placed synapses, the value returned by `decor::place()` will be incorrect.
* When the locset involves simulation-construction time parameters (e.g. per-sim random seeds, which we plan on supporting), it is not even possible in principle to determine the indices of the corresponding synapses, without independently evaluating all the placement expressions against the corresponding mprovider object, which would have to be obtained via a new interface.

The flat nature of decor is at odds with the requirement that there be an `mprovider` object in hand with which to evaluate the given locset expressions.

A related issue at a higher level arises when describing population/projection style network descriptions: when connection endpoints require both a cell from a population and a synapse on that cell, there must be a language by which the relevant synapse or subset of synapses can be identified independently of a particular cell.

A first rough proposal:
* Targets (in particular, synapses on cable cells) are associated with one (or possible more) labels. These might be text labels, or numeric. For cable cell decor, these would constitute part of the synapse description.
* Connection targets as returned by `recipe::connections_on` are given in terms of target gid, target synapse label, and possibly a third parameter which is interpreted specifically by a cable-kind specific handler and which describes how to resolve ambiguity when the label does not correspond to exactly one end-point.
* A translation phase resolves these label descriptions of targets into concrete target ids after cell groups have instantiated the cells.

A similar problem arose with probe ids: probe ids are assigned sequentially on a given cell, but each id may correspond to zero or more actual probes, depending upon the corresponding locset evaluation. This was less of an issue, because the metadata for the probes associated with a probe id is intended to be queried by the application after a simulation object is created and the corresponding cells are fully realized.

noraabiakar:
New structs and types:
* `cell_tag_type` (std::string): for labeling placeable items on a cell. The label refers to a number of items placed on a locset, equal to the number of locations in a locset. The number of locations in not always known to the user, so the previous way of using indices for items was no longer sufficient. 
* `lid_selection_policy`: for allowing a user to select a single item from a group of items sharing a label. Currently only `round_robin` and `assert_univalent` are supported. 
* `cell_local_label_type` and `cell_global_label_type`: for identifying the target and source of a connection or gap_junction connection. 
* `cell_label_ranges`, and `cell_labels_and_gids`: for propagating information about the labeled items on the cell from the cell groups back to the simulation and communicator. 
* `label_resolution_map` and `resolver`: for selecting an item (and retaining state) from a labeled group of items on a cell according to a user-selected policy.

Changes to the model-initialization: 
* The `communicator` now needs `label_resolution_maps` constructed from the cell group data in order to build the `connections` vectors. 
* The `simulation_state` object handles the transfer of the information when it is constructed. 
* Spike exchange at runtime remains unchanged, because `communicator::connections` remains unchanged. 

Changes to cells, cell_groups and recipe:
* `decor::place` expects a third label parameter, no longer returns an `lid_range`.
* `lif`, `source`, and `benchmark` cells need source/target labels in their constructors. 
* A `cell_group` needs to save data about the gid/labels/lid_ranges of each cell, to propagate back to the `communicator` constructor. 
* Connections/gap junction connections are formed between {label, policy} pairs on cells instead of indices. 
* `num_sources`, `num_targets`, `num_gap_junction_sites` deleted from `recipe`.

Additional changes:
* Add mpi wrapper for exchanging vectors of strings. 
* Fix unit tests. 
* Fix python wrapper.
* Fix C++ examples.
* Fix python examples.
* Fix docs. 
* Add unit tests. 

Fixes #1394


noraabiakar:
bors try

noraabiakar:
bors try

noraabiakar:
bors try

arbor/fvm_lowered_cell.hpp
```diff
@@ -211,6 +211,13 @@ struct fvm_lowered_cell {
         std::vector<deliverable_event> staged_events,
         std::vector<sample_event> staged_samples) = 0;
 
+    virtual cell_labeled_ranges detector_data() const = 0;
+    virtual cell_labeled_ranges synapse_data() const = 0;
+    virtual cell_labeled_ranges gap_junction_data() const = 0;
+
+    virtual fvm_size_type num_detectors(cell_gid_type gid) const = 0;
+    virtual fvm_size_type num_synapses(cell_gid_type gid) const = 0;
+
```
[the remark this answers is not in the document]
noraabiakar: I guess you mean `mc_cell_group` instead of `mc_cable_cell`?
I would agree, but we actually need the `cell_labelled_ranges` in `fvm_lowered_cell_impl` in order to resolve the gap junction labels.

arbor/cable_cell.cpp
```diff
@@ -88,7 +88,12 @@ struct cable_cell_impl {
             placed<Item> p{l, lid++, item};
             mm.push_back(p);
         }
-        return lid_range(first, lid);
+        auto range = lid_range(first, lid);
+        auto& lid_ranges = labeled_lid_ranges.get<Item>();
+        if (lid_ranges.count(label)) {
+            throw arb::cable_cell_error(util::pprintf("Duplicate label detected \"{}\"", label));
+        }
```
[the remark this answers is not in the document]
noraabiakar: Okay, that makes sense.

arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -468,12 +525,16 @@ void fvm_lowered_cell_impl<Backend>::initialize(
 
     // Fill src_to_spike and cv_to_cell vectors only if mechanisms with post_events implemented are present.
     post_events_ = mech_data.post_events;
-    auto max_detector = post_events_ ? util::max_value(nsources) : 0;
+    auto max_detector = 0;
+    if (post_events_) {
+        auto it = std::max_element(num_detectors_.begin(), num_detectors_.end(), [](const auto& lhs, const auto& rhs) -> bool {return lhs.second < rhs.second;});
```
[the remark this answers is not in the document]
noraabiakar: thanks :)

arbor/label_resolver.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <arbor/assert.hpp>
+#include <arbor/common_types.hpp>
+
+namespace arb {
+
+// Data required for {gid, label} to lid resolution.
+// (gids, sizes) and (labels, ranges) are expected to have the same size.
+// `sizes` is a partitioning vector for associating a gid, with a (label, range) pair.
+struct cell_labeled_ranges {
```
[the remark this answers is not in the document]
noraabiakar: I should have checked! I'll use `cell_label_range` to avoid issues in the future :)

arbor/label_resolver.cpp
```diff
@@ -0,0 +1,90 @@
...
+    arb_assert(gids.size() == sizes.size());
+};
+
+void cell_labeled_ranges::append(cell_labeled_ranges other) {
+    using util::append;
+    append(gids, other.gids);
+    append(sizes, other.sizes);
+    append(labels, other.labels);
+    append(ranges, other.ranges);
+}
```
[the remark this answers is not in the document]
noraabiakar: Right, I tried to add a `util::append` capable of moving the vectors in (using `std::make_move_iterator`) but I ran into some issues there (related to appending `util::transform_view` - data was being moved instead of copied). Granted, I'm not an expert on move semantics yet. Do you think adding this method is worthwhile?

arbor/fvm_lowered_cell.hpp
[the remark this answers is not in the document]
halfflat: I did indeed mean `mc_cell_group`, thanks :)
That resolution though is required only during the initialization phase; there's no need to hold on to it down in `fvm_lowered_cell_impl` after that, right?

arbor/label_resolver.cpp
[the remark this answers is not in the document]
halfflat: If we can get the semantics straight, it would be handy, but it might not be worth the effort (i.e., perhaps just use `std::` functionality.)

doc/python/cell.rst
```diff
@@ -9,19 +9,104 @@ The types defined below are used as identifiers for cells and members of cell-lo
...
+   Enumeration used for selecting an individual item from a group of items sharing the
+   same label.
+
+   .. attribute:: round_robin
+
+      Iterate over the items of the group in a round-robin fashion.
+
+   .. attribute:: assert_univalent
+
+      Assert that ony one item is available in the group. Throws an exception if the assertion
```
schmitts, line 23: ony -> only

arbor/include/arbor/event_generator.hpp
```diff
@@ -68,9 +68,10 @@ struct empty_generator {
     event_seq events(time_type, time_type) {
         return {nullptr, nullptr};
     }
-    std::vector<cell_lid_type> targets() {
+    std::vector<cell_local_label_type> targets() {
         return {};
     };
+    void init(const std::vector<cell_lid_type>&) {};
```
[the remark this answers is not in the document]
noraabiakar: Regarding the different semantics between the schedule-based and explicit generators, I have changed it so that both will cycle through the range of possible lids according to the selection policy.

noraabiakar:
bors try

noraabiakar:
bors try

halfflat (changes requested):
Main concerns are about the redundancy of having cell groups write the cell label gid info, and the semantics where labels are associated with more than one range.

arbor/fvm_lowered_cell.hpp
```diff
@@ -193,17 +193,34 @@ struct probe_association_map {
...
+    // Handles for accessing lowered cell.
+    std::vector<target_handle> target_handles;
+
+    // Maps probe ids to probe handles and tags.
+    probe_association_map probe_map;
+
+    // Structs required for {gid, label} to lid resolution
+    cell_label_range source_data;
+    cell_label_range target_data;
+    cell_label_range gap_junction_data;
```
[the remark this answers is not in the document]
noraabiakar: I kept it in for the unit tests, and I thought it wouldn't hurt performance. I can remove those tests though.

noraabiakar:
bors try

noraabiakar:
bors try

doc/python/interconnectivity.rst
```diff
@@ -7,24 +7,28 @@ Interconnectivity
...
 
     .. function:: connection(source, destination, weight, delay)
 
         Construct a connection between the :attr:`source` and the :attr:`dest` with a :attr:`weight` and :attr:`delay`.
 
     .. attribute:: source
 
-        The source end point of the connection (type: :class:`arbor.cell_member`, which can be initialized with a (gid, index) tuple).
+        The source end point of the connection (type: :class:`arbor.cell_global_label`, which can be initialized with a
+        (gid, label) or a (gid, label, policy) tuple. If the policy is not indicated, the default
```
halfflat, line 23: The constructor doesn't have a 3-tuple form as currently defined, so it'd be `(gid, (label, policy))`.
Given that gap junctions currently require a univalent policy, should the default still be round_robin? Another option might be to force an explicit policy, even though it's pretty verbose.

doc/tutorial/network_ring.rst
```diff
@@ -97,20 +104,24 @@ are connecting the cells **(8)**, returning a configurable number of cells **(6)
...
-source and one target endpoint.
+previous cell has a gid ``(gid-1)%self.ncells``. The connection has a weight of 0.1 μS and a delay of 5 ms.
+The first two arguments to :py:class:`arbor.connection` are the **source** and **target** of the connection.
+
+The **source** is a :py:class:`arbor.cell_global_label` object containing a cell index ``gid``, the source label
+corresponding to a valid detector label on the cell and an optional selection policy (for choosing a single detector
+out of potentially many detectors grouped under the same label - remember, in this case the number of detectors labeled
+'detector' is 1).
+The :py:class:`arbor.cell_global_label` can be initialized with a ``(gid, label)`` tuple, in which case the selection
+policy is the default :py:attr:`arbor.selection_policy.round_robin`; or a ``(gid, label, policy)`` tuple.
```
halfflat, line 115: As noted above, the constructor currently takes only a 2-tuple form, not a 3-tuple.

python/recipe.hpp
```diff
@@ -74,18 +74,6 @@ class py_recipe_trampoline: public py_recipe {
...
-    }
-
-    arb::cell_size_type num_targets(arb::cell_gid_type gid) const override {
-        PYBIND11_OVERLOAD(arb::cell_size_type, py_recipe, num_targets, gid);
-    }
-
-    arb::cell_size_type num_gap_junction_sites(arb::cell_gid_type gid) const override {
-        PYBIND11_OVERLOAD(arb::cell_size_type, py_recipe, num_gap_junction_sites, gid);
-    }
-
```
halfflat, line 88: `num_sources`, `num_targets`, and `num_gap_junction_sites` should be removed from `py_recipe` too.
A number of recipes in the python examples still have `num_xxx` methods defined.

arbor/label_resolution.cpp
```diff
@@ -0,0 +1,154 @@
...
+        // Get the state of the round_robin iterator.
+        auto& rr_state = state_map[iden.gid][iden.label.tag][iden.label.policy];
+        auto idx = std::get<round_robin_state>(rr_state).state;
+
+        // Update the state of the round_robin iterator.
+        rr_state = round_robin_state((idx+1) % range_set.size());
+
+        // Get the lid at the current index.
+        return range_set.at(idx);
+    }
```
halfflat, line 141: As an indentation thing, because these inner blocks demarcated by braces are one scope deeper than that of the switch statement, it'd be clearer if they were indented (for the same reason essentially that we _don't_ indent case labels in switches).

e.g.
```
switch (...) {
case foo:
    // in scope of 'switch' here.
    {
        // in scope of inner block here.
        ...
    }
case bar: // etc,
}
```

halfflat (changes requested):
Most comments are minor; otherwise concerned about copying resolution maps across event generators, and mooting the possibility of making the state objects for lid selection policies responsible for their own iteration.

doc/python/interconnectivity.rst
noraabiakar: I think we can change the default to assert_univalent. Do yo

halfflat: I think it's fine to default to univalent. The verbosity of specifying a policy explicitly (and similarly with other enum-using interfaces) is something we can look at again in the future.

arbor/simulation.cpp
```diff
@@ -198,20 +224,23 @@ simulation_state::simulation_state(
...
-            // Check validity of event_generator targets
+            // Resolve event_generator targets
             auto event_gens = rec.event_generators(gid);
-            auto num_targets = rec.num_targets(gid);
-            for (const auto& g: event_gens) {
-                for (const auto& t: g.targets()) {
-                    if (t >= num_targets) {
-                        throw arb::bad_event_generator_target_lid(gid, t, num_targets);
+            for (auto& g: event_gens) {
+                g.resolve_label([target_resolution_map, event_resolver, gid](const cell_local_label_type& label) mutable {
```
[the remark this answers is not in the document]
noraabiakar: I think they should each get their own resolver state. I'll implement the shared pointer approach.

doc/concepts/interconnectivity.rst
```diff
@@ -51,11 +57,14 @@ A recipe lets you define which sites are connected to which.
...
 
-      1. Create a **gap junction site** on two separate cells as part of their
+      1. Create labeled **gap junction sites** on two separate cells as part of their
          :ref:`cell descriptions <modelcelldesc>` in the :ref:`recipe <modelrecipe>`.
-      2. Declare the Gap Junction connections in the recipe: between a peer and local **gap junction site**
-         identified using :gen:`cell_member` and a conductance in μS. Two of these connections are needed,
-         on each of the peer and local cells.
+         Each labeled group of gap junctions may contain multiple items on possibly multiple locations
+         on the cell.
+      2. Declare the Gap Junction connections in the recipe *on the local cell*: from a peer **gap junction site**
```
brenthuisman, line 64: What is the significant of the word 'peer'? I don't see it used before.

doc/concepts/interconnectivity.rst
```diff
@@ -51,11 +57,14 @@ A recipe lets you define which sites are connected to which.
...
+         identified using a :gen:`global_label`; to a local **gap junction site** identified using a
+         :gen:`local_label` (:gen:`gid` of the site is implicitly known); and a conductance in μS.
```
brenthuisman, line 66: "implicitly known": can you add that it is known by way of the ``gid`` of the recipe function call?

doc/concepts/recipe.rst
```diff
@@ -78,11 +64,11 @@ Are recipes always necessary?
 Yes. However, we provide a python :class:`single_cell_model <py_single_cell_model>`
 that abstracts away the details of a recipe for simulations of  single, stand-alone
 :ref:`cable cells<modelcablecell>`, which absolves the users from having to create the
-recipe themselves. This is possible because the number of cells, spike targets, spike sources
-and gap junction sites is fixed and known, as well as the fact that there can be no connections
-or gap junctions on a single cell. The single cell model is able to fill out the details of the
-recipe under the hood, and the user need only provide the cell description, and any probes they
-wish to place on the cell.
+recipe themselves. This is possible because the number of cells is fixed and known,
+and it is guaranteed that there can be no connections or gap junctions in a model of a
```
brenthuisman, line 68: Well, this is not true in the neuroscientific sense, right? Autapses etc. The guarantee is that Arbor does not support it, right?

doc/python/interconnectivity.rst
```diff
@@ -45,20 +49,20 @@ Interconnectivity
             import arbor
 
             def connections_on(gid):
-               # construct a connection from the 0th source index of cell 2 (2,0)
-               # to the 1st target index of cell gid (gid,1) with weight 0.01 and delay of 10 ms.
-               src  = arbor.cell_member(2,0)
-               dest = 1 # gid of the destination is is determined by the argument to `connections_on`
+               # construct a connection from the "detector" source label on cell 2 (2,0)
```
brenthuisman, line 52: (2,0) can be removed.

brenthuisman (approved):
I reviewed the docs: minor comments.

doc/concepts/interconnectivity.rst
noraabiakar: synaptic connections have a source and a target, because they are directed. Gap junction connections have a local site and a peer site, because they are bidirectional.

doc/concepts/recipe.rst
noraabiakar: Yes

doc/concepts/interconnectivity.rst
brenthuisman: Thanks, I see! Maybe this deserves to be mentioned explicitly, for those not too current on graph-theory.

arbor/label_resolution.cpp
```diff
@@ -118,36 +123,48 @@ label_resolution_map::label_resolution_map(const cell_labels_and_gids& clg) {
...
+
+    // Construct state if if doesn't exist
+    if (!state_map_[iden.gid][iden.label.tag].count(iden.label.policy)) {
+        state_map_[iden.gid][iden.label.tag][iden.label.policy] = construct_state(iden.label.policy);
     }
-    default: throw arb::bad_connection_label(iden.gid, iden.label.tag);
+
+    auto lid = std::visit([range_set](auto& state) { return state.update(range_set); }, state_map_[iden.gid][iden.label.tag][iden.label.policy]);
+    if (!lid) {
+        throw(arb::bad_connection_label(iden.gid, iden.label.tag, lid.error().what()));
```
halfflat, line 165: Minor formatting: `throw xyz;` instead of `throw(xyz);`

arbor/label_resolution.hpp
```diff
@@ -3,12 +3,19 @@
...
+#include <arbor/util/expected.hpp>
+
+#include "util/partition.hpp"
 
 namespace arb {
 
+struct resolution_error: arb::arbor_exception {
+    resolution_error(const std::string& msg);
+};
+using lid_hopefully = arb::util::expected<cell_lid_type , resolution_error>;
```
halfflat, line 17: `resolution_error` needn't be derived from an exception class; it can just be a wrapper around a string (or other data). Or even just a bare string. As we don't throw them, inheriting from `arbor_exception` is misleading.

halfflat (commented):
Only a couple of minor changes requested as marked. Happy to edit PR in place, if you're fine with that!

noraabiakar:
bors try

halfflat (approved):
Great! Thanks for going through all these revisions!

PR #1505, opened 2021-04-28
Propagate python unit test failure to github actions. 

noraabiakar:
* Fix python unit tests
* Return exit code 1 if the python unit tests fail to ensure github actions fail too


halfflat (approved):
Great! Thanks for the catch and the fix.

PR #1510, opened 2021-04-29
Add default hidden visibliity to pyarb shared lib.

### LINKED ISSUE #1509: Visibility warnings when building pyarb
We are getting warnings such as the following when building the Python library:
```
[192/199] Building CXX object python/CMakeFiles/pyarb_obj.dir/simulation.cpp.o
In file included from ../python/simulation.cpp:11:
../python/pyarb.hpp:57:8: warning: 'pyarb::probe_meta_cvt_map' declared with greater visibility than the type of its field 'pyarb::probe_meta_cvt_map::map_' [-Wattributes]
   57 | struct probe_meta_cvt_map {
      |        ^~~~~~~~~~~~~~~~~~
```

IIRC at some point we had default visibility hidden on `pyarb`, perhaps inherited from the `pybind11` module target, but currently we are not compiling with `-fvisibility=hidden`. Pybind11 documentation strongly recommends that we do so.

halfflat:
Fixes #1509.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -45,6 +45,7 @@ set(pyarb_source
 # use by both the Python wrapper target (pyarb) and for the
 # unit tests of the C++ components in the Python wrapper.
 add_library(pyarb_obj OBJECT ${pyarb_source})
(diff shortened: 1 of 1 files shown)


bcumming (approved):
looks good!

PR #1512, opened 2021-04-30
Add Fujitsu ARM-SVE CI test

### LINKED ISSUE #1498: [CI/bors] ARM testing
1. Can we have bors run our testsuite on the A64FX machines at ault? @haampie perhaps you know if this is possible and how we should go about this?
2. Alternatively, we should forego of SVE testing and settle for a Qemu GA.

bcumming:
Add a CI test for ARM-SVE

- compiles and runs the unit tests with vectorization enabled for the `armv8.2-a+sv` architecture
- uses a dedicated gitlab runner on ault.cscs.ch

Fixes #1498


bcumming:
bors try

ci/test-ault-fujitsu.sh
```diff
@@ -0,0 +1,23 @@
+#! /bin/bash
+
```
schmitts, line 2: set -euxo pipefail

bcumming: I don't understand the comment?

bcumming:
bors try

ci/test-ault-fujitsu.sh
schmitts: Cf. https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

bcumming: ok, good to know. We don't need the `-x`, because gitlab prints those out nicely for us.
`set -euo pipefail` looks a good option.

bcumming:
bors try.

brenthuisman:
Last job failed with "Requested node configuration is not available". Is that a problem with gitlab runners/allocation at ault?

haampie:
There's a couple `SLURM_*` variables set for all jobs that shouldn't apply to the fujitsu job

bcumming:
bors try

brenthuisman:
Side comment: how does one review an issue like this? I can't really verify/run anything myself, just see if `bors try` succeeds, which you already have.

bcumming:
Good question.

I would propose reading the test code to see if looks good, then verify that it passes CI (like normal, I suppose)

ci/gitlab-cscs.yml.only-ault
```diff
@@ -0,0 +1,11 @@
+stages:
```
noraabiakar, line 1: Do we still need this file?

cmake/CompilerOptions.cmake
```diff
@@ -6,12 +6,6 @@ include(CMakePushCheckState)
 set(CXXOPT_DEBUG "-g")
 set(CXXOPT_CXX11 "-std=c++11")
 
-if(CMAKE_CXX_COMPILER_ID MATCHES "XL")
```
noraabiakar, line 9: Is this no longer an issue?

ci/gitlab-cscs.yml
```diff
@@ -87,6 +121,15 @@ multi node release:
     SLURM_NTASKS: 2
     ALLOCATION_NAME: arbor-ci-release-$CI_PIPELINE_ID
 
+test-ault-fujitsu:
```
noraabiakar, line 124: Minor nitpick: I think replacing the `-` with spaces would match the other test names better. Also the `_` in `notify_github_*`)

ci/gitlab-cscs.yml
```diff
@@ -87,6 +121,15 @@ multi node release:
...
+    needs: []
+    tags:
```
noraabiakar, line 126: Do we need: `only: ['master', 'staging', 'trying']` here?

noraabiakar (commented):
I have a few questions and one suggestion, otherwise looks good!

ci/gitlab-cscs.yml
brenthuisman: Would that mean you can't run a `bors try` on any PR?

noraabiakar: I think our PRs get tested on the `trying` branch.

ci/gitlab-cscs.yml.only-ault
bcumming: no, we don't. That was a backup while I was debugging the Ault pipeline.

ci/gitlab-cscs.yml
haampie: see https://bors.tech/documentation/getting-started/

cmake/CompilerOptions.cmake
bcumming: We stopped support for the IBM XL compiler long ago, so those flags will be the least of your concerns if you try to compile with it today. I just did that clean up when I was modifying something else in the same file.

ci/gitlab-cscs.yml
bcumming: I changed them all to underscore, because on balance that was the most widely used convention in the files.

bcumming:
bors try

ci/gitlab-cscs.yml
noraabiakar: I was mostly wondering why the settings for `test_ault_fujitsu` are different from `single_node_release` and `multi_node_release`.

bcumming: The `test_ault_fuji` test is performed in a single stage by a runner that executes a script that configures, builds, then runs the test on an ARM node.
The tests on Daint are performed in stages, where Arbor is first built in a container, then later stages that allocate nodes and run the tests.
In an ideal world we could put the daint tests in one yml file, and the ault tests in another, because they are not related at all. But Gitlab runners force us to put all the tests in a single file (unlike GitHub Actions).

noraabiakar: Thanks for the explanation :)

noraabiakar (approved):

PR #1513, opened 2021-05-03
Fix communicator source gid bounds check.

halfflat:
* Correct bounds check order in communicator.
* Use `.at()` in test recipe to catch failures in bound checking.

Fixes bug highlighted in PR #1501.


thorstenhater (approved):
👍

PR #1515, opened 2021-05-03
Clarify/rewrite pw_elements zip.

### LINKED ISSUE #1514: Problems with `pw_elements` zip.
There is trouble in the `pw_elements` zip routine
* Sanitized build found a bug in the `void` version (see PR #1501).
* It doesn't behave sensibly in the presence of zero-width elements.
* The desired semantics aren't really described anywhere.

halfflat:
* Use a struct with meaningful field names instead of a `std::pair` to represent elements in `pw_elements`.
* Describe explicitly the semantics of the zip operation.
* Update implementation and add unit tests to suit.

Fixes #1514.

### DIFF
--- a/arbor/morph/embed_pwlin.cpp
+++ b/arbor/morph/embed_pwlin.cpp
@@ -93,9 +93,9 @@ mcable_list data_cmp(const branch_pw_ratpoly<1, 0>& f, unsigned bid, double val,
     mcable_list L;
     const auto& pw = f.at(bid);
     for (const auto& piece: pw) {
-        auto extents = piece.first;
-        auto left_val = piece.second(0);
-        auto right_val = piece.second(1);
+        auto extents = piece.interval;
+        auto left_val = piece.element(0);
+        auto right_val = piece.element(1);
 
         if (!op(left_val, val) && !op(right_val, val)) {
             continue;
@@ -275,11 +275,11 @@ embed_pwlin::embed_pwlin(const arb::morphology& m) {
             segment_cables_[seg.id] = mcable{bid, pos0, pos1};
         }
 
-        double length_0 = parent==mnpos? 0: data_->length[parent].back().second[1];
+        double length_0 = parent==mnpos? 0: data_->length[parent].back().element[1];
         data_->length[bid].push_back(0., 1, rat_element<1, 0>(length_0, length_0+branch_length));
 
-        double area_0 = parent==mnpos? 0: data_->area[parent].back().second[2];
-        double ixa_0 = parent==mnpos? 0: data_->ixa[parent].back().second[2];
+        double area_0 = parent==mnpos? 0: data_->area[parent].back().element[2];
+        double ixa_0 = parent==mnpos? 0: data_->ixa[parent].back().element[2];
 
         for (auto i: util::count_along(segments)) {
             auto prox = segments[i].prox;
--- a/arbor/util/piecewise.hpp
+++ b/arbor/util/piecewise.hpp
@@ -72,11 +72,20 @@ struct pw_elements {
     using size_type = pw_size_type;
     static constexpr size_type npos = pw_npos;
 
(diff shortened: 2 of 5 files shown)


arbor/morph/embed_pwlin.cpp
```diff
@@ -93,9 +93,9 @@ mcable_list data_cmp(const branch_pw_ratpoly<1, 0>& f, unsigned bid, double val,
     mcable_list L;
     const auto& pw = f.at(bid);
     for (const auto& piece: pw) {
-        auto extents = piece.first;
-        auto left_val = piece.second(0);
-        auto right_val = piece.second(1);
+        auto extents = piece.interval;
+        auto left_val = piece.element(0);
+        auto right_val = piece.element(1);
```
thorstenhater, line 98: Here the call operator is used....

arbor/morph/embed_pwlin.cpp
```diff
@@ -275,11 +275,11 @@ embed_pwlin::embed_pwlin(const arb::morphology& m) {
             segment_cables_[seg.id] = mcable{bid, pos0, pos1};
         }
 
-        double length_0 = parent==mnpos? 0: data_->length[parent].back().second[1];
+        double length_0 = parent==mnpos? 0: data_->length[parent].back().element[1];
```
thorstenhater, line 278: ... and here the subscript operator. Any particular reason?

arbor/util/piecewise.hpp
```diff
@@ -305,8 +319,16 @@ template <> struct pw_elements<void> {
         else return partn.index(x);
     }
 
-    using const_iterator = indexed_const_iterator<pw_elements<void>>;
-    using iterator = const_iterator;
+    // Return iterator pair spanning elements whose corresponding closed intervals contain x.
+    std::pair<iterator, iterator> equal_range(double x) const {
```
thorstenhater, line 323: This function could use a comment or two; it's not immediately obvious why shifting the iterators is needed and correct the way it is.

arbor/util/piecewise.hpp
```diff
@@ -382,125 +404,113 @@ namespace impl {
     struct piecewise_pairify {
         std::pair<A, B> operator()(
             double left, double right,
```
thorstenhater, line 406: Arguments `left` and `right` are not used?

thorstenhater (changes requested):
A few questions/comments. I also would like to see a consistency check upon construction.

arbor/morph/embed_pwlin.cpp
halfflat: And here, it is accessing the interpolant value directly (for length, there are two elements, and `[1]` refers to the value at the end). The difference here is that we want the exact value used for interpolation without relying upon any further calculations, to avoid any potential rounding issues.

arbor/util/piecewise.hpp
halfflat: They're named just for clarity, but this could indeed just be `(double, double, const pw_element<A>& a_elem, const pw_element<B>& b_elem)`.

arbor/morph/embed_pwlin.cpp
halfflat: Here it is evaluating the polynomial at the ends…

halfflat: (But, this could also in principle use the stored interpolant values directly.)

thorstenhater: Ok

arbor/morph/embed_pwlin.cpp
thorstenhater: 👍

arbor/util/piecewise.hpp
thorstenhater: Maybe `[[maybe_unused]]` is helpful to indicate that fact.

thorstenhater (approved):
All good

Issue #1263, opened 2020-12-01
Probes for LIF cells.


halfflat:
LIF cells currently are unprobeable. Apart from the functionality itself, probe implementations on more than one cell kind will help ensure that the probe APIs remains general.

Proposed probe types:
* Membrane voltage.
* (Maybe) exponential moving average inter-spike delay.

thorstenhater:
This issue requires a bit of thought:
LIF cells are integrated by analytical solution between steps, so probing in between requires interpolating 
into said solution.

PR #1517, opened 2021-05-04
Add leaky-integrate and fire mechanism

schmitts:
A leaky-integrate and fire mechanism is added. Compared to the builtin LIF cells, it allows to attach e.g. point processes. Also cf. #1263.

### DIFF
--- a/mechanisms/CMakeLists.txt
+++ b/mechanisms/CMakeLists.txt
@@ -23,7 +23,7 @@ make_catalogue(
   NAME default
   SOURCES "${CMAKE_CURRENT_SOURCE_DIR}/default"
   OUTPUT "CAT_DEFAULT_SOURCES"
-  MECHS exp2syn expsyn expsyn_stdp hh kamt kdrmt nax nernst pas
+  MECHS exp2syn expsyn expsyn_stdp hh kamt kdrmt nax nernst pas lif
   ARBOR "${PROJECT_SOURCE_DIR}"
   STANDALONE FALSE
   VERBOSE ${ARB_CAT_VERBOSE})
--- a/mechanisms/default/lif.mod
+++ b/mechanisms/default/lif.mod
@@ -1,0 +1,62 @@
+: Leaky-integrate and fire mechanism
+: When crossing the spike threshold, the membrane is pulled
+: towards the reset potential with a (possibly) high conductance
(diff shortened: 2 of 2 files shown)


brenthuisman:
I'm going to paraphrase @halfflat's message in the channel a bit. Btw, let's try to keep such discussion close to the source, it's useful for anyone not reading the channel to know how the PR evolved. 

> This LIF mechanism relies upon NMODL semantics that we'd prefer not to support, in that it requires an update to the mechanism state at the point where the current is computed.
> 
> On the other hand, we're not going to have an alternate scheme for supporting this until arblang is ready, and in particular, arblang with support for state machines governing dynamic regimes. If it's packaged with the library, it feels like we're blessing something that we want ultimately to deprecate.

@schmitts You wrote that you want to test synapse models with an easy neuron model. That sounds like something we should support. Can we identify something that is missing from Arbor's LIF cell?

@bcumming @halfflat Would a catalogue with testing-mechs be an option? Perhaps living as a separate repo, once the mech ABI is merged? I think we need to enable Arbor users here.

bcumming:
In the slack channel I pointed out that its inclusion could cause confusion. It is not a typical density channel, which are used to model current over a region of the membrane, either leaky or due to ion exchange. These can be distributed over the cell surface. On the other hand, this model uses this interface to approximate a simpler LIF model, and can't be composed like other density channels. So, while it is a clever way to simulate a LIF cell, its presence in the library would cause confusion to users familiar with multi-compartment modelling.

The motivation behind its development was to use it on a morphology modeled as a single compartment, to replicate a LIF cell to which multiple arbitrary synapses. In this sense it is a workaround to a feature not supported by Arbor, so we could have a discussion about whether to add that feature.

brenthuisman:
Hi @schmitts, I'll close this PR now because although it's not released yet (but soon!), the mechanism ABI should allow for precisely these kinds of features to live somewhere where we're not required to support them. Currently we're thinking of a mechanism repo in `arbor-contrib`, and a bit of advertisement on the website on all the things we (you :)) have placed in there so far. Should it become clear that multiple users depend on particular items in `arbor-contrib`, then of course we will reconsider.

How does that sound?

schmitts:
Fine with me.

PR #1524, opened 2021-05-11
Modcc: Invalid pointer bug

noraabiakar:
After reviewing the failures in PR #1521, I found and fixed a bug in the modcc function inliner: the `rhs` pointer is potentially changed then the old value is reused.


noraabiakar:
Fix is already in PR #1521

PR #1538, opened 2021-05-19
Update URLs to arbor-sim.org

brenthuisman:
Also update our test-badges to our actual workflow in basic.yml.


bcumming (approved):
Looks good!

PR #1542, opened 2021-05-21
Delete symlinks from repository for Windows devs

Helveg:
It's right next to its target `profstats.py` as well. Symlinks get interpreted as directories on Windows, continually get flagged as "unstaged changes", cause merge conflicts and freeze up git tools on Windows. The unstaged changes also can't be discarded by any tools, only the CLI can nuke them with `git reset --hard` which can lead to loss of local changes for Windows devs.


brenthuisman (changes requested):
This symlink (which is the only one in the codebase) is superfluous anyways, so fine by me. Perhaps it's nice remove extensions from all excutable files in `scripts` (and mark `compare_spike_trains.py` executable), this is a convention that we now seems to be applying inconsistently.

Helveg:
What changes exactly do you request?

schmitts:
@Helveg Is this misbehaviour a known shortcoming of WSL? I've tried to find a bug report/issue online but haven't found anything.

Helveg:
I can find _some_ examples at least: https://stackoverflow.com/questions/64489501/wsl2-linux-relative-symlinks-broken-when-accessed-from-windows-only-for-the

brenthuisman:
@Helveg There is [an old convention](https://google.github.io/styleguide/shellguide.html#s2.1-file-extensions) that says that any script marked executable has its extension chopped off. In the `scripts` dir this is not done consistently, so I proposed that you take this cleanup as an excuse to chop them all off (and check they are marked executable). Does this help make it clear?

halfflat:
I'm very much in favour of removing the extensions from executable programs used as such (e.g. from the shell, or from `exec`) — as the convention says, the implementation language should be irrelevant.

brenthuisman:
@Helveg I think I pushed the permissions correctly. Github UI does not seem to show this though.

Let me know if you're satisfied I didn't mess your PR up and I'll merge.

Helveg:
Wholly satisfied, go ahead

brenthuisman (approved):

PR #1544, opened 2021-05-25
Add CI for the in-repo Spack package

### LINKED ISSUE #1534: Add CI for Arbor's Spack package
Arbor's Spack package https://github.com/arbor-sim/arbor/blob/master/spack/package.py needs CI to ensure that it is ready to be PRed to upstream Spack with every release of Arbor. A github action that pulls Spack develop branch, updates the Arbor package and performs a build (`spack dev-build`) is probably sufficient.

schmitts:
Adds script and workflow to build the in-repo Spack package (closes #1534)

### DIFF
--- a/.github/workflows/spack.yml
+++ b/.github/workflows/spack.yml
@@ -9,14 +9,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-latest, ubuntu-latest]
+        os: [ubuntu-latest]
         python-version: [3.8]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
         with:
             path: arbor
-      - name: Build Arbor's Spack package against the latest release
-        run: arbor/scripts/build_spack_package.sh arbor latest_release
       - name: Build Arbor's Spack package against the develop branch
         run: arbor/scripts/build_spack_package.sh arbor develop
--- a/.github/workflows/spack.yml
+++ b/.github/workflows/spack.yml
@@ -1,0 +1,22 @@
+name: Spack
+
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest, ubuntu-latest]
+        python-version: [3.8]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+            path: arbor
+      - name: Build Arbor's Spack package against the latest release
+        run: arbor/scripts/build_spack_package.sh arbor latest_release
+      - name: Build Arbor's Spack package against the develop branch
+        run: arbor/scripts/build_spack_package.sh arbor develop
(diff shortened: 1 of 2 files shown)


bcumming:
Can you please add a description of the changes in the PR, and link to any issues that it address/fixes in the commit message.

And you might need to apply the spack changes to a clean branch from master, because I don't think that you meant to touch 120 files!

brenthuisman (commented):
The job runs on commits to master (which is only done for a release). `build_spack_package` however get the latest release itself (I think we might be pushing to master before the release is made, in which case this script will be testing the old release) or the `develop` branch (which we do not use at the moment).

Would it be simpler to test the creation of the package against the current commit (which should be at `.`)? It can still fire upon commits to master, which indeed should be a release.

Also, do we want to install the here created Spack package and run tests on it?

schmitts:
@brenthuisman The script checks against the `develop` branch of Spack or a release of `Spack`. It always tests against commits to `master` of Arbor.

brenthuisman:
Ah, I see now, I missed you're checking out both repos. Still, I see no need to checkout Arbor twice (first with the Checkout Action, then later manually). The action is enough, no?

schmitts:
> Ah, I see now, I missed you're checking out both repos. Still, I see no need to checkout Arbor twice (first with the Checkout Action, then later manually). The action is enough, no?

@brenthuisman You are right. Should fixed be now.

brenthuisman:
Great!

Then my last question for discussion remains: should we run tests on the package or is that overkill?

schmitts:
I would like to get this PR in asap and add tests when #1566 is done. How does that sound to you?

schmitts:
The latest https://github.com/schmitts/arbor/actions/runs/895777892 is good.

brenthuisman:
There is now a conflict, `spack/package.py` was changed in another merged PR. Can you update?

brenthuisman:
Maybe I need to update my prescription, but has `build_spack_package.sh` disappeared? I don't see it in `scripts/` and it is required to run this action, correct?

schmitts:
Should be back. Problem atm is that the build on Mac against the latest Spack release seems to fail. Not sure how to continue. It just may be that the latest Spack release has a problem that got fixed in the develop branch. I would say that we test only against Spack develop.

brenthuisman:
Looking at the last 3 runs, the latest fails at macos + develop branch, not latest release. (The one before that fails on an invalid hostname: `http://`, which I don't quite understand, and the one before that on account of the script being missing.) Right? Do you have a mac to test? If not, I'll ask if someone to do so.

schmitts:
You're right. It fails against develop. No, I don't have a Mac at hand.

bcumming:
Should we restrict ourselves to testing Spack on Linux for now?

Particularly given that the spack builds take so long, due to spack building all of the dependencies.

schmitts:
> Should we restrict ourselves to testing Spack on Linux for now?
> 
> Particularly given that the spack builds take so long, due to spack building all of the dependencies.

Testing on both Linux and Spack/develop only would be fine with me.

brenthuisman (approved):
lgtm!

PR #1556, opened 2021-05-26
Correct label parse error message in complex expressions

bcumming:
The C++ label parse code was returning the error type in the value part of a `parse_hopefully<std::any>`,
and not in the error slot.


noraabiakar (approved):
Looks good!

PR #1558, opened 2021-05-27
WIP: include scripts in documentation instead of copy'n'paste

schmitts:

### DIFF
--- a/doc/tutorial/single_cell_recipe.rst
+++ b/doc/tutorial/single_cell_recipe.rst
@@ -21,25 +21,9 @@ The cell
 We can immediately paste the cell description code from the
 :ref:`previous example <tutorialsinglecell-cell>` where it is explained in detail.
 
-.. code-block:: python
-
-    import arbor
-
-    # (1) Create a morphology with a single (cylindrical) segment of length=diameter=6 μm
-    tree = arbor.segment_tree()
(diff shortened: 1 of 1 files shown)


bcumming:
I like it.
This means that we can test the code examples used in the documentation during CI, so that we don't have stale / out of date docs.
Does this require a specific version of Sphinx, or other dependencies?

schmitts:
>  Does this require a specific version of Sphinx, or other dependencies?

I see literalinclude mentioned at least back to 2010. I don't think it needs anything special.

schmitts:
What's the opinion on smaller code blocks, e.g., https://docs.arbor-sim.org/en/latest/python/profiler.html - these could also be moved to files and included. I'm in favour of doing so to have them CIable as @bcumming pointed out.

brenthuisman:
I'm a bit worried about line numbers going out of date. That's harder to spot than the code itself, at least for me. What do you think?

schmitts:
> I'm a bit worried about line numbers going out of date. That's harder to spot than the code itself, at least for me. What do you think?

I agree. Maybe one should add a hint in the script like "This script is included in documentation. Adapt line numbers if touched.".

brenthuisman:
I'm incorporating this into https://github.com/arbor-sim/arbor/pull/1575, because there I already needed to go through the tutorials.

PR #1562, opened 2021-05-28
Add example of a single dendrite to show influence of cv policy

schmitts:


python/example/single_cell_cable.py
```diff
@@ -0,0 +1,213 @@
...
+
+class Cable(arbor.recipe):
+    def __init__(self, probes,
+                 Vm, length, radius, cm, rL, g,
+                 stimulus_start, stimulus_duration, stimulus_amplitude,
+                 cv_policy_max_extent):
+        """
+        probes -- list of probes
+
+        Vm -- membrane leak potential
```
noraabiakar, line 19: Can we use `μm`, `Ω.cm` and `S/cm^2` here?

python/example/single_cell_cable.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def num_sources(self, gid):
+        assert gid == 0
+        return 0
+
+    def cell_kind(self, gid):
+        assert gid == 0
+        return arbor.cell_kind.cable
+
+    def cell_description(self, gid):
```
noraabiakar, line 63: I think this method could benefit from some comments, because it'll show up in the tutorial.

python/example/single_cell_cable.py
```diff
@@ -0,0 +1,213 @@
...
+
+    plot.figure.savefig("single_cell_cable_result.svg")
+
+    # calculcate the idealized conduction velocity, cf. cable equation
+    rm = get_rm(args.g*1/(0.01*0.01))
+    taum = get_taum(args.cm, rm)
+    lambdam = get_lambdam(args.radius*1e-6, rm, args.rL*0.01)
+    vcond_ideal = get_vcond(lambdam, taum)
+
+    # take the last at second probe
```
noraabiakar, line 205: The last time?

python/example/single_cell_cable.py
```diff
@@ -0,0 +1,213 @@
...
+                        help="start of stimulus in ms", type=float, default=10)
+    parser.add_argument('--stimulus_duration',
+                        help="duration of stimulus in ms", type=float, default=0.1)
+    parser.add_argument('--stimulus_amplitude',
+                        help="amplitude of stimulus in nA", type=float, default=1)
+
+    parser.add_argument('--cv_policy_max_extent',
+                        help="maximum extent of control volume in micrometer", type=float, default=10)
+
+    args = parser.parse_args()
```
noraabiakar, line 167: A few comments here would to explain what is going on here would also be really helpful for the tutorial.

noraabiakar (changes requested):
Looks good, thanks! And sorry for the late review. I have suggested some additions to add some clarity to the tutorial page.

noraabiakar (approved):

PR #1564, opened 2021-05-31
Clean-up and extend mech cat iterator interface.

### LINKED ISSUE #1547: `catalogue` iteration and the 'MechCatIterator`
While using the catalogues there are some UX issues due to the limited available information on the `catalogue` and `mechanism_info` objects:

The way to iterate over a catalogue feels incomplete. It is iterable, which is good (so `iter(cat)` works), but as a dictionary style object it is strange that the `values`, `keys` and `items` iterators are unavailable. The obtained `MechCatIterator` is also a very strange object AND attribute? Why is it available as an attribute of a `catalogue`? As an attribute it is completely unusable from Python:

```python
>>> cat.MechCatIterator
<class 'arbor._arbor.catalogue.MechCatIterator'>
>>> list(cat.MechCatIterator)
TypeError: 'pybind11_type' object is not iterable
>>> list(iter(cat.MechCatIterator))
TypeError: 'pybind11_type' object is not iterable
>>> next(cat.MechCatIterator)
TypeError: 'pybind11_type' object is not an iterator
>>> cat.MechCatIterator.__iter__
<instancemethod __iter__ at 0x7f9b6e45a160>
>>> cat.MechCatIterator.__iter__()
TypeError: __iter__(): incompatible function arguments. The following argument types are supported:
    1. (self: arbor._arbor.catalogue.MechCatIterator) -> arbor._arbor.catalogue.MechCatIterator

Invoked with:
>>> iter(cat.MechCatIterator)
TypeError: 'pybind11_type' object is not iterable
>>> iter(cat.MechCatIterator())
TypeError: arbor._arbor.MechCatIterator: No constructor defined!
```

It would be nicer if this `MechCatIterator` was not available in the `__dict__` of a `catalogue` or was at least prefixed with an underscore if absolutely required.

thorstenhater:
Closes #1547.

- Move `MechCat*Iter` to module instead of catalogue class
- Add `keys`, `values`, and `items`
- Fix a typo.

### DIFF
--- a/python/mechanism.cpp
+++ b/python/mechanism.cpp
@@ -95,7 +95,7 @@ void register_mechanisms(pybind11::module& m) {
         .def_readonly("ions", &arb::mechanism_info::ions,
             "Ion dependencies.")
         .def_readonly("linear", &arb::mechanism_info::linear,
-            "True if a synapse mechanism has linear current contributions so that multiple instances on the same compartment can be coalesed.")
+            "True if a synapse mechanism has linear current contributions so that multiple instances on the same compartment can be coalesced.")
         .def("__repr__",
                 [](const arb::mechanism_info& inf) {
                     return util::pprintf("(arbor.mechanism_info)"); })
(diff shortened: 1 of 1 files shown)


bcumming (approved):
A belated lgtm!

PR #1569, opened 2021-06-01
Remove under construction disclaimer

schmitts:


schmitts:
Either the information is already complete enough to justify having it on a separate page or not. If the info is good enough, there's no need for a disclaimer. If the info is not good enough I would hide the page completely. But that's probably a matter of taste. I find "under construction" disclaimers quite off-putting. More so as they tend to stay forever.

brenthuisman:
I suppose I am more OK with showing where the gaps are.

How about I add in a few lines to this PR? Instead of removing the warnings and adding nothing, I'd like to add in a little something, and I should :) I can also fold that text into https://github.com/arbor-sim/arbor/pull/1575 if you prefer.

schmitts:
Good!

schmitts:
Will be solved within #1575.

PR #1573, opened 2021-06-02
Improve mechanism painting documentation

schmitts:


doc/python/mechanisms.rst
```diff
@@ -48,6 +48,34 @@ mechanism that is to be painted or placed on the cable cell.
         # by comma-separated list of parameter after the base mechanism name.
         rev = arbor.mechanism('nernst/F=96485,x=ca')
 
+    Mechanisms can then be painted to a region. Different mechanism can be painted on top of each other.
```
brenthuisman, line 51: Second "mechanism" should be plural. To which condition is 'then' referring to? Perhaps nothing in particular, in which case the word can be left out.

brenthuisman (changes requested):

brenthuisman (approved):

PR #1595, opened 2021-06-23
Python unit test clean-up

noraabiakar:
* Remove `test_simulator.py`: was originally removed in #1503 then accidentally merged again in #1504.
* Fix `test_decor.py` and add it to the to the unit test runner.
* Run the python unit tests in verbose mode in order to display which tests were skipped and the reason.


python/test/unit_distributed/runner.py
```diff
@@ -21,6 +21,7 @@
     import test_contexts_arbmpi
     import test_contexts_mpi4py
     import test_domain_decompositions
+    import simulator
```
halfflat, line 24: I don't see anything that matches this module name; should this line be here?

noraabiakar: That's supposed to be `test_simulator`

noraabiakar: And it's for completeness. AFAICT, the `try` clause will always fail unless there's an `options.py` file in the same directory as `test_*` files and the `runner.py` file, which isn't the case at the moment.

halfflat (approved):

PR #1596, opened 2021-06-24
Bump to v0.5.3-dev, bump spack file to v0.5.2

brenthuisman:


spack/package.py
```diff
@@ -12,11 +12,11 @@ class Arbor(CMakePackage, CudaPackage):
 
     homepage = 'https://arbor-sim.org'
     git      = 'https://github.com/arbor-sim/arbor.git'
-    url      = 'https://github.com/arbor-sim/arbor/releases/download/v0.5/arbor-v0.5-full.tar.gz'
+    url      = 'https://github.com/arbor-sim/arbor/releases/download/v0.5.2/arbor-v0.5.2-full.tar.gz'
     maintainers = ['bcumming', 'brenthuisman', 'haampie', 'halfflat']
 
     version('master', branch='master', submodules=True)
-    version('0.5', 'd0c8a4c7f97565d7c30493c66249be794d1dc424de266fc79cecbbf0e313df59', url='https://github.com/arbor-sim/arbor/releases/download/v0.5/arbor-v0.5-full.tar.gz')
+    version('0.5.2', '290e2ad8ca8050db1791cabb6b431e7c0409c305af31b559e397e26b300a115d', url='https://github.com/arbor-sim/arbor/releases/download/v0.5.2/arbor-v0.5.2-full.tar.gz')
```
noraabiakar, line 19: I think we should add version 0.5.2 on a seperate line, not replace the previous version. 
Also we should update this file too: https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/arbor/package.py

brenthuisman: Once we agree on and merge `package.py` here, I'll make a PR with the updates at the Spack repo.

Re: versions: good point. Two follow-up questions:

1. Do we deprecate older versions? At least until we reach v1.0?
2. Having multiple versions means we might have different dependencies. Reading the [Spack packaging docs](https://spack.readthedocs.io/en/latest/packaging_guide.html#dependencies), you can add a version to the `when` key passed to `depends_on()`, but it's not clear to me if that refer to the version of the dep, or of this package. Does anyone know?

noraabiakar: I think we shouldn't remove versions from the spack file unless we have a good reason. But let's bring this up on slack. As for the dependencies, why do you ask? We don't have any new dependencies.

brenthuisman: Removing them is not the same as deprecating them (you pass in `deprecated=True` to `version()`.

Deps: not now, correct. Was for future reference.

bcumming (approved):
lgtm

PR #1559, opened 2021-05-28
Revert "Add default hidden visibliity to pyarb shared lib. (#1510)"

espenhgn:
This reverts commit e8b08ca8d9b36a5c8e89ab60fd8876d0453fecbd. 

Closes #1557.

PR #1599, opened 2021-06-25
Compile all arbor source files with `-fvisibility=hidden`

### LINKED ISSUE #1557: `RuntimeError: Error parsing the label: ` in python functions and example files
**Describe the bug**
Python examples are using seemingly deprecated location label syntax, for instance:
```
% cd <arbor>/python/example
% python single_cell_detailed_recipe.py single_cell_detailed.swc
  File "/Users/espenhagen/Repositories/arbor/python/example/single_cell_detailed_recipe.py", line 29, in <module>
    labels['soma'] = '(tag 1)'
RuntimeError: 
----- internal error -------------------------------------------
Error parsing the label: 'soma' = '(tag 1)'

bad any cast

Please file a bug report with this full error message at:
    github.com/arbor-sim/arbor/issues
----------------------------------------------------------------
```

Several such cases may also be present as defaults in arbor itself, for example: 
```
>>> help(arbor.cv_policy_single)

Help on built-in function cv_policy_single in module arbor._arbor:

cv_policy_single(...) method of builtins.PyCapsule instance
    cv_policy_single(domain: str = '(all)') -> arbor._arbor.cv_policy
```
hence
```
>>> arbor.cv_policy_single()
RuntimeError: Invalid region description: '(all)' is neither a valid region expression or region label string.
```

**Additional context**
conda env/installation:
```
conda create -n arbor2 python=3.9 pip numpy pandas seaborn cmake c-compiler cxx-compiler nlohmann_json sphinx svgwrite -c conda-forge -y
conda activate arbor2
cd <arbor>
git checkout a304d5ac7
pip install -v .
... 
% python -c "import arbor; print(arbor.__version__, arbor.__file__)"
0.5.1-dev /Users/espenhagen/miniforge3/envs/arbor2/lib/python3.9/site-packages/arbor/__init__.py # seems correct
```

noraabiakar:
The visibility of symbols in the python shared library is set to 'hidden'. However, checking the dynamic symbol table of the generated `.so` file reveals that all the symbols of the static arbor libraries are still visible. This causes some issues on the Mac M1. To resolve, all source files should be compiled with `fvisibility=hidden`.

Fixes #1557
Fixes #1559

### DIFF
--- a/mechanisms/generate_catalogue
+++ b/mechanisms/generate_catalogue
@@ -120,7 +120,7 @@ const mechanism_catalogue& global_${catalogue}_catalogue() {
 
 #ifdef STANDALONE
 extern "C" {
-    __attribute__ ((visibility ("default"))) const void* get_catalogue() {
+    [[gnu::visibility("default")]] const void* get_catalogue() {
         static auto cat = arb::build_${catalogue}_catalogue();
         return (void*)&cat;
     }
--- a/mechanisms/generate_catalogue
+++ b/mechanisms/generate_catalogue
@@ -120,7 +120,7 @@ const mechanism_catalogue& global_${catalogue}_catalogue() {
 
 #ifdef STANDALONE
 extern "C" {
-    const void* get_catalogue() {
+    __attribute__ ((visibility ("default"))) const void* get_catalogue() {
         static auto cat = arb::build_${catalogue}_catalogue();
         return (void*)&cat;
     }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -255,6 +255,7 @@ separate_arguments(ARB_CXX_FLAGS_TARGET_FULL)
 
 target_compile_options(arbor-private-deps INTERFACE ${ARB_CXX_FLAGS_TARGET_FULL})
 target_compile_options(arborenv-private-deps INTERFACE ${ARB_CXX_FLAGS_TARGET_FULL})
+target_compile_options(arborio-private-deps INTERFACE ${ARB_CXX_FLAGS_TARGET_FULL})
 
 # Profiling and test features
 #-----------------------------
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -246,6 +246,11 @@ else()
     set_arch_target(ARB_CXXOPT_ARCH ${ARB_ARCH})
     set(ARB_CXX_FLAGS_TARGET_FULL ${ARB_CXX_FLAGS_TARGET} ${ARB_CXXOPT_ARCH})
 endif()
+
+# Compile with `-fvisibility=hidden` to ensure that the symbols of the generated
+# arbor static libraries are hidden from the dynamic symbol tables of any shared
+# libraries that link against them.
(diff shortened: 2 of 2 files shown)


noraabiakar:
bors try

noraabiakar:
bors try

noraabiakar:
bors try

noraabiakar:
bors try

kanzl:
Works for me on M1 Macbook

mechanisms/generate_catalogue
```diff
@@ -120,7 +120,7 @@ const mechanism_catalogue& global_${catalogue}_catalogue() {
 
 #ifdef STANDALONE
 extern "C" {
-    const void* get_catalogue() {
+    __attribute__ ((visibility ("default"))) const void* get_catalogue() {
```
thorstenhater, line 123: Since C++17 we are allowed to use `[[gnu::visibility("default")]]` as part of the standard. The `__attribute__` is technically an extension by GCC.

thorstenhater (changes requested):
Looks good, also works on my non-M1 MacBook. A tiny standards nitpick and we are good to merge.

mechanisms/generate_catalogue
noraabiakar: But that would only work for GCC. According to this: https://gcc.gnu.org/wiki/Visibility, we should add this entire block:
```
#if defined _WIN32 || defined __CYGWIN__
  #ifdef BUILDING_DLL
    #ifdef __GNUC__
      #define DLL_PUBLIC __attribute__ ((dllexport))
    #else
      #define DLL_PUBLIC __declspec(dllexport) // Note: actually gcc seems to also supports this syntax.
    #endif
  #else
    #ifdef __GNUC__
      #define DLL_PUBLIC __attribute__ ((dllimport))
    #else
      #define DLL_PUBLIC __declspec(dllimport) // Note: actually gcc seems to also supports this syntax.
    #endif
  #endif
  #define DLL_LOCAL
#else
  #if __GNUC__ >= 4
    #define DLL_PUBLIC __attribute__ ((visibility ("default")))
    #define DLL_LOCAL  __attribute__ ((visibility ("hidden")))
  #else
    #define DLL_PUBLIC
    #define DLL_LOCAL
  #endif
#endif
```
I think we need that top block for WSL, but I'm not sure and I don't have a system to test it on.

thorstenhater: That page predates the C++17 attribute (last change 2016). We also can savely assume GCC > 4 (else C++14 would be tricky). And finally, we do not care about Windows right now. In endeffect it comes down to what you have in the PR.

Regarding your comment, this is working in clang and gcc. To quote my last comment `__attribute__` is a GCC extension in itself. However, Clang shares most of GCC's extensions. Reference: https://gcc.godbolt.org/z/Edev9PGnK

thorstenhater (approved):

espenhgn:
Hi. As I raised #1557 I tested this PR out locally, and it appears to have sorted out the problem(s) I encountered there. Thanks!

PR #1602, opened 2021-06-29
mpi.py updated for new synapse labelling

brenthuisman:


noraabiakar (changes requested):
Looks good! I missed this file because I didn't know it was an example. Can we rename it to something other than `mpi.py`? Maybe `distributed_network_ring.py`? Also, is it possible to add it to the CI?

brenthuisman:
You are right. Since I like to be able to end up with meaningful grouping when sorting directories alphabetically (alphanumerically?), I propose `network_ring_mpi.py`.

noraabiakar (approved):

PR #1609, opened 2021-07-29
Update spack package to include fmt for arbor@0.5.3:

bcumming:
Update spack package to specify fmt as a dependency, as required by the new mechanism ABI.


halfflat (approved):
Looks fine!

PR #1610, opened 2021-07-29
Docs mechabi

brenthuisman:
* Bit of review on the copy of the mech abi docs.
* Convert api/abi docs to C Domain directives.
* 'the library' -> Arbor

Questions:
* Why do the comments on the fields of `arb_ion_info` end in questions marks?
* Mechanism Plugin: I removed the word `plugin` because it was not consistently used, but it maybe useful to describe it as such. Should we consistently use that term to separate user provided code from Arbor (formerly known as 'the library')?

### DIFF
--- a/doc/internals/mechanism_abi.rst
+++ b/doc/internals/mechanism_abi.rst
@@ -3,15 +3,20 @@
 Mechanism ABI
 =============
 
-Here you will find the information needed to connect Arbor to mechanism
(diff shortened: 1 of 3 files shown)


thorstenhater (approved):
lgtm

thorstenhater:
The question marks are indicating that these are binary flags.

thorstenhater:
I cannot comment on the CSS changes.

PR #1611, opened 2021-07-29
Add dependency on fmt for Spack build

schmitts:


bcumming:
Another PR was already merged with this change, and #1612  fixes an issue with the original, so that the `fmt` dependency is only used if building `v0.5.3` or later

PR #1615, opened 2021-07-31
Bug/assorted static analysis

thorstenhater:
Fixes a series of smaller issues found by PVS studio.

### DIFF
--- a/arbor/partition_load_balance.cpp
+++ b/arbor/partition_load_balance.cpp
@@ -188,7 +188,7 @@ domain_decomposition partition_load_balance(
                 group_elements.push_back(cell.id);
             } else {
                 if (group_elements.size() + super_cells[cell.id].size() > group_size && !group_elements.empty()) {
-                    groups.push_back({k, group_elements, backend});
+                    groups.push_back({k, std::move(group_elements), backend});
                     group_elements.clear();
                 }
                 for (auto gid: super_cells[cell.id]) {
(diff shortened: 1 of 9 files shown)


arborio/cableio.cpp
```diff
@@ -584,7 +584,7 @@ eval_map named_evals{
     {"gap-junction-site", make_call<>(make_gap_junction_site,
                               "'gap-junction-site' with 0 arguments")},
     {"ion-reversal-potential-method", make_call<std::string, arb::mechanism_desc>(make_ion_reversal_potential_method,
-                                          "'ion-reversal-potential-method' with 2 ""arguments (ion:string mech:mechanism)")},
+                                                                                  "'ion-reversal-potential-method' with 2 arguments (ion:string mech:mechanism)")},
```
halfflat, line 587: Do we need this to be so indented?

arbor/partition_load_balance.cpp
```diff
@@ -188,7 +188,7 @@ domain_decomposition partition_load_balance(
                 group_elements.push_back(cell.id);
             } else {
                 if (group_elements.size() + super_cells[cell.id].size() > group_size && !group_elements.empty()) {
-                    groups.push_back({k, std::move(group_elements), backend});
+                    groups.push_back({k, group_elements, backend});
```
halfflat, line 191: This one is a false positive from static analysis: we clear `group_elements` after move, so that it is in a valid and known (empty) state.

halfflat (changes requested):
Good catches all, save for the `std::move` one, which is a false positive.

arbor/partition_load_balance.cpp
thorstenhater: Correct, for STL containers and `clear` this is ok. I'll revert this.

halfflat (approved):

PR #1619, opened 2021-08-03
Cache Spack for GHA

schmitts:
- make spack ci run use arbor version information
- Cache Spack for GHA

### DIFF
--- a/scripts/build_spack_package.sh
+++ b/scripts/build_spack_package.sh
@@ -54,4 +54,5 @@ spack repo add $SPACK_CUSTOM_REPO
 
 cp $ARBOR_DIR/spack/package.py $SPACK_CUSTOM_REPO/packages/arbor
 cd $ARBOR_DIR
-spack dev-build arbor@with-package-from-repo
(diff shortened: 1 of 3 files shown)


scripts/build_spack_package.sh
```diff
@@ -46,12 +46,20 @@ case $SPACK_VERSION in
 		echo "SPACK_VERSION" must be \"latest_release\" or \"develop\"
 		exit 1
 esac
+
+mkdir ~/.spack
```
bcumming, line 50: Should this be the path `~/.spack-ci` referenced in the other files?

schmitts: No, `~/.spack` is the transient configuration. `~/.spack-ci` is restored from the cache. We should probably rename it to `~/.spack-cache`.

bcumming: Thanks. `spack-cache` is a better name.

bcumming (approved):
Looks good, thanks!

PR #1622, opened 2021-08-11
Better Python installation instructions

### LINKED ISSUE #1405: Python installation through setup.py improvements
* [x] NeuroML should be enabled by default in `setup.py`, as well as any other file formats that may be supported.
  * auditwheel correctly bundles libxml when Arbor is compiled with NeuroML=ON
* ~[ ] modcc should be build and installed. build-catalog eventually as well.~
  * ~https://stackoverflow.com/questions/4840182/setup-py-and-adding-file-to-bin~
  * Outdated
* [x] Should `setup.py` default to ARB_VECTORIZE=ON?
  * This requires AVX
  * After discussion: No.
* [ ] CMake option to install the Python package as user package, and use `python3 -m site --user-site` in such case, overriding any supplied prefix. What we do now [only works by coincidence on the major distros](https://docs.python.org/3/install/index.html#alternate-installation), not by design: there is no such thing as a user-platform-package-dir.

brenthuisman:
Python packaging...

## Cause

`sysconfig.get_path('platlib',vars={'base':pfx,'platbase':pfx})` is not guaranteed to produce a correct path for non-system installations (the global site-packages directory as Python calls it). For user installations, the user site-packages directory is used, which can be obtained with `python -m site --user-site`. Confusingly, when in a Python or Conda virtual environment, Python is not guaranteed to have the `site` package (although it did on my system during the typing of this comment) or is not guaranteed to respect the venv boundary (I can confirm). Moreover, `platlib` directory obtained with `sysconfig` only has meaning for the global site-packages directory. Although for Ubuntu, Fedora, CentOS and Arch this coincides with how they have apparently patched Python, this is again not a guarantee. OpenSuse indeed have no platlib for local installs, so Arbor is copied to a place (`~/.local/lib64`) that system Python does not use (which is `~/.local/lib`). `pip` does get the right path in all the above circumstances (global, user, venv site-package dirs), and I am too tired of chasing this issue that I am not going to bother finding out how it determines that path. It somehow knows which item to choose from the list of global site-package dirs you obtain with `site.getsitepackages()`.

Our `sysconfig.get_path('platlib',vars={'base':pfx,'platbase':pfx})` only produced the right path on the tested platform by coincidence.

## Solution

I've modified to docs to always use `pip` to install the Arbor Python package, because that is the best or only way to get the most Pythonic installation experience. Only on the CMake page have I left a hint of what to feed ARB_PYTHON_LIB_PATH, and that by using CMake, you accept unpythonicness.

## Sources:

1. https://docs.python.org/3/install/index.html#alternate-installation
2. https://stackoverflow.com/questions/122327/how-do-i-find-the-location-of-my-python-site-packages-directory
3. https://old.reddit.com/r/openSUSE/comments/p12l3t/getting_the_sitepackages_dir_on_tumbleweed/h8azyho/
4. https://bugs.python.org/issue21197

Closes #1405

### DIFF
--- a/doc/install/build_install.rst
+++ b/doc/install/build_install.rst
@@ -456,7 +456,7 @@ use ``ARB_PYTHON_LIB_PATH`` to specify the location where the Python module is t
     cmake .. -DARB_WITH_PYTHON=on -DARB_PYTHON_PATH_LIB=/custom/path
 
 .. note::
-    The location of libraries under a prefix in only guaranteed to be standard for Pythons global library location.
+    The location of libraries under a prefix in only guaranteed to be standard for Python's global library location.
     Therefore, correct installation of the Python package to any other location using ``CMAKE_INSTALL_PREFIX``,
     such as user directory (e.g. `~/.local`), a Python or Conda virtual environment, may result in installation to a wrong path.
     
@@ -483,7 +483,8 @@ use ``ARB_PYTHON_LIB_PATH`` to specify the location where the Python module is t
     # Obtain arbor
     git clone --recursive git@github.com:arbor-sim/arbor.git
 
-    # Get the user site-package directory
+    # Manually set the prefix under which the python package will be installed.
+    # In this case, the first directory found by querying Python's list of site-package directories.
     pyprefix=`python3 -c 'import site; print(site.getsitepackages()[0])'`
 
     # Setup CMake
@@ -491,7 +492,7 @@ use ``ARB_PYTHON_LIB_PATH`` to specify the location where the Python module is t
     cd build
     cmake ../arbor -DARB_WITH_PYTHON=on       \       # enable python support.
                    -DARB_USE_BUNDLED_LIBS=on  \       # use bundled versions of deps.
-                   -DCMAKE_INSTALL_PREFIX="$pyprefix" # set custom installation path.
+                   -DARB_PYTHON_LIB_PATH="$pyprefix"  # set Python installation path.
 
     # Build and install
     make -j4
--- a/doc/install/build_install.rst
+++ b/doc/install/build_install.rst
@@ -444,7 +444,6 @@ system with the executable in ``/usr/bin/python3.8``:
 
     cmake .. -DARB_WITH_PYTHON=ON -DPYTHON_EXECUTABLE=/usr/bin/python3.8
 
-
 By default the Python package will be installed in the appropriate sub-directory
 inside ``CMAKE_INSTALL_PREFIX``, determined by querying Python's sysconfig library.
 For example ``${CMAKE_INSTALL_PREFIX}/lib/python3.9/site-packages/``.
@@ -456,11 +455,21 @@ use ``ARB_PYTHON_LIB_PATH`` to specify the location where the Python module is t
 
     cmake .. -DARB_WITH_PYTHON=on -DARB_PYTHON_PATH_LIB=/custom/path
 
-If CMake is run in a `venv` or Conda environment, set ``CMAKE_INSTALL_PREFIX`` to the
-base path of the venv. The example below shows a workflow that creates a virtual
-environment, then installs Arbor inside the environment.
+.. note::
+    The location of libraries under a prefix in only guaranteed to be standard for Pythons global library location.
+    Therefore, correct installation of the Python package to any other location using ``CMAKE_INSTALL_PREFIX``,
+    such as user directory (e.g. `~/.local`), a Python or Conda virtual environment, may result in installation to a wrong path.
+    
+    ``python3 -m site --user-site`` (for user installations) or a path from ``python3 -c 'import site; print(site.getsitepackages())'``
+    (for virtual environment installation) can be used in combination with ``ARB_PYTHON_LIB_PATH``.
+
+    In addition, installation via ``pip`` or ``python setup.py`` is guaranteed to find the right path. Please refer to the
+    :ref:`Python installation instruction <in_python_custom>`.
+
 
 .. code-block:: bash
+
+    # A demonstration using ARB_PYTHON_LIB_PATH
 
     # Set up your venv.
     mkdir myenv
@@ -469,21 +478,15 @@ environment, then installs Arbor inside the environment.
     source env/bin/activate
 
     # Install dependencies
-    pip install numpy
+    pip3 install numpy
 
     # Obtain arbor
     git clone --recursive git@github.com:arbor-sim/arbor.git
 
-    # Determine the prefix path for your installation
-    # method 1: set it explicitly
-    export pyprefix=$(pwd)/env
-    # method 2: query python directly by instpecting the output of sysconfig.
-    # Note that the ending of the form lib/python3.x/site-packages must be dropped.
-    # E.g if the output was /home/xxx/myenv/env/lib/python3.9/site-packages, we
-    # want to set pyprefix=/home/xxx/myenv/env/
-    python -c "import sysconfig; print(sysconfig.get_path('platlib'))"
+    # Get the user site-package directory
+    pyprefix=`python3 -c 'import site; print(site.getsitepackages()[0])'`
 
-    # Configure Arbor
+    # Setup CMake
     mkdir build
     cd build
     cmake ../arbor -DARB_WITH_PYTHON=on       \       # enable python support.
@@ -496,6 +499,7 @@ environment, then installs Arbor inside the environment.
 
     # Test it out!
     python -c "import arbor; print(arbor.__config__)"
+
 
 The Arbor Python wrapper has optional support for mpi4py, though
 it is not required to use Arbor with Python and MPI.
--- a/doc/install/python.rst
+++ b/doc/install/python.rst
@@ -11,7 +11,9 @@ Arbor's Python API will be the most convenient interface for most users.
 Getting Arbor
 -------------
 
-Every point release of Arbor is pushed to the Python Package Index. The easiest way to get Arbor is with
+Every point release of Arbor is pushed to the Python Package Index.
+For x86-64 Linux and MacOS plaftorms, we provide binary wheels.
+The easiest way to get Arbor is with
 `pip <https://packaging.python.org/tutorials/installing-packages>`_:
 
 .. code-block:: bash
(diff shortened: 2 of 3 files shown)


doc/install/build_install.rst
```diff
@@ -456,34 +455,38 @@ use ``ARB_PYTHON_LIB_PATH`` to specify the location where the Python module is t
...
 
-    # Determine the prefix path for your installation
-    # method 1: set it explicitly
-    export pyprefix=$(pwd)/env
-    # method 2: query python directly by instpecting the output of sysconfig.
-    # Note that the ending of the form lib/python3.x/site-packages must be dropped.
-    # E.g if the output was /home/xxx/myenv/env/lib/python3.9/site-packages, we
-    # want to set pyprefix=/home/xxx/myenv/env/
-    python -c "import sysconfig; print(sysconfig.get_path('platlib'))"
+    # Get the user site-package directory
```
bcumming, line 486: How about:
```
# Manually set the prefix under which the python package will be installed.
# In this case, use the user site-package directory found by querying Python.
```

doc/install/build_install.rst
```diff
@@ -456,34 +455,38 @@ use ``ARB_PYTHON_LIB_PATH`` to specify the location where the Python module is t
 
     cmake .. -DARB_WITH_PYTHON=on -DARB_PYTHON_PATH_LIB=/custom/path
 
-If CMake is run in a `venv` or Conda environment, set ``CMAKE_INSTALL_PREFIX`` to the
-base path of the venv. The example below shows a workflow that creates a virtual
-environment, then installs Arbor inside the environment.
+.. note::
+    The location of libraries under a prefix in only guaranteed to be standard for Pythons global library location.
```
bcumming, line 459: `Pythons` -> `Python's`

doc/install/python.rst
```diff
@@ -66,77 +71,74 @@ Every time you make changes to the code, you'll have to repeat the second step.
...
 * ``--gpu``: Enable GPU support for NVIDIA GPUs with nvcc using ``cuda``, or with clang using ``cuda-clang`` (both require cudaruntime).
   Enable GPU support for AMD GPUs with hipcc using ``hip``. By default set to ``none``, which disables gpu support.
-* ``--vec``: Enable vectorization. This might require choosing an appropriate architecture using ``--arch``. Note that on x86-64 platforms compilation will fail if you enable vectorization, but the CPU or ``--arch`` does not support any form of AVX.
-* ``--arch``: CPU micro-architecture to target. By default this is set to ``native``.
-
-If calling ``setup.py`` the flags must come after ``install`` on the command line,
-and if being passed to pip they must be passed via ``--install-option``. The examples
-below demonstrate this for both pip and ``setup.py``.
+* ``--vec``: Enable vectorization. This requires choosing an appropriate architecture using ``--arch``.
+  Note that on x86-64 platforms compilation will fail if you enable vectorization, but the CPU or ``--arch`` does not support any form of AVX.
```
bcumming, line 84: Instead, redirect the readers to existing documentation for the architecture flag:

https://docs.arbor-sim.org/en/latest/install/build_install.html#architecture

doc/install/python.rst
```diff
@@ -66,77 +71,74 @@ Every time you make changes to the code, you'll have to repeat the second step.
...
+    If you had Arbor installed already, you may need to remove it first before you can (re)compile
+    it with the flags you need.
+
 .. Note::
     Detailed instructions on how to install using CMake are in the
     :ref:`Python configuration <install-python>` section of the
     :ref:`installation guide <in_build_install>`.
-    CMake is recommended for developers, integration with package managers such as
-    Spack and EasyBuild, and fine-grained control over compilation and installation.
+    CMake is recommended if you need more control over Arbors compilation, plan to use Arbor with C++,
```
bcumming, line 140: `control over Arbors compilation` -> `control over compilation and/or installation`

bcumming (changes requested):

bcumming (changes requested):
A side comment: has the `import site; print(site.getsitepackages()[0])` command been tested on a range of different OS combinations?

doc/install/build_install.rst
```diff
@@ -456,34 +455,38 @@ use ``ARB_PYTHON_LIB_PATH`` to specify the location where the Python module is t
...
-    # Determine the prefix path for your installation
-    # method 1: set it explicitly
-    export pyprefix=$(pwd)/env
-    # method 2: query python directly by instpecting the output of sysconfig.
-    # Note that the ending of the form lib/python3.x/site-packages must be dropped.
-    # E.g if the output was /home/xxx/myenv/env/lib/python3.9/site-packages, we
-    # want to set pyprefix=/home/xxx/myenv/env/
-    python -c "import sysconfig; print(sysconfig.get_path('platlib'))"
+    # Get the user site-package directory
+    pyprefix=`python3 -c 'import site; print(site.getsitepackages()[0])'`
```
Helveg, line 487: So `import site; print(site.getsitepackages()[0])` is the new query? IIRC there was a problem with `virtualenv`s and the `site` module. Is this case covered? 

You also mention that only `pip` finds the correct location in all cases, but we're not querying pip here?

The 2 commands you mention provide different outputs for me:
```
> python -m site --user-site
/home/robin/.local/lib/python3.9/site-packages
```
```
python -c "import site; print(site.getsitepackages()[0])"
/home/robin/.pyenv/versions/vvv/lib/python3.9/site-packages
```

Helveg (commented):
Slightly confused, I suppose everyone is 😂

doc/install/build_install.rst
brenthuisman: Very confused, which is my I propose that Pythonic installation just uses pip, which should always work out as expected.

Indeed, those commands produce different outputs, because you may want to install the package differently. Either as a local user package `-m site --user-site`, or in the venv you're in `-c "import site; print(site.getsitepackages()[0])"`. At least, I assume those are the two main options that a Python user is interested in, because a system install into the default CMake prefix.

The point of this PR is to make clear that using CMake to install the Python package is going to be very unpythonic, and if that is what you want, you should really use pip. Does the documentation succeed in conveying that message?

brenthuisman:
> A side comment: has the `import site; print(site.getsitepackages()[0])` command been tested on a range of different OS combinations?

No, and it is there merely as a suggestion. The fact that none of these are guaranteed to work is very much the point. That line being embedded in an example made that clear enough, I hoped, but if it is not, I can add an extra remark that there is zero guarantee it does what you think it might be doing.

doc/install/build_install.rst
Helveg: OK, I think it does. I googled "arbor python installation" and basically got `pip3 install arbor` which is good 👍 Then I went to the docs homepage and read through the "Get Arbor" text which also pointed me to the Python installation and `pip3 install arbor` or to build from source, where these paths seem to make their entry and the note reasonably explains that the path might be tricky to get right but the explanation of `-DARB_PYTHON_PATH_LIB` is there.

doc/install/build_install.rst
brenthuisman: `site.getsitepackages()` does not in fact return _user_ site-package directories (ever?), but I'll make it clearer it's up to the user to understand what they're feeding in here.

doc/install/build_install.rst
brenthuisman: :+1:

Of course, let me know if you find out a better way.

bcumming:
The old command, `import sysconfig; print(sysconfig.get_path('platlib'))` was tested very extensively, and had been verified to work with virtualenvs, system installs, conda, homebrew, spack, on MacOs and various Linux distros.

If there is a case where it doesn't work, why replace it with another approach that is also imperfect?

brenthuisman:
It is not replaced. It is explained that according to the official docs and a particular case it does not do what we thought it was doing.

bcumming:
> It is not replaced. It is explained that according to the official docs and a particular case it does not do what we thought it was doing.

It is directly replaced in the example that we provide for a method to get the appropriate installation directory. The previous version had been tested extensively by many of us in a broad range of environments, and had not been observed to break. So, while like every approach that isn't `pip install...`, it has been demonstrated to be very stable.

So the justification for making this change is still a mystery to me: why is the new approach better? Does it work around an observed problem? Or is it a stylistic change?

doc/install/python.rst
```diff
@@ -66,77 +71,73 @@ Every time you make changes to the code, you'll have to repeat the second step.
...
 * ``--mpi``: Enable MPI support (requires MPI library).
 * ``--gpu``: Enable GPU support for NVIDIA GPUs with nvcc using ``cuda``, or with clang using ``cuda-clang`` (both require cudaruntime).
   Enable GPU support for AMD GPUs with hipcc using ``hip``. By default set to ``none``, which disables gpu support.
-* ``--vec``: Enable vectorization. This might require choosing an appropriate architecture using ``--arch``. Note that on x86-64 platforms compilation will fail if you enable vectorization, but the CPU or ``--arch`` does not support any form of AVX.
-* ``--arch``: CPU micro-architecture to target. By default this is set to ``native``.
-
-If calling ``setup.py`` the flags must come after ``install`` on the command line,
-and if being passed to pip they must be passed via ``--install-option``. The examples
-below demonstrate this for both pip and ``setup.py``.
+* ``--vec``: Enable vectorization. This requires choosing an appropriate architecture using ``--arch``.
```
bcumming, line 83: This statement is not really correct. If you enable vectorization on a platform that supports vectorization, and don't set `--arch`, the correct vectorized code will be generated.

Instead, you could mention that

> The `--arch` argument, documented below, may also have to be set appropriately to generated vectorized code.

bcumming (changes requested):
I am still confused about why the approach, which we had tested and shown produced the least surprising results in the past, and is used internally by our CMake to generate the full installation path for Python packages, is replaced in the example with another approach which is apparently also not perfect, and has not been tested widely.

doc/install/build_install.rst
```diff
@@ -456,39 +455,44 @@ use ``ARB_PYTHON_LIB_PATH`` to specify the location where the Python module is t
 
     cmake .. -DARB_WITH_PYTHON=on -DARB_PYTHON_PATH_LIB=/custom/path
 
-If CMake is run in a `venv` or Conda environment, set ``CMAKE_INSTALL_PREFIX`` to the
-base path of the venv. The example below shows a workflow that creates a virtual
-environment, then installs Arbor inside the environment.
+.. note::
+    The location of libraries under a prefix in only guaranteed to be standard for Python's global library location.
+    Therefore, correct installation of the Python package to any other location using ``CMAKE_INSTALL_PREFIX``,
+    such as user directory (e.g. `~/.local`), a Python or Conda virtual environment, may result in installation to a wrong path.
```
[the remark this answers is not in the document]
brenthuisman: So the issue is that under the prefix, the `/lib[arch]` dir is up to the implementation/Python distro. We don't know whether it will be `/lib` or `/lib64`, hence the warning in the Python docs. The only way to be sure of the path, is to not assume a directory layout under the prefix and ask Python for the full path, for instance with `python3 -m site --user-site`, and set `ARB_PYTHON_LIB_PATH`. (I did not know this, until I ran into such a Python distro.)

So, I think it's important to advise the user that setting a prefix is not guaranteed to work for installs except in `/usr`, where Python _does_ promise the directory layout we all know and love. The only way to be sure it's installed in a path that is used, is to 1. use pip, not CMake; or 2. prod at Python yourself to find the particular path it uses and set it with `ARB_PYTHON_LIB_PATH`. The issue with 2. is that there is no single function to obtain this path (also depends on whether you want a user, venv, system install).

doc/install/python.rst
brenthuisman: :+1:

bcumming (approved):

PR #1624, opened 2021-08-12
Categorise ASSIGNED RANGE variables as STATE-ish.

### LINKED ISSUE #1623: NMODL `ASSIGNED` variables appear as `PARAMETER` in `mech_info`
Some fallout from the ABI change
```
In [1]: import arbor as A

In [2]: m = A.default_catalogue()['hh']

In [3]: m.globals
Out[3]: {}

In [4]: m.parameters
Out[4]:
{'gnabar': {units: 'S / cm2', default: 0.12, min: -1.79769e+308, max: 1.79769e+308},
 'el': {units: 'mV', default: -54.3, min: -1.79769e+308, max: 1.79769e+308},
 'q10': {units: '1', default: nan, min: -1.79769e+308, max: 1.79769e+308},                      # <<<<<< Why?????
 'gl': {units: 'S / cm2', default: 0.0003, min: -1.79769e+308, max: 1.79769e+308},
 'gkbar': {units: 'S / cm2', default: 0.036, min: -1.79769e+308, max: 1.79769e+308}}

In [6]: m.state
Out[6]:
{'m': {units: '1', default: nan, min: -1.79769e+308, max: 1.79769e+308},
 'h': {units: '1', default: nan, min: -1.79769e+308, max: 1.79769e+308},
 'n': {units: '1', default: nan, min: -1.79769e+308, max: 1.79769e+308}}
```
if you take a look at the hh.mod  you see `q10` is `RANGE ASSIGNED`, and neither `PARAMETER` nor `STATE` .
The cause is that the tables are -- and were before! -- constructed quite naively by traversing the variables
and
```
v.is_array() &&  v.is_state() => state_table
v.is_array() && !v.is_state() => parameter_table
```
Before it did not matter, since `field_table` collected all of them. Even now it's more of an aesthetic issue,
but `mechinfo` exposes these variables as tunable `parameters`. 

So, we have the options of categorising them as `STATE` or adding a new category

- `STATE` is _close_ to what they are, but this would mean these variable can be probed like state vars.
-  in particular, the GUI will expose them as such via `mech_info`
- adding a new category though means changing the ABI
- we can make that a soft change by adding a `bool internal` field to the info, but it's still a change.

thorstenhater:
Fixes #1623.

Also use the opportunity to clean up the infoprinter.

### DIFF
--- a/modcc/printer/printerutil.hpp
+++ b/modcc/printer/printerutil.hpp
@@ -105,6 +105,9 @@ struct module_variables_t {
     std::vector<VariableExpression*> arrays;
 };
 
+
+
+
 // Scalar and array variables with local linkage.
 
 module_variables_t local_module_variables(const Module&);
--- a/modcc/printer/printerutil.hpp
+++ b/modcc/printer/printerutil.hpp
@@ -105,9 +105,6 @@ struct module_variables_t {
     std::vector<VariableExpression*> arrays;
 };
 
-
-
-
 // Scalar and array variables with local linkage.
 
 module_variables_t local_module_variables(const Module&);
--- a/arbor/backends/gpu/shared_state.cpp
+++ b/arbor/backends/gpu/shared_state.cpp
@@ -243,7 +243,7 @@ struct chunk_writer {
 };
 }
 
-void shared_state::set_field(mechanism& m, const std::string& key, const std::vector<arb_value_type>& values) {
+void shared_state::set_parameter(mechanism& m, const std::string& key, const std::vector<arb_value_type>& values) {
     if (values.size()!=m.ppack_.width) throw arbor_internal_error("mechanism parameter size mismatch");
     const auto& store = storage.at(m.mechanism_id());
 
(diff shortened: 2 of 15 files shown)


thorstenhater:
bors try

thorstenhater:
bors try

thorstenhater:
bors try

noraabiakar:
bors try

modcc/printer/printerutil.hpp
```diff
@@ -105,6 +105,9 @@ struct module_variables_t {
     std::vector<VariableExpression*> arrays;
 };
 
+
```
noraabiakar, line 108: Is this extra whitespace on purpose?

noraabiakar (approved):
Looks good! 
One minor comment but I'm approving.

modcc/printer/printerutil.hpp
thorstenhater: No, that was an accident. Cleaned-up.

PR #1626, opened 2021-08-16
Test for presence of AVX when enabling vectorization on x86-64 platforms

brenthuisman:
Gives you the error at config-time, rather than build-time.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -260,6 +260,19 @@ else()
     set(ARB_CXX_FLAGS_TARGET_FULL ${ARB_CXX_FLAGS_TARGET} ${ARB_CXXOPT_ARCH})
 endif()
 
+execute_process(
+    COMMAND sh "-c" "gcc -march=${ARB_ARCH} -dM -E - < /dev/null | grep 'AVX' | wc -l"
+    OUTPUT_VARIABLE GCC_AVX_PRESENT
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(
+    COMMAND sh "-c" "gcc -march=${ARB_ARCH} -dM -E - < /dev/null | grep 'x86_64' | wc -l"
+    OUTPUT_VARIABLE GCC_TARGET_ARCH_IS_X86_64
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+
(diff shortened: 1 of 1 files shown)


bcumming:
Do you have an example of the type of error that this is addressing?

From what I can tell, this condition will always evaluate to true for any CPU architecture more recent than Nehalem.

brenthuisman:
Westmere/Sandybridge is the cutoff. Some virtualization methods may not transmit the correct arch/options either. Perhaps most relevant: `x86-64-v2`, which Red Hat uses as the default arch, does not support AVX. I kinda use it for precompiled binaries.

bcumming:
Enabling vectorization without specifying at target architecture is risky, particularly as you point out, if you are doing it in a virtual environment or an OS that uses an ancient default architecture.

The proposed solution will break for targets that are not x86 for some versions of clang/gcc, and with some compiler wrappers (it will break the spack build, for example), and when using custom target-specific flags.

Our CMake logic around architectures is already quite complex, and to support configure-time detection would require significant additional complexity, that would probably break in more weird and wonderful ways.

brenthuisman:
The PR checks for x86-64, and only then fails if the requested arch (including `native`) supports AVX according to gcc/clang. So, there should be no breakage for non x86-64 targets.

Ancient is not the word I would use: Intel were producing 'new' Atoms until 2018 or so without AVX, and also AMD had those extensions later than Intel (2014 they added them to their top-of-the-line cpus as far as I could establish).

bcumming (changes requested):
This will not work in all cases, because it tests the output of passing `-march=ARB_ARCH` to the chosen compiler/wrapper, which is not always passed to the compiler, and it focuses only on x86 targets where we assume AVX* vector extensions.

A more robust solution would be to test whether the compiler can compile Arbor SIMD code with the CXX flags used to compile mechanisms.

Steps to achieve this:
1. Use the flags `ARB_CXX_FLAGS_FULL`
  * which is set after this line: https://github.com/arbor-sim/arbor/blob/master/CMakeLists.txt#L269
2. Test whether the compiler can compile SIMD
  * A `cmake/dummy-simd.cpp` file with one line that includes the Arbor simd header might be sufficient
  * use `try_compile`, see https://github.com/arbor-sim/arbor/blob/master/cmake/CompilerOptions.cmake#L103
  * make a stand-alone test in the `cmake` path

PR #1627, opened 2021-08-18
Add dry run benchmark cell model for testing communication scaling

bcumming:
Add a modified version of the benchmark cell example that uses dry run scaling.


example/drybench/readme.md
```diff
@@ -0,0 +1,50 @@
+# Dryrun Example
+
+A miniapp that demonstrates how to use dry-run mode to simulate the effect of communication scaling
+with benchmark cells. This is useful for evaluating the communication overheads as a model
+is weak-scaled over MPI.
+
+
```
thorstenhater, line 7: This newline could be replaced by a more in-depth description. 
- Dry-run does not set up real MPI ranks, instead a single rank is used.
- The remaining ranks are fake and just generate spikes for the local tile to process.

thorstenhater (changes requested):
Just a question on the Readme, rest looks fine.

example/drybench/readme.md
bcumming: You are right that "communication overheads" would include MPI communication, which this benchmark does not.
Instead, if allows us to focus on the overheads of spike and event handing as the number of MPI ranks increases, without having to use MPI.

thorstenhater: Should I add a comment regarding this to the PR, or do you want to do that yourself. As it stands it's the only blocker.

thorstenhater: @bcumming 👋

thorstenhater (approved):
Thanks 👍

brenthuisman:
Tests are not passing however, problem?

noraabiakar:
bors try

PR #1629, opened 2021-08-19
Fix ambiguous Region/Locset expressions

thorstenhater:
Fixes an issue where `join` (and likely `intersect`) would not work due to the fact that
`nil` values could not be coerced to `region` or `locset`. This occurred while loading 
a NML file in the GUI.

While investigating it was found that certain calls, eg `(join () ())` are ambiguous and cannot
be resolved without changes in no proportion to the gain. The cause here is that
we can resolve `()` as region or a locset and at the same time `join` maps a list of
locsets or regions to a locset or region, consequently `(join () ())` can be both.

For future reference, in `label_parse.cpp::eval` we look for the first match in the `eval_map` 
of a function name that can be successfully evaluated. However, this might not be the best match
(and the ordering depends on the internals of `eval_map`).  We _could_ check all successful
evaluations, but as eval is recursive, this idea has some obvious issues.

- Add a set of test cases to check the behaviour
- Remove the function `(nil)` from the the DSL
- Remove `nil` and `()` as literals
- Add functions `(region-nil)` and `(locset-nil)`

### DIFF
--- a/arbor/morph/locset.cpp
+++ b/arbor/morph/locset.cpp
@@ -41,7 +41,7 @@ mlocation_list thingify_(const nil_& x, const mprovider&) {
 }
 
 std::ostream& operator<<(std::ostream& o, const nil_& x) {
-    return o << "nil";
+    return o << "(locset-nil)";
 }
 
 // An explicit location.
--- a/arbor/morph/region.cpp
+++ b/arbor/morph/region.cpp
@@ -41,7 +41,7 @@ mextent thingify_(const nil_& x, const mprovider&) {
 }
 
 std::ostream& operator<<(std::ostream& o, const nil_& x) {
-    return o << "nil";
+    return o << "(region-nil)";
 }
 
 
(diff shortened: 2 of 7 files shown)


bcumming (changes requested):
Looks good, and I think it is the simplest solution to not treat `nil` as either region or locset.

some small fixes:
1. the `doc/scripts/gen-labels.py` script needs to be updated on line 166 to replace `(nil)` with `(region-nil)`.
2. add unit tests that verify to that errors are produced correctly, specifically at line 351 in `test/unit/test_s_expr.cpp`:

```C++
352                      "()",           // an empty/nil definition
353                      "(join (terminal) ())",  // locset join with nil
354                      "(join (terminal) nil)", // locset join with nil
355                      "(join () (tag 3))",     // region join with nil
```

thorstenhater:
bors try

bcumming (approved):

PR #1633, opened 2021-08-20
Check that mechanisms have the right kind

noraabiakar:
- Expose `arb_mechanism_kind` to `mechanism_info`.
- While building mechanism data in `fvm_build_mechanism_data`, check that synapses, painted region dynamics and reversal potential methods have the expected `arb_mechanism_kind`. 
- Add unit test.

### DIFF
--- a/arbor/fvm_layout.cpp
+++ b/arbor/fvm_layout.cpp
@@ -858,14 +858,17 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties& 
         const std::string& name = entry.first;
         mechanism_info info = catalogue[name];
 
-        std::vector<double> param_dflt;
         fvm_mechanism_config config;
-        config.kind = arb_mechanism_kind_density;
+        if (info.kind != arb_mechanism_kind_density) {
+            throw cable_cell_error("Expected density mechanism, got "+ name);
+        }
+        config.kind = info.kind;
 
         std::vector<std::string> param_names;
         assign(param_names, util::keys(info.parameters));
         sort(param_names);
 
+        std::vector<double> param_dflt;
         std::size_t n_param = param_names.size();
         param_dflt.reserve(n_param);
         config.param_values.reserve(n_param);
@@ -963,6 +966,10 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties& 
         const std::string& name = entry.first;
         mechanism_info info = catalogue[name];
 
+        if (info.kind != arb_mechanism_kind_point) {
+            throw cable_cell_error("Expected point mechanism, got "+ name);
+        }
+
         post_events |= info.post_events;
         std::size_t n_param = info.parameters.size();
         std::size_t n_inst = entry.second.size();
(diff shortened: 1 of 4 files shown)


arbor/fvm_layout.cpp
```diff
@@ -858,14 +858,17 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties&
         const std::string& name = entry.first;
         mechanism_info info = catalogue[name];
 
-        std::vector<double> param_dflt;
         fvm_mechanism_config config;
-        config.kind = arb_mechanism_kind_density;
+        if (info.kind != arb_mechanism_kind_density) {
+            throw cable_cell_error("Expected density mechanism, got "+ name);
```
thorstenhater, line 863: For all these errors it might help the user if the actual `kind` is included in the message, eg
```
Expected density mechanism, got <NAME> which is of kind <KIND>.
```
It's nothing that can't be checked with a Python REPL is 30s, but I think it would be appreciated.

thorstenhater (changes requested):
Looks good, but please add a `kind` property to the Python interface
of `mechinfo` and extend the docs accordingly.

arbor/fvm_layout.cpp
```diff
@@ -838,6 +838,16 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties&
         }
     };
 
+    // Mechanism kind as string, for error messages.
+    auto mech_kind_to_string = [](arb_mechanism_kind t) {
```
thorstenhater, line 842: Just an option for compression: This is essentially the same as in the Python wrapper.

test/unit/test_fvm_layout.cpp
```diff
@@ -129,6 +129,60 @@ namespace {
     }
 } // namespace
 
+TEST(fvm_layout, compatible_mechanisms) {
```
thorstenhater, line 132: Another option for compression:
All checks have the same layout, except the decoration step. You might prefer to pull that out into a lambda
like this for example
```cxx
    // S is not needed except for my laziness to look up the proper type.
    template<typename P, typename S> check_failure(S& system, P painter) {
        auto system = two_cell_system();

        painter(system);

        auto cells = system.cells();
        check_two_cell_system(cells);
        fvm_cv_discretization D = fvm_cv_discretize(cells, gprop.default_parameters);

        EXPECT_THROW(fvm_build_mechanism_data(gprop, cells, D), arb::cable_cell_error);
    }
```
and write this 
```cxx
do_check(system, [](auto& sys) {system.descriptions[0].decorations.place(system.builders[0].location({1, 0.4}), "hh", "syn0"); });
```
making the code more compact and saves some duplication, but uses more complex features.
Your call on whether you like it or not.

arbor/fvm_layout.cpp
noraabiakar: Yes, it just didn't seem "important" enough to make it into a public header file like `mechinfo.hpp`. Do you have any idea on where it should go?

thorstenhater: Usually I would put it next to the enum it converts, but that's in `mechanism_abi.h`. Options I see here
- make it a C function in the ABI header (I don't love this, but maybe useful)
- put it in `mechinfo.hpp` as a free function (:/ goes into the global namespace)
- put it as a member into `mechinfo` (also not in love with this)
- have it in `util/abi_helpers.hpp` or similar.
I tend to either option 1. or 4.

test/unit/test_fvm_layout.cpp
noraabiakar: Okay, I like it.

arbor/fvm_layout.cpp
noraabiakar: Since it's not a huge duplication, and it's in separate parts of the code (core lib/python bindings), and there's no clear option on where it should go, should we keep the duplication for now? If you still think it's better to take it out, I'll make a `util/abi_helpers.hpp`

thorstenhater (approved):
👍

PR #1636, opened 2021-08-27
Attempt fixing a64fx build

bcumming:
Reduce number of parallel build jobs to reduce risk of OOM in CI builds.


bcumming:
bors try

bcumming:
bors try

thorstenhater (approved):

PR #1638, opened 2021-09-01
Add description of arbor and NEURON's  `nernst` application rules to the docs.

noraabiakar:


doc/fileformat/nmodl.rst
```diff
@@ -109,3 +109,35 @@ Arbor-specific features
...
+
+2. In NEURON, there is a rule which is evaluated under the hood to determine whether or
+not the reversal potential of an ion remains constant or is calculated using ``nernst``.
+The rule is documented `here <https://neuron.yale.edu/neuron/static/new_doc/modelspec/programmatic/ions.html>`_
+and can be summarized as follows:
+
+  If the internal or external concentration of an ion is **written**, and its reversal
+  potential is **read but not written**, then the nernst equation is used **continuously**
+  during the simulation to update the reversal potential of the ion.
+  And if the internal or external concentration of an ion is **read**, and its reversal
```
thorstenhater, line 136: Do we have a way of saying that in Arbor?

doc/fileformat/nmodl.rst
```diff
@@ -109,3 +109,35 @@ Arbor-specific features
...
+  potential is **read but not written**, then the nernst equation is used **once** at the
+  beginning of the simulation to caluclate the reversal potential of the ion, and then
+  remains constant.
+  Otherwise, the reversal potential is set by the user and remains constant.
+
```
thorstenhater, line 141: There is one unclarity in the quoted NEURON docs, which I would love to do away with:
As written, it's unclear whether the rule applies to the local NMODL file only or the global
view is considered. 

Example
>  Assume a mechanism `A` that reads `eX` and writes to `Xi` is added. By the rules above, `nernst` is used.
>  Now we add mechanism `B` that writes to `eX`.
I _assume_ this results in `nernst` not being used, but is that correct?

Also, where is this rule applied by NEURON? Is having one CV overlap between `A` and `B` enough to change global behaviours? Or just for the overlap? (Does it matter?)

thorstenhater (changes requested):
Hi, 

looks good, but I have two smaller questions that might be helpful in clarifying some rules.

doc/fileformat/nmodl.rst
noraabiakar: The Arbor policy is stated above and is simpler. Reversal potentials of an ion remain at the user supplied initial value, or they are altered by a mechanism. The link provided gives more details on how to create/use such a mechanism.

doc/fileformat/nmodl.rst
noraabiakar: The global view is considered. So for a given ion, you need to look at all the mechanisms that use/write it and reason about the effect. In the above example, `A` reads the `eX` value that is written by `B` and if `B` reads `Xi`, it reads the value written by `A`. I will make this clearer. 
As for whether `A` and `B` need to be on the same section for it to apply; I think yes. The docs indicate that this behavior is evaluated per `section`. Though I would need to check to be certain.

doc/fileformat/nmodl.rst
thorstenhater: I do not debate the simplicity nor the wisdom of Arbor's policy here, but I expect people coming from Neuron
wanting to re-create what they know (again: no judgement), so offering help in this direction is in the direct
interest of our users.

doc/fileformat/nmodl.rst
thorstenhater: So, tweaking the decomposition of the morphology into sections in NEURON potentially changes the treatment
of a globally used ion species? Or is that just for the overlapping CVs?

Sorry for asking so many questions on NEURON, but I expect to have to deal with them for quite some time in the 
future.

doc/fileformat/nmodl.rst
noraabiakar: What kind of help are you thinking? Is the info in (1) missing something?

doc/fileformat/nmodl.rst
noraabiakar: So in Neuron, you can only add mechanisms on sections, not segments (CVs). I will run some examples to see what exactly happens in Neuron and get back to this question.

doc/fileformat/nmodl.rst
thorstenhater: Something like an example maybe, how one would query the `cable_cell` or `decor` for the relevant mechanisms
and their properties via `mech_info` to decide whether Neuron would have used `nernst`.

That would require some infrastructure, as we need to interrogate the `cable_cell` object whether two regions
overlap (we need the morphology and possibly even the discretised version here). Depending on the final 
judgement on Neurons behaviour (see below) that is.

doc/fileformat/nmodl.rst
noraabiakar: Okay so based on my experiments: 
NEURON's rules for using nernst for the calculation of the reversal potential of an ion as stated above in this PR are applied per section (similar to arbor which can control the reversal potential of an ion at the granularity of a region). Since mechanisms can only be inserted on a section (not part of it) and since a NEURON segment (CV) can only belong to one section, there is no possibility for segment (CV) overlap that could cause confusion.
However, the rules do not consider a single mechanism in isolation on a section. All of the mechanisms on a section are considered when the rule is applied for that section.

doc/fileformat/nmodl.rst
noraabiakar: This is a bit out of the scope of this PR in my opinion.
Also, it is good to know what NEURON does, but I don't think we should encourage it by showing how to replicate it in Arbor.

thorstenhater: I partially agree about the scope issue. However, for verification/reproduction it seems important to have this somewhere. Similarly, for onboarding new users who actively seek a Neuron-like experience (again, not judging) it might be convenient to.

doc/fileformat/nmodl.rst
thorstenhater: Copy from private discussion:
> Arbor and Neuron treat this fundamently diffent. In Neuron's model, Nernst can be used on a per-section basis in alternation with fixed reversal potentials. For Arbor, it is used globally, per ion species.

@halfflat on the motivation for not allowing it per region in Arbor,
- CVs could be partially covered by the Nernst and const methods for the same ion
- This cannot happen in Neuron, as each CV belongs to exactly one Section

thorstenhater: Let's add a comment about this final difference ARB ./. NRN ie

> In Arbor, `nernst` is applied globally at the level of ion species, while in Neuron different parts of the morphology may use different mechanisms for the revpot. This is the result of the difference in the models of discretisation (in Arbor `regions` may partially cover CVs).

and then I am happy :D

PR #1641, opened 2021-09-06
Fix class members in cable example

schmitts:


thorstenhater:
Hi,

in principle this looks good, but for the review: could you please a one-line description?

thorstenhater (changes requested):
Looks good, but please add a short explanation on why and what happens here. It helps the reviewers to categorise your PR and start looking for context.

schmitts:
@thorstenhater ok now?

PR #1652, opened 2021-09-09
Add example based on LFPykit example Example_Arbor_swc.ipynb

### LINKED ISSUE #1036: Prepare extracellular potential demo for LFPy (Python)
Arbor may be able to be well utilized as a back-end for [LFPy](https://github.com/LFPy/LFPy). As a first step, we will make a simple demo in Python that computes the potential at a fixed set of electrode sites from a single morphologically detailed neuron.

Model NEURON python code: [pure_neuron_example.py.txt](https://github.com/arbor-sim/arbor/files/4629768/pure_neuron_example.py.txt)

Once #1034 is merged, we'll need to:
* expose trans membrane current probes and sample data in Python interface, together with metadata;
* use the single-cell Python API to model and run a neuron such as in the example code;
* and in anticipation of running multiple neurons, we will need to implement exact sample timing.

espenhgn:
This PR adds an example showcasing predictions of extracellular potentials from a detailed cell model on .swc format using LFPykit (https://lfpykit.readthedocs.io). 

Closes #1036.


brenthuisman (changes requested):
Can you add the script to `scripts/run_python_examples.sh`? Don't forget to add `python -m pip install -r requirements.txt` there too.

brenthuisman (approved):
lgtm!

PR #1654, opened 2021-09-09
Fix simulation::add_sampler cell_group dispatch

### LINKED ISSUE #1653: Sampler exceptions in mixed `cell_kind` simulations
When a model contains multiple `cell_kind`s, some of which don't support sampling, an exception is always thrown from `simulation::add_sampler`. 
The culprit: 
```
sampler_association_handle simulation_state::add_sampler(
        cell_member_predicate probe_ids,
        schedule sched,
        sampler_function f,
        sampling_policy policy)
{
    sampler_association_handle h = sassoc_handles_.acquire();

    foreach_group(
        [&](cell_group_ptr& group) { group->add_sampler(h, probe_ids, sched, f, policy); });

    return h;
}
``` 
The `add_sample` function is dispatched to all `cell_group`s (not only the responsible one), which will throw an exception if sampling is not supported.

noraabiakar:
Add check on dispatch to `cell_group::add_sampler` in `simulation_state::add_sampler` to avoid erroneous exceptions.

Fixes #1653

### DIFF
--- a/arbor/benchmark_cell_group.cpp
+++ b/arbor/benchmark_cell_group.cpp
@@ -98,9 +98,6 @@ void benchmark_cell_group::add_sampler(sampler_association_handle h,
                                    cell_member_predicate probe_ids,
                                    schedule sched,
                                    sampler_function fn,
-                                   sampling_policy policy)
-{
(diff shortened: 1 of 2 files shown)


halfflat (approved):
Looks good!

PR #1657, opened 2021-09-09
Segfault on instantiating mechanisms on empty regions

thorstenhater:
More, subtler fallout of the ABI patchset: When mechanisms are added to empty regions, various, layered segfaults
appear.
1. `set_parameter` will receive a `nullptr` upon linear search, throwing `no such parameter`
2. Guarding this for finite `width` uncovers further need in the ABI interface methods
3. This is caused by early exit in `instantiate` on `width == 0`
4. Getting rid of this is better, as we can treat vacuous mechanism painting like all others...
5. ...but causes another segfault, since we default index arrays to `cv_pos.back()` but `cv_pos` is empty.

This patch fixes all of the above and in addition removes mechanisms without support before reifying them
on the FVM cells. Relevant tests are added.

### DIFF
--- a/arbor/backends/multicore/shared_state.cpp
+++ b/arbor/backends/multicore/shared_state.cpp
@@ -522,9 +522,6 @@ void shared_state::instantiate(arb::mechanism& m, unsigned id, const mechanism_o
         m.ppack_.ion_states[idx] = { oion->iX_.data(), oion->eX_.data(), oion->Xi_.data(), oion->Xo_.data(), oion->charge.data() };
     }
 
-    // If there are no sites (is this ever meaningful?) there is nothing more to do.
-    // if (m.ppack_.width==0) return;
-
     // Initialize state and parameter vectors with default values.
     {
         // Allocate bulk storage
@@ -570,7 +567,7 @@ void shared_state::instantiate(arb::mechanism& m, unsigned id, const mechanism_o
         store.indices_ = iarray(count*index_width_padded, 0, pad);
         chunk_writer writer(store.indices_.data(), index_width_padded);
         // Setup node indices
-        m.ppack_.node_index = writer.append(pos_data.cv, pos_data.cv.empty() ? NAN : pos_data.cv.back());
+        m.ppack_.node_index = writer.append(pos_data.cv, pos_data.cv.empty() ? 0 : pos_data.cv.back());
         auto node_index = util::range_n(m.ppack_.node_index, index_width_padded);
         // Make SIMD index constraints and set the view
         store.constraints_ = make_constraint_partition(node_index, m.ppack_.width, m.iface_.partition_width);
--- a/arbor/backends/multicore/shared_state.cpp
+++ b/arbor/backends/multicore/shared_state.cpp
@@ -567,7 +567,8 @@ void shared_state::instantiate(arb::mechanism& m, unsigned id, const mechanism_o
         store.indices_ = iarray(count*index_width_padded, 0, pad);
         chunk_writer writer(store.indices_.data(), index_width_padded);
         // Setup node indices
-        m.ppack_.node_index = writer.append(pos_data.cv, pos_data.cv.empty() ? 0 : pos_data.cv.back());
+        auto pad_val = pos_data.cv.empty() ? 0 : pos_data.cv.back();
+        m.ppack_.node_index = writer.append(pos_data.cv, pad_val);
         auto node_index = util::range_n(m.ppack_.node_index, index_width_padded);
         // Make SIMD index constraints and set the view
         store.constraints_ = make_constraint_partition(node_index, m.ppack_.width, m.iface_.partition_width);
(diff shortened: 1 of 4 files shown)


arbor/fvm_layout.cpp
```diff
@@ -943,7 +943,7 @@ fvm_mechanism_data fvm_build_mechanism_data(const cable_cell_global_properties&
         }
 
         update_ion_support(info, config.cv);
-        M.mechanisms[name] = std::move(config);
+        if (!config.cv.empty()) M.mechanisms[name] = std::move(config);
```
noraabiakar, line 946: Why only density mechanisms, not also point mechanisms?

arbor/backends/multicore/shared_state.cpp
```diff
@@ -567,8 +567,7 @@ void shared_state::instantiate(arb::mechanism& m, unsigned id, const mechanism_o
         store.indices_ = iarray(count*index_width_padded, 0, pad);
         chunk_writer writer(store.indices_.data(), index_width_padded);
         // Setup node indices
-        m.ppack_.node_index = writer.append(pos_data.cv, pos_data.cv.back());
-
+        m.ppack_.node_index = writer.append(pos_data.cv, pos_data.cv.empty() ? 0 : pos_data.cv.back());
```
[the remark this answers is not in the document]
thorstenhater: No, but I see why this is confusing. It's just an unfortunate quirk of the `writer` interface. You have to supply a padding value, for which the default is `cv.back()`, but `cv` is empty, so we need to summon something from the void. This value is not used, though, as `cv` is empty and thus no value is appended.

It's probably better to wrap the whole statement in a conditional and set `node_index = nullptr`.

arbor/backends/multicore/shared_state.cpp
```diff
@@ -430,17 +430,20 @@ std::size_t extend_width(const arb::mechanism& mech, std::size_t width) {
...
     for (arb_size_type i = 0; i<m.mech_.n_parameters; ++i) {
         if (key==m.mech_.parameters[i].name) {
             data = m.ppack_.parameters[i];
+            found = true;
             break;
         }
     }
 
-    if (!data) throw arbor_internal_error(util::pprintf("no such parameter '{}'", key));
+    if (!found) throw arbor_internal_error(util::pprintf("no such parameter '{}'", key));
```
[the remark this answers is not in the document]
thorstenhater: Yes, there's at least two ways `data` can be nil here and only one coincides with the `not found` case. Classic boolean blindness.

thorstenhater:
bors try

arbor/backends/multicore/shared_state.cpp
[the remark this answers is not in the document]
thorstenhater: I am worried that using nil here might cause confusion, so I explained the why instead.

arbor/fvm_layout.cpp
thorstenhater: Done

test/unit/test_fvm_lowered.cpp
```diff
@@ -535,6 +535,34 @@ TEST(fvm_lowered, derived_mechs) {
...
+    auto cell = builder.make_cell();
+
+    cell.decorations.paint(reg::nil(), "test_kin1");
+    cell.decorations.paint(reg::nil(), "custom_kin1");
+
+    cable1d_recipe rec(cable_cell{cell});
+    rec.catalogue() = make_unit_test_catalogue();
+    rec.catalogue().derive("custom_kin1", "test_kin1", {{"tau", 20.0}});
+
+    std::vector<double> samples[3];
```
noraabiakar, line 558: This isn't used.

test/unit/test_fvm_lowered.cpp
```diff
@@ -535,6 +535,34 @@ TEST(fvm_lowered, derived_mechs) {
...
+    auto ctx = make_context(resources);
+    auto decomp = partition_load_balance(rec, ctx);
+    simulation sim(rec, decomp, ctx);
+    sim.run(30.0, 1.f/1024);
```
noraabiakar, line 562: Should we be expecting or asserting something here?

test/unit/test_abi.cpp
```diff
@@ -187,4 +234,53 @@ TEST(abi, gpu_initialisation) {
...
+    arb::gpu::shared_state shared_state(ncell, ncell, 0,
+                                        cv_to_intdom, cv_to_intdom,
+                                        gj, vinit, temp, diam, src_to_spike,
+                                        1);
+
+    arb::mechanism_layout layout;
+    layout.weight.assign(ncv, 1.);
+    for (arb_size_type i = 0; i<ncv; ++i) layout.cv.push_back(i);
+
+    shared_state.instantiate(mech, 42, {}, layout);
```
noraabiakar, line 282: Same here, should we be checking that this example doesn't throw? Or examining the result of the instantiation in some way?

noraabiakar (changes requested):
Looks good, thanks for fixing this. I just have some questions about the unit tests.

test/unit/test_fvm_lowered.cpp
thorstenhater: True, copy-pasta left-overs.

test/unit/test_abi.cpp
thorstenhater: NO_THROW is correct, the result is nil.

noraabiakar (approved):

PR #1658, opened 2021-09-09
Fix on-components on empty cable bug

### LINKED ISSUE #1645: Amplified stimulus with current clamp at edge of cylinder
When placing a current clamp at the edge of a cylinder, the stimulus response is greater compared to placing the current clamp to "edge - epsilon".

The attached example puts out a plot `cylinder_edge.svg`.

Modify line 88 to see the effect.

[cylinder_edge.py.gz](https://github.com/arbor-sim/arbor/files/7121434/cylinder_edge.py.gz)

halfflat:
* Filter results of `ls::on_components(1, reg)` to avoid duplicate or coincident locations arising from terminal zero-length cables in region argument.

Fixes #1645

### DIFF
--- a/arbor/morph/locset.cpp
+++ b/arbor/morph/locset.cpp
@@ -353,7 +353,7 @@ mlocation_list thingify_(const on_components_& n, const mprovider& p) {
         }
         else if (n.relpos==1) {
             double diameter = 0;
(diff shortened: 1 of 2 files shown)


noraabiakar (approved):
Looks good!

PR #1670, opened 2021-09-16
Feature/python clean memory

max9901:
The have the ability to clear the memory through the python API making it possible to run very long simulations without running out of memory. 

In the Simulation class that is mapped to the python interface the  `::clear()` function call is added. 
We would love to have this as a feature in core arbor. 

cheerz,
EMC team.


bcumming:
Thanks @max9901!

What is the reason for making random123 a sub-module?

max9901:
I found it easier to compile + this way its easier to update. 
And i was actually wondering why it wasn't a submodule in the first place.

bcumming:
Let's take it out of this PR, because it is best to make each PR about a single topic, which for this PR is adding support for clearing spike and sample information through the Python interface.

You can create an issue proposing that we make random123 a submodule, with the reasons why, and we can discuss it with the team.

max9901:
Cleaned up this PR. ✔️

thorstenhater (changes requested):
Hi @max9901 

thanks for adding this, I have some small requests:
1. Please add some documentation, somewhere around here https://docs.arbor-sim.org/en/latest/python/simulation.html?highlight=sample#arbor.simulation.reset
2. A test in the Python suite would also be useful, eg one to prove that after `clear` there's no more data in the sampler.
3. I would prefer a different name, maybe `reset_samplers`, to make the effect more clear to the user. What do you think?
3. Optional: If **you** consider it useful, maybe add a more selective call to clear only individual samplers.

PR #1672, opened 2021-09-20
Fixed MRO and code duplication in `setup.py`

Helveg:
The duplication arose from slightly more complicated composition than usual but could be resolved by proper MRO, it took me a few passes from its original form as well before I figured out the situation is not as complicated as it seems.

`super()` is equal to `super(cls, self)` where `cls` is obtained from the current function's class scope. If arg 2  is an object it will start an MRO search of the object, but skipping classes before the 1st arg type is encountered. In our case with the following base classes: `(_command_template, base_command_class)` this means we can define our base logic in `_command_template` and use `super()` calls there to traverse the MRO in `(base_command_class,)`, so by simply defining `class install_command(_command_template, install)` and `bdist_wheel_command(_command_template, bdist_wheel)` our MRO issue is solved and the duplication removed 👍

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,9 +67,7 @@ class _command_template:
     """
     Override a setuptools-like command to augment the command line options.
     Needs to appear before the command class in the class's argument list for
-    correct MRO. Will assume that the 2nd class is the base command class unless
-    a ``base`` kwarg is explicitly defined in the class's argument list.
-
+    correct MRO.
     Examples
     --------
 
@@ -79,8 +77,11 @@ class _command_template:
           pass
 
 
-      class complex_command(_command_template, mixin1, install, base=install):
-          pass
+      class complex_command(_command_template, mixin1, install):
+          def initialize_options(self):
+              # Both here and in `mixin1`, a `super` call is required
+              super().initialize_options()
+              # ...
 
     Usage
     -----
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -62,15 +62,42 @@ def check_cmake():
...
+      class complex_command(_command_template, mixin1, install):
+          def initialize_options(self):
+              # Both here and in `mixin1`, a `super` call is required
+              super().initialize_options()
+              # ...
+
+    Usage
+    -----
+
+        python3 setup.py install --mpi --arch=skylake
```
brenthuisman, line 90: Might be a better idea to refer to `doc/install/python.rst`, it's more complete.

brenthuisman (changes requested):
Wheel build passes [1] and I tested the output on my machine with success, so lgtm apart from a minor comment.

[1] https://github.com/brenthuisman/arbor/actions/runs/1253580549

setup.py
Helveg: There was a dev comment there before, which I replaced by a docstring on the private member `_command_template`; It won't be picked up by `sphinx` by default, but if one ever makes a dev-doc with `:private-members:` this would be included; if in that context you still want to refer, could you suggest an edit?

brenthuisman: I meant the `Usage` section in particular, which is not complete here, and does not really need to be, as long as that documentation lives in `doc/install/python.rst`. Autodoc won't be used for some time to come, due to the incompatibilities with Pybind.

Helveg: Ok, so I remove the `Usage` section and the old dev comment about the `install_command` is not included anymore? It was a bit out of place anyway imho.

brenthuisman: Yes, let's remove Usage (or refer to the docs). However, the rest of the docstring can be left in place, the example is worth keeping I would say.

brenthuisman (approved):

PR #1673, opened 2021-09-20
Makejobs arg

Helveg:
Added a `--makejobs=4` user option to `setup.py` to control how many jobs should be started to `make` the project during the installation.

Depends on #1672 to avoid duplicate defenition of the arg and merge issues.


brenthuisman (changes requested):
Can you advertise the flag in `doc/install/python.rst`? Under the Advanced Options header.

Default of 2 is OK with me.

brenthuisman (approved):

PR #1677, opened 2021-09-21
Added an empty catalogue

### LINKED ISSUE #1676: Add an empty catalogue
For functional operations with catalogues it would be useful to have a builtin empty catalogue.

Helveg:
The empty catalogue can be used to simplify catalogue combination. closes #1676.

Does it need to be documented anywhere or does `pybind` do that?


arbor/include/arbor/mechcat.hpp
```diff
@@ -112,6 +112,7 @@ class mechanism_catalogue {
 // Reference to global default mechanism catalogue.
 
 const mechanism_catalogue& global_default_catalogue();
+const mechanism_catalogue& global_empty_catalogue();
```
thorstenhater, line 115: Ok!

mechanisms/CMakeLists.txt
```diff
@@ -31,11 +31,22 @@ make_catalogue(
...
+  CXX_FLAGS_TARGET "${ARB_CXX_FLAGS_TARGET_FULL}"
+  STANDALONE FALSE
+  VERBOSE ${ARB_CAT_VERBOSE})
+
 # Join sources
 set(arbor_mechanism_sources
   ${CAT_BBP_SOURCES}
   ${CAT_ALLEN_SOURCES}
   ${CAT_DEFAULT_SOURCES}
+  ${CAT_EMPTY_SOURCES}
```
thorstenhater, line 49: But you are making this way harder than it should be. Just implement
```cxx
const mechanism_catalogue& global_empty_catalogue() {
  static mechanism_catalogue empty = mechanism_catalogue{};
  return empty;
}
```

thorstenhater (changes requested):
Hi,

I see the need for this, especially in the light of lifetime management concerning catalogues,
but I wonder if it would not be simpler to expose the default constructor of `mechanism_catalogue` to Python. In any case, please add docs and tests (if you want help with one of those, please ask).

mechanisms/CMakeLists.txt
Helveg: Where should this code go?

Helveg:
Help with both docs and test would be appreciated:

* Where should this be documented?
* Where can I add tests and how can I run tests locally?

mechanisms/CMakeLists.txt
thorstenhater: This should go into `mechcat.cpp`, like the rest of the `global_XYZ_catalogue` functions. 

As for the extensions:
- this returns a constant reference thus immutable
- if you want to extend it, you'll have to make a mutable copy

Helveg: I don't understand the part of the extensions, will this work:

```python
e1 = arbor.empty_catalogue()
e2 = arbor.empty_catalogue()
e1.extend(smth)
assert len(e2) == 0
```

thorstenhater: Like this?
```py
In [1]: import arbor as A

In [2]: c = A.default_catalogue()

In [4]: len([n for n in c])
Out[4]: 9

In [5]: c.extend(A.default_catalogue(), 'cc/')

In [6]: len([n for n in c])
Out[6]: 18
```

thorstenhater: After reading through the sources again, I think the easiest course to take is to expose the default constructor of `mechanism_catalogue` to python. See here
https://github.com/arbor-sim/arbor/blob/f13f786a35f07c6eb00f7d6d18c07a2c16664f7a/python/mechanism.cpp#L155
only the copy constructor is bound, simply add the default one in the same fashion.

As for tests, go to hereabouts 
https://github.com/arbor-sim/arbor/blob/f13f786a35f07c6eb00f7d6d18c07a2c16664f7a/python/test/unit/test_catalogues.py#L73
and add tests for the behaviours you rely on, here's the list I would currently see 
1. `A.catalogue()` has no mechs
2. `A.catalogue().extend(other)` has `other`'s mechs
3. and so on

The docs here
https://github.com/arbor-sim/arbor/blob/master/doc/python/mechanisms.rst#mechanism-catalogues
need a line or two on the constructors.

Helveg: I dont' understand this in the slightest 😇 What I can make of Line 155 in f13f786 is that the constructor is already exposed? And I don't see anything mentioning `copy`. Perhaps it's better you implement this.

thorstenhater: Welcome to C++ :) A constructor with a signature like `A(const A&)` is called a copy ctor (for `A`).
We have also the additional pleasure of working with pybind11 here, so 
```cxx
pybind11::class_<arb::mechanism_catalogue> cat(m, "catalogue");
cat.def(pybind11::init<const arb::mechanism_catalogue&>())
```
means: on the class `cat`, define a constructor (by way of `init`) that takes a `const arb::mechanism_catalogue&`
for arguments. Effectively a copy ctor.
```cxx
cat.def(pybind11::init())
```
should do the trick.

thorstenhater:
Running tests locally from `build`
- `make tests && bin/unit` for C++
- `PYTHONPATH=./python:$PYTHONPATH python ../python/test/unit/runner.py -v2`

thorstenhater (approved):
Looks good, thanks for going through with this.

PR #1678, opened 2021-09-22
Name and describe arguments to constructors where ambiguous.

### LINKED ISSUE #1666: Awkward pybind signatures
**Describe the bug**
`arbor.mechanism`'s set of accepted arguments are a bit awkward because the signature with positional arguments has anonymous parameter names. It binds `arg0` instead of `name`:

```
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. arbor._arbor.mechanism(arg0: str)
    2. arbor._arbor.mechanism(name: str, params: Dict[str, float])
```

I suppose this occurs for all functions with similar signatures?

Shouldn't the signatures be:

    1. arbor._arbor.mechanism(name: str)
    2. arbor._arbor.mechanism(name: str, params: Dict[str, float])

I can do `arbor.mechanism("name", {})` and `arbor.mechanism(name="name", params={})` and `arbor.mechanism("name")` but not `arbor.mechanism(name="name")`

brenthuisman:
Fixes #1666

### DIFF
--- a/python/cells.cpp
+++ b/python/cells.cpp
@@ -267,7 +267,7 @@ void register_cells(pybind11::module& m) {
     pybind11::class_<arb::cv_policy> cv_policy(m, "cv_policy",
             "Describes the rules used to discretize (compartmentalise) a cable cell morphology.");
     cv_policy
-        .def(pybind11::init([](const std::string& expression) { return arborio::parse_cv_policy_expression(s).unwrap(); }),
+        .def(pybind11::init([](const std::string& expression) { return arborio::parse_cv_policy_expression(expression).unwrap(); }),
             "expression"_a, "A valid CV policy expression")
         .def_property_readonly("domain",
                                [](const arb::cv_policy& p) {return util::pprintf("{}", p.domain());},
--- a/python/cells.cpp
+++ b/python/cells.cpp
@@ -267,7 +267,8 @@ void register_cells(pybind11::module& m) {
     pybind11::class_<arb::cv_policy> cv_policy(m, "cv_policy",
             "Describes the rules used to discretize (compartmentalise) a cable cell morphology.");
     cv_policy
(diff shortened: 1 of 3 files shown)


Helveg (approved):
Looks good on the Python side, can use `arbor.mechanism(name=...)` and fixed docstring:

```
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. arbor._arbor.mechanism(name: str)
    2. arbor._arbor.mechanism(name: str, params: Dict[str, float])
```

python/cells.cpp
```diff
@@ -594,7 +594,9 @@ void register_cells(pybind11::module& m) {
         .def(pybind11::init(
             [](const arb::morphology& m, const label_dict_proxy& labels, const arb::decor& d) {
                 return arb::cable_cell(m, labels.dict, d);
-            }), "morphology"_a, "labels"_a, "decor"_a)
+            }),
+            "morphology"_a, "labels"_a, "decor"_a,
+            "Construct with a morphology.")
```
noraabiakar, line 599: This constructs with a morphology, label dictionary and decor.

python/cells.cpp
```diff
@@ -267,7 +267,8 @@ void register_cells(pybind11::module& m) {
     pybind11::class_<arb::cv_policy> cv_policy(m, "cv_policy",
             "Describes the rules used to discretize (compartmentalise) a cable cell morphology.");
     cv_policy
-        .def(pybind11::init([](const std::string& s) { return arborio::parse_cv_policy_expression(s).unwrap(); }))
+        .def(pybind11::init([](const std::string& s) { return arborio::parse_cv_policy_expression(s).unwrap(); }),
+            "s"_a, "A valid CV policy expression")
```
noraabiakar, line 271: Maybe we can give this an argument name more descriptive than "s"?

noraabiakar (changes requested):
Looks good thanks! I have a few small suggestions.

noraabiakar (approved):

PR #1681, opened 2021-09-23
Fix modcc simd generation

noraabiakar:
Fixes 2 bugs where modcc expects a range variable but gets a non-range variable instead. 
1. `PowBinaryExpressions` were being handled separately from other `BinaryExpressions` and as a result, some crucial analysis was being skipped on the analysis of the lhs and rhs arguments. This PR incorporates the analysis and printing of `PowBinaryExpression` into `BinaryExpression`.
2. Printing code for a masked `AssignmentExpression` was not handling the case of a non-range variable on the rhs of the assignment correctly. In that case, an explicit cast to a vector is needed, which is added in this PR.

### DIFF
--- a/modcc/printer/cexpr_emit.cpp
+++ b/modcc/printer/cexpr_emit.cpp
@@ -377,6 +377,9 @@ void SimdExprEmitter::visit(AssignmentExpression* e) {
     auto lhs_pfxd = id_prefix(e->lhs()->is_identifier());
 
     if (lhs->is_variable() && lhs->is_variable()->is_range()) {
+        // input_mask_ will only appear in PROCEDURE and FUNCTION calls which
+        // can only assign to VariableExpression (STATE and ASSIGNED block), not
+        // IndexedVariable or any other expression
         if (!input_mask_.empty()) {
             mask = "S::logical_and(" + mask + ", " + input_mask_ + ")";
         }
@@ -390,11 +393,11 @@ void SimdExprEmitter::visit(AssignmentExpression* e) {
         bool cast = e->rhs()->is_number();
         if (cast) out_ << "simd_cast<simd_value>(";
         e->rhs()->accept(this);
+        if (cast) out_ << ")";
 
         out_ << ")";
-
-        if (cast) out_ << ")";
-    } else {
+    }
+    else {
         out_ << "S::where(" << mask << ", ";
         e->lhs()->accept(this);
         out_ << ") = ";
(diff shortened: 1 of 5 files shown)


noraabiakar:
bors try

modcc/printer/cexpr_emit.cpp
```diff
@@ -385,25 +376,41 @@ void SimdExprEmitter::visit(AssignmentExpression* e) {
...
             mask = "S::logical_and(" + mask + ", " + input_mask_ + ")";
         }
-        if(is_indirect_)
-            out_ << "indirect(" << lhs_pfxd << "+index_, simd_width_) = ";
-        else
-            out_ << "indirect(" << lhs_pfxd << "+i_, simd_width_) = ";
 
-        out_ << "S::where(" << mask << ", ";
+        std::string index = is_indirect_ ? "index_" : "i_";
+        out_ << "indirect(" << lhs_pfxd << "+" << index << ", simd_width_) = S::where(" << mask << ", ";
```
thorstenhater, line 395: I wonder whether it is helpful to write `mask` to the generated code

1. avoid deeply nested statements
2. performance, as the mask is used more than once
3. also simplify the templating

ie
```cxx
        if (!input_mask_.empty()) {
            mask = "S::logical_and(" + mask + ", " + input_mask_ + ")";
        }
       out_  << 'auto mask = {mask};'
```

thorstenhater (changes requested):
one question, rest is good to go

modcc/printer/cexpr_emit.cpp
noraabiakar: This is a bit more involved than it appears:
1. This assignment function is called for each `AssignementExpression`. At the moment the following code: 
```
if (mask) {
  a = 3
  b = 2
}
```
will generate something like 
```
S::where(mask, a) = 3;
S::where(mask, b) = 2;
```
If we simply assign the mask into a temporary variable, something like this would be generated instead: 
```
auto mask_0 = mask; 
S::where(mask_0, a) = 3;
auto mask_1 = mask;
S::where(mask_1, b) = 2;
```
Having only one of these temporary variables instead of two requires a bit more logic in the code. 

2. The reason why this wasn't attempted originally is because I thought the current way was more illustrative of the meaning of the code. Consider the following: 
```
if (mask_0) {
  if (mask_1) {
     a = 3
     b = 2
  } else {
     a = 2
     b = 3
  }
}
```
This should generate something like: 
```
S::where(mask_0 && mask_1, a) = 3
S::where(mask_0 && mask_1, b) = 2
S::where(mask_0 && !mask_1, a) = 2
S::where(mask_0 && !mask_1, b) = 3
```
We should only ever have `and`, `or` and `not` ops between named variables as the "mask" of a `S::where` expression. Of course, because of SVE we can't use `&&`, `||` and `!` which complicates the printing, but the underlying idea is still valid.

thorstenhater (approved):
I accept your argument, as said, it is a double-edged sword.

PR #1684, opened 2021-09-24
Allow to limit Poisson schedule with a stop time (closes #1617)

### LINKED ISSUE #1617: Add an optional stop time to the Poisson schedule
The `regular_schedule` can be configured with a start and stop time. Within this time window, events are generated. The `poisson_schedule` also allows for a start time but lacks a stop time.

Feature request: add an optional stop time to the Poisson schedule

schmitts:
Allow to limit Poisson schedule with a stop time using the same
semantics as the regular schedule.

### DIFF
--- a/python/schedule.cpp
+++ b/python/schedule.cpp
@@ -26,6 +26,7 @@ std::ostream& operator<<(std::ostream& o, const explicit_schedule_shim& e) {
 
 std::ostream& operator<<(std::ostream& o, const poisson_schedule_shim& p) {
     return o << "<arbor.poisson_schedule: tstart " << p.tstart << " ms"
+             << ", tstop " << util::to_string(p.tstop) << " ms"
              << ", freq " << p.freq << " kHz"
              << ", seed " << p.seed << ">";
 };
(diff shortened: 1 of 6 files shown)


noraabiakar (changes requested):
Looks good thanks! I just have a small request to add an assertion.

noraabiakar (approved):
Thanks!

python/schedule.cpp
```diff
@@ -256,6 +268,8 @@ void register_schedules(py::module& m) {
             "The expected frequency [kHz].")
         .def_readwrite("seed", &poisson_schedule_shim::seed,
             "The seed for the random number generator.")
+        .def_property("tstop", &poisson_schedule_shim::get_tstop, &poisson_schedule_shim::set_tstop,
+            "No events delivered after this time [ms].")
```
noraabiakar, line 272: `std::ostream& operator<<(std::ostream& o, const poisson_schedule_shim& p)` (line 27 of this file) needs to be amended to print out the `tstop` time as well.

schmitts: Will do.

PR #1686, opened 2021-09-27
Added a debug mode to `build-catalogue`

Helveg:
A `--debug` flag is added so that users can inspect the out-of-tree generated C++ code of their catalogue builds.

PR #1689, opened 2021-09-28
Allow to keep the directory where the catalogue was built

schmitts:
E.g. for modifiying generated sources but still using the
convenience of the build-catalogue script.

### DIFF
--- a/scripts/build-catalogue.in
+++ b/scripts/build-catalogue.in
@@ -44,6 +44,10 @@ def parse_arguments():
                         type=str,
                         help='Directory name where *.mod files live.')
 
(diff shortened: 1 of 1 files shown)


Helveg:
Heh, duplicate of #1686

Helveg:
Does #1686 work for you, do you like its implementation? I can make changes

schmitts:
> Does #1686 work for you, do you like its implementation? I can make changes

Thanks. I'll test asap.

schmitts:
@Helveg I tested #1686 and it works fine for me.

PR #1695, opened 2021-10-04
Declare `setuptools` package

### LINKED ISSUE #1683: `setup.py` doesn't declare arbor package
**Describe the bug**

```python

        # ...

            os.makedirs(self.build_temp)
```

Helveg:
Fixes #1683

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,13 @@
+[options]
+package_dir=
+    =python
+packages=find:
+
+[options.packages.find]
+where=python
+exclude=
+  test
+  example
+
 [metadata]
 description-file = python/readme.md
--- a/setup.py
+++ b/setup.py
@@ -144,11 +144,6 @@ class cmake_build(build_ext):
         build_directory = os.path.abspath(self.build_temp)
         # The path where the package will be copied after building.
         lib_directory = os.path.abspath(self.build_lib)
-        # The path where the Python package will be compiled.
-        source_path = build_directory + '/python/arbor'
-        # Where to copy the package after it is built, so that whatever the next phase is
-        # can copy it into the target 'prefix' path.
-        dest_path = lib_directory + '/arbor'
 
         opt = cl_opt()
         cmake_args = [
@@ -186,13 +181,6 @@ class cmake_build(build_ext):
         cmake_cmd = ['cmake', '--build', '.'] + build_args
         subprocess.check_call(cmake_cmd,
                               cwd=self.build_temp)
-
-        # Copy from build path to some other place from whence it will later be installed.
-        # ... or something like that
-        # ... setuptools is an enigma monkey patched on a mystery
-        if not os.path.exists(dest_path):
-            os.makedirs(dest_path, exist_ok=True)
-        self.copy_tree(source_path, dest_path)
 
 setuptools.setup(
     name='arbor',


Helveg:
@brenthuisman @bcumming is there some specific wheel build that you need to trigger? I produced wheels from it locally.

brenthuisman:
Github actions is configured to trigger on Git tags starting with `v` and the `ciwheel` branch. I'll make a push to the latter.

brenthuisman:
Wheel (and specifically manywheel) production is a bit opaque (to me). The [cibuildwheel Github Action](https://cibuildwheel.readthedocs.io) made things simpler at the price of making them more opaque, but I couldn't find a way to generate Macos wheel any other way. Locally I've succeeded making manylinux2014 wheels using the official [PyPA docker image](quay.io/pypa/manylinux2014_x86_64) and the following script:
```
#!/bin/bash
set -e -u -x

yum -y install libxml2-devel

rm -rf /opt/python/cp35-cp35m

for PYBIN in /opt/python/*/bin; do
    "${PYBIN}/pip" install wheel cmake
    export PATH="${PYBIN}":$PATH
    which cmake
    "${PYBIN}/python" ./arbor/setup.py bdist_wheel # --neuroml --vec --arch=sandybridge
done

# Bundle external shared libraries into the wheels
/opt/python/cp39-cp39/bin/pip install auditwheel
for whl in ./dist/*.whl; do
    /opt/python/cp39-cp39/bin/auditwheel repair "$whl"
done
```

(That's in a dir with Arbor checked out; I start the docker image with `podman run -v /path/to/arbor/and/above/script:Z  -ti manylinux2014_x86_64 /bin/bash`).

Let me know if that helps. Meanwhile I'll give your new commit a push to the `ciwheel` branch again. I think you could also do it yourself in your own Github repo btw, the trigger is the branchname `ciwheel`.

Helveg:
It seems from some vague hints in their docs and from `cibuildwheel --help` that `package_dir` needs to be set from the command line, let's try that then :) I will try to trigger things myself on my fork:

```
$ cibuildwheel --help
usage: cibuildwheel [-h] [--platform {auto,linux,macos,windows}] [--archs ARCHS] [--output-dir OUTPUT_DIR]
                    [--config-file CONFIG_FILE] [--print-build-identifiers] [--allow-empty] [--prerelease-pythons]
                    [package_dir]

Build wheels for all the platforms.

positional arguments:
  package_dir           Path to the package that you want wheels for. Must be a subdirectory of the working directory.
                        When set, the working directory is still considered the 'project' and is copied into the
                        Docker container on Linux. Default: the working directory.

optional arguments:
  -h, --help            show this help message and exit
  --platform {auto,linux,macos,windows}
                        Platform to build for. For "linux" you need docker running, on Mac or Linux. For "macos", you
                        need a Mac machine, and note that this script is going to automatically install MacPython on
                        your system, so don't run on your development machine. For "windows", you need to run in
                        Windows, and it will build and test for all versions of Python. Default: auto.
  --archs ARCHS         Comma-separated list of CPU architectures to build for. When set to 'auto', builds the
                        architectures natively supported on this machine. Set this option to build an architecture via
                        emulation, for example, using binfmt_misc and QEMU. Default: auto. Choices: auto, auto64,
                        auto32, native, all, x86_64, i686, aarch64, ppc64le, s390x, universal2, arm64, x86, AMD64
  --output-dir OUTPUT_DIR
                        Destination folder for the wheels.
  --config-file CONFIG_FILE
                        TOML config file for cibuildwheel. Defaults to pyproject.toml, but can be overridden with this
                        option.
  --print-build-identifiers
                        Print the build identifiers matched by the current invocation and exit.
  --allow-empty         Do not report an error code if the build does not match any wheels.
  --prerelease-pythons  Enable pre-release Python versions if available.
```

Helveg:
@brenthuisman with your script I get the following CMake error (to be expected because the `manylinux` images don't have `libpython.so`) 

```
CMake Error at /opt/_internal/cpython-3.7.12/lib/python3.7/site-packages/cmake/data/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Python3 (missing: Python3_INCLUDE_DIRS Python3_LIBRARIES
  Development Development.Module Development.Embed) (found suitable version
  "3.8.12", minimum required is "3.6.0")
Call Stack (most recent call first):
  /opt/_internal/cpython-3.7.12/lib/python3.7/site-packages/cmake/data/share/cmake-3.21/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  /opt/_internal/cpython-3.7.12/lib/python3.7/site-packages/cmake/data/share/cmake-3.21/Modules/FindPython/Support.cmake:3166 (find_package_handle_standard_args)
  /opt/_internal/cpython-3.7.12/lib/python3.7/site-packages/cmake/data/share/cmake-3.21/Modules/FindPython3.cmake:485 (include)
  CMakeLists.txt:301 (find_package)
```

brenthuisman:
I basically stumbled upon a working combination of moving parts for the wheel builds, documentation is not great. Some questions:

* [PEP517](https://www.python.org/dev/peps/pep-0517/) says:
> Here we define a new style of source tree based around the pyproject.toml file defined in PEP 518, extending the [build-system] table in that file with one additional key, build-backend.

By "builders", you mean "Build backend interfaces"?

* PEP517 implies `pyproject.toml`:
> If the pyproject.toml file is absent, or the build-backend key is missing, the source tree is not using this specification, and tools should revert to the legacy behaviour of running setup.py (either directly, or by implicitly invoking the setuptools.build_meta:__legacy__ backend).

In [PEP518](https://www.python.org/dev/peps/pep-0518/) we read:
> There are two issues with setup.cfg used by setuptools as a general format. One is that they are .ini files which have issues as mentioned in the configparser discussion above. The other is that the schema for that file has never been rigorously defined and thus it's unknown which format would be safe to use going forward without potentially confusing setuptools installations.

In the [PyPA packaging guide](https://packaging.python.org/tutorials/packaging-projects/) we read:
> Static metadata (setup.cfg): guaranteed to be the same every time. This is simpler, easier to read, and avoids many common errors, like encoding errors.
and
> setup.cfg is the configuration file for setuptools. It tells setuptools about your package (such as the name and version) as well as which code files to include. Eventually much of this configuration may be able to move to pyproject.toml.

It seems we need to keep writing package config details spread over 3 files, although at some point in the future, we may not have to. So far for clarity :cry: 

Anyway, I think you wanted to add the build backend, which I guess is a CMake build backend or some code steering CMake, which would then make Arbor current with PEP517, correct? And thereby more robust in the face of future changes to the many parts that make up Python packaging. OK. Arbor surely isn't the first project that uses CMake, so can we find some example packages?

PR #1702, opened 2021-10-06
Better handling of errors during dynamic catalogue loading

thorstenhater:
- Add platform specific handling of shared object loading
  - handle, dl_open, dl_get_symbol, dl_close
  - exceptions 
  - currently POSIX only (BSD, Mac and Linux)
    - These have well-behaved `const char*` error messages 
- All of the above may be depending on the platform, so wrap into a header `util/dl.hpp` and select correct platform by CPP. 
  - Currently Mac and Linux are detected, everything else throws a compile error.
  - BSD should be trivial, but we cannot test.
- Refine error handling when loading catalogues
  - Test for file existence first
  - catch platform dependent errors from `dl.hpp` and wrap them in `bad_catalogue_error`

### DIFF
--- a/arbor/arbexcept.cpp
+++ b/arbor/arbexcept.cpp
@@ -124,8 +124,16 @@ range_check_failure::range_check_failure(const std::string& whatstr, double valu
 {}
 
 file_not_found_error::file_not_found_error(const std::string &fn)
-    : arbor_exception(pprintf("Could not find file '{}'", fn)),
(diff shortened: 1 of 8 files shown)


thorstenhater:
> Might need to namespace the expected error: `arbor._arbor.FileNotFoundError`.

Issue is the semantics of `register_exception`. It will create a _new Py Exception in the module_.
We need however the global one.

PR #1710, opened 2021-10-10
Swap && and || operator precedence in modcc

noraabiakar:
The operator precedence for `&&` (land) and `||` (lor) was swapped in the lexer, resulting in incorrect parenthesis in the code generation step.


schmitts:
Do we want to add tests or will it all be void with arblang soon?

Issue #1535, opened 2021-05-19
`label_dict` does not behave like a `dict`


Helveg:
* Misses parts of the `dict` interface:
  * `fromkeys`
  * `items`
  * `keys`
  * `values`
  * `pop`
  * `popitem`
  * `get`
  * `setdefault`
  * `update`
  * `clear`
  * `__reversed__`
  * `__contains__`
* The `arbor.label_dict` function does not accept the same arguments as the `dict` constructor (e.g. doesn't know what to do with a `zip` object, it does not even accept another `label_dict` as arg)
* Converting a `label_dict` to a `dict` yields an empty dictionary, or an error:
```python
>>> labels = arbor.label_dict({'midpoint': '(location 0 0.5)'})
>>> print(labels, dict(labels))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: dictionary update sequence element #0 has length 8; 2 is required
```
```
(label_dict (region  "apic" (tag 4)) (region  "dend" (tag 3)) (region  "axon" (tag 2)) (region  "soma" (tag 1))) {}
```

PR #1711, opened 2021-10-12
Feature/label dict

apeyser:
label_dict does not behave like a dict #1535

So I added a few members to the python label_dict and functionality to label_dict_proxy.


apeyser:
Updates pybind11 to 2.8 as well.

PR #1716, opened 2021-10-14
Have option to set thread count to local maximum

### LINKED ISSUE #1692: Defaults: creating a context saturating local cpu
For new users, and users of the scientific Python ecosystem, having to set the number of threads manually for local use of Arbor is a bit surprising. In this issue I'd like to discuss how we can give users a better default while keeping the amount of magic down for more advanced use of Arbor.

Something like `arbor.context(use_all_threads=True)`, or `arbor.all_local_threads_context()` instead of `arbor.context()`. `arbor.single_cell_model` would also use that context or setting.

For reference, [there was an attempt](https://github.com/arbor-sim/arbor/pull/1415). I think the code there is perfectly valid.

brenthuisman:
* Pyarb specific.
* `proc_allocation_shim()` throws error if user sets threads to zero.
* `arbor.context` accepts `avail_threads`, which sets the number of threads to `arbenv::thread_concurrency()`
  * This introduces a dependency on arbenv for Pyarb.
* Associated docs and tests added.

Fixes #1692


brenthuisman:
Yes, I shall.

arbor/arbexcept.cpp
```diff
@@ -37,6 +37,11 @@ bad_global_property::bad_global_property(cell_kind kind):
     kind(kind)
 {}
 
+zero_thread_requested_error::zero_thread_requested_error(unsigned nbt):
+    arbor_exception(pprintf("threads must be a positive integer")),
```
noraabiakar, line 41: Since this PR seems to be restricted to pyarb changes, I think this exception would work better as a `pyarb_error` (defined in python/error.hpp)

python/context.cpp
```diff
@@ -48,13 +50,16 @@ struct proc_allocation_shim {
         gpu_id = py2optional<int>(gpu, "gpu_id must be None, or a non-negative integer", is_nonneg());
     };
 
-    void set_num_threads(int threads) {
-        pyarb::assert_throw([](int n) { return n>0; }(threads), "threads must be a positive integer");
-        num_threads = threads;
+    void set_num_threads(unsigned threads) {
+        if (0==threads) {
+            throw arb::zero_thread_requested_error(threads);
+        } else {
```
noraabiakar, line 56: `else` branch not needed: 
```
if (threads==0) {
    throw arb::zero_thread_requested_error(threads);
}
num_threads = threads;
```

test/unit/test_local_context.cpp
```diff
@@ -3,6 +3,7 @@
 #include "../gtest.h"
 #include "distributed_context.hpp"
 
+#include <arbor/context.hpp>
```
noraabiakar, line 6: Can be removed

python/test/unit/test_contexts.py
```diff
@@ -19,7 +19,7 @@ def test_default_allocation(self):
         self.assertEqual(alloc.threads, 1)
         self.assertEqual(alloc.gpu_id, None)
         self.assertFalse(alloc.has_gpu)
-
+
```
noraabiakar, line 22: Can we remove the extra indentation here?

python/context.cpp
```diff
@@ -127,6 +132,45 @@ void register_contexts(pybind11::module& m) {
...
+                    }
+                    if (auto c = py2optional<mpi_comm_shim>(mpi, mpi_err_str)) {
+                        return context_shim(arb::make_context(alloc, c->comm));
+                    }
+    #endif
+
+                    return context_shim(arb::make_context(alloc));
+                } else {
+                    throw pyarb_error(
+                        util::pprintf("{} is not a valid value.", threads));
```
noraabiakar, line 166: Maybe a slightly more descriptive error message such as: 
"Attempt to set threads to {}. The only valid thread options are a positive integer greater than 0, or 'avial_threads'."

python/context.cpp
```diff
@@ -127,6 +132,45 @@ void register_contexts(pybind11::module& m) {
...
+                }
+            }),
+            "threads"_a, "gpu_id"_a=pybind11::none(), "mpi"_a=pybind11::none(),
+            "Construct a distributed context with arguments:\n"
+            "  threads: The number of threads available locally for execution, 1 by default.\n"
```
noraabiakar, line 171: "threads: a string option describing the number of threads. Only "avail_threads" is supported."
Or something similar.

python/context.cpp
```diff
@@ -127,6 +132,45 @@ void register_contexts(pybind11::module& m) {
             "  threads: The number of threads available locally for execution, 1 by default.\n"
             "  gpu_id:  The identifier of the GPU to use, None by default. Only available if arbor.__config__['gpu']==True.\n"
             "  mpi:     The MPI communicator, None by default. Only available if arbor.__config__['mpi']==True.\n")
+        .def(pybind11::init(
```
noraabiakar, line 135: There's some duplication here which I think we can avoid by extracting the common code into a function: 
```
context_shim create_context(unsigned threads, pybind11::object gpu, pybind11::object mpi) {
    const char* gpu_err_str = "gpu_id must be None, or a non-negative integer";

#ifndef ARB_GPU_ENABLED
    if (!gpu.is_none()) {
        throw pyarb_error("Attempt to set an GPU communicator but Arbor is not configured with GPU support.");
    }
#endif

    auto gpu_id = py2optional<int>(gpu, gpu_err_str, is_nonneg());
    arb::proc_allocation alloc(threads, gpu_id.value_or(-1));

#ifndef ARB_MPI_ENABLED
    if (!mpi.is_none()) {
        throw pyarb_error("Attempt to set an MPI communicator but Arbor is not configured with MPI support.");
    }
#endif

    const char* mpi_err_str = "mpi must be None, or an MPI communicator";
    if (can_convert_to_mpi_comm(mpi)) {
        return context_shim(arb::make_context(alloc, convert_to_mpi_comm(mpi)));
    }
    if (auto c = py2optional<mpi_comm_shim>(mpi, mpi_err_str)) {
        return context_shim(arb::make_context(alloc, c->comm));
    }
    return context_shim(arb::make_context(alloc));
}
```

Then, we can define the init functions like: 
```
.def(pybind11::init(
    [](unsigned threads, pybind11::object gpu, pybind11::object mpi){
        return create_context(threads, gpu, mpi);
    }), ... )
.def(pybind11::init(
    [](std::string threads, pybind11::object gpu, pybind11::object mpi){
        if ("avail_threads" == threads) {
            return create_context(arbenv::thread_concurrency(), gpu, mpi);
        }
        throw pyarb_error(util::pprintf("{} is not a valid value.", threads));
    }), ... )
```

python/context.cpp
```diff
@@ -48,13 +50,16 @@ struct proc_allocation_shim {
         gpu_id = py2optional<int>(gpu, "gpu_id must be None, or a non-negative integer", is_nonneg());
     };
 
-    void set_num_threads(int threads) {
-        pyarb::assert_throw([](int n) { return n>0; }(threads), "threads must be a positive integer");
-        num_threads = threads;
+    void set_num_threads(unsigned threads) {
```
noraabiakar, line 53: This function needs to be documented.

doc/python/hardware.rst
```diff
@@ -88,6 +88,7 @@ The Python wrapper provides an API for:
     .. attribute:: threads
 
         The number of CPU threads available, 1 by default.
+        Passing ``"avail_threads"`` (as string) will query and use the maximum number of threads the system makes available.
```
noraabiakar, line 91: This is actually not implemented.

noraabiakar (changes requested):
Looks good! I just have a few suggestions.

arbor/arbexcept.cpp
brenthuisman: I see your point, but I'm not entirely happy with 1. this split anyway and 2. the particular contents of `python/error.hpp`. First of all, `python/pyarb.cpp` already contained one error translator `file_not_found_error`, which is a much clearer pattern: you register errors raised in C++ code, and translate them as you wish. Since the code of this particular change is C++ (and really all our code), even though restricted to being raised within Pyarb for the moment, it makes more sense to me to follow the same pattern. The `pyarb_error` in `python/error.hpp` should really be removed, it relies on implicit translation by `pybind11` and it's uses broken up into more meaningful exception classes.

If you agree, I can do that in a follow up PR. What do you think?

python/context.cpp
brenthuisman: You mean `set_num_threads()`? You mean, a note in the docs `arbor.proc_allocation` that the minimum number is 1?

doc/python/hardware.rst
brenthuisman: Isn't that what this PR is about? Locally passing `"avail_threads"` gives me the expected number of threads.

python/context.cpp
brenthuisman: Thanks, I had been breaking my head over not being able to overload these constructors. You took the last part of `create_context()` outside of the `#else` block though, is that intentional? `can_convert_to_mpi_comm` and `mpi_comm_shim` are not available if not `ARB_MPI_ENABLED`, so it can't build. Maybe I misunderstand?

arbor/arbexcept.cpp
thorstenhater: The use of translation was a test and especially aimed at FileNotFound, because users expect FNF instead of RuntimeError. But yes, now that this is in for a while and caused no issues, we should port all our errors.

noraabiakar: I see what you're saying, but I don't think the solution is to move all of our custom errors to the arbor library. The errors that are used only in the `pyarb` library should be defined there, either in `python/error.hpp` or somewhere else. 
I agree that we should translate our exceptions into python exceptions whenever possible, but that has nothing to do with where the exceptions are defined. 
I think it's better to follow the current convention of using `pyarb_error` instead of `arbor_exception` in this case.

python/context.cpp
noraabiakar: I mean that the class is not fully documented: https://docs.arbor-sim.org/en/latest/python/hardware.html#arbor.proc_allocation 
I'm now noticing it's not just `set_num_threads`, but also `get_num_threads` and `get_gpu_id`. 

Also, you've made changes to `context`, so the documentation for that too needs updating: https://docs.arbor-sim.org/en/latest/python/hardware.html#arbor.context

python/context.cpp
noraabiakar: Yes, you're right. They need to stay in the `#else` block, my bad!

arbor/arbexcept.cpp
brenthuisman: There are already custom Arbor errors in `arbor/arbexcept.*pp` , right? They are not specific to a part of the API, else they would not live in one file. I don't think Python needs to be an exception (ha), other parts of Arbor may very well raise a `zero_thread_requested_error` at some point. I thought that having all errors in one place is so you can quickly find out if the exception you are raising already fits an existing class. 

Now, if the exception was very specific to Python I would agree with you, but this is about user input, which is not specific to Python.

I prefer to start (well, continue!) writing new exceptions in this pattern, and tackle existing `pyarb_error`s in the near future.

noraabiakar: We're looking at 2 libraries
1. `arbor` is the core arbor C++ library and has it's own exceptions in `arbor/arbexcept.*pp`.
2. `pyarb` is the python wrapper library, it needs `arbor` as a dependency but they are separate libraries. It has it's own exceptions in `python/error.*pp`.

An error that is only used in `pyarb` doesn't need to live in `arbor`. 
Separating the errors keeps the code clear, you can look at the type of the error if it's `pyarb_error` you know it's been raised from `pyarb`, if it's `arbor_exception` it's been raised from `arbor`. 
There's no need to accumulate all errors in the same file in the `arbor` lib. We can replicate what we have in the `arbor` lib in the `pyarb` lib as long as we keep that separation between the libraries clear.

python/context.cpp
brenthuisman: The Python class documentation is on lines 106-123, not on the shim. The Python binding combines `set_num_threads` and `get_num_threads` into `threads`, I've update that and `doc/python/hardware.rst` to reflect 1 is the minimum. Context docs is also corrected (whoops!)

doc/python/hardware.rst
brenthuisman: Yes!

arbor/arbexcept.cpp
brenthuisman: It is not a Python error, but an input error. It just happens to be only raised by the Python binding for the moment. In general, I think they should overlap as much as possible, not as little as possible.

I think "Python errors" should be seen as exceptions raised by Python machinery, e.g. C++ is unlikely to ever need a `TabError`.

noraabiakar: I think this is a small issue and I don't want to keep going around in circles. The current setup is to use `pyarb_error` in the `pyarb` library and I think this error belong there. It's up to you to change or not.

thorstenhater (approved):
👍

noraabiakar:
bors try

noraabiakar (approved):

PR #1735, opened 2021-10-26
Generate correct simd code for reading `peer_index` and `v_peer`

### LINKED ISSUE #1734: Incorrect indexing of `v_peer` in the generated vectorized code for gap-junction mechanisms.
The vectorized code of gap_junction mechanisms contains a bug. 
The `peer_index` (used to obtain the index of the peer CV of a gap-junction mechanism), as well as `v_peer` which is indexed by `peer_index`, are being read according to the vectorization `constraint_category` of `node_index` and `v` respectively.
This is not correct, the `constraint_category` of `peer_index` is assumed to be `index_constraint::none`. No vectorization optimizations should be applied.

noraabiakar:
Use `constraint_category` only when reading/writing data sources indexed by `node_index`. Every other indexed data source assumes that the underlying index belongs to the category `index_constraint::none`. 
Fixes #1734.


bcumming (approved):
Looks good.

Issue #1731, opened 2021-10-20
Formal dependency version management


brenthuisman:
## Goal

1. Avoid updating dependencies without having checked that they match (e.g. git submodule hash and Spack file version tag), that they are in fact available (git submodules are updated more frequently than Spack files) and taking into account an important runtime environment of Arbor: [the EBrains Jupyterlab](https://lab.ebrains.eu) (e.g. it currently has Python 3.6, so Arbor can't drop compatibility for it until that changes).

2. Ensure that Spack builds of releases versions of Arbor are reproducible, by which I mean that building any Arbor version other than `master`, we fix the (range of) versions of dependencies.

## Rationale

1. We've [just established](https://github.com/arbor-sim/arbor/issues/1717#issuecomment-947552886) that a dependency had a git submodule commit hash that did not match the version in the Spack file. A solution without reverting is not trivial: dependencies in Spack files can only rely on stated versions in their respective Spack file, which may be trailing release tags in the git repo, and then lead Spack to resolve to the master branch which is present in many Spack files, which led to confusion. It's important to avoid these events and that building Arbor does what it says on the tin ;)

2. a. As the Spack ecosystem ages, dep requirements of older versions is going to significantly diverge from current dep version requirements. We don't test older versions with current dependencies as a matter of course, and I'm not sure we are going to. Freezing older versions at specific dependency versions may be the only practical way we can insure those older versions remain usable/behave predictably/have been actually tested. In addition: we're not going to be in control of all build envs (EBrains). I'm not fond of relying on users/builders speccing the env out in excruciating detail, because I'm not sure it'll be done correctly (it's hard for me!); freezing the deps of older versions seems like a better idea to ensure spack install my_package@some_old_version results in something stable over time. Concrete use-case: processing pipelines that researchers store and re-use perhaps for decades. I'm not sure what's optimal here yet either, so any opinions on how to keep such pipelines working over longer timespans is appreciated!
  b. Many packages define a master version, which will get you their current master branch. Together with the fact that Spack file updates trail project releases by some time, this may mean in practice that specific versions may be pulling in unspecific dependency versions. That I am not fond of. When building our package's master branch I'm OK with pulling in deps master branches, but not for any version release.

## Discussion

Even for Spack CI of the master branch, the problem under 2.b can be a bit of a headache, because due to the update lag, Spack CI may be testing other versions of dependencies than regular CI. I propose to not address this right now, but merely bring it to your attention. In case someone feels strongly that we must address this, please do.

[...]

PR #1741, opened 2021-10-27
Weekly cron job checking for submodule updates

brenthuisman:
* Create a script that generates `diff.log` if git submodules have tags more recent that the commit they're currently at.
* A Github Action runs this script monthly.
* _If_ updates found, Github Action creates an Issue like so: https://github.com/brenthuisman/arbor/issues/9
* Actually updating the git submodule and Spack file are still manual

First step in addressing #1731


scripts/submodule-diff.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
```
schmitts, line 2: Consider adding `set -euxo pipefail` (https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail)

scripts/submodule-diff.sh
```diff
@@ -0,0 +1,11 @@
...
+git submodule foreach 'git describe HEAD --tags' > pre.log
+git submodule foreach 'git fetch'
+git submodule foreach 'git describe `git log --branches -1 --pretty=format:"%H"` --tags --abbrev=0' > post.log
+if diff pre.log post.log ; then
+    echo "No submodule updates found"
+else
+    diff pre.log post.log -U 10000 > diff.log
+fi
+rm pre.log && rm post.log
```
schmitts, line 11: I'd delete unconditionally

scripts/submodule-diff.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+git submodule foreach 'git describe HEAD --tags' > pre.log
+git submodule foreach 'git fetch'
+git submodule foreach 'git describe `git log --branches -1 --pretty=format:"%H"` --tags --abbrev=0' > post.log
```
schmitts, line 5: default to `$()` instead of backticks (doesn't matter much here, because no nesting, but it should be the preferred style in any case)

scripts/submodule-diff.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+git submodule foreach 'git describe HEAD --tags' > pre.log
```
schmitts, line 3: maybe do `| tee pre.log` here and below to ease debugging?

.gitmodules
```diff
@@ -1,9 +1,12 @@
 [submodule "google-benchmark"]
 	path = ext/google-benchmark
 	url = https://github.com/google/benchmark
+    branch = main
```
schmitts, line 4: whitespace error here and below

schmitts (commented):
Some minor comments inline

scripts/submodule-diff.sh
brenthuisman: I think for this use-case it isn't what we want, because it would exit the script and Github Action early and we would not get output for any submodules that were not checked yet.

scripts/submodule-diff.sh
brenthuisman: :+1:

scripts/submodule-diff.sh
brenthuisman: My shell foo isn't great, is that style compatible across shells? I like backticks because of readability, and have not run into problems with them, but I don't use the shell much.

On line 5, there is some nesting going on, in any case. Although technically maybe it isn't done by the shell?

scripts/submodule-diff.sh
brenthuisman: Also for `post.log` then?

.gitmodules
brenthuisman: What's the error? This file is generated, and I would assume `git` knows how ;)

scripts/submodule-diff.sh
schmitts: I would still put it. We want to exit whenever a command fails. Consider 4 fails (silently). Then we would get an empty diff.

scripts/submodule-diff.sh
```diff
@@ -0,0 +1,12 @@
...
+
+git submodule foreach 'git describe HEAD --tags' > current_state_of_git_submodules_in_arbor_repo.log | tee current_state_of_git_submodules_in_arbor_repo.log
+git submodule foreach 'git fetch'
+git submodule foreach 'git describe `git log --branches -1 --pretty=format:"%H"` --tags --abbrev=0' > upstream_state_of_git_submodules.log | tee upstream_state_of_git_submodules.log
+if diff current_state_of_git_submodules_in_arbor_repo.log upstream_state_of_git_submodules.log ; then
+    echo "No submodule updates found"
+else
+    diff current_state_of_git_submodules_in_arbor_repo.log upstream_state_of_git_submodules.log -U 10000 > diff.log
+fi
+rm current_state_of_git_submodules_in_arbor_repo.log
```
schmitts, line 11: No need to delete the files, but idc so much.

scripts/submodule-diff.sh
schmitts: add a small summary of what the script does.

scripts/submodule-diff.sh
brenthuisman: It's to prevent clutter in case you use it locally.

scripts/submodule-diff.sh
brenthuisman: `diff` returns nonzero upon any difference. So no, here we don't want to exit upon any 'fail', we'd ever only get the difference reported for a single module.

I'll add a summary!

scripts/submodule-diff.sh
schmitts: You cannot nest with backticks only. $() is preferred and POSIX standard: https://stackoverflow.com/a/9406350

scripts/submodule-diff.sh
schmitts: You're right (Cf. http://mywiki.wooledge.org/BashFAQ/105). But what about error handling, e.g., if `git submodule foreach 'git fetch'` fails?

brenthuisman: When I disable network and run the script, the `fetch` loop terminates early (after the first remote being inaccessible) and the diff is generated based on unupdated module remotes. That is not terribly useful, but also not terribly bad, assuming that by running this weekly, such a failure every now and then will do no more harm than postpone us being notified a week late of updates.

Is that acceptable?

schmitts (approved):
LGTM

scripts/submodule-diff.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# Check any git submodule remotes for updates, and print difference with current Arbor repo state to `diff.log`
+
+git submodule foreach 'git describe HEAD --tags' > current_state_of_git_submodules_in_arbor_repo.log | tee current_state_of_git_submodules_in_arbor_repo.log
```
noraabiakar, line 4: Why do we need both `| tee` and `>` ?

noraabiakar (commented):
If there are differences between local and upstream submodules, and the automatically generated issue is not addressed, does that mean that every week an identical issue will be opened?

scripts/submodule-diff.sh
brenthuisman: To both display the output on stdout and have it in the file.

brenthuisman:
> If there are differences between local and upstream submodules, and the automatically generated issue is not addressed, does that mean that every week an identical issue will be opened?

Correct.

scripts/submodule-diff.sh
noraabiakar: I just tried it out locally and it didn't display the output on stdout. But `git submodule foreach 'git describe HEAD --tags' | tee current_state_of_git_submodules_in_arbor_repo.log` does what you describe.

noraabiakar:
I think a new issue every week might be too much, what do you think?

scripts/submodule-diff.sh
brenthuisman: Whoops :)

brenthuisman:
Weekly is perhaps too frequent, I've changed it to a monthly job.

noraabiakar (approved):

PR #1747, opened 2021-11-04
Update/fix tinyopt README.md

halfflat:
tinyopt README.md had an example which was not updated to the 1.0 API. Fixed upstream, and this is a copy.


schmitts (approved):

noraabiakar (approved):
Thanks!

PR #1748, opened 2021-11-08
Test separately built catalogues

bcumming:
* add a test that tests that a catalogue built separately can be loaded via the Python interface
* further simplification of dynamic library support
  * move all platform-specific code into the cpp implementation and out of header.


arbor/util/dylib.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include <arbor/arbexcept.hpp>
+
+#include "util/dylib.hpp"
+#include "util/strprintf.hpp"
+
+namespace arb {
+namespace util {
+
+struct dl_handle {
```
thorstenhater, line 14: This is redundant now, it was meant as an opaque handle, but if it's not exposed, it can go.

scripts/test-catalogue.py
```diff
@@ -0,0 +1,24 @@
...
+args = P.parse_args()
+catname = args.catname
+
+print(catname)
+
+if not os.path.isfile(catname):
+    print('ERROR: unable to open catalogue file')
+    sys.exit(1)
+
+print(ctypes.CDLL(catname))
```
thorstenhater, line 22: This is also redundant: Debugging left-overs.

arbor/util/dylib.cpp
```diff
@@ -0,0 +1,59 @@
...
+    return {result};
+}
+
+void dl_close(dl_handle& handle) {
+    dlclose(handle.dl);
+    handle.dl = nullptr;
+}
+
+namespace impl{
+void* dl_get_symbol(const std::string& fn, const std::string& symbol) {
```
thorstenhater, line 42: What's the reason for the change from 
```cxx
void* dl_get_symbol(void* handel, const std::string& symbol)
```
? Now, fetching a symbol will go through `dlopen` everytime.

scripts/test-catalogue.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
+
+import argparse
+import ctypes
```
thorstenhater, line 4: This can also go.

thorstenhater (changes requested):

arbor/util/dylib.cpp
bcumming: The library was only being used to return a symbol from a library, for which the `handle` type was not required.

This way the handle type is not required in the public interface. It could be reintroduced later, if we ever have more elaborate use cases for dynamic libraries.

noraabiakar:
This PR updates pybind11, can we add that to the PR message with the version info?

bcumming:
@noraabiakar  I can't understand why the PR shows pybind11 being updated: the pybind11 commit in the branch and in master are exactly the same, and the changes shown in the "Files changed" tab show pybind11 being updated to that same commit. 

So I don't think anything is actually being changed. It looks like GitHub is getting confused by the slightly convoluted history of the commits in this PR.

arbor/util/dylib.cpp
thorstenhater: Good point.

arbor/util/dylib.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+// Find and return a symbol from a dynamic library with filename.
+// Throws dl_error on error.
+template<typename T>
+T dl_get_symbol(const std::string& filename, const std::string& symbol) {
+    return reinterpret_cast<T>(impl::dl_get_symbol(filename, symbol));
+}
+
+} // namespace util
+} // namespace arbo
```
thorstenhater, line 26: `arbo`?

thorstenhater (changes requested):
Apart from a tiny typo and the pybind divergence everything's well.

thorstenhater (approved):
pybind11 malaise clarified offline, all good.

PR #1750, opened 2021-11-09
Add python gap junction example

kanzl:
Add gap junction example analog to existing C++ gj example.

Creates chains of cells (soma + one dendrite) that connect via gap junctions. These chains are linked together with synapses.


noraabiakar (approved):
Looks good! thanks a lot.

PR #1755, opened 2021-11-12
modcc: exit solvers early if errors encountered.

### LINKED ISSUE #1722: Solver 'sparse': error for constant value in ODE
The error "internal compiler error: zero row in matrix solver" occurs if an ODE of the type `dp/dt = c`, with `c` constant, is to be solved. No issues occur for `dp/dt = -p` or `dp/dt = 0`, for instance. The following mechanism code serves to reproduce the error in arbor v0.5.2:
```
NEURON {
    POINT_PROCESS test
}

STATE {
    z
    p
}

INITIAL {
    z = 0
    p = 0
}

BREAKPOINT {
    SOLVE state METHOD cnexp
}

DERIVATIVE state {

    p' = 1
    z' = p
}
```

noraabiakar:
The `finalize` stage of the solvers can throw unhelpful exceptions during the Gauss-Jordan elimination phase. The real, useful  error messages are usually recorded earlier but not propagated to the user. To report back the real errors to the user, don't perform the final stage of the solvers if any errors have already been encountered. 

Fixes #1722.

### DIFF
--- a/modcc/solvers.cpp
+++ b/modcc/solvers.cpp
@@ -448,6 +448,7 @@ void SparseSolverVisitor::visit(ConserveExpression *e) {
 }
 
 void SparseSolverVisitor::finalize() {
+    if (has_error()) return;
 
     if (solve_variant_ == solverVariant::steadystate && !conserve_) {
         error({"Conserve statement(s) missing in steady-state solver", {}});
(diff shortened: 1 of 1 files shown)


bcumming (approved):
lgtm

PR #1761, opened 2021-11-15
Fix broken links in `probe_sample.rst`

Helveg:
There was no module directive on the page and most `class` and `func` directives were broken.


Helveg:
From a local build of the doc it seems adding the `module` directive helped, but `arbor.probe` is still broken, possibly undocumented?

bcumming (approved):

PR #1773, opened 2021-11-22
simd: sve needs runtime vector length

boeschf:
variable runtime based array size for copy
- use static thread_local vector to minimize performance penalty
- set compile time constants to 0 for SVE

### DIFF
--- a/arbor/include/arbor/simd/simd.hpp
+++ b/arbor/include/arbor/simd/simd.hpp
@@ -175,7 +175,7 @@ namespace detail {
 
         indirect_indexed_expression& operator=(V s) {
             static thread_local std::vector<typename simd_traits<ImplIndex>::scalar_type> idx(width);
-            ids.resize(width);
+            idx.resize(width);
             ImplIndex::copy_to(index.value_, idx.data());
             for (unsigned i = 0; i < width; ++i) {
                 p[idx[i]] = s;
--- a/arbor/include/arbor/simd/simd.hpp
+++ b/arbor/include/arbor/simd/simd.hpp
@@ -174,7 +174,7 @@ namespace detail {
         {}
 
         indirect_indexed_expression& operator=(V s) {
-            typename simd_traits<ImplIndex>::scalar_type idx[width];
+            typename simd_traits<ImplIndex>::scalar_type idx[simd_traits<ImplIndex>::width];
             ImplIndex::copy_to(index.value_, idx);
             for (unsigned i = 0; i < width; ++i) {
                 p[idx[i]] = s;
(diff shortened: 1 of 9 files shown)


boeschf:
bors try

noraabiakar:
bors try

noraabiakar:
bors try

noraabiakar:
This seems to be failing? https://gitlab.com/cscs-ci/arbor-sim/arbor/-/jobs/1807473999

boeschf:
bors try

boeschf:
bors try

noraabiakar:
CI on piz daint through gitlab is down. I spoke to @haampie about it and hopefully things will be fixed soon.

boeschf:
bors try

arbor/include/arbor/simd/simd.hpp
```diff
@@ -174,8 +175,9 @@ namespace detail {
         {}
 
         indirect_indexed_expression& operator=(V s) {
-            typename simd_traits<ImplIndex>::scalar_type idx[simd_traits<ImplIndex>::width];
-            ImplIndex::copy_to(index.value_, idx);
+            static thread_local std::vector<typename simd_traits<ImplIndex>::scalar_type> idx(width);
```
bcumming, line 178: Why not make this a stack variable?
`thread_local` variables have to be `static`, so the `static` is redundant?

bcumming (commented):
I have a question about why we use a thread local variable

arbor/include/arbor/simd/simd.hpp
boeschf: yes `thread_local` is static in a block scope by default, so it is redundant here (I can remove it but l usually leave it for clarity). A couple of reasons why it can't be on the stack:
- `width` is a runtime argument and may change -> heap allocation is required
- `static` storage duration helps to reduce the number of allocations (you pay the cost only once if the width doesn't change or shrinks)
- `thread_local` is needed because I assumed this function needs to be thread safe. I am not entirely sure whether this is actually the case or whether each thread manages its own instance.

bcumming: Sorry for the belated reply on this.

A `thread_local std::vector` isn't appropriate here because:
* every call to this function will require checking a global variable to test whether this is the first time this thread has called the function, to initialize `idx` if need be. See the amount of boilerplate here: https://godbolt.org/z/z9rPWf58K
* this should be possible with stack storage!

Isn't it possible to allocate a stack array with the a compile time width that corresponds to the largest possible `width` value?

Issue #1369, opened 2021-02-12
Documentation: tutorial structure, howtos


brenthuisman:
## Proposal

First, to bring a bit more structure to the tutorials, here is what Nora and I came up with:

### Tutorial

Have a few topics which build up in complexity. Reduce the introduction of concepts already introduced, unless it can be done very briefly, otherwise link to the (section in) the tutorial where it was introduced. Make the "Concepts covered in this example:" box more focussed, but do not *implicitly* rely on previously intro'ed concepts.

Topics:
* cell
    * simple cell
    * complex cell
    * cell from a database? (allen)
* network
    * simple network (take a simple cell already intro'd)
      * https://github.com/arbor-sim/arbor/pull/1313
    * complex network (mix multiple types of cells, take ones intro'd)
    * populations of cells
* probe
    * use C++ probe-demo for inspiration
    * how to gap junction?
    * how to probe everything on a cell?
* IO
    * An example of NeuroML input
    * SWC modes
* Performance
    * Run with mpi, on gpu. Can be brief.
    * built in profiling/metering tools
    * a custom domain decomposition
    * Perhaps tutorial where we use profiling to find a bottleneck, which we then solve with a customized domain decomp or hints?
      * https://github.com/brenthuisman/arbor/tree/ex_hardware

### Howtos/Examples

We couldn't come up with a good threshold for when something stops being a tutorial and starts being a howto. Here are a few proposed definitions:

* A howto is everything that doesnt need/deserve documentation.
* Dont need to go into every detail in a howto
* A howto is not a contrived example but a possible real world one
* A tutorial covers Arbor concepts, a howto does not.

For the time being, we propose to solve it by advertising a link to `/examples` (on GH) front and center on the Tutorial overview page. A howto will not have a page in the docs, but will have a link and short description on the Tutorial overview. They will be referred to as examples. I will exercise editorial prerogative to promote an example to tutorial. It may not be very important to make the distinction sharper than this, but, if you feel it needs to be, and you have a good criterion, then please reply ;)

Proposed Examples (insofar they don't already
* examples
    * An example that shows how to read a json description from Allen DB and map it onto Arbor
      * https://github.com/brenthuisman/arbor/tree/ex_allen
      * Describing the provenance of the data does require some detailing, as well as setting up a complex decor. Maybe it can be our most complex cell tutorial?
    * run sth from a online database. A non-Allen example?
    * An example that shows how to create a single cell model in Arbor-GUI
[...]

PR #1781, opened 2021-11-30
Allen tutorial

brenthuisman:
* Adds tutorial demonstrating working with the Allen database.

* Addresses #1369


python/example/single_cell_allen.py
```diff
@@ -0,0 +1,174 @@
...
+import matplotlib.patches as patches
+import seaborn as sns
+import numpy as np
+import pandas as pd
+
+# Check arbor is loaded fine and print some config diagnostics
+print(A.config())
+
+# cable parameters
+class parameters:
```
thorstenhater, line 16: If you are worried about versions here, the pre-3.7 way of things is using a NamedTuple. Otherwise, a plain tuple does the trick.

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,273 @@
...
+The morphology
+--------------
+
+:ref:`In an earlier tutorial <tutorialsinglecellswc-cell>` we've seen how an ``swc`` file can be loaded **(1)**
+and how the labels can be set **(2)**:
+
+.. code-block:: python
+
+  # (1) Load the cell morphology.
+  morphology = arbor.load_swc_allen('single_cell_allen.swc', no_gaps=False)
```
thorstenhater, line 49: Outdated.

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,273 @@
...
+    the ion according to the nernst equation
+
+Step **(5)** overrides the global properties for all *regions* for which the fit parameters file specifies adapted
+values. Regional properties are :func:`painted `arbor.paint`, and are painted over the defaults.
+
+Step **(6)** sets the regional reversal potentials.
+
+Step **(7)** assigns the regional mechanisms.
+
+Now that the electrodynamics are all set up, let's move on to the experimental setup.
```
thorstenhater, line 193: electro-phys ;)

python/example/single_cell_allen.py
```diff
@@ -0,0 +1,174 @@
...
+
+# (15) Extract data
+data, _ = sim.samples(hdl)[0]
+ts = data[:, 0]
+us = data[:, 1]
+spikes = np.array([t for (_, t) in sim.spikes()])
+
+# (16) Plot
+fg, ax = plt.subplots()
+c_arbor, c_spike, *_ = sns.color_palette()
```
thorstenhater, line 166: Should add the stimulus as well.

thorstenhater (changes requested):
Some early comments. A general observation after seeing more of these examples:
Next release planning should be focussing on ergonomics.

python/example/single_cell_allen.py
brenthuisman: Namedtuples are not mutable, so not a great fit with the way they're preallocated in the dict. I prefer to use a class in order to have names for the fields to keep the code readable. I didn't notice a slowdown w.r.t the dataclass, if that is your concern.

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,273 @@
...
+:ref:`In an earlier tutorial <tutorialsinglecellswc-cell>` we've seen how an ``swc`` file can be loaded **(1)**
+and how the labels can be set **(2)**:
+
+.. code-block:: python
+
+  # (1) Load the cell morphology.
+  morphology = arbor.load_swc_allen('single_cell_allen.swc', no_gaps=False)
+  # (2) Label the region tags found in the swc with the names used in the parameter fit file.
+  # In addition, label the midpoint of the soma.
+  labels = arbor.label_dict({
```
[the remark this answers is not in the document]
brenthuisman: Actually, I've not updated the `.rst` yet :)

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,170 @@
...
+   3. Load a parameter fit file and apply it to a :class:`arbor.decor`.
+   4. Building a :class:`arbor.cable_cell` representative of the cell in the model.
+   5. Building a :class:`arbor.recipe` reflective of the cell stimulation in the model.
+   6. Running a simulation and visualising the results.
+
+Obtaining the model
+-------------------
+
+We need a model for which morphological data is available. We'll take
+`a cell from the mouse's visual cortex <http://celltypes.brain-map.org/experiment/electrophysiology/488683425>`_.
```
thorstenhater, line 24: No HTTPS :/ Also, to isolate us against mishaps we should cache the downloaded file, eg in arbor-contrib.

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,170 @@
...
+
+In the "Biophysical - all active" model (zip) file you'll find:
+
+* a directory with mechanisms (modfiles). For your convenience, Arbor already comes with the
+  :func:`arbor.allen_catalogue`, containing mechanisms used by the Allen Brain Atlas.
+* an ``swc`` file describing the morphology of the cell.
+* ``fit_parameters.json`` describing a set of optimized mechanism parameters. This information will need to be
+  loaded into the Arbor simulation.
+
+We will replicate the "Sweep 35" experiment, which applies a current of 150 nA for a duration of 1 s.
```
thorstenhater, line 38: Style: put quantities in some markup.

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,170 @@
...
+   :dedent:
+   :lines: 80-114
+
+Step **(4)** creates an empty :class:`arbor.decor`.
+
+Step **(5)** assigns global (cell-wide) properties using :func:`arbor.decor.set_property`. In addition, initial
+internal and external calcium concentrations are set, and configured to be mediated by the Nernst equation.
+
+.. note::
+    Setting the calcium reversal potential to be mediated by the Nernst equation has to be done manually, in order to mirror
```
thorstenhater, line 95: mediated -> determined

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,170 @@
...
+    `an implicit Neuron behavior <https://neuron.yale.edu/neuron/static/new_doc/modelspec/programmatic/ions.html>`_,
+    for which the fit parameters were obtained. This behavior can be stated as the following rule:
+
+    If the internal or external concentration of an ion is written, and its reversal potential is read but not
+    written, then the nernst equation is used continuously during the simulation to update the reversal potential of
```
thorstenhater, line 100: Nernst

doc/tutorial/single_cell_allen.rst
thorstenhater: This is still physiology, not electro-dynamics

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,170 @@
...
+Let's look at the result! In step **(16)** we first load the reference generated with Neuron and the AllenSDK.
+Then, we extract Arbor's output, accessible after the simulation ran at
+:class:`arbor.single_cell_model.traces`. Then, we plot them, together with the :class:`arbor.single_cell_model.spikes` in step **(17)**.
+
+.. literalinclude:: ../../python/example/single_cell_allen.py
+   :language: python
+   :dedent:
+   :lines: 129-
+
+.. figure:: single_cell_allen_result.svg
```
thorstenhater, line 149: Add comment: green dots = Spikes

python/example/single_cell_allen.py
```diff
@@ -0,0 +1,139 @@
...
+import matplotlib.pyplot as plt
+
+# (3) A function that parses the Allen parameter fit file into components for an arbor.decor
+# NB. Needs to be adjusted when using a different model
+def load_allen_fit(fit):
+    with open(fit) as fd:
+        fit = json.load(fd)
+
+    # cable parameters convenience class
+    class parameters:
```
thorstenhater, line 17: This is not idiomatic Python. You are setting members to a _class_ not to an _instance_?
```
In [19]: class A:
    ...:     a = 42
    ...:
In [22]: o = A()
In [23]: o.a
Out[23]: 42
In [24]: p = A()
In [25]: p.a
Out[25]: 42
In [26]: A.a = 23
In [27]: p.a
Out[27]: 23
In [28]: o.a
Out[28]: 23
```

It works here though, for reasons known only to Python... Still I wouldn't want to advertise this
as proper. Either use `dataclass` like my original code does or namedtuple, see 
https://docs.python.org/3/library/collections.html#namedtuple-factory-function-for-tuples-with-named-fields
and examples therein.

python/example/single_cell_allen.py
```diff
@@ -0,0 +1,139 @@
...
+                    raise Exception(f"Unknown key: {name}")
+                continue
+            else:
+                raise Exception(f"Illegal combination {mech} {name}")
+        mechs[(region, mech)][name] = value
+
+    regs = [(r, vs) for r, vs in param.items()]
+    mechs = [(r, m, vs) for (r, m), vs in mechs.items()]
+
+    default = parameters()
```
thorstenhater, line 54: also this is much simpler using DC/NT.

python/example/single_cell_allen.py
```diff
@@ -0,0 +1,139 @@
...
+                                'dend': '(tag 3)',
+                                'apic': '(tag 4)',
+                                'midpoint': '(location 0 0.5)'})
+
+    # (3) A function that parses the Allen parameter fit file into components for an arbor.decor
+    dflt, regions, ions, mechanisms = load_allen_fit(fit)
+
+    # (4) Instantiate an empty decor.
+    decor = arbor.decor()
+    # (5) assign global electro-physical parameters
```
thorstenhater, line 85: physiological.

python/example/single_cell_allen.py
```diff
@@ -0,0 +1,139 @@
...
+        region = kv['section']
+        for k, v in kv.items():
+            if k == 'section':
+                continue
+            ion = k[1:]
+            ions.append((region, ion, float(v)))
+
+    return default, regs, ions, mechs
+
+def make_cell(swc, fit, epas_is_param=True):
```
thorstenhater, line 70: This parameter `epas...` should not be here, it's a bridge between Arbor versions I used to test multiple versions.

python/example/single_cell_allen.py
```diff
@@ -0,0 +1,139 @@
...
+model.probe('voltage', '"midpoint"', frequency=200)
+
+# (14) Install the Allen mechanism catalogue.
+model.properties.catalogue.extend(arbor.allen_catalogue(), "")
+
+# (15) Run simulation
+model.run(tfinal=1400, dt=0.005)
+
+# (16) Load and scale reference
+reference = 1000.0*pandas.read_csv('single_cell_allen_neuron_ref.csv')['U/mV'].values[:-1] - 14.0
```
thorstenhater, line 130: You can load or at least motivate the -14.

doc/tutorial/single_cell_allen.rst
thorstenhater: This function does not exist anymore.

python/example/single_cell_allen.py
thorstenhater: They do not need to be mutable
```
d['a'].foo = 42 # FAIL
d['b'] = (1, 2, 3) # FINE
```

python/example/single_cell_allen.py
thorstenhater: No?

thorstenhater (changes requested):

doc/tutorial/single_cell_allen.rst
brenthuisman: I prefer to reserve code-like markup for code, filenames, variable names, and use good old quotes as they were intended. Or do you have some other markup in mind?

doc/tutorial/single_cell_allen.rst
thorstenhater: Never trust ;)

doc/tutorial/single_cell_allen.rst
thorstenhater: Isn't there a `math` role we have? At least in Sphinx? Would tie in nicely with usage in LaTeX

doc/tutorial/single_cell_allen.rst
```diff
@@ -0,0 +1,273 @@
...
+  # (2) Label the region tags found in the swc with the names used in the parameter fit file.
+  # In addition, label the midpoint of the soma.
+  labels = arbor.label_dict({
+      'soma': '(tag 1)',
+      'axon': '(tag 2)',
+      'dend': '(tag 3)',
+      'apic': '(tag 4)',
+      'midpoint': '(location 0 0.5)'})
+
+Step **(1)** loads the ``swc`` file using :func:`arbor.load_swc_allen`. Since the ``swc`` specification is informal, a few different interpretations exist, and we use the appropriate one. The interpretations are described ::ref`here <morph-formats>`.
```
thorstenhater: Also: swc allen.

doc/tutorial/single_cell_allen.rst
brenthuisman: I get your point, but it's the Allen DB. If those URLs become invalid, millions of papers will cry out in terror and be suddenly silent right along with us.

doc/tutorial/single_cell_allen.rst
brenthuisman: There is [a math role](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#math)! But are names math? "Sweep 35" is the name of an experiment, which I'd write like that in a paper too. I guess we could think of it as a variable name (it is also a row in a table) if you insist on changing it from quotes.

doc/tutorial/single_cell_allen.rst
thorstenhater: True, we could make it N-1 and we use this particular example quite often (L2L, here, Posters, ...). But fine if you think it's too much.

doc/tutorial/single_cell_allen.rst
thorstenhater: I was thinking about $150 nA$

doc/tutorial/single_cell_allen.rst
brenthuisman: All references corrected in subsequent commits.

doc/tutorial/single_cell_allen.rst
brenthuisman: We supply the bits we use for the example. As with all references: let's fix them as they break.

python/example/single_cell_allen.py
brenthuisman: Sure!

python/example/single_cell_allen.py
brenthuisman: Actually, we should load it from the file.

doc/tutorial/single_cell_allen.rst
brenthuisman: Done.

python/example/single_cell_allen.py
brenthuisman: But it requires a less readable way of assignment, and that's the primary goal of this/any example. I'd like to keep the class, the function returns in about a second here, which is plenty fast. We can add the `@dataclass` decorator back when we drop 3.6.

doc/tutorial/single_cell_allen.rst
brenthuisman: Ach so. Nah, that's a TeXism. I think there's nothing unclear about how it is now, numeric characters may appear in text, nothing wrong with that ;)

thorstenhater: It was a comment on style and it's not tied to TeX, but to proper typesetting.

python/example/single_cell_allen.py
thorstenhater: Sure, but that's not really the issue. You are using a `class` where you should be using an `object` of the `class`.
I'd argue that if you need mutable throw-away collections the Python way is to use either a List or a Dict.

brenthuisman: Afaik this is the only way (pre dataclass) to let you get/set the fields as member, which is all that I am after. Lists or Dicts will make the code less clear.

doc/tutorial/single_cell_allen.rst
brenthuisman: Inline number should conform to the typesetting of the text they appear in, I would say.

python/example/single_cell_allen.py
brenthuisman:
```
    class parameters:
        def __init__():
            self.cm = None
            self.tempK = None
            self.Vm = None
            self.rL = None
```

thorstenhater: mystery solved, we **add** instance variables _over_ class variables, which weird, but works.

brenthuisman: See new commit!

```
    class parameters:
        def __init__(self, **kwargs):
            self.cm = None
            self.tempK = None
            self.Vm = None
            self.rL = None
            self.__dict__.update(kwargs)
```

python/example/single_cell_allen.py
```diff
@@ -0,0 +1,142 @@
...
+        fit = json.load(fd)
+
+    # cable parameters convenience class
+    class parameters:
+        def __init__(self, **kwargs):
+            self.cm = None
+            self.tempK = None
+            self.Vm = None
+            self.rL = None
+            self.__dict__.update(kwargs)
```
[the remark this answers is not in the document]
brenthuisman: Ugly, or elegant? Let the reader decide ;)

thorstenhater (approved):

PR #1788, opened 2021-12-09
Construct `domain_decomposition` given a list of `group_descriptions`.

noraabiakar:
- Implement a new constructor for `domain_decomposition` from a vector of `group_descriptions`. The new constructor performs various checks on the given `group_descriptions` and throws an appropriate exception if any problem is found. It is the only available constructor for `domain_decomposition`.
- Make the various members of `domain_decomposition` private and implement their getters.  
- Expose the `domain_decomposition` constructor to the python interface through a `partition_by_group` function.
- Add unit tests. 
- Add documentation.


noraabiakar:
bors try

noraabiakar:
bors try

noraabiakar:
bors try

noraabiakar:
bors try

noraabiakar:
bors try

arbor/partition_load_balance.cpp
```diff
@@ -222,5 +222,110 @@ domain_decomposition partition_load_balance(
...
+
+    std::vector<cell_gid_type> local_gids;
+    for (const auto& g: groups) {
+        local_gids.insert(local_gids.end(), g.gids.begin(), g.gids.end());
+    }
+    cell_size_type num_local_cells = local_gids.size();
+
+    auto global_gids = ctx->distributed->gather_gids(local_gids);
+
+    domain_decomposition d;
```
thorstenhater, line 253: This seems quite old-school (Cxx20 will gives us dedicated initializers, but still).
Does this look better to you?
```
  return { num_domains, domain_id, 
           num_local_cells, num_global_cells, 
           groups, partition_gid_domain(global_gids, num_domains)};
```

thorstenhater (changes requested):
A few minor points and a high-level question. 

This still is based on round-robin distribution, correct? So there is no way
for the user to bunch up multiple groups on a single node, eg a CPU and a 
GPU group. It seems like an easy extension that'll be requested immediately ;)

arbor/simulation.cpp
```diff
@@ -186,6 +186,7 @@ simulation_state::simulation_state(
     task_system_(ctx.thread_pool),
     local_spikes_({thread_private_spike_store(ctx.thread_pool), thread_private_spike_store(ctx.thread_pool)})
 {
+    check_domain_decomposition(rec, ctx, decomp);
```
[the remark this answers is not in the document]
noraabiakar: In the C++ interface, `domain_decomposition` is a struct, and users can modify any part of that struct after construction. The check needs to happen here I think.

arbor/include/arbor/arbexcept.hpp
```diff
@@ -77,11 +77,65 @@ struct gj_unsupported_lid_selection_policy: arbor_exception {
 
 // Domain decomposition errors:
 
-struct gj_unsupported_domain_decomposition: arbor_exception {
-    gj_unsupported_domain_decomposition(cell_gid_type gid_0, cell_gid_type gid_1);
+struct dom_dec_invalid_gj_cell_group: arbor_exception {
```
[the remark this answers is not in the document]
noraabiakar: Okay

doc/cpp/domdec.rst
```diff
@@ -58,6 +59,22 @@ describes the cell groups on the local MPI rank.
         computational cost, hence it may not produce a balanced partition for
         models with cells that have a large variance in computational costs.
 
+.. cpp:function:: domain_decomposition partition_by_group(const recipe& rec, const arb::context& ctx, const std::vector<group_description>& groups)
+
+    Construct a :cpp:class:`domain_decomposition` that  assigns the groups described by
+    the provided list of :cpp:class:`group_description` to the local hardware resources
+    of the calling rank.
+
+    The function expects to be called by each rank in a distributed simulation with the
```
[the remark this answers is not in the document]
noraabiakar: I can add an example similar to the one in the python docs.

noraabiakar:
This isn't round-robin partitioning, it's completely custom partitioning. And there should be no reason why we can't bunch up multiple groups on the same node. For example: 
```
  n_cells = 15
  recipe = my_recipe(n_cells)
  
  # Code for rank 0
  groups = [arbor.group_description(arbor.cell_kind.cable, [0, 2, 4, 6, 8], arbor.backend.gpu),
                  arbor.group_description(arbor.cell_kind.cable, [1, 3, 5, 7, 9], arbor.backend.multicore)] ​

  ​decomp = arbor.partition_by_group(recipe, context, groups)

  # Code for rank 1
  groups = [arbor.group_description(arbor.cell_kind.cable, [10], arbor.backend.multicore),
                  arbor.group_description(arbor.cell_kind.cable, [11], arbor.backend.multicore),
                  arbor.group_description(arbor.cell_kind.cable, [12, 13, 14], arbor.backend.gpu)] ​

  ​decomp = arbor.partition_by_group(recipe, context, groups)

`​``
Rank 0 has 2 cell groups of 5 cells each, one running on the CPU and one on the GPU. 
Rank 1 has 3 cell groups, the first two containing one cell each and running on the CPU, the third containing 3 cells running on the GPU.

File: doc/python/domdec.rst [hunk 5]
```diff
@@ -38,6 +40,64 @@ describes the cell groups on the local MPI rank.
...
+   The validity is only checked when a :class:`simulation` object is constructed using that :class:`domain_decomposition`.
+
+   .. Note::
+        This function is intended for users who have a good understanding of the computational
+        cost of simulating the cells in their network and want fine-grained control over the
+        partitioning of cells across ranks. It is recommended to start off by using
+        :func:`partition_load_balance` and switch to this function if the observed performance across
+        ranks is unbalanced.
+
+   .. Note::
```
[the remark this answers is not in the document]
noraabiakar: Which sizes?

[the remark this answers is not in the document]
thorstenhater: Where CPU and GPU break even in terms of performanec.

arbor/partition_load_balance.cpp
thorstenhater: I think I like the C++20 way and right now both variants are terrible ;)

thorstenhater:
Ah, my misunderstanding then! Yes, by having each rank declaring its groups this works.
However, did I also miss a check against the gids being bound twice?

arbor/simulation.cpp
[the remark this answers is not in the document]
thorstenhater: We could stop that from happening. Is there ever a need for it?

noraabiakar:
There's a check for gids being bound twice [here](https://github.com/arbor-sim/arbor/pull/1788/files#diff-8dc23d1405f7d07e262e646d0a6087da3b345fed5f18d928aee8580b7794eb51R318)

doc/python/domdec.rst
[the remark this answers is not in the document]
noraabiakar: Depends on the cells. So yes, difficult to say.

arbor/simulation.cpp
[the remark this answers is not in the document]
noraabiakar: We could now that we have this function. It essentially does the same thing.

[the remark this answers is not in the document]
noraabiakar: I implemented the suggested change, but then I realized that I prefer to keep things as is for 2 reasons: 
- The struct is quite a jumble of things, and having to form it with a list initializer or a constructor will make the code more difficult to read.
- The constructor of `domain_decomposition` doesn't require the rec, or ctx. The partitioning algorithms do. So even if we make the changes to `domain_decomposition`, we still wouldn't be able to call `check_domain_decomposition` in the constructor of the class. 

What do you think?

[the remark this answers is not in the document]
thorstenhater: My reasoning was this:
- domain decompositions should be immutable after creation
- they should only be constructed via the algorithms (hard-ish to enforce, but doable)
- it's on the algorithms to check validity, they have also access to the relevant items (ctx et al)

Alternatively we could do this:
- simulations only work on checked_dom_dec
- checked_dom_dec can only be made by the check function

doc/python/domdec.rst
```diff
@@ -12,7 +12,9 @@ Load balancers
 
 Load balancing generates a :class:`domain_decomposition` given an :class:`arbor.recipe`
 and a description of the hardware on which the model will run. Currently Arbor provides
-one load balancer, :func:`partition_load_balance`, and more will be added over time.
+one load balancer, :func:`partition_load_balance`; and a function for creating
+custom decompositions, :func:`partition_by_group`.
+More load balancers will be added over time.
```
thorstenhater, line 17: `may` rather than `will`?

doc/python/domdec.rst
```diff
@@ -38,6 +40,64 @@ describes the cell groups on the local MPI rank.
...
+   The function expects to be called by each rank in a distributed simulation with the
+   selected groups for that rank. For example, in a simulation of 10 cells on 2 MPI ranks
+   where cells {0, 2, 4, 6, 8} of kind :class:`cable_cell` are expected to be in a single group executed
+   on the GPU on rank 0; and cells {1, 3, 5, 7, 9} of kind :class:`lif_cell` are expected to be in a single
+   group executed on the CPU on rank 1:
+
+   Rank 0 should run:
+
+   .. code-block:: python
+
```
thorstenhater, line 57: Reading these examples of decompositions are actually used, make me wonder whether they should be
moved to the recipe.

Reasoning:
1. Partition-load-balance: needs the recipe anyhow, so could be used like this:
```cxx
struct arb::recipe {
  virtual domain_decomposition decompose(ctx) { return partition_load_balance(...); }
};
```
Concrete recipes might want to override this, but it would remove one non-obvious step from
the construction of a simulation, which -- I think -- our users would appreciate.
2. partition_by_group: needs a lot of the arguments provided by the cell group
  - num_cells and cell_kind by gid for the group makeup
  - recipe and context as before

That said, it's also awkward, since the MPI rank/size is needed and suddenly recipe
becomes dependent on MPI :/ Potentially we could work around that by passing rank/size
to the `decompose` method.

doc/python/domdec.rst
```diff
@@ -38,6 +40,64 @@ describes the cell groups on the local MPI rank.
...
+        decomp = arbor.partition_by_group(recipe, context, groups)
+
+   The function doesn't perform any checks on the validity of the generated :class:`domain_decomposition`.
+   The validity is only checked when a :class:`simulation` object is constructed using that :class:`domain_decomposition`.
+
+   .. Note::
+        This function is intended for users who have a good understanding of the computational
+        cost of simulating the cells in their network and want fine-grained control over the
+        partitioning of cells across ranks. It is recommended to start off by using
+        :func:`partition_load_balance` and switch to this function if the observed performance across
```
thorstenhater, line 93: Any suggestions on how this could be observed?

doc/python/domdec.rst
```diff
@@ -38,6 +40,64 @@ describes the cell groups on the local MPI rank.
         computational cost, hence it may not produce a balanced partition for
         models with cells that have a large variance in computational costs.
 
+.. function:: partition_by_group(recipe, context, groups)
+
+   Construct a :class:`domain_decomposition` that assigns the groups described by
+   the provided list of :class:`group_description` to the local hardware of the calling rank.
```
thorstenhater, line 46: It would be helpful to provide a list of constraints here to be upheld:
- GJ connections force cells into one group
- a group must be homogeneous in its cell_kind
- ...

thorstenhater (changes requested):

doc/python/domdec.rst
noraabiakar: I think we need to add more, such as your suggested random partitioning, or round-robin. This is us acknowledging that the available functions aren't sufficient. That being said, I can change it to may.

doc/python/domdec.rst
noraabiakar: The simulation runtime wouldn't scale well. But I see your point. I'll add a bit more info here.

doc/python/domdec.rst
noraabiakar: Yes, I'll add the GJ connections rule. The groups are homogeneous in cell-kind by design.

doc/python/domdec.rst
noraabiakar: I don't think that the `domain_decomposition` should be part of the `recipe`. The `recipe` is used to describe the scientific model: the cells, the connections, the probes etc. The `domain_decomposition` is not related to the model description, it's more technical, it relates the the available hardware resources and to implementation details (how cells are grouped, how expensive they are to simulate). I think it is a good idea to keep these 2 things separate. 
That being said, we can make the `domain_decomposition` immutable after construction, and enforce that the functions that create these objects check their validity after creation.

doc/python/domdec.rst
thorstenhater: Yes, but it's a constraint that needs to be observed by eg partition_load_balance and thus I would suspect a reader might be interested in that fact.

doc/python/domdec.rst
noraabiakar: Still, that won't fix the issue that `domain_decomposition` is part of the public interface, so there's nothing stopping the users from using the constructors of the class to create custom `domain_decomposition`s and bypassing the partitioning functions.

noraabiakar: We could delete all constructors and have the partitioning functions be `friend`s. But I don't like that so much.

thorstenhater: Yeah, using `friend` would be the way to go, but as ever icky.

arbor/simulation.cpp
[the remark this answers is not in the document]
bcumming: For some context, the C++ `domain_decomposition` struct was designed for users to specify their own decomposition, or implement model-specific decomposition functions.
I don't think we want to restrict ourselves to the algorithms provided by Arbor, due to the difficulty of writing and maintaining algorithms that are applicable to all problems.

A stand-alone validator for domain decompositions.

In Python land the story is different: it isn't possible to construct a domain decomposition in Python because it requires writing a gid mapping function, which would mean supporting Python callbacks, which have serious consequences for parallelisation.

So, if a domain decomposition is created by one of the algorithms provided via the Python interface, they should validate the decomposition.

But I don't think we need to make this a hard-coded requirement on the C++ side.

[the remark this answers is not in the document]
bcumming: For some context, the C++ domain_decomposition struct was designed for users to specify their own decomposition, or implement model-specific decomposition functions.
I don't think we want to restrict ourselves to the algorithms provided by Arbor, due to the difficulty of writing and maintaining algorithms that are applicable to all problems.

A stand-alone validator for domain decompositions.

In Python land the story is different: it isn't possible to construct a domain decomposition in Python because it requires writing a gid mapping function, which would mean supporting Python callbacks, which have serious consequences for parallelisation.

So, if a domain decomposition is created by one of the algorithms provided via the Python interface, they should validate the decomposition.

But I don't think we need to make this a hard-coded requirement on the C++ side.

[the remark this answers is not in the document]
thorstenhater: Hey @bcumming thanks for the context.

> For some context, the C++ domain_decomposition struct was designed for users to specify their own decomposition, or implement model-specific decomposition functions.
I don't think we want to restrict ourselves to the algorithms provided by Arbor, due to the difficulty of writing and maintaining algorithms that are applicable to all problems.

That's great, so this data (`domain_decomposition`) is a piece of user-supplied input and we have seen (ref one-sided gap-junctions) that it is very well possible to produce strange behaviour if used incorrectly. I think it's good practice to validate this such that
1. it is checked early in the flow in order to get feedback to the user sooner
2. it cannot be changed afterwards without re-validation
3. Preferably enforce 2. using the type system

The current state is to check upon construction of the simulation, which is fine, but it delays user feedback quite a bit.
That was the core of my suggestion.

> But I don't think we need to make this a hard-coded requirement on the C++ side.

Even in C++ code decompositions made by us should be valid. Else the validation scheme is broken and needs to be fixed.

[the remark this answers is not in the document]
noraabiakar: I think the key point in Ben's comment is: 
> the C++ domain_decomposition struct was designed for users to specify their own decomposition, or implement model-specific decomposition functions.
I don't think we want to restrict ourselves to the algorithms provided by Arbor, due to the difficulty of writing and maintaining algorithms that are applicable to all problems.

If we agree that `domain_decomposition` should remain as it is in order to allow users to write their own partitioning algorithms, then we have given the users the tool to check their structs as early on as they want using the public `check_domain_decomposition` function. If the users neglect to do that, it will be checked again before the object is used in the constructor of `simulation`. 

However, we could argue that the users can use the newly added `partition_by_group` to create their own custom `domain_decomposition` objects and therefore the `domain_decomposition` struct doesn't need to remain as it is. The key bit of information that the user would be interested in (the different cell groupings on each rank) is passed as an argument to `partition_by_group` which then fills out the rest of the struct members. So, we could make the `domain_decomposition` struct more opaque to the users. We could make `partition_by_group` the only constructor of `domain_decomposition` (so `partition_load_balance` would need to change a bit), and we can perform the correctness check in the constructor.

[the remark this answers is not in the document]
thorstenhater: 👍 That's essentially what I had in mind, but was unsure whether `partition_by_group` would be the only viable (smart) constructor.

doc/cpp/domdec.rst
```diff
@@ -7,136 +7,127 @@ Domain decomposition
...
+                {arb::cell_kind::lif,   {1, 3, 5, 7, 9}, arb::backend_kind::multicore}
+            };
+            auto decomp = arb::domain_decomposition(recipe, context, groups);
+
+        .. _domdecnotes:
+        .. Important::
+            Constructing a balanced :cpp:class:`domain_decomposition` quickly
+            becomes a difficult task for large and diverse networks. This is why
+            arbor provides :ref:`load balancing algorithms<domdecloadbalance>`
+            that automatically generate a :cpp:class:`domain_decomposition` from
```
thorstenhater, line 85: Typo: should be generate*s*

arbor/include/arbor/domain_decomposition.hpp
```diff
@@ -31,26 +32,37 @@ struct group_description {
...
+public:
+    domain_decomposition() = delete;
+    domain_decomposition(const recipe& rec, const context& ctx, const std::vector<group_description>& groups);
+
+    int gid_domain(cell_gid_type gid) const;
+    int num_domains() const;
+    int domain_id() const;
+    cell_size_type num_local_cells() const;
+    cell_size_type num_global_cells() const;
+    const std::vector<group_description>& groups() const;
```
thorstenhater, line 45: While reading the code I saw these two things used frequently
```cxx
auto g  = dd.groups()[ix];
auto n = dd.groups().size();
```
Maybe add this as final polish
```cxx
class domain_decomposition {
public:
  size_t num_groups();
  group_description& group(size_t);
  const group_description& group(size_t) const;
```

thorstenhater (changes requested):
A typo and an optional suggestion for some polish.

doc/cpp/domdec.rst
noraabiakar: it's load balancing algorithms in plural. I know we only have one at the moment, but the plan has always been to add more.

arbor/include/arbor/domain_decomposition.hpp
noraabiakar: I'll only add `const group_description& group(size_t) const;`. We don't want the groups to be mutable after construction.

doc/cpp/domdec.rst
thorstenhater: Ok, that was me reading `partition_load_balance`. The curse of a prior.

noraabiakar:
bors try

thorstenhater (approved):
:+1:

PR #1797, opened 2022-01-07
Py/setuptools wheel fixes

brenthuisman:
Changes setup.py to build binary wheels with nml enabled. See https://github.com/arbor-sim/arbor/actions/runs/1668149223

Todo:
* [x] Add `nml` status to `config()`. ('bundled' added as well.)
* [x] Add `build-wheels.sh` to `scripts/` which can be used to generate manylinux wheels using docker/podman. This script is not used by the Github Action, and should be kept in sync with it (`.github/workflows/ciwheel.yml`).
* [ ] Run tests on the wheel. Currently blocked by `setup.py` not correctly installing the build-catalog script.
* [x] Adjust docs (renamed some install options, and the defaults)
* [ ] If #1784 is merged, this PR is obsolete.

### DIFF
--- a/.github/workflows/ciwheel.yml
+++ b/.github/workflows/ciwheel.yml
@@ -6,6 +6,9 @@ on:
     tags:
       - v*
 
+  schedule:
+    - cron: '0 2 * * 0' # run at 2 AM every sunday
+
 jobs:
   build_binary_wheels:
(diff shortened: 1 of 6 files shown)


brenthuisman:
We've decided to go with skbuild, moving changes into #1784.

PR #1798, opened 2022-01-12
Add release cycle info to docs

brenthuisman:


thorstenhater (changes requested):
Looks good, but please replace concrete names/short hands with functional descriptions.
So
- BH -> release manager
- AK -> Scrum entity

Reduces confusion ;)

thorstenhater (approved):

PR #1807, opened 2022-01-20
Remove footgun: Catalogue Lifetimes

### LINKED ISSUE #1446: Mechanism catalogue lifetimes in Python and C++
Currently, mechanism catalogue lifetime in Python has to be managed semi-manually, by tying it to e.g. the recipe object.
Similarly, it is the user code responsibility to manage catalogue lifetimes in C++ — a reference counted solution would be much safer.

thorstenhater:
Cable Cell Recipes (actually global properties) now hold on to catalogues via value.
Many users have found and reported crashes because of the old behaviour.
This change also make most examples a bit simpler, as the default catalogue is now 
the ... er ... default and there is no need anymore to keep hold of the catalogue separately.

In conclusion
- the old pointer never saved memory (that would mean separate global_prop's on the same node).
- users kept their catalogue around as a value anyhow and ran into crashes if they did not.
- we traded a bit of performance (constructing the default_catalogue w/ each global_props) for quite
  some convenience (now props just works for many simple cases).
- also 90% of cases seem to use this pattern
  ```py
   cat = arb.default_catalogue()
   cat.extend(other_cat)
   props.register(cat)
   ```
   which now becomes this simpler sequence
   ```py
    props.catalogue.extend(other_cat)
   ```

# Changes

- `cable_cell_global_properties` now holds a catalogue directy
- fixes for tests, examples, and python bindings accordingly

# Linked Issues
Fixes #1446


python/test/unit/test_catalogues.py
```diff
@@ -13,11 +13,11 @@ def __init__(self):
         self.tree.append(arb.mnpos, (0, 0, 0, 10), (1, 0, 0, 10), 1)
         self.props = arb.neuron_cable_properties()
         try:
-            self.cat = arb.default_catalogue()
-            self.props.register(self.cat)
+            self.props.catalogue = arb.load_catalogue('dummy-catalogue.so')
```
bcumming, line 16: Fun fact: when building with meson on Mac OS shared libraries have the `dylib` file extension, not `so`.
Something that I have generalised in the meson branch.

bcumming (approved):
This is easier to understand, and much less dangerous.

python/test/unit/test_catalogues.py
thorstenhater: HA! We would have had that in CMake but I force-overrode that behaviour.

noraabiakar:
Tutorial needs a fix: https://github.com/arbor-sim/arbor/blob/master/doc/tutorial/single_cell_detailed_recipe.rst lines 85/86 (mention of registering the catalogue).

thorstenhater:
@noraabiakar thanks, these literal includes escaped me. 
The upside is that these have become simpler and smaller

PR #1814, opened 2022-01-25
v0.6-rc

brenthuisman:
* bump version to v0.6-rc
* Update `MANIFEST.in` and `scripts/run_python_examples.sh`
* Update Release instructions `docs/contrib/release.rst`
* Fix missing quotes in `python/example/single_cell_nml.py`
* Add testing to sdist creation in `ciwheel.yml`
* Rename `basic.yml` to `test-everything.yml`
   * Remove Python examples step from matrix (mostly because plotting takes much CPU on macos)
   * change to build /w nml
   * add test of pip builds (`pip install` in addition to CMake builds)
      * Python examples are tested here, once is enough.


.github/workflows/test-everything.yml
```diff
@@ -92,11 +92,15 @@ jobs:
         # See https://github.com/open-mpi/ompi/issues/6518
         OMPI_MCA_btl: "self,tcp"
     steps:
-      - name: Get clang/gcc 8
+      - name: Get clang/gcc 8, libxml2, Linux
```
noraabiakar, line 95: Can we rename to "Linux: get clang/gcc 8, libxml2" (and "MacOS: get libxml2" below)

.github/workflows/test-everything.yml
```diff
@@ -182,3 +186,19 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install Python packages
+        run:  pip install numpy setuptools scikit-build ninja cmake
+      - name: Clone w/ submodules
+        uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - name: Build and install Arbor using pip
+        run: python3 -m pip install --verbose --install-option="-DARB_VECTORIZE=ON" --install-option="-DARB_ARCH=native" .
```
noraabiakar, line 200: Do we verify that vectorization is actually enabled?

noraabiakar (approved):
A couple of minor comments, but otherwise looks good.

.github/workflows/test-everything.yml
brenthuisman: Sure!

.github/workflows/test-everything.yml
brenthuisman: We do now :)

noraabiakar: Cool :)

PR #1819, opened 2022-01-27
🐍 Be more lenient when accepting args to file I/O

thorstenhater:
PyArbor I/O routines now try to cast `filename` arguments to a string.
This allows passing `pathlib.Path` objects.

### DIFF
--- a/python/cable_cell_io.cpp
+++ b/python/cable_cell_io.cpp
@@ -13,7 +13,10 @@
 
 namespace pyarb {
 
-arborio::cable_cell_component load_component(const std::string& fname) {
+namespace py = pybind11;
+
+arborio::cable_cell_component load_component(py::object fn) {
+    const std::string fname = py::str(fn);
     std::ifstream fid{fname};
     if (!fid.good()) {
         throw pyarb_error("Can't open file '{}'" + fname);
@@ -26,13 +29,13 @@ arborio::cable_cell_component load_component(const std::string& fname) {
 };
 
 template<typename T>
-void write_component(const T& component, const std::string& fname) {
-    std::ofstream fid(fname);
+void write_component(const T& component, py::object fn) {
+    std::ofstream fid(std::string{py::str(fn)});
     arborio::write_component(fid, component, arborio::meta_data{});
 }
 
-void write_component(const arborio::cable_cell_component& component, const std::string& fname) {
-    std::ofstream fid(fname);
+void write_component(const arborio::cable_cell_component& component, py::object fn) {
+    std::ofstream fid(std::string{py::str(fn)});
     arborio::write_component(fid, component);
 }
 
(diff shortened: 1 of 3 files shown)


python/cable_cell_io.cpp
```diff
@@ -13,7 +13,10 @@
 
 namespace pyarb {
 
-arborio::cable_cell_component load_component(const std::string& fname) {
+namespace py = pybind11;
+
+arborio::cable_cell_component load_component(py::object fn) {
+    const std::string fname = py::str(fn);
```
bcumming, line 19: Here we are using the `py::object` as though it can be converted to a `std::string`, without checking.

The errors when a type that can't be converted to string, or converts to an odd string, is passed.

How about writing a little helper function that tests the type and converts it to a string, or throws an exception with a helpful message?

bcumming (changes requested):
One small suggestion to improve robustness.

python/cable_cell_io.cpp
bcumming: ... option 1!

brenthuisman: The Pythonic way is to try and handle errors (which can just be "Unknown input provided, aborting...").

python/cable_cell_io.cpp
```diff
@@ -9,14 +9,18 @@
...
 
-arborio::cable_cell_component load_component(const std::string& fname) {
+namespace py = pybind11;
+
+arborio::cable_cell_component load_component(py::object fn) {
+    const auto fname = util::to_path(fn);
     std::ifstream fid{fname};
     if (!fid.good()) {
-        throw pyarb_error("Can't open file '{}'" + fname);
+        throw pyarb_error(util::pprintf("Can't open file '{}'", fname));
```
brenthuisman, line 23: I'd be happy to have a new-style typed error here, but I'll leave that up to you.

python/cable_cell_io.cpp
```diff
@@ -43,40 +47,40 @@ void register_cable_loader(pybind11::module& m) {
           "Load arbor-component (decor, morphology, label_dict, cable_cell) from file.");
 
     m.def("write_component",
-          [](const arborio::cable_cell_component& d, const std::string& fname) {
-            return write_component(d, fname);
+          [](const arborio::cable_cell_component& d, py::object fn) {
+            return write_component(d, fn);
           },
           pybind11::arg_v("object", "the cable_component object."),
           pybind11::arg_v("filename", "the name of the file."),
```
brenthuisman, line 54:
```suggestion
          pybind11::arg_v("filename", "the path of the file."),
```
and below too.

python/mechanism.cpp
```diff
@@ -197,7 +198,7 @@ void register_mechanisms(pybind11::module& m) {
     m.def("default_catalogue", [](){return arb::global_default_catalogue();});
     m.def("allen_catalogue", [](){return arb::global_allen_catalogue();});
     m.def("bbp_catalogue", [](){return arb::global_bbp_catalogue();});
-    m.def("load_catalogue", [](const std::string& fn){return arb::load_catalogue(fn);});
+    m.def("load_catalogue", [](pybind11::object fn) { return arb::load_catalogue(util::to_string(fn)); });
```
brenthuisman, line 201:
```suggestion
    m.def("load_catalogue", [](pybind11::object fn) { return arb::load_catalogue(util::to_path(fn)); });
```
?

python/util.hpp
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <pybind11/pybind11.h>
+
+#include "strprintf.hpp"
+
+namespace pyarb {
+namespace util {
```
brenthuisman, line 8: I'm not sure it's wise to have two different namespaces with the same name, despite being properly nested. Won't it lead to surprises?

brenthuisman (changes requested):

python/cable_cell_io.cpp
thorstenhater: That's a good point, I should remember my own demands.

python/util.hpp
thorstenhater: I think that's fine, even common, and otherwise we wouldn't need nested namespaces or even namespaces at all. More confusing would be -- to me -- `pyarb::py_utils` or similar.

python/mechanism.cpp
brenthuisman: When is `std::fs`` due?

python/util.hpp
brenthuisman: OK

python/cable_cell_io.cpp
brenthuisman: Go on :)

python/mechanism.cpp
thorstenhater: C++17, it's supported by all compilers, **but** MacOS pre-10.15 doesn't have the proper syscalls, so there **no** compiler has it. @espenhgn had issues with that and we reverted.

espenhgn: Hi. I've since gotten another machine (w. macOS 11.6). I've had no issues compiling Arbor-v0.6 on that. If you now keep 10.15 support alive only for me, feel free to revert back.

brenthuisman (approved):

python/mechanism.cpp
brenthuisman: We've prevented `std::filesystem` from entering the Arbor code base for this reason, and it is good to know that we now can :)

PR #1821, opened 2022-01-28
Users may not give dt < 0.

thorstenhater:
Fixes an obvious bug w/ `dt <= 0`.

### DIFF
--- a/arbor/simulation.cpp
+++ b/arbor/simulation.cpp
@@ -515,7 +515,7 @@ void simulation::reset() {
 
 time_type simulation::run(time_type tfinal, time_type dt) {
     if (dt <= 0.0) {
-        throw arbor_exception("Finite time-step must be supplied.");
+        throw domain_error("Finite time-step must be supplied.");
     }
     return impl_->run(tfinal, dt);
 }
--- a/arbor/simulation.cpp
+++ b/arbor/simulation.cpp
@@ -514,6 +514,9 @@ void simulation::reset() {
 }
 
 time_type simulation::run(time_type tfinal, time_type dt) {
+    if (dt <= 0.0) {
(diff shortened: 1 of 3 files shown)


arbor/simulation.cpp
```diff
@@ -514,6 +514,9 @@ void simulation::reset() {
 }
 
 time_type simulation::run(time_type tfinal, time_type dt) {
+    if (dt <= 0.0) {
+        throw arbor_exception("Finite time-step must be supplied.");
```
brenthuisman, line 518: Maybe an exception a tad more type specific? There's no `user_input_error` in `arbexcept?pp`, maybe this is a good occasion to add it.

brenthuisman (changes requested):

arbor/simulation.cpp
bcumming: `pyarb_error` in `python/error.hpp` might be the ticket.

thorstenhater: In the end everything is the user's fault, but I think it's not very polite to tell them.
I decided to rip off `std` and make `domain_error` a thing.

brenthuisman (approved):

arbor/simulation.cpp
bcumming: Here is the thing: shouldn't we be using that error type for all such errors in our interface?
Instead of creating a new error type, to be used for the one case in this change, we could make a separate issue to systematically address such errors.

thorstenhater: Sure, let's consider this the obvious start. We have checks on eg context and domain decomposition, but I am not aware of more place where this could/should go.

thorstenhater:
I opened a feature request for the process here
https://github.com/arbor-sim/arbor/issues/1826

Issue #1752, opened 2021-11-09
Bad Any Cast Error in C++ examples


kanzl:
**Describe the bug**

Can't run the C++ examples on my M1 Mac.
The benchmark miniapp produces a bad any cast error and the other examples get error messages similar to this:

`exception caught in gap junction miniapp:
recipe::get_cell_kind(gid=6) -> cell_kind::cable does not match the cell type provided by recipe::get_cell_description(gid=6)`

thorstenhater:
Currently this is -- surprisingly -- related to our linking options. A temporary fix is to add
`-fvisibility=hidden` to the linker options of the examples.

boeschf:
Hi @kanzl, I have a PR prepared for your issue. In order for me to check all cases, could you tell me how you invoked the example exactly? thanks

PR #1824, opened 2022-02-01
export API

boeschf:
- addresses #1752
- exports all symbols so arbor can be compiled as set of shared libraries

In order to export all symbols correctly, I have added one macro per library and one global macro. The content of the macros is determined at configure time depending on build variant (static/shared), compiler, and platforms (linux, mac os) and goes into the library's include directory as `export.hpp` when installed (at build time it resides at cmake's temporary build directory). The per-library macro is named `ARB_LIBNAME_API` and goes in front of to-be-exported symbols. The global macro is `ARB_SYMBOL_VISIBLE`. Below, I sketch out a few examples where these macros would be needed.

In this PR I tried to find all places where any annotation is required. Most of them are in the public headers (and corresponding sources) but some are also added in internal headers, which were required for the unit tests to link properly.

`header.hpp`
```
#pragma once 

#include <arbor/export.hpp>

// inline class
// no exports are needed in general (there are exceptions :) )
struct A {
    int foo() { return 42; }
    friend std::ostream& operator<<(std::ostream& o, A const & a) { o << a.foo();  return o; }
};

// template classes are inline
// no exports are needed
template<typename T>
class B {
    T m;
    B(T&& t) : m{std::move(t)} {}
};

// class declaration, implementation in source
// class needs to be exported (note this will make all member symbols visible)
class ARB_ARBOR_API C {
    void bar();
    friend std::ostream& operator<<(std::ostream& o, C const & a);
};

// inline exception class
// even though it is inline, one must make it visible! Otherwise, thrown exceptions
// will cause havoc.
class ARB_SYMBOL_VISIBLE some_error : public std::runtime_error { };

// any class which is wrapped into a std::any/arb::util::any or similar or is put into a variant
// and is part of the public interface (will be referenced outside the library) must have visibiltiy!
class ARB_SYMBOL_VISIBLE D { int a; };

// inline free function
// no need to export
void baz(int i) { std::cout << i << std::endl; }

// free function
// needs to be exported
ARB_ARBOR_API void fuu();

// (extern) global variables
// need to be exported
ARB_ARBOR_API int g;
ARB_ARBOR_API extern int h;
```

`source.cpp`
```
// no export at point of definition needed for member functions
// of an exported class
void C::bar() {
    throw some_error("haha");
}

// friend functions need to be exported though
ARB_ARBOR_API std::ostream& operator<<(std::ostream& o, C const& c) {
    o << 'c';
    return o;
}

// free functions need here an annotation, too!
ARB_ARBOR_API void fuu() {
    std::cout << 42 << std::endl;
}

// need to be annotated, as well
ARB_ARBOR_API int g = 10;
ARB_ARBOR_API int h = 11;
```


brenthuisman:
Could you test the cibuildwheel workflow on this change? You can overwrite the `ciwheel` branch to trigger the action.

modcc/CMakeLists.txt
```diff
@@ -30,17 +30,32 @@ set(libmodcc_sources
 set(modcc_sources modcc.cpp)
 
 add_library(libmodcc STATIC ${libmodcc_sources})
-target_include_directories(libmodcc PUBLIC .)
+target_link_libraries(libmodcc PUBLIC arbor-public-headers)
```
noraabiakar, line 33: As far as I understand, to export the symbols of the various libraries we need to build arbor as a shared library using the CMake option `-DBUILD_SHARED_LIBS=ON`. I tried this and noticed that `arbor`, `arborio`, `arborenv` and `arbor-sup` all get built as shared libraries, but `libmodcc` is built as a static library. I'm far from a CMake pro, if you could point out where that gets decided that would be really helpful.

modcc/expression.hpp
```diff
@@ -68,10 +69,10 @@ symbol_ptr make_symbol(Args&&... args) {
 }
 
 // helper functions for generating unary and binary expressions
-expression_ptr unary_expression(Location, tok, expression_ptr&&);
-expression_ptr unary_expression(tok, expression_ptr&&);
-expression_ptr binary_expression(Location, tok, expression_ptr&&, expression_ptr&&);
-expression_ptr binary_expression(tok, expression_ptr&&, expression_ptr&&);
+ARB_LIBMODCC_API expression_ptr unary_expression(Location, tok, expression_ptr&&);
+//ARB_LIBMODCC_API expression_ptr unary_expression(tok, expression_ptr&&);
```
noraabiakar, line 73: Can also be removed.

noraabiakar (changes requested):
I have some minimal suggestions and questions. Otherwise looks good!

arbor/include/arbor/domain_decomposition.hpp
```diff
@@ -4,6 +4,7 @@
 #include <functional>
 #include <vector>
 
+#include <arbor/export.hpp>
```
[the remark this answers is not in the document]
boeschf: ok, I can do that

arbor/profile/profiler.cpp
```diff
@@ -373,19 +373,19 @@ std::ostream& operator<<(std::ostream& o, const profile& prof) {
...
-void profiler_enter(region_id_type) {}
-profile profiler_summary();
-void profiler_print(const profile& prof, float threshold) {};
-profile profiler_summary() {return profile();}
-region_id_type profiler_region_id(const std::string&) {return 0;}
-std::ostream& operator<<(std::ostream& o, const profile&) {return o;}
+ARB_ARBOR_API void profiler_leave() {}
+ARB_ARBOR_API void profiler_enter(region_id_type) {}
+ARB_ARBOR_API profile profiler_summary();
+//void profiler_print(const profile& prof, float threshold) {};
```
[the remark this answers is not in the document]
boeschf: ok

modcc/CMakeLists.txt
boeschf: that's because I forced it to be static here https://github.com/boeschf/arbor/blob/f38aece0d9bb1d8a14e92022e3354083ef80447e/modcc/CMakeLists.txt#L32
There were some problems building it dynamically (I have to check what happens exactly). The library gets used to make the `modcc` compiler (only?) so I thought it would be ok to leave it static.

boeschf: `add_library` without `STATIC` or `SHARED` are compiled as static library by default unless you set `-DBUILD_SHARED_LIBS=ON`

arbor/include/arbor/domain_decomposition.hpp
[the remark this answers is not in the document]
noraabiakar: Thanks :)

example/bench/CMakeLists.txt
```diff
@@ -2,3 +2,4 @@ add_executable(bench EXCLUDE_FROM_ALL bench.cpp)
 add_dependencies(examples bench)
 
 target_link_libraries(bench PRIVATE arbor arborenv arbor-sup ${json_library_name})
+set_target_properties(bench PROPERTIES CXX_VISIBILITY_PRESET hidden)
```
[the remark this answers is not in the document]
boeschf: I removed the hidden visibility again. Now, I could not get rid of some warnings on macos when static libraries are built, however, the tests pass. One way to avoid the warnings would be to compile with default visibility in the special case of macos+static. If the warnings are not bothering anyone, we could go with what we have.

boeschf:
bors try

boeschf:
bors try

test/unit/test_unique_any.cpp
```diff
@@ -164,7 +164,7 @@ TEST(unique_any, any_cast_ptr) {
     EXPECT_EQ(any_cast<std::string>(&ai), nullptr);
     unique_any empty;
     EXPECT_EQ(any_cast<int>(&empty), nullptr);
-    EXPECT_EQ(any_cast<int>((util::unique_any*)nullptr), nullptr);
+    //EXPECT_EQ(any_cast<int>((util::unique_any*)nullptr), nullptr);
```
noraabiakar, line 167: Should this stay commented?

test/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 find_package(Threads REQUIRED)
 
-add_library(gtest EXCLUDE_FROM_ALL STATIC gtest-all.cpp)
+#add_library(gtest EXCLUDE_FROM_ALL STATIC gtest-all.cpp)
```
noraabiakar, line 3: Can this be deleted?

boeschf: yes

noraabiakar (approved):

noraabiakar:
bors try

PR #1829, opened 2022-02-03
Update gitmodule `fmt` to 8.1.1

brenthuisman:


bcumming:
Are we bumping the minimum supported version of fmt, or just the version of fmt?

See my comment here in the following link for rationale: https://github.com/arbor-sim/arbor/issues/1731#issuecomment-1074978147

PR #1830, opened 2022-02-03
Update gitmodule Pybind11 to v2.9.1

brenthuisman:


bcumming:
What is the motivation for updating the submodule?

brenthuisman:
The rationale discussed was that certain deps (Google benchmark *ahem*) were terribly out of date, and that we wanted to 1. know about new versions (hence #1741) and then see if there's any breakage (me following up for the first time on output of #1741). Ofc, we don't have to merge, we can simply note that there's not build issue. I'm more than fine with that; the build conflicts on Ebrains are to be expected if we trail the latest versions too closely.

So, if you agree, I can add the info that builds pass with v2.9.1 to https://github.com/arbor-sim/arbor/wiki/Testing-and-version-management, and add that document + the policy in #1731 to the docs as a new page (which was the intent of #1731).

bcumming:
Do we need to make a note that pybind11 2.9.1 is explicitly supported?

We have a minimum supported version of each dependency, e.g. pybind 2.8.1.
This means that we support all versions of pybind11 greater than or equal to 2.8.1, i.e. `py-pybind11@2.8.1:` is spack-ese.

So if we say in our docs that Arbor supports minimum pybind11 2.8.1, that implies 2.9.1 is supported.

PR #1831, opened 2022-02-03
Update gitmodule `google-benchmark` to v1.6.1

brenthuisman:


bcumming:
This probably isn't too important: google benchmark is used internally by the backend devs to perform microbenchmarks during development. It's not required to support or test Arbor, so can we leave it as is and let the devs update it if they have the need?

PR #1838, opened 2022-02-14
Bit more on Spack, fix in tutorial

brenthuisman:
* Bit more on Spack and how or why to use it, plus an instruction that is required on some systems.
* Fix in network_ring tutorial

### DIFF
--- a/doc/install/spack.rst
+++ b/doc/install/spack.rst
@@ -23,16 +23,16 @@ A powerful feature for users of scientific software is Spack's `Environment feat
 Issues when using Spack
 -----------------------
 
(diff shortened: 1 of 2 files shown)


brenthuisman:
Yeah, elaborating a bit on the use is a good idea! I am not familiar with the build-env feature however. Would you mind typing your proposal out and [adding them to this PR](https://docs.arbor-sim.org/en/latest/contrib/pr.html#collaborating-on-a-pr)?

thorstenhater:
bors try

noraabiakar (approved):
lgtm!

PR #1841, opened 2022-02-17
Allow to clip membrane voltage

### LINKED ISSUE #1843: Restrict membrane within user defined range
Typical computational neuroscience models, like LIF and AdEx, restrict the membrane voltage to below the threshold. To allow to replicate results, e.g., the AdEx Naud patterns (https://dx.doi.org/10.1007%2Fs00422-008-0264-7), it is crucial that the membrane voltage does not exceed the threshold.

#1841 implements clipping of the membrane to user defined ranges.

This functionality is also vital for modelling (analog) neuromorphic circuits that have an intrisic range (like up to 1.8 V).

schmitts:
Typical computational neuroscience models, like LIF and AdEx, restrict the membrane voltage to below the threshold. To allow to replicate results, e.g., the AdEx Naud patterns (https://dx.doi.org/10.1007%2Fs00422-008-0264-7), it is crucial that This functionality is also vital for modelling (analog) neuromorphic circuits that have an intrisic range (like up to 1.8 V). The membrane voltage does not exceed the threshold.

This implements clipping of the membrane to user defined ranges.

Fixes #1843

Edit @thorstenhater: copied description from #1843.

### DIFF
--- a/arbor/fvm_lowered_cell_impl.hpp
+++ b/arbor/fvm_lowered_cell_impl.hpp
@@ -100,6 +100,9 @@ private:
     // Non-physical voltage check threshold, 0 => no check.
     value_type check_voltage_mV_ = 0;
 
+    std::optional<double> membrane_min_;
+    std::optional<double> membrane_max_;
+
     // Flag indicating that at least one of the mechanisms implements the post_events procedure
     bool post_events_ = false;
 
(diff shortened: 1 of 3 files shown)


bcumming:
When making PRs, don't forget to include a description and a link back to any issues associated with the PR so we can follow the relevant discussions.

thorstenhater:
Thanks @schmitts , 

we are going to close in favour of a more general solution (text below from #1843, please continue
the discussion there)

> As of now this hangs in limbo: Clipping Um breaks the cable model, yet there's some need for it.
Your list reads like you need the full cable model plus membrane potential manipulation, so I suggest
we add it in the following way:

    Um can be written from mechanisms (that gives both clamping and clipping) via the ABI
    This mechanism kind is run after the cable step (similar to revpot mech which run at the top)

PR #1846, opened 2022-02-24
add a better error.

thorstenhater:
Since we have support for ACC loading the errors thrown by `paint` are no longer
sufficiently informative. Even before that it was already hard to correlate source 
and locus of painting errors.

Example
```
$> python3 main.C51A.py
  File "main.C51A.py", line 33, in <module>
    cell = A.cable_cell(morph, labels, decor)
RuntimeError: cable_cell: cable (cable 6 0 1) overpaints
```
This adds more details to the error message
```
$> python3 main.C51A.py
  File "main.C51A.py", line 33, in <module>
    cell = A.cable_cell(morph, labels, decor)
RuntimeError: cable_cell: Setting property 'ion-reversal-potential' on region '(region "soma_group")' overpaints at '(cable 6 0 1)'
```

### DIFF
--- a/arbor/cable_cell_param.cpp
+++ b/arbor/cable_cell_param.cpp
@@ -152,37 +152,4 @@ void decor::set_default(defaultable what) {
             what);
 }
 
-std::ostream& operator<<(std::ostream& os, const paintable& item) {
-    std::visit(
-        [&] (const auto& p) {
-            using T = std::decay_t<decltype(p)>;
-            if constexpr (std::is_same_v<init_membrane_potential, T>) {
-                os << "init-membrane-potential";
-            }
-            else if constexpr (std::is_same_v<axial_resistivity, T>) {
-                os << "axial-resistivity";
-            }
-            else if constexpr (std::is_same_v<temperature_K, T>) {
-                os << "temperature-kelvin";
-            }
-            else if constexpr (std::is_same_v<membrane_capacitance, T>) {
-                os << "membrane-capacitance";
-            }
-            else if constexpr (std::is_same_v<init_int_concentration, T>) {
-                os << "ion-internal-concentration";
-            }
-            else if constexpr (std::is_same_v<init_ext_concentration, T>) {
-                os << "ion-external-concentration";
-            }
-            else if constexpr (std::is_same_v<init_reversal_potential, T>) {
-                os << "ion-reversal-potential";
-            }
-            else if constexpr (std::is_same_v<density, T>) {
-                os << "density:" << p.mech.name();
-            }
-        },
-        item);
-    return os;
-}
-
 } // namespace arb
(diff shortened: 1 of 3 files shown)


arbor/cable_cell_param.cpp
```diff
@@ -152,4 +152,37 @@ void decor::set_default(defaultable what) {
             what);
 }
 
+std::ostream& operator<<(std::ostream& os, const paintable& item) {
```
noraabiakar, line 155: FYI, we have a helper in `util` to make this code less verbose: 
```
auto paintable_visitor = arb::util::overload(
     [&](const init_membrane_potential& p) { os << "init-membrane-potential" },
     [&](const axial_resistivity & p) { os << "axial-resistivity"; },
     [&](const temperature_k & p){ os << "temperature-kelvin"; }, 
     ...
);
std::visit(paintable_visitor, item);
```
However, we don't use it in other spots in this file, so it might be better to stick to the same style.

arbor/cable_cell_param.cpp
```diff
@@ -152,4 +152,37 @@ void decor::set_default(defaultable what) {
...
+    std::visit(
+        [&] (const auto& p) {
+            using T = std::decay_t<decltype(p)>;
+            if constexpr (std::is_same_v<init_membrane_potential, T>) {
+                os << "init-membrane-potential";
```
noraabiakar, line 160: We have been printing everything in the decor part of the code as s-expressions, so maybe it's worth sticking to that style? Also we could print the value while we're at it.
Something like `(init-membrane-potential -65)`.

noraabiakar (changes requested):
Looks good thanks! One small suggestion on what we can print.

arbor/cable_cell_param.cpp
thorstenhater: I deliberated on the same question, here's what I swayed me against
1. telling _what_ might be enough information 
2. ...and not overloading the reader, the line is quite long already.
3. also, where to draw the line? `(density ('name' (...)))`? Or print

arbor/cable_cell_param.cpp
thorstenhater: I find the `if constexpr (...)` style slightly better, if it's the same to you.

arbor/cable_cell_param.cpp
noraabiakar: Fair enough. Should we keep it out of the public header as it's not printing the "full" information about the class?

thorstenhater: Ok, good point to think about. It just prints the outer DSL constructor in s-exp-speak.
At some point we had opened an issue on a 'general object printing facility', maybe it's
a good time to take that a bit seriously?

For now, I'll change `<<` to `print_paintable` and hide it.

brenthuisman (approved):
Excellent.

noraabiakar (approved):

Issue #1847, opened 2022-02-24
Add export API developer documentation


noraabiakar:
#1824 added annotations to export symbols in the public (and sometimes private) interface. The description of the PR contains some instructions on how to add the annotations and when. It would be better to have these instructions in the developer documentation.

thorstenhater:
Hey, this seems to have dodged closing in #1849

PR #1849, opened 2022-02-24
export doc

boeschf:
Adding internal developer documentation with respect to the new API macros for symbol visibility (issue #1847).


doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
+.. _export:
+
+Exporting Symbols
+=================
+
+The arbor libraries are compiled with hidden visibility by default. Thus, symbols which are part of
+the public API need to be marked explicitely as visible. Arbor provides a couple of macros to
```
brenthuisman, line 7:
```suggestion
the public API need to be marked explicitly as visible. Arbor provides a couple of macros to
```

doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
...
+annotate functions and classes which are defined in the header file ``export.hpp`` in each library's
+include directory, i.e. ``include/arbor/export.hpp``. These header files are generated at configure
+time based on the build variant, compiler and platform.
+
+Macro Descripiton
+-----------------
+
+.. c:macro:: ARB_LIBNAME_API
+
+    Here ``LIBNAME`` is a placeholder for the library's name: ``ARB_ARBOR_API`` for the main arbor
```
brenthuisman, line 17:
```suggestion
    Here "``LIBNAME``" is a placeholder for the library's name: ``ARB_ARBOR_API`` for the core Arbor
```

doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
...
+    library, ``ARB_ARBORIO_API`` for arborio, etc. This macro is intended to annotate functions,
```
brenthuisman, line 18:
```suggestion
    library, ``ARB_ARBORIO_API`` for Arborio, etc. This macro is intended to annotate functions,
```

doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
...
+    classes and structs which need to be accessible when using the library. Note, that it expands to
```
brenthuisman, line 19:
```suggestion
    classes and structs which need to be accessible when interfacing with the library. Note that it expands to
```
What's 'it' in 'it expands'?

doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
...
+    different values when arbor is being built vs. when arbor is being used by an application. Below
```
brenthuisman, line 20:
```suggestion
    different values when Arbor is being built vs. when Arbor is being used by an application. Below
```

doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
...
+    we list the places where the macro needs to be added or can be safely omitted (we assume all of
+    the symbols below are part of the public API).
+
+    .. code-block:: cpp
+        :caption: header.hpp
+
+        // free function declaration
```
brenthuisman, line 27: What's a 'free function'?

doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
...
+        ARB_ARBOR_API void foo();
+
+        // free function (inline)
```
brenthuisman, line 30: 'inline' refers to not being exposed/hidden? If so, prefer 'hidden'. If not, educate me :)

doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
...
+        void bar(int i) { /* ... */ }
+
+        // function template (inline)
+        template<typename T>
+        void baz(T i) { /* ... */ }
+
+        // class declaration
+        // note: this will make all member symbols visible
```
brenthuisman, line 38: member symbols will become visible as per class visibility directives or do you mean that those are overridden as well?

doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
...
+-----------------
+
+.. c:macro:: ARB_LIBNAME_API
+
+    Here ``LIBNAME`` is a placeholder for the library's name: ``ARB_ARBOR_API`` for the main arbor
+    library, ``ARB_ARBORIO_API`` for arborio, etc. This macro is intended to annotate functions,
+    classes and structs which need to be accessible when using the library. Note, that it expands to
+    different values when arbor is being built vs. when arbor is being used by an application. Below
+    we list the places where the macro needs to be added or can be safely omitted (we assume all of
+    the symbols below are part of the public API).
```
brenthuisman, line 22: public = not hidden? Earlier you've used 'visible' as opposite for 'hidden'. Prefer consistency (I see the GCC page does not set a good example either).

doc/internals/export.rst
```diff
@@ -0,0 +1,98 @@
...
+
+
+.. c:macro:: ARB_SYMBOL_VISIBLE
+
+    Objects which are type-erased and passed across the library boundaries sometimes need runtime
+    type information (rtti). In particular, exception classes and classes stored in ``std::any`` or
+    similar need to have the correct runtime information attached. Hidden visibility strips away
+    this information which leads to all kind of unexpected behaviour. Therefore, all such classes
+    must be annotated with this macro which guarantees that the symbol is always visible. Note, it
+    is not enough to use the first macro for these cases.
```
brenthuisman, line 84: To avoid confusion: when you say that the above macro isn't enough, do you mean that the following is an additional specifier or that this should be used instead? Looking at the example, probably instead, but the language could be ambiguous.

brenthuisman (changes requested):
A few small changes + questions.

doc/internals/export.rst
boeschf: inline means, the function body (the definition) of the function resides in the header (it is compiled anew for each usage of Arbor). Conversely non-inline functions are compiled once and are part of the pre-compiled library (only the declaration is in the header).

doc/internals/export.rst
boeschf: not a class member

doc/internals/export.rst
boeschf: yes I mean: should be used instead. Will clarify

doc/internals/export.rst
boeschf: I don't quite understand. All members of a class (member functions and members) will be implicitly annotated with hidden=default (i.e. visible) if the class is made visible through the macro.

doc/internals/export.rst
boeschf: public in the sense of accessible (vs. private, implementation details). Public/private is an orthogonal concept to hidden/visible. Both public and private symbols are affected by visibility. I could replace "public API" with user-facing API, would that help? I don't think I've used public elsewhere to describe non-hidden - that would indeed be confusing.

doc/internals/export.rst
brenthuisman: OK!

doc/internals/export.rst
brenthuisman: Of course, the absence of `ARB_ARBOR_API` ticked me off.

Then there are two differences with the 'free function': absence of the annotation and the presence of a body. Whether a free function is inline or not, does not affect how to annotate, right? If so, maybe leave it out.

doc/internals/export.rst
brenthuisman:
```
class B {
   int myint;
};
```
`myint` is hidden because class members are hidden by default. `B` is hidden from the public API, because it isn't annotated to be visible. Right?
```
class ARB_ARBOR_API B {
   int myint;
};
```
I would expect `myint` to be hidden, even though having annotated `B` to be visible. Is that so?

doc/internals/export.rst
boeschf: it does affect annotation:
- inline free function (definition in header): don't annotate (is visible no matter what)
- free function (declaration in header, definition in source): must annotate both declaration and definition (is hidden otherwise)

brenthuisman: Can you put those bullets in the documentation? Very clear :)

doc/internals/export.rst
brenthuisman: It's clear now that I am not an expert. My knowledge is limited to private/public/protected properties/methods, which are sometimes described as visibility, and exported symbols (or symbols not exported, which you might describe as hidden).
It seems that hidden/visible has another specific meaning which I struggle to disentangle from what one learns in every first year C++ course.

Is it possible to have a line or two right at the beginning of the page that clarify the difference?

boeschf:
I rewrote part of the text and removed the word "inline" and replaced with more pertinent declaration/definition.

doc/internals/export.rst
boeschf: I did add 2 bullets in the text - let me know if that's ok

doc/internals/export.rst
boeschf: done

doc/internals/export.rst
```diff
@@ -0,0 +1,134 @@
...
+
+.. Note::
+    When linking an application with **static** Arbor libraries the linker may issue warnings (particularly on macos).  Thus, if you encounter problems, try building shared Arbor libraries (cmake option ``-DBUILD_SHARED_LIBS=ON``) instead.
+
+Macro Descripiton
+-----------------
+
+.. c:macro:: ARB_LIBNAME_API
+
+    Here ``LIBNAME`` is a placeholder for the library's name: ``ARB_ARBOR_API`` for the core Arbor
```
brenthuisman, line 31:
```suggestion
    Here "``LIBNAME``" is a placeholder for the library's name: ``ARB_ARBOR_API`` for the core Arbor
```
Think you accidentally deleted them.

doc/internals/export.rst
```diff
@@ -0,0 +1,134 @@
...
+    classes and structs which need to be accessible when interfacing with the library. Note that it
+    expands to different values when Arbor is being built vs. when Arbor is being used by an
+    application.
+    
+    Below we list the places where the macro needs to be added or can be safely omitted (we assume
+    all of the symbols below are part of the public/user-facing API).  In general, annotation is
+    required
+    
+    * for **declarations in header files** which are not definitions
+    * for **defintions** of functions, friend functions and (extern) variables **in source files**
```
brenthuisman, line 42:
```suggestion
    * for **definitions** of functions, friend functions and (extern) variables **in source files**
```

doc/internals/export.rst
brenthuisman: Yes!

doc/internals/export.rst
brenthuisman: Thanks!

doc/internals/export.rst
```diff
@@ -0,0 +1,134 @@
...
+        void baz(T i) { /* ... */ }
+
+        // class definition
+        // class member declaration
+        class ARB_ARBOR_API A {
+            A();
+            friend std::ostream& operator<<(std::ostream& o, A const & a);
+        };
+
+        // class defintion
```
brenthuisman, line 74:
```suggestion
        // class definition
```

doc/internals/export.rst
```diff
@@ -0,0 +1,134 @@
...
+        // class member defintion
```
brenthuisman, line 75:
```suggestion
        // class member definition
```
:)

doc/internals/export.rst
```diff
@@ -0,0 +1,134 @@
...
+        class B {
+            B() { /* ... */ }
+        };
+
+        // template class defintion
```
brenthuisman, line 80:
```suggestion
        // template class definition
```

doc/internals/export.rst
```diff
@@ -0,0 +1,134 @@
...
+        ARB_ARBOR_API extern int h;
+
+
+    .. code-block:: cpp
+        :caption: source.cpp
+
+        // free function definition
+        ARB_ARBOR_API void foo() { /* ... */ }
+
+        // class member defintion (will be visible since A is visible)
```
brenthuisman, line 98:
```suggestion
        // class member definition (will be visible since A is visible)
```

doc/internals/export.rst
```diff
@@ -0,0 +1,134 @@
...
+        A::A() { /* ... */ }
+
+        // friend function defintion
```
brenthuisman, line 101:
```suggestion
        // friend function definition
```

brenthuisman (changes requested):
I approve, just a few instances of a typo, and some ticks that dropped. Thanks!

brenthuisman (approved):

PR #1854, opened 2022-02-28
Make brunel.py setup faster.

thorstenhater:
Small tweak, gets 10% setup time reduction for 500 cells. Will be more w/ larger networks.

### DIFF
--- a/python/example/brunel.py
+++ b/python/example/brunel.py
@@ -21,13 +21,14 @@ Call with parameters, for example:
 # Samples m unique values in interval [start, end) - gid.
 # We exclude gid because we don't want self-loops.
 def sample_subset(gid, start, end, m):
-    gen = numpy.random.RandomState(gid+42)
-    s = set()
-    while len(s) < m:
-        val = gen.randint(low=start,high=end)
-        if val != gid:
-            s.add(val)
-    return s
+    end = min(end, start + m)
+    idx = numpy.arange(start, end + 1)
+    if start < gid < end:
+        numpy.delete(idx, gid - start)
+    else:
+        numpy.delete(idx, end - start)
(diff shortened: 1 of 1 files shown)


python/example/brunel.py
```diff
@@ -21,13 +21,14 @@
...
-    s = set()
-    while len(s) < m:
-        val = gen.randint(low=start,high=end)
-        if val != gid:
-            s.add(val)
-    return s
+    end = min(end, start + m)
+    idx = numpy.arange(start, end + 1)
+    if start < gid < end:
+        numpy.delete(idx, gid - start)
```
brenthuisman, line 27:
```suggestion
        idx = numpy.delete(idx, gid - start)
```
`.delete()` is not in-place. Also, due to the `+1` above, `end` will occur in the list, right? The function description says it returns on interval `[start,end)`.

python/example/brunel.py
```diff
@@ -21,13 +21,14 @@
...
+    else:
+        numpy.delete(idx, end - start)
```
brenthuisman, line 29:
```suggestion
        idx = numpy.delete(idx, end - start)
```

brenthuisman (changes requested):
Not too sure about including `end` in case idx is in range.

python/example/brunel.py
thorstenhater: Yes. We over-allocate one item and delete one. Makes net zero.

brenthuisman: Length, yes, but at the cost of returning values in interval `[start, end]`, no? Maybe it's not a problem?

thorstenhater: As stands this is completly busted. It'll return something in range `[start, start + m)`.

brenthuisman (approved):

PR #1859, opened 2022-03-10
Fix load_component for label-dict

noraabiakar:
The `cable_component::component` property in pyarb returns a `cable_cell_variant` object which is an alias for`std::variant<arb::morphology, arb::label_dict, arb::decor, arb::cable_cell>`.
`pybind11` is usually able to present an `std::variant` object to a python user as the underlying type of the object, but only if there exists a `pybind11::class` representation of said type. 
In this case, `pybind11` doesn't have a class representation for `arb::label_dict`, it has instead a class representation for `pyarb::label_dict_proxy`. This discrepancy makes it not possible to use `load_component` to load a `label_dict` using the pyarb library.  This PR fixes the issue.

### DIFF
--- a/python/cable_cell_io.cpp
+++ b/python/cable_cell_io.cpp
@@ -5,12 +5,14 @@
 #include <iomanip>
 
 #include <arbor/cable_cell.hpp>
+#include <arbor/util/any_visitor.hpp>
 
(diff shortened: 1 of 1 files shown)


thorstenhater (approved):
👍

PR #1866, opened 2022-03-22
modcc: generate missing `node_index` read needed for reading time `t` in the mechanisms

noraabiakar:
Rectify the assumption that the `node_index` is read in every mechanism call back where time is also read.


thorstenhater (approved):
👍

thorstenhater:
Fix

PR #1867, opened 2022-03-23
Support markers in neurolucida ascii files

bcumming:
* add support for parsing markers defined in asc files
* add hitherto unknown "High" leaf branch type


lukasgd (approved):
Thanks for this, works with [BluePyOpt/examples/l5pc/morphology/C060114A7.asc](https://github.com/BlueBrain/BluePyOpt/blob/8e158e618d9569f6ec0097ab8018ca10af860365/examples/l5pc/morphology/C060114A7.asc).

Issue #1749, opened 2021-11-08
Delayed processing of pre-synaptic spikes


jlubo:
I need an interface for delayed processing of pre-synaptic spikes. The enhancement should enable a pre-synaptic spike to trigger multiple events in the same post-synaptic neuron at different delay times (in my case, specifically, these are a postsynaptic potential after short delay and an increase in calcium concentration after longer delay, as described [here](https://www.nature.com/articles/s42003-021-01778-y)).

To summarize, it would be great to have something like the `POST_EVENT` block with its `time` argument, just that it should be called at a certain time after a *pre-synaptic* spike.

jlubo:
I don't think so - those synapses would have to communicate somehow (the calcium is needed for changes in synaptic weight). 

Also, even if such a workaround were possible, there would be a lot of redunancy so I guess a proper implementation would be better in terms of performance.

jlubo:
We found that the following workaround does the job. 

We used two connections to the same postsynaptic neuron with different (pseudo) weights `1` and `-1` and different delay times `d0` and `d1`:
```
[arbor.connection((src,"spike_detector1"), "syn_exc_calcium_plasticity", 1, d0), \
arbor.connection((src,"spike_detector1"), "syn_exc_calcium_plasticity", -1, d1)]
```
which can be processed in one and the same synapse mechanism `syn_exc_calcium_plasticity` as follows:
```
NET_RECEIVE(weight) {
	if (weight > 0) {
		: postsynaptic potential evoked by presynaptic spike
	}
	else {
		: increase of calcium amount evoked by presynaptic spike
	}
}
```
Should we close this issue?

schmitts:
I think we can close the issue and think about writing an AEP for allowing additional payloads/slots in connections.

thorstenhater:
As this is related to a PR is this still needed? If not, please dismiss this and the PR.

PR #1868, opened 2022-03-24
New policy 'round_robin_halt'

jlubo:
Introduces the new policy `round_robin_halt`. This enables to halt at the current item until the policy `round_robin` is called again.

Related to issue #1749.

### DIFF
--- a/arbor/common_types_io.cpp
+++ b/arbor/common_types_io.cpp
@@ -8,6 +8,8 @@ ARB_ARBOR_API std::ostream& operator<<(std::ostream& o, lid_selection_policy pol
     switch (policy) {
     case lid_selection_policy::round_robin:
         return o << "round_robin";
+	case lid_selection_policy::round_robin_halt:
+        return o << "round_robin_halt";
     case lid_selection_policy::assert_univalent:
         return o << "univalent";
     }
--- a/arbor/include/arbor/common_types.hpp
+++ b/arbor/include/arbor/common_types.hpp
@@ -70,6 +70,7 @@ struct lid_range {
 
 enum class lid_selection_policy {
     round_robin,
+	round_robin_halt,
     assert_univalent // throw if the range of possible lids is wider than 1
 };
 
--- a/arbor/label_resolution.cpp
+++ b/arbor/label_resolution.cpp
@@ -126,6 +126,11 @@ lid_hopefully round_robin_state::update(const label_resolution_map::range_set& r
     return lid;
 }
 
+lid_hopefully round_robin_halt_state::update(const label_resolution_map::range_set& range_set) {
+    auto lid = range_set.at(state);
+    return lid;
+}
+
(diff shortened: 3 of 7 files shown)


brenthuisman:
Hi @jlubo, sorry for taking a while to respond. Is it feasible to add a unit test for this?

brenthuisman:
Great, thanks! I see an error (twice) in the run:
```
ERROR: test_multiple_connections (test.unit.test_multiple_connections.TestMultipleConnections)
----------------------------------------------------------------------
  File "/home/runner/work/arbor/arbor/python/test/fixtures.py", line 30, in wrapper
    return func(*args, **kwargs)
  File "/home/runner/work/arbor/arbor/python/test/fixtures.py", line 30, in wrapper
    return func(*args, **kwargs)
  File "/home/runner/work/arbor/arbor/python/test/fixtures.py", line 30, in wrapper
    return func(*args, **kwargs)
  File "/home/runner/work/arbor/arbor/python/test/unit/test_multiple_connections.py", line 116, in test_multiple_connections
    data_mem, _ = sim.samples(handle_mem)[0]
IndexError: list index out of range
```

This is because you are collecting all results on one node. Unfortunately, that won't work; through `samples` you can only access results that were computed on the node your script is running, see also [this page in the docs](https://docs.arbor-sim.org/en/latest/tutorial/network_ring_mpi.html#the-results). At the moment, there is no other solution than to collect the results yourself: i.e. return the results and collect, or write to disk and read back. See the above documentation for how you could handle it.

I admit this isn't great, but for the moment there is no way around it.

jlubo:
Thanks @brenthuisman for pointing that out. To avoid the error in MPI setups, I've now added a condition that prevents running the related assertion (which is anyway not crucial for this test) if the data is not available on the current node. In the future this condition may be removed if #1892 is solved.

jlubo:
The problem has been quite hard to wrap my head around... It seems that the tests (as of commit [417bd5e](https://github.com/arbor-sim/arbor/pull/1868/commits/417bd5e00c74aa0a55cdb971012068990b3e1294)) did sometimes not pass because the relationship between `lid` and `state` in `label_resolution.cpp` can differ across machines. On my local machine, for example, if `round_robin_state.state == 0` then `lid.value() == 1`. On the machine running the PR check, however, `round_robin_state.state == 0` corresponded to `lid.value() == 0`, causing different synapses to be targeted and leading to different results.

This problem is now (as of commit [a8627a5](https://github.com/arbor-sim/arbor/pull/1868/commits/a8627a57030d07fde79a80b6c999edec80c8b905)) avoided by considering instances of synapse mechanisms that are indistinguishable upon initialization (when the `lid` indices are determined) before evolving differently. In order to test the effect of the `round_robin_halt` policy, it is necessary to observe such differences between the instances of the same mechanism. I have now implemented them by using a synapse mechanism that features plasticity (`expsyn_stdp`).

boeschf (approved):

PR #1872, opened 2022-03-25
random123 submodule added

brenthuisman:
Due to limited history in the random123 repo, it's checked out at the last (and only) released version, 1.14.0.

As per https://github.com/arbor-sim/arbor/issues/1731.


brenthuisman:
bors try

brenthuisman:
bors try

bcumming:
It looks like the `spack/package.py` has to be udpated to use random123 provided by spack.

brenthuisman:
@bcumming Yes! However, the version that exists on Github is not yet in the Spackage. I'll make a PR for it.

thorstenhater (commented):
I remember @noraabiakar had to patch some files in order for GPU support to work.
How was this resolved?

brenthuisman:
Apparently by just updating `random123`. Those patches were for an earlier version and we said that if bors passes we don't need them anymore. I'll ask @noraabiakar to make sure.

brenthuisman:
@noraabiakar writes that we should test AMD GPUs, for which there is no CI at the moment. @bcumming are you able to run those tests?

thorstenhater:
@boeschf has volunteered to run this.

PR #1873, opened 2022-03-25
Add support for epoch callbacks

### LINKED ISSUE #1870: Progress measure
It would good to be able to measure, or display, the progress of a running simulation. 

This could, for example, happen through a string that contains the current progress percentage and that is written to an output stream specified by an argument of `arbor.simulation.run()`. A more sophisticated solution would be a Python function that would be called periodically and that receives the progress percentage as an argument.

As @schmitts pointed out, inspiration could also come from Brian2: https://brian2.readthedocs.io/en/stable/user/running.html#progress-reporting.

bcumming:
Adds support for the user to add a generic callback that is to be called:
1. before the first time step
2. at the end of every epoch

One implementation of such a callback that prints a progress bar to the screen is provided, and wrapped for the Python front end. Example output of the progress bar:

```
 57% [||||||||||||||||||||||||||||                      ]           575ms
```

Fixes #1870

### DIFF
--- a/arbor/simulation.cpp
+++ b/arbor/simulation.cpp
@@ -578,24 +578,30 @@ simulation::~simulation() = default;
 ARB_ARBOR_API epoch_function epoch_progress_bar() {
     struct impl {
         double t0 = 0;
-        unsigned calls = 0;
-
-        const char* PBSTR = "||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||";
-        unsigned PBWIDTH = 50;
+        bool first = true;
 
         void operator() (double t, double tfinal) {
-            if (calls==0) {
+            constexpr unsigned bar_width = 50;
+            static const std::string bar_buffer(bar_width+1, '-');
+
+            if (first) {
+                first = false;
                 t0 = t;
             }
 
             double percentage = (tfinal==t0)? 1: (t-t0)/(tfinal-t0);
             int val = percentage * 100;
-            int lpad = percentage * PBWIDTH;
-            int rpad = PBWIDTH - lpad;
-            printf("\r%3d%% [%.*s%*s]  %12ums", val, lpad, PBSTR, rpad, "", (unsigned)t);
-            fflush(stdout);
+            int lpad = percentage * bar_width;
+            int rpad = bar_width - lpad;
+            printf("\r%3d%% |%.*s%*s|  %12ums", val, lpad, bar_buffer.c_str(), rpad, "", (unsigned)t);
 
-            ++calls;
+            if (t==tfinal) {
+                // Print new line and reset counters on the last step.
+                printf("\n");
+                t0 = tfinal;
+                first = true;
+            }
+            fflush(stdout);
         }
     };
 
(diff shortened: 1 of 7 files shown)


arbor/simulation.cpp
```diff
@@ -558,10 +565,41 @@ void simulation::set_local_spike_callback(spike_export_function export_callback)
...
 
 simulation::~simulation() = default;
 
+epoch_function epoch_progress_bar() {
+    struct impl {
+        double t0 = 0;
+        unsigned calls = 0;
+
+        const char* PBSTR = "||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||";
+        unsigned PBWIDTH = 50;
```
jlubo, line 584:
```suggestion
        unsigned PBWIDTH = 50;
        const std::string PBSTR = std::string(2*PBWIDTH, '|');
```
Using the "repeat character" constructor of `std::string`.

jlubo (changes requested):
The callback and the Python wrapper seem to work well, and the progress bar looks fine!

See below for two minor suggestions. Also, documentation is still missing.

jlubo (changes requested):

arbor/simulation.cpp
```diff
@@ -558,10 +565,41 @@ void simulation::set_local_spike_callback(spike_export_function export_callback)
...
+            if (calls==0) {
+                t0 = t;
+            }
+
+            double percentage = (tfinal==t0)? 1: (t-t0)/(tfinal-t0);
+            int val = percentage * 100;
+            int lpad = percentage * PBWIDTH;
+            int rpad = PBWIDTH - lpad;
+            printf("\r%3d%% [%.*s%*s]  %12ums", val, lpad, PBSTR, rpad, "", (unsigned)t);
+            fflush(stdout);
```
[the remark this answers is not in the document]
bcumming: :+1: 

I tweaked it to print a new line and reset the counter so that code like this:
```python
sim.run(1000, 0.1)
sim.run(2000, 0.1)
sim.run(3000, 0.1)
```
will generate the following output:
```
100% |--------------------------------------------------|          1000ms
100% |--------------------------------------------------|          2000ms
100% |--------------------------------------------------|          3000ms
```

arbor/simulation.cpp
jlubo: Seems like the best solution!

jlubo (approved):

arbor/simulation.cpp
```diff
@@ -558,10 +565,47 @@ void simulation::set_local_spike_callback(spike_export_function export_callback)
     impl_->local_export_callback_ = std::move(export_callback);
 }
 
+void simulation::set_epoch_callback(epoch_function epoch_callback, bool global) {
```
thorstenhater, line 568: `global` is redundant? Or is this compatibility with `spike_callback`?

thorstenhater (changes requested):
One tiny comment, rest is good.

arbor/simulation.cpp
bcumming: Good point. That is a hang over from an earlier draft of the epoch callback.
I realised that the `global` information is redundant, and the caller can decide whether to attach a callback on one or more distributed contexts for themselves, e.g.:
https://github.com/arbor-sim/arbor/pull/1873/files#diff-5827156d5c4d2c924b0def9bbf4e51db85e1ea4a0c45c75328ca38dab6ea8e12R190

thorstenhater (approved):
👍

PR #1882, opened 2022-04-12
Sort (in python) the mechanism names (for convenience).

### LINKED ISSUE #1879: `cable_global_properties.catalogue` assignment changes iteration order, is surprising
**Describe the bug**
When assigning to the `catalogue` property of a `cable_global_properties` object, it seems to create and store a copy of the original object, which has some surprising side effects: it invalidates and shuffles the otherwise seemingly deterministic order of the mech cat iterator:

```
>>> list(arbor.load_catalogue("lib.so"))
["A", "B", "C", "D"]
>>> list(arbor.load_catalogue("lib.so"))
["A", "B", "C", "D"]
>>> list(arbor.load_catalogue("lib.so"))
["A", "B", "C", "D"]
>>> cat = arbor.load_catalogue("lib.so"); print(list(cat)); print(list(cat)); print(list(cat))
["A", "B", "C", "D"]
["A", "B", "C", "D"]
["A", "B", "C", "D"]
>>> props = arbor.cable_global_properties(); props.catalogue = cat; print(list(props.catalogue))
["C", "B", "D", "A"]
```

**Context**
Whilst @thorstenhater will probably argue that it is a bag, or a set, or a dragon, and order is not guaranteed 😉 it is still an open trapdoor since users (read: me (and @noraabiakar )) build up an assumption through use of the catalogues and its iterators, that the order is preserved. When then accessing the "same" catalogue through `recipe.properties.catalogue` for example, another iteration order is obtained. This led to catastrophy in the benchmarks, where the recorded samples and mechanism label were mismatched when zipped together.

Please do not propose to shuffle the return results of the mech cat iterator 👉 👈

thorstenhater:
Return mechanism names in sorted order.

Fixes #1879

### DIFF
--- a/python/mechanism.cpp
+++ b/python/mechanism.cpp
@@ -113,7 +113,9 @@ void register_mechanisms(pybind11::module& m) {
     pybind11::class_<arb::mechanism_catalogue> cat(m, "catalogue");
 
     struct mech_cat_iter_state {
-        mech_cat_iter_state(const arb::mechanism_catalogue &cat_, pybind11::object ref_): names(cat_.mechanism_names()), ref(ref_), cat(cat_) { }
+        mech_cat_iter_state(const arb::mechanism_catalogue &cat_, pybind11::object ref_): names(cat_.mechanism_names()), ref(ref_), cat(cat_) {
+            std::sort(names.begin(), names.end());
+        }
         std::vector<std::string> names;      // cache the names else these will be allocated multiple times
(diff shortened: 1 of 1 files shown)


brenthuisman:
The bug report states that order is not preserved when a catalog is assigned. This PR sorts them, which is not the same as preserving order. Maybe the bug reporter no longer requires this?

brenthuisman (approved):
lgtm

Issue #1650, opened 2021-09-08
AEP: Inhomogeneous parameters


halfflat:
This is a proposal to supply functionality similar to the[ inhomogeneous parameters](https://neuroml.org/NeuroML2CoreTypes/Cells.html#inhomogeneousParameter) of NeuroML2 and which can be used to describe and simulate models such as described in #1582.

Use case is basically to support models where a (range) parameter of a density mechanism can vary as a function of position.

### Proposal

When a density mechanism is painted on a cable cell in a decor, we allow zero or more _inhomogeneous scaling expressions_ that apply a position-dependent scaling factor to one of the mechanism's range parameters. These scaling expressions are given by a parameter name and an _iexpr_ that describes the scaling factor.

Example:
```
    decor D;
    iexpr x = …;
    iexpr y = …;
    
    D.paint(region1, "hh", scale_parameter("gnabar", x), scale_parameter("gkbar", y));
    D.paint(region2, mechanism_desc("hh").set("gnabar", 1.3), scale_parameter("gnabar", x));
```

**Why not put these expressions in `mechanism_desc`?** On one hand, using a scalar scaling parameter means we don't need to worry about the variety of units that a mechanism parameter might be expressed in. On the other, it allows mechanism description objects to remain as simple collections of a mechanism name together with a set of parameter value overrides, which will also suit the proposed rationalization of global parameter settings mooted in #1226.

#### Inhomogeneous expressions

Inhomogeneous scaling expressions are of class type `arb::iexpr` and can be constructed from static methods in this class. There are two sorts of `iexpr` constructions: arithmetical expressions that combine scalar values and other inhomogeneous expressions; and scaling expressions that are functions of position in the morphology. Scaling expressions combine some sort of description of a distance, or displacement, or radius, or similar quantity with an implicit or explicit _scale_ factor (implicitly in units of μm⁻¹, at least until we have unit support in the library) that turns it into a scalar value.

In the following descriptions of scaling expressions, d(*x*, *y*) denotes the minimum path length between two morphological points *x* and *y*, and *y*-*x* denotes the signed displacement from *x* to *y*, which is equal to d(*x*, *y*) if *x*≤*y*, -d(*x*, *y*) if *y*≤*x*, or else is undefined if *x* and *y* are not comparable in the morphological partial order.

Not all scaling expressions are defined over the full morphology: the region in which they are well-defined will be called the _domain_ of the expression.

Proposed scaling expressions:
[...]

PR #1887, opened 2022-04-22
Inhomogeneous parameters

AdhocMan:
Implementation of #1650.

This PR adds inhomogeneous expressions (`iexpr`) and a wrapper struct `scaled_mechanism` (naming suggestions are welcome), to allow painting of mechanisms (currently limited to density), where selected parameters can be scaled by `iexpr` evaluated along a cell. The wrapper allows for the current mechanism API to remain unchanged.
Any `iexpr` can be converted from / to  S-Expressions and added to the label dictionary.

A python example:
```
  # Create mechanism with custom conductance (range)
  m = arbor.mechanism('pas', {'g': 0.1})

  decor = arbor.decor()
  # paint a scaled density mechanism, where 'g' is scaled by the radius along the cell.
  decor.paint('"soma"', arbor.scaled_mechanism(arbor.density(m), {'g': '(add 2.0 (radius 0.5) (exp (distance 3.0 (terminal))))'}))

```

### DIFF
--- a/arbor/cable_cell.cpp
+++ b/arbor/cable_cell.cpp
@@ -165,7 +165,7 @@ struct cable_cell_impl {
       mextent cables = thingify(reg, provider);
       auto &mm = get_region_map(prop.t_mech);
 
-      std::unordered_map<std::string, std::shared_ptr<iexpr_interface>> im;
+      std::unordered_map<std::string, iexpr_ptr> im;
       for(const auto& it : prop.scale_expr) {
           im.insert_or_assign(it.first, thingify(it.second, provider));
       }
--- a/arbor/cable_cell.cpp
+++ b/arbor/cable_cell.cpp
@@ -1,3 +1,4 @@
+#include <memory>
 #include <sstream>
 #include <unordered_map>
 #include <variant>
@@ -15,9 +16,6 @@
 #include "util/strprintf.hpp"
 
 namespace arb {
-
-using region_map = std::unordered_map<std::string, mcable_list>;
-using locset_map = std::unordered_map<std::string, mlocation_list>;
 
 using value_type = cable_cell::value_type;
 using index_type = cable_cell::index_type;
@@ -138,8 +136,9 @@ struct cable_cell_impl {
         return region_map.get<T>();
     }
 
-    mcable_map<density>& get_region_map(const density& desc) {
-        return region_map.get<density>()[desc.mech.name()];
+    mcable_map<std::pair<density, iexpr_map>> &
+    get_region_map(const density &desc) {
+      return region_map.get<density>()[desc.mech.name()];
     }
 
     mcable_map<init_int_concentration>& get_region_map(const init_int_concentration& init) {
@@ -158,8 +157,32 @@ struct cable_cell_impl {
         return region_map.get<init_reversal_potential>()[init.ion];
     }
 
-    template <typename Property>
-    void paint(const region& reg, const Property& prop) {
+    void paint(const region& reg, const density& prop) {
+        this->paint(reg, scaled_mechanism<density>(prop));
+    }
+
+    void paint(const region &reg, const scaled_mechanism<density> &prop) {
+      mextent cables = thingify(reg, provider);
+      auto &mm = get_region_map(prop.t_mech);
+
+      std::unordered_map<std::string, std::shared_ptr<iexpr_interface>> im;
+      for(const auto& it : prop.scale_expr) {
+          im.insert_or_assign(it.first, thingify(it.second, provider));
+      }
+
+      for (auto c : cables) {
+        // Skip zero-length cables in extent:
+        if (c.prox_pos == c.dist_pos)
+          continue;
+
+        if (!mm.insert(c, {prop.t_mech, im})) {
+          throw cable_cell_error(util::pprintf("cable {} overpaints", c));
+        }
+      }
+    }
+
+    template <typename TaggedMech>
+    void paint(const region& reg, const TaggedMech& prop) {
         mextent cables = thingify(reg, provider);
         auto& mm = get_region_map(prop);
 
--- a/arbor/cable_cell.cpp
+++ b/arbor/cable_cell.cpp
@@ -161,24 +161,23 @@ struct cable_cell_impl {
         this->paint(reg, scaled_mechanism<density>(prop));
     }
 
-    void paint(const region &reg, const scaled_mechanism<density> &prop) {
-      mextent cables = thingify(reg, provider);
-      auto &mm = get_region_map(prop.t_mech);
+    void paint(const region& reg, const scaled_mechanism<density>& prop) {
+        mextent cables = thingify(reg, provider);
+        auto& mm = get_region_map(prop.t_mech);
 
-      std::unordered_map<std::string, iexpr_ptr> im;
-      for(const auto& it : prop.scale_expr) {
-          im.insert_or_assign(it.first, thingify(it.second, provider));
-      }
-
-      for (auto c : cables) {
-        // Skip zero-length cables in extent:
-        if (c.prox_pos == c.dist_pos)
-          continue;
-
-        if (!mm.insert(c, {prop.t_mech, im})) {
-          throw cable_cell_error(util::pprintf("cable {} overpaints", c));
+        std::unordered_map<std::string, iexpr_ptr> im;
+        for (const auto& [fst, snd]: prop.scale_expr) {
+            im.insert_or_assign(fst, thingify(snd, provider));
         }
-      }
+
+        for (const auto& c: cables) {
+            // Skip zero-length cables in extent:
+            if (c.prox_pos == c.dist_pos) continue;
+
+            if (!mm.insert(c, {prop.t_mech, im})) {
+                throw cable_cell_error(util::pprintf("cable {} overpaints", c));
+            }
+        }
     }
 
     template <typename TaggedMech>
--- a/arbor/fvm_layout.cpp
+++ b/arbor/fvm_layout.cpp
@@ -932,13 +932,12 @@ fvm_mechanism_data fvm_build_mechanism_data(
 
         param_maps.resize(n_param);
 
-        for (auto& on_cable: cables) {
-            const auto& mech = on_cable.second.mech;
-            verify_mechanism(info, mech);
-            mcable cable = on_cable.first;
-            const auto& set_params = mech.values();
+        for (const auto& [cable, density_iexpr]: cables) {
+            const auto& mech = density_iexpr.first.mech;
+            const auto& scale_expr = density_iexpr.second;
 
-            const auto& scale_expr = on_cable.second.second;
+            verify_mechanism(info, mech);
+            const auto& set_params = mech.values();
 
             support.insert(cable, 1.);
             for (std::size_t i = 0; i<n_param; ++i) {
--- a/arbor/fvm_layout.cpp
+++ b/arbor/fvm_layout.cpp
@@ -900,7 +900,7 @@ fvm_mechanism_data fvm_build_mechanism_data(
 
     std::unordered_map<std::string, mcable_map<double>> init_iconc_mask;
     std::unordered_map<std::string, mcable_map<double>> init_econc_mask;
-    std::shared_ptr<iexpr_interface> unit_scale = thingify(iexpr::scalar(1.0), cell.provider());
+    iexpr_ptr unit_scale = thingify(iexpr::scalar(1.0), cell.provider());
 
     // Density mechanisms:
     for (const auto& [name, cables]: assignments.get<density>()) {
@@ -928,7 +928,7 @@ fvm_mechanism_data fvm_build_mechanism_data(
         }
 
         mcable_map<double> support;
-        std::vector<mcable_map<std::pair<double, std::shared_ptr<iexpr_interface>>>> param_maps;
+        std::vector<mcable_map<std::pair<double, iexpr_ptr>>> param_maps;
 
         param_maps.resize(n_param);
 
--- a/arbor/fvm_layout.cpp
+++ b/arbor/fvm_layout.cpp
@@ -5,9 +5,12 @@
 #include <unordered_set>
 #include <unordered_map>
 #include <vector>
+#include <utility>
+#include <memory>
 
 #include <arbor/arbexcept.hpp>
 #include <arbor/cable_cell.hpp>
+#include <arbor/iexpr.hpp>
 #include <arbor/math.hpp>
 #include <arbor/morph/mcable_map.hpp>
 #include <arbor/morph/mprovider.hpp>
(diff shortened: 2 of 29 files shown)


doc/concepts/labels.rst
```diff
@@ -608,6 +618,120 @@ Region expressions
...
+      :width: 600
+      :align: center
+
+      Example of an interpolation expression. **Left**: A single proximal location locset. **Middle**: Terminal locations as distal locset.
+      **Right**: Area, where the expression evaluates to non-zero values.
+
+.. label:: (interpolation prox_value:real prox_reg:region dist_value:real dist_reg:region)
+
+    Interpolates between the region ``prox_reg`` in proximal diretion and the region ``dist_reg`` in distal direction
+    with the assosiated unitless values ``prox_value`` and ``dist_value``. If evaluated inside either region, returns the corresponding value.
```
thorstenhater, line 689: Typo: `associated`

doc/concepts/decor.rst
```diff
@@ -166,7 +167,24 @@ Take for example the built-in mechanism for passive leaky dynamics:
...
+
+.. code-block:: Python
+
+    # Create mechanism with custom conductance (range)
+    m = arbor.mechanism('pas', {'g': 0.1})
+
+    decor = arbor.decor()
+    # paint a scaled density mechanism, where 'g' is scaled by the radius along the cell.
+    decor.paint('"soma"', arbor.scaled_mechanism(arbor.density(m), {'g': '(radius 0.5)'}))
+
```
brenthuisman, line 186: Maybe add a slightly more complex example (e.g. involving interpolation/distance).

doc/concepts/labels.rst
```diff
@@ -608,6 +618,120 @@ Region expressions
...
+    Interpolates between the closest point in proximal direction in locset ``prox_loc`` and the closest point in 
+    distal direction ``dist_loc`` with the assosiated unitless values ``prox_value`` and ``dist_value``.
+    Evaluates to zero, if no point is located in each required direction.
+
+    .. figure:: ../gen-images/iexpr_interp.svg
+      :width: 600
+      :align: center
+
+      Example of an interpolation expression. **Left**: A single proximal location locset. **Middle**: Terminal locations as distal locset.
+      **Right**: Area, where the expression evaluates to non-zero values.
```
thorstenhater, line 684: Please add another comment why this region is non-zero
```
B <- root -> [A -> B] -> Terminal
```
but not this
```
[B <- root -> A] -> B -> Terminal
```

doc/concepts/labels.rst
```diff
@@ -608,6 +618,120 @@ Region expressions
...
+.. label:: (div (iexpr | real) (iexpr | real) [... (iexpr | real)])
+
+    Division of at least two inhomogeneous expressions or real numbers.
+    The expression is evaluated from the left to right, dividing the first element by each divisor in turn.
+
+.. label:: (exp value:iexpr)
+
+    The exponential function of the inhomogeneous expression ``value``.
+
+.. label:: (exp value:real)
```
thorstenhater, line 722: Could merge this and `log`

arbor/cable_cell.cpp
```diff
@@ -154,8 +153,32 @@ struct cable_cell_impl {
         return region_map.get<init_reversal_potential>()[init.ion];
     }
 
-    template <typename Property>
-    void paint(const region& reg, const Property& prop) {
+    void paint(const region& reg, const density& prop) {
+        this->paint(reg, scaled_mechanism<density>(prop));
+    }
+
+    void paint(const region &reg, const scaled_mechanism<density> &prop) {
```
thorstenhater, line 160: Style: `region& reg`

arbor/cable_cell.cpp
```diff
@@ -154,8 +153,32 @@ struct cable_cell_impl {
...
+      mextent cables = thingify(reg, provider);
+      auto &mm = get_region_map(prop.t_mech);
+
+      std::unordered_map<std::string, std::shared_ptr<iexpr_interface>> im;
+      for(const auto& it : prop.scale_expr) {
+          im.insert_or_assign(it.first, thingify(it.second, provider));
+      }
+
+      for (auto c : cables) {
```
thorstenhater, line 169: const& c?

arbor/cable_cell.cpp
```diff
@@ -154,8 +153,32 @@ struct cable_cell_impl {
...
+    void paint(const region& reg, const density& prop) {
+        this->paint(reg, scaled_mechanism<density>(prop));
+    }
+
+    void paint(const region &reg, const scaled_mechanism<density> &prop) {
+      mextent cables = thingify(reg, provider);
+      auto &mm = get_region_map(prop.t_mech);
+
+      std::unordered_map<std::string, std::shared_ptr<iexpr_interface>> im;
+      for(const auto& it : prop.scale_expr) {
```
thorstenhater, line 165: `it` => `[fst, snd]`

arbor/fvm_layout.cpp
```diff
@@ -293,7 +296,7 @@ ARB_ARBOR_API fvm_cv_discretization fvm_cv_discretize(const cable_cell& cell, co
 
             mcable span{bid, parent_refpt, cv_refpt};
             double resistance = embedding.integrate_ixa(span, D.axial_resistivity[0].at(bid));
-            D.face_conductance[i] = 100/resistance; // 100 scales to µS.
+            D.face_conductance[i] = 100/resistance; // 100 scale to µS.
```
thorstenhater, line 299: ?

arbor/include/arbor/cable_cell_param.hpp
```diff
@@ -221,6 +225,19 @@ struct ARB_SYMBOL_VISIBLE ion_reversal_potential_method {
     mechanism_desc method;
 };
 
+template <typename TaggedMech>
+struct ARB_SYMBOL_VISIBLE scaled_mechanism {
```
thorstenhater, line 229: Do we need that template?

arbor/include/arbor/iexpr.hpp
```diff
@@ -0,0 +1,117 @@
...
+
+
+private:
+    iexpr(iexpr_type type, std::any args): type_(type), args_(std::move(args)) {}
+
+    iexpr_type type_;
+    std::any args_;
+};
+
+ARB_SYMBOL_VISIBLE std::ostream& operator<<(std::ostream& o, const iexpr& e);
```
thorstenhater, line 94: Maybe ARB_API is enough here?

arbor/include/arbor/morph/mprovider.hpp
```diff
@@ -20,6 +22,7 @@ struct ARB_ARBOR_API mprovider {
     // Throw exception on missing or recursive definition.
     const mextent& region(const std::string& name) const;
     const mlocation_list& locset(const std::string& name) const;
+    const std::shared_ptr<iexpr_interface>& iexpr(const std::string& name) const;
```
thorstenhater, line 25: Maybe at this point add `iexpr_ptr = std::shared_ptr<iexpr_interface>`
(other places in arbor do it this way)

thorstenhater (changes requested):
Hi,

some minor comments, rest looks good.

doc/concepts/labels.rst
```diff
@@ -98,6 +98,16 @@ the region of cables that have radius less than 0.5 μm
...
 .. _labels-expressions:
 
+.. glossary::
+  iexpr
+    An iexpr is an inhomogeneous expression, that can be evaluated at any point on a cell.
+
+Some common inhomogeneous expression:
+
+* radius: The radius of the cell at a location.
+* distance: Distance to a locset or region from a location.
```
thorstenhater, line 108: Distance here means:
Along the morphology, measured from the closest common ancestor (=minimum distance)

AdhocMan:
The code style issues should be addressed now. I've also tried to improve the documentation with custom images for distance and interpolation expressions.
What I'm not sure about, is removing the template argument from the `scaled_mechanism` struct. It depends on how likely we are going to extend this feature to other types of mechanisms, if at all. If not, naming it `scaled_density` might be better.

arbor/include/arbor/cable_cell_param.hpp
thorstenhater: To elaborate: 
1. Compile Times. This is a public header and also used in many places inside Arbor, so templates here can add a lot of compile time.
2. Communicating Intent. Currently, this is only ever used with DensityMechanisms, so this template gives the wrong impression of bein general.

doc/concepts/decor.rst
thorstenhater: Seconded.

doc/concepts/decor.rst
```diff
@@ -166,7 +167,24 @@ Take for example the built-in mechanism for passive leaky dynamics:
 
 .. _cablecell-ions:
 
-4. Ion species
+.. _cablecell-scaled-mechs:
+
+4. Scaled mechanisms
+~~~~~~~~~~~~~~~~~~~~~
+:ref:`Inhomogenous expressions  <labels-iexpr>` can be applied to mechanisms, to scale parameters on a cable cell.
```
thorstenhater, line 174: Hey, a sentence or two on why and how this is desirable would go a long way. Something like:

> It is useful to have a quantity smoothly varying along the dendrite, for example with the distance from the soma. 

Then use this idea in your examples.

thorstenhater, line 174: Also, if you want to keep `iexpr`, motivate why they are called `inhomogeneous`.

doc/concepts/decor.rst
```diff
@@ -166,7 +167,24 @@ Take for example the built-in mechanism for passive leaky dynamics:
...
+These expressions are evaluated at mid-points of the internal cable partition of the cell and multiplied with the set parameter value.
```
thorstenhater, line 175: Finally his could be made clearer and best split into two sentences.
1. These expressions yield a _scaling factor_ that can be applied to a base value of a parameter (currently only density mechanisms).
2. Example
3. The point of evaluation is the mid point...

Since 3. is a important detail, but does not add to the understanding  of the concrete use.

doc/concepts/labels.rst
```diff
@@ -98,6 +98,16 @@ the region of cables that have radius less than 0.5 μm
 
 .. _labels-expressions:
 
+.. glossary::
+  iexpr
+    An iexpr is an inhomogeneous expression, that can be evaluated at any point on a cell.
+
+Some common inhomogeneous expression:
```
thorstenhater, line 105: These are somewhat redundant, since the full list follows directly.

doc/concepts/labels.rst
```diff
@@ -608,6 +618,126 @@ Region expressions
...
+.. label:: (proximal-distance scale:real loc:locset)
+
+    The minimum distance in proximal direction from the points within the locset ``loc``. The scaling parameter ``scale`` has unit :math:`{\mu m}^{-1}` 
+    and is multiplied with the distance, such that the result is unitless.
+
+    .. figure:: ../gen-images/iexpr_prox_dis.svg
+      :width: 600
+      :align: center
+
+      Example of a proximal-distance expression with a single input location. **Left**: Input location. **Right**: Area, where the expression evaluates to non-zero values.
```
thorstenhater, line 657: No `,` behind `Area`

doc/concepts/labels.rst
```diff
@@ -608,6 +618,126 @@ Region expressions
 
       Two regions (left and middle) and their intersection (right).
 
+
+.. _labels-iexpr:
+
+Inhomogeneous Expressions
```
thorstenhater, line 624: Just a thought for all these expression having a `scale` parameter.
Wouldn't it be convenient to create overloads where `scale=1` which
I assume might be the most common case. I might be wrong, though.
Example
```LISP
(distance scale:real loc:locset)
(distance loc:locset) ; same as (distance scale=1 loc:locset)

**Review by thorstenhater — CHANGES_REQUESTED** [review]

File: doc/concepts/labels.rst (hunk 15 above)
Comment (AdhocMan) [reply doc/concepts/labels.rst:105]: I tried to keep it in the same format as the description of regions and locsets above, where simple examples are given in the same way. I don't mind removing it, if you prefer less redundancy over consistency.

File: doc/concepts/labels.rst (hunk 17 above)
Comment (AdhocMan) [reply doc/concepts/labels.rst:624]: I'm not sure if `scale=1` is going to be a common case. The advantage of always requiring a scale is that it forces the user to think about units, potentially preventing some mistakes. Maybe we could wait and see, if this will be a user request or what scales are common.

File: doc/concepts/labels.rst (hunk 15 above)
Comment (thorstenhater) [reply doc/concepts/labels.rst:105]: I don't like redundancy, no ;)

File: arbor/include/arbor/cable_cell_param.hpp (hunk 9 above)
Comment (AdhocMan) [reply arbor/include/arbor/cable_cell_param.hpp:229]: As discussed, I'll leave the template argument, expecting to extend this to other types of mechanisms later on.

File: doc/concepts/decor.rst (hunk 13 above)
Comment (AdhocMan) [reply doc/concepts/decor.rst]: I've added your suggested sentence and changed to example to a distance expression.

File: doc/concepts/decor.rst (hunk 14 above)
Comment (AdhocMan) [reply doc/concepts/decor.rst]: I've reformulated the text and changed the example to use a distance expression.

File: doc/concepts/labels.rst (hunk 17 above)
Comment (thorstenhater) [reply doc/concepts/labels.rst:618]: :tada: Thanks

**Review by thorstenhater — CHANGES_REQUESTED** [review]
One last (I promise!) clarification in the docs.

File: doc/concepts/decor.rst [hunk 18]
```diff
@@ -166,7 +167,25 @@ Take for example the built-in mechanism for passive leaky dynamics:
 
 .. _cablecell-ions:
 
-4. Ion species
+.. _cablecell-scaled-mechs:
+
+4. Scaled mechanisms
+~~~~~~~~~~~~~~~~~~~~~
+:ref:`Inhomogenous expressions  <labels-iexpr>` can be applied to mechanisms, to scale parameters on a cable cell, such that a parameter is inhomogenous along the cell.
```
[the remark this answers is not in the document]
AdhocMan: How about:
```
Mechanism parameters are usually homogeneous along a cell. However, sometimes it is useful to scale parameters based on inhomogeneous properties.
:ref:`Inhomogeneous expressions  <labels-iexpr>` provide a way to describe a desired scaling formula, which for example can include the cell radius or the distance to a given set of locations.
The name is inspired by NeuroML's https://docs.neuroml.org/Userdocs/Schemas/Cells.html#schema-inhomogeneousparameter.
Such an expression is evaluated along the cell and yields a scaling factor, which is multiplied with the base value of the selected parameter.
Internally, this evaluation and scaling is done at mid-points of the cable partition of the cell.
Currently, only parameters of :ref:`density mechanisms <cablecell-density-mechs>` can be scaled.
```

thorstenhater:
I'll leave the commit message and merge to you @AdhocMan

Issue #1802, opened 2022-01-20
Make `cable_global_properties` readable


Helveg:
Like the `decor`, `cable_global_properties` is a one way bag, and what is put in can't be read out.

**Describe the feature you need**
Read operations of `set_property` and `set_ion`, and retrieving the catalogue one has registered.

**Explain what it is supposed to enable**
This is a problem for recipe composition (#1680). One can't read or merge the global properties of subrecipes.

**Additional context**
Arborio project needs to merge the recipe of the IO and CC

llandsmeer:
This would be useful for example in the implementation of 'top-level' recipes that merge a set of recipes into a single recipe: https://gist.github.com/llandsmeer/83a50458ddb134eb87e94cd3c3a77b27
This was surprisingly easy to implement, except for the distribution of global_properties() -  because that function does not take a `gid` parameter.

thorstenhater:
The catalogue now is a plain data member.

PR #1890, opened 2022-05-05
Add introspection for global properties.

thorstenhater:
Add accessors for the following items in `global_properties`
- ion_species (name -> valence) RO
- ion_data (name -> initial values) RO
- ion_rev_pot (name -> rev pot mech) RO
- Vm, rL, TK, Cm: RW

Fixes: #1802


boeschf (approved):
This looks good to me from an implementation standpoint. I am not quite sure what potential side effects would be, if any. Since this is a useful feature I approve.

PR #1891, opened 2022-05-11
Add point probes to demo

schmitts:
Extend the probe demo example with point probes sampling the conductance of two synapses.

### DIFF
--- a/example/probe-demo/probe-demo.cpp
+++ b/example/probe-demo/probe-demo.cpp
@@ -197,7 +197,7 @@ void vector_sampler(arb::probe_metadata pm, std::size_t n, const arb::sample_rec
         auto* value_range = any_cast<const arb::cable_sample_range*>(samples[i].data);
         assert(value_range);
         const auto& [lo, hi] = *value_range;
-        assert(n_entities==hi-low);
+        assert(n_entities==hi-lo);
 
         for (unsigned j = 0; j<n_entities; ++j) {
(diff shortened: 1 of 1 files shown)


thorstenhater (changes requested):
Thanks for adding this! A C++ism and then we are good to merge.

thorstenhater (approved):
👍

example/probe-demo/probe-demo.cpp
```diff
@@ -173,21 +186,29 @@ void scalar_sampler(arb::probe_metadata pm, std::size_t n, const arb::sample_rec
...
-        assert(n_cable==value_range->second-value_range->first);
-
-        for (unsigned j = 0; j<n_cable; ++j) {
-            arb::mcable where = (*cables_ptr)[j];
-            std::cout << samples[i].time << ", "
-                      << where.prox_pos << ", "
-                      << where.dist_pos << ", "
-                      << value_range->first[j] << '\n';
+        const auto& [lo, hi] = *value_range;
+        assert(n_entities==hi-low);
```
brenthuisman, line 200:
```suggestion
        assert(n_entities==hi-lo);
```

brenthuisman (changes requested):
I believe this is the showstopper.

schmitts:
> I believe this is the showstopper.

I was compiling in `release` that nulls the asserts ...

brenthuisman (approved):

Issue #1484, opened 2021-04-12
Cable cell probing and sampling


brenthuisman:
After closing #1353, we did not make a new ticket to track having "Cable cell probing and sampling" pages across concepts/API. This page should describe experimental/artificial in and outputs.

### Status
* Concepts is blank
* C++ has more than Python API, including schedules, which are on Python's Recipe page
* Should "Event generator and schedules" be on this page?
* A new example?

thorstenhater:
done by PR #1898

PR #1898, opened 2022-05-17
probe id -> probeset id, and clarification in docs

### LINKED ISSUE #1877: Unclear from docs why `sim.samples` of a single handle produces array of datasets
The docs here https://docs.arbor-sim.org/en/latest/python/simulation.html#recording-samples explain it as:

> and returns the recorded data as a list, with one entry for **each probe** associated with the **probe id** that was used in step 2 above

But all the text above mentions probes and probe ids as a singular thing, produced by an index entry in the return array of `recipe.probes`; so why does my `handle = sim.sample(probe_id, ...)` call, with all singular references to probe and handle suddenly produce multiple datasets?

Answer: Because each `probe(s?)` can be expanded into multiple recording sites because the `locset` can point to multiple locations. 

That isn't really covered in the doc there. And makes the `data, meta = sim.samples(handle)[0]` susceptible to become cargo cult pattern: Don't know why, but I copy pasted this from the docs.

brenthuisman:
* Rename `probe id` and `probe id` to `probeset id` and `probeset_id` to clarify that they are associated to a (loc)set.
* Clarify documentation. 

Fixes #1877, #1484


doc/python/probe_sample.rst
```diff
@@ -5,14 +5,14 @@ Cable cell probing and sampling
 
 .. module:: arbor
 
-Cable cell probe addresses are defined analogously to their counterparts in
+Cable cell probeset addresses are defined analogously to their counterparts in
 the C++ API (see :ref:`cablecell-probes` for details). Sample data recorded
 by the Arbor simulation object is returned in the form of a NumPy array,
 with the first column holding sample times, and subsequent columns holding
 the corresponding scalar- or vector-valued sample.
```
Helveg, line 12: So do I understand correctly that in the case of scalar valued probes we get an `(N, 2)` shaped numpy array, and in the case of an `M`-long vector valued probe we'd get an `(N, 1 + M)` shaped array? Might be worth to show 2 examples of that here. Plus, we'd repeat that `K` times for each probe in the probeset?

doc/python/probe_sample.rst
```diff
@@ -5,14 +5,14 @@ Cable cell probing and sampling
...
 
 Location expressions will be realised as zero or more specific sites on
-a cell; probe addresses defined over location expressions will describe zero,
+a cell; probeset addresses defined over location expressions will describe zero,
```
Helveg, line 15: One keeps repeating the "probeset addresses defined over locations expressions", but is there any other way of defining them without location expressions that I am missing? Shouldn't this just be a statement: Probesets are defined over a location expression and will describe zero, one, or more probes...

doc/python/recipe.rst
```diff
@@ -72,10 +72,10 @@ Recipe
 
     .. function:: probes(gid)
 
-        Returns a list specifying the probe addresses describing probes on the cell ``gid``.
+        Returns a list specifying the probeset addresses describing probes on the cell ``gid``.
         Each address in the list is an opaque object of type :class:`probe` produced by
```
Helveg, line 76: Mismatch of the definitions:

> returns a list specifying probeset adresses

> each adress in the list is of type probe

Helveg: Each address in the list is a list of opaque objects of type probe. Right?

doc/python/simulation.rst
```diff
@@ -114,34 +114,34 @@ over the local and distributed hardware resources (see :ref:`pydomdec`). Then, t
...
     .. function:: remove_all_samplers()
 
         Disable all sampling processes and remove any associated recorded data.
 
     .. function:: samples(handle)
 
-        Retrieve a list of sample data associated with the given ``handle``.
-        There will be one entry in the list per probe associated with the :term:`probe id` used when the sampling was set up.
+        Retrieve a list of sample data associated with the given :term:`handle`.
+        There will be one entry in the list per probe associated with the :term:`probeset id` used when the sampling was set up.
```
Helveg, line 144: If we're doing ```:term:`probeset_id` ```then there's a lot of missed occurences that are just `probeset_id`

doc/python/simulation.rst
```diff
@@ -274,29 +274,37 @@ Definitions
         current, mechanism state, and a number of other quantities, measured either over the whole cell,
         or at specific sites (see :ref:`pycablecell-probesample`).
 
-        Probes are described by probe addresses, and the collection of probe addresses for a given cell is
+    probeset address
+        Probes are described by probeset addresses, and the collection of probeset addresses for a given cell is
```
Helveg, line 278: I wouldn't say that probes are described by probeset addresses. Maybe "probesets are described by probeset addresses" but I really still don't see the distinction between the two, what's a probeset address? The `cell_member`? That is usually already called the _probeset id_.

doc/python/simulation.rst
```diff
@@ -274,29 +274,37 @@ Definitions
...
-        particular measuring schedule, and then having a handle with which to access the recorded probe data later on.
+        particular measuring schedule, and then having a :term:`handle` with which to access the recorded probe data later on.
 
     sample
         A record of data corresponding to the value at a specific *probe* at a specific time.
 
+    handle
+        A handle for reaching sampling data associated to a sampler, which is associated to a probeset.
+        Setting a sampler (through :py:func:`simulation.sample`) returns handles. Sampled data can be retrieved
+        by passing the handle associated to a sampler (associated to a probe) to :py:func:`simulation.samples`.
```
Helveg, line 306: Good stuff! 👍

Helveg (changes requested):
I didn't go over the code changes, just the RST. I skipped the 2 cpp docs because they made my head spin. Unless this level of terseness makes cpp devs happy, Id' opt for a full rewrite of `cpp/probe_sample.rst`, there's so much technicalities and abstract justifications of the design mixed into the text that I have a hard time finding information in it. I'd start by introducing the concepts, with simple examples, and then layering in the more detailed things towards the end of the doc.

I believe `probeset address` should be replaced by `probeset`. Is it still required to include the term `address` for anything?

brenthuisman:
Thanks for the review! I've taken most into account.

The `probeset address` is a distinction only relevant for the C++ API, you won't need to know what they are when using Python. I agree the distinction makes it a bit muddy, but at least on the C++ side it needs to stay. I think I've eradicated every mention of probeset addresses from the Python pages, which I think must have started as a copy-paste of the C++ API. Please have a look, maybe there's still some stuff left that now doesn't make sense anymore.

The definitions were moved to the concept section (logical, right?) so the `probeset address` def does not confuse.

I left your comments on `probeset address` on the cpp pages open, because @thorstenhater will have a quick skim and it's good to have him confirm.

doc/concepts/probe_sample.rst
```diff
@@ -3,8 +3,52 @@
...
+    probeset address
+        Probesets are located at a probeset addresses, and the collection of probeset addresses for a given cell is
+        provided by the :py:class:`recipe` object. One address may correspond to more than one probe:
+        as an example, a request for membrane voltage on a cable cell at sites specified by a location
+        expression will generate one probe for each site in that location expression.
+
+        See :ref:`pycablecell-probesample` for a list of objects that return a probeset address.
+
+    probeset id
+        A designator for one or more probes as specified by a recipe. The *probeset id* is a
```
Helveg, line 26:
```suggestion
        A designator for a probeset specified by a recipe. The *probeset id* is a
```

doc/concepts/probe_sample.rst
```diff
@@ -3,8 +3,52 @@
...
+        probeset addresses returned by the recipe for that gid.
+
+    metadata
+        Each probe has associated metadata describing, for example, the location on a cell where the
+        measurement is being taken, or other such identifying information. Metadata for the probes
+        associated with a :term:`probeset id` can be retrieved from the simulation object, and is also provided
+        along with any recorded samples.
+
+    sampler
+        A sampler is something that receives probe data. It amounts to setting a particular :term:`probe` to a
```
Helveg, line 37:
```suggestion
        A sampler is something that receives probeset data. It amounts to setting a particular :term:`probeset` to a
```

doc/python/probe_sample.rst
brenthuisman: `sim.samples(handle)` will return a list of length `K`. In each element, you'll find a tuple containing `metadata` and `data`, where `metadata` will be a string describing the location, and `data` a `numpy.ndarray`.

Helveg (changes requested):
Found some more sections where probe and probeset were used interchangeably. Other than that it's becoming clearer I think :) The new terms seem to work at conveying the message.

doc/concepts/probe_sample.rst
```diff
@@ -3,8 +3,52 @@
...
+        expression will generate one probe for each site in that location expression.
+
+        See :ref:`pycablecell-probesample` for a list of objects that return a probeset address.
+
+    probeset id
+        A designator for one or more probes as specified by a recipe. The *probeset id* is a
+        :py:class:`cell_member` referring to a specific cell by gid, and the index into the list of
+        probeset addresses returned by the recipe for that gid.
+
+    metadata
```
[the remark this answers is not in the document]
brenthuisman: Per probe. The `handle` is to the probeset, which is why `samples(handle)` gets you a list of length `K`. Each element ( `(metadata, data)` tuple) is to one particular probe associated to one particular location in the locset on which the probe was placed.

I've added some words.

doc/python/recipe.rst
```diff
@@ -72,10 +72,10 @@ Recipe
 
     .. function:: probes(gid)
 
-        Returns a list specifying the probe addresses describing probes on the cell ``gid``.
-        Each address in the list is an opaque object of type :class:`probe` produced by
-        cell kind-specific probe address functions. Each probe address in the list
-        has a corresponding probe id of type :class:`cell_member`: an id ``(gid, i)``
+        Returns a list specifying the probesets describing probes on the cell ``gid``.
+        Each element in the list is an opaque object of type :class:`probe` produced by
```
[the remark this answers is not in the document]
brenthuisman: It's the Python version of cpps `probe_info`, which is what probes must return. So, I guess it's a bit of an internal interface, which is why it's not documented yet (nor has a useful docstring). I actually also don't know exactly how you'd make your own probes, but I suppose it will always need to return a meaningful `probe_info`/`probe`. I'll add a definition with a hint to `probe_info`. In future we can round it out if need be.

Helveg (approved):
Nice endeavour :)

doc/cpp/probe_sample.rst
```diff
@@ -56,7 +56,7 @@ Mechanism state queries however will throw a ``cable_cell_error`` exception
 at simulation initialization if the requested state variable does not exist
 on the mechanism.
 
-Cable cell probe addresses that are described by a ``locset`` may generate more
+Cable cell probeset addresses that are described by a ``locset`` may generate more
```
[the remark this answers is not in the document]
thorstenhater: Technically it might be worse: `Locset`s are what's usually called a multiset, ie a set of unique keys
(called the _support_ of the multiset) and an associated multiplicity for each key. Probes should be
defined on the support, thus each location is probed only once. So, the number of points in the `locset`
might be larger than the number of unique points.

doc/concepts/probe_sample.rst
```diff
@@ -3,8 +3,55 @@
...
 
-.. TODO::
-    WIP!
+Definitions
+***********
+
+.. glossary::
+
+    probe
+        A measurement that can be performed on a cell. Each cell kind will have its own sorts of probe;
```
thorstenhater, line 12: `;` -> `.`

doc/concepts/probe_sample.rst
```diff
@@ -3,8 +3,55 @@
...
+        current, mechanism state, and a number of other quantities, measured either over the whole cell,
+        or at specific sites (see :ref:`pycablecell-probesample`).
+        
+    probeset
+        A set of probes. Probes are placed on locsets, and therefore may describe more than one probe.
+
+    probeset address
+        Probesets are located at a probeset address, and the collection of probeset addresses for a given cell is
+        provided by the :py:class:`recipe` object. One address may correspond to more than one probe:
+        as an example, a request for membrane voltage on a cable cell at sites specified by a location
```
thorstenhater, line 23: location expression -> locset expression

doc/concepts/probe_sample.rst
```diff
@@ -3,8 +3,55 @@
...
+        expression will generate one probe for each site in that location expression.
+
+        See :ref:`pycablecell-probesample` for a list of objects that return a probeset address.
+
+    probeset id
+        A designator a probeset as specified by a recipe. The *probeset id* is a
```
thorstenhater, line 29: add `for`

doc/concepts/probe_sample.rst
```diff
@@ -3,8 +3,55 @@
...
+        probeset addresses returned by the recipe for that gid.
+
+    metadata
+        Each probe has associated metadata describing, for example, the location on a cell where the
+        measurement is being taken, or other such identifying information. Metadata for the probes
+        associated with a :term:`probeset id` can be retrieved from the simulation object, and is also provided
+        along with any recorded samples.
+
+    sampler
+        A sampler is something that receives probeset data. It amounts to setting a particular :term:`probeset` to a
```
thorstenhater, line 40: Here `probeset` is used once w/ a 'term` prefix and once w/o. Accident?

doc/cpp/probe_sample.rst
```diff
@@ -8,8 +8,7 @@ Cable cell probing and sampling
 Cable cell probes
 -----------------
 
-Various properties of a a cable cell can be sampled via one of the cable cell
-specific probe address described below. They fall into two classes: scalar
+Various properties of a cable cell can be sampled. They fall into two classes: scalar
 probes are associated with a single real value, such as a membrane voltage
 or mechanism state value at a particular location; vector probes return
 multiple values corresponding to a quantity sampled over a whole cell.
```
thorstenhater, line 14: Replace 'over a whole cell' w/ 'at a set of points on the cell'

doc/cpp/probe_sample.rst
```diff
@@ -277,7 +276,7 @@ If the mechanism is not defined at a particular site, that site is ignored.
 
 *  Sample value: ``double``. State variable value.
 
-*  Metadata: ``mlocation``. Location as given in the probe address.
+*  Metadata: ``mlocation``. Location as given in the probeset address.
```
thorstenhater, line 279: Note to self: get rid of these 'm' prefixes.

doc/cpp/probe_sample.rst
```diff
@@ -387,16 +386,16 @@ as a way of passing additional metadata about a probe to any sampler
...
 possibly with different value types.
 
-Probe addresses are decoupled from the cell descriptions themselves —
+probeset addresses are decoupled from the cell descriptions themselves —
 this allows a recipe implementation to construct probes independently
 of the cells themselves. It is the responsibility of a cell group implementation
-to parse the probe address objects wrapped in the ``any address`` field.
+to parse the probeset address objects wrapped in the ``any address`` field.
 
 The _k_th element of the vector returned by ``get_probes(gid)`` is
```
thorstenhater, line 394: Note: this is a strange choice. It's very brittle; imagine a user constructing the vector from an unordered container.

doc/python/probe_sample.rst
```diff
@@ -5,14 +5,13 @@ Cable cell probing and sampling
 
 .. module:: arbor
 
-Cable cell probe addresses are defined analogously to their counterparts in
+Cable cell probesets are defined analogously to their counterparts in
 the C++ API (see :ref:`cablecell-probes` for details). Sample data recorded
```
thorstenhater, line 9: This comment is very helpful :D

doc/python/recipe.rst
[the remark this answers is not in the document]
thorstenhater: > Making your own probes 

As in 'add a new quantity to be probed'? You'll need to go down into the C++ code and recompile.

thorstenhater (changes requested):
A few minor comments.

thorstenhater:
A general observation on this part of the docs: A graph of the process could help.

doc/concepts/probe_sample.rst
brenthuisman: Yes.

doc/cpp/probe_sample.rst
brenthuisman: Shall I do it in this PR? Touches a few files.

doc/cpp/probe_sample.rst
brenthuisman: I believe some chatter on Gitter elaborated on this. Should I add some of that here?

doc/python/probe_sample.rst
brenthuisman: It is a veiled excuse for the apparent convolution of the design. As per Gitter, the choices made stem from consideration, so I wanted to at least link to the outcomes of that consideration. I like that C++ and Python API are largely similar, but this is a good candidate for deviation. It is already, e.g. the `probe_tag` is nowhere to be found in Python. The `probe` class is not anything you actually can use to build with.

Should I do better than give excuses here? That would be to have people be able to create probes using only Python, and to maybe simplify things. Reading Gitter is sounds like that may be more involved than it seems, but maybe not? Maybe in a followup PR, if and when users ask for it?

brenthuisman:
Which process is that exactly? Of creating your own probe?

doc/python/recipe.rst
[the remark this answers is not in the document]
thorstenhater: How about adding 
> `probe`: opaque handle, unused in Python as of now.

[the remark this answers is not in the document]
thorstenhater: I see this was done above, nice.

doc/python/probe_sample.rst
thorstenhater: You know me ... a graphic? A crude approximation
```
O-------------         Cell morphology
^    ^    ^            Probe locset
|    |    |
L0 L1 L2               Locations, eg (cable 0 0.23)
U0 V0 W0               Values, one column per Location
U1                     one entry per sampling time point
U2
...
```

brenthuisman:
Shortcut to the requested schematic: https://github.com/arbor-sim/arbor/pull/1898/files#diff-119b7780b97222f316ae7b9bf31a04ac8d477cc2c07f1dcbcd33722b14b23222

doc/python/probe_sample.rst
thorstenhater: Thanks for adding the graphic, but I think we could be even clearer by splitting it in two
1. Showing the relation of handle, probeset, probes, and their locations
2. Showing the relation between the sampled data and the probeset

Sorry for being picky here, but we inflict multiple layers of callbacks and handles on our
users (for good reasons), so giving them a hand is needed. 

That said, I made one of the images, so maybe that plus the more schematic one you added
is enough. Tell me what you think.

Also, since I cannot comment on full files, feedback for the graphic here
1. Typo: experession
2. Did you check array format? I thought I remember columns of time, value. As shown it indicates rows. Maybe this would be clearer if shown as a table.
3. The `->` relation is unclear
  - handle -> probeset: _obtain_ a probeset from simulation
  - probeset -> probe, probe -> data|meta: ownership
  - meta -> locset, data -> ndarray: is a

On 3. I would suggest showing 'has a' and 'is a' relations like this; example for probe
```
+---------------+
| Probe         |
| meta: locset  | 
| data: ndarray |
+---------------+
```
The handle -> probeset -> probe could remain.

thorstenhater (changes requested):
Being picky on the graphics, sorry. I make amends as soon as I figure out how to upload an SVG here.

thorstenhater:
probing
There!

brenthuisman:
> Great image. The only thing that's omitted is the scalar vs vector valued probing

What probe returns vector data? I don't think I actually ever saw one...

Helveg:
> > Great image. The only thing that's omitted is the scalar vs vector valued probing
> 
> What probe returns vector data? I don't think I actually ever saw one...

Me neither, but the docs mentioned it, so I thought they exist.

brenthuisman:
Strictly speaking, it's up to the (C++) probe to define what it returns, and also whether that is one or more values per timestamp (or a timestamp at all). Although it can be any structure, as far as I know the ones we have now return an `ndarray`, as shown in the example.

thorstenhater:
In Python we do not have user-defined callbacks. Thus `ndarray` and its format is fixed, it'll always have
one column for time and each of the probe locations and one row per sampling point.

thorstenhater (changes requested):
Thanks for improving this. Over experimenting with the API and reading the code in
more depth for this review it became clear that there is still some potential
confusion.

For reference, I'll set up an example
```py
class recipe(A.recipe):
    # ...
    
    def probes(self, _):
        if gid == 0:
            return [A.cable_probe_ion_diff_concentration_cell("na"),
                    A.cable_probe_membrane_voltage("(location 0 0)"), ]
        else:
            return [A.cable_probe_ion_diff_concentration_cell("ca"), 
                    A.cable_probe_membrane_voltage("(terminal)"), ]
        
    # ...

R = recipe()
S = A.simulation(R())

ts = A.regular_schedule(0.1)
hdls = [S.sample((0, 0), ts), # Sodium in gid 0 
        S.sample((0, 1), ts), # U_m in gid 0 soma
        S.sample((1, 0), ts), # Calcium in gid 1
        S.sample((1, 1), ts), # U_m in gid 1 terminals
]

sim.run(tfinal=0.5)

for hdl in hdls:
    for data, meta in S.samples(hdl):
       pass
```

Here we extract three _vector_ probes and one _scalar_. For the sake of the
example assume two cells; cell 0 has `N` segments and cell 1 `M` segments of which `T` are
terminals. The schedule evaluates to five points `[0, 0.1, 0.2, 0.3, 0.4]`. We
were to tabulate the result we would see something like this

| Handle | Meta | Data | Quantity | Vector |
| ------ | ---- | ---- | -------- | ------ |
| (0, 0) | `[(cable ..), ...]` length `N` | NDArray[`N + 1`, 5] | `Nad` | Y |
| (0, 1) | `(cable ..)` | NDArray[2, 5] | `Um` | N | 
| (1, 0) | `[(cable ..), ...]` length `M` | NDArray[`M + 1`, 5] | `Cad` | Y |
| (1, 1) | `(cable ..)` | NDArray[2, 5] | `Um` | N |

Regarding the improved figure, thanks for keeping up with it. In the light of the above
we should rework s.t.
- `meta` is a _list_ of locations iff a vector probe is sampled else a single location
- `data` is a table with one column per location above plus one for the time
- why is a handle associated with a set? In my experiments never seemed to be that? Is there a fundamental difference Py/C++? If so we need a huge red warning banner
- what does the dashed backedge do?
- why does every probe have a connection from a single handle?

Feel free to use and adapt the above example.

Further it should be clarified
- `_cell` probes are vector probes, everything is a scalar
- how the locations for `_cell` probes are made
- `handle` associates has `(cell, num of probe on cell)`
- asking for a nonexistent probe `(23, 42)` will get something empty not an error
- even if you -- in Python at least -- probe on locsets with cardinality > 1 you'll get one location only (BUG?)
- thus it's more accurate `cable_probe_XYZ(location)` instead of `locset`
- no user callbacks in Python

brenthuisman:
I had to refresh my memory a bit.

1. I can update the image and description to reflect the fact that, depending on the type of probe, a probe may result in vector data or point data. I suppose I could borrow your example and paste it in. I believe this addresses your first two questions (re `meta`, `data`).
2. Why a handle is associated with a set: I can't answer the existential question in here, but I can confirm it is so, to the extent I tested it, solely with Python. Looking at the code, I had no reason to believe this was different in C++. Could you share more details on your experience to the contrary?
3. The dashed backedge is meant to demonstrate the association of a handle to a probeset. I will agree it can be improved by changing the arrow into a bi-directional one. Would that satisfy you?
4. Every probe has a connection from a single handle: the shaded box around it indicates a list, but better would be to add that explicitly in the caption.

Can the further clarifications wait for a second PR (in the interest of not letting this one slide any further than it has)?

thorstenhater:
I'd love to make it correct on the first possible moment, since we risk users coming up with misconceptions otherwise.
I am really confused by the issue that non-cell-type probes seem to be scalar, meaning that the intuitive 
```py
probe(locset('(terminal)'), membrane_voltage)
```
does not in fact return a list of voltages at all terminal points, but rather at one of those points.

Could you confirm this behaviour? Just tweaking the single_cell_recipe example should be sufficient.
If confirmed, we should potentially consider this a bug.

> I can update the image and description to reflect the fact that, depending on the type of probe, a probe may result in vector data or point data. I suppose I could borrow your example and paste it in. I believe this addresses your first two questions (re meta, data).

The point here is that rather than returning a `[(meta, values_over_time)]` a vector probe gives `(list_of_locations, values_for_locations_over_time)`. Difference being in the shape of the data.

brenthuisman:
Let's start at  `simulation.sample()` and go both ways.

1. `simulation.sample()` returns a `sampler_association_handle`, https://github.com/arbor-sim/arbor/blob/master/python/simulation.cpp#L137
2. `sampler_association_handle` is a integer number, https://github.com/arbor-sim/arbor/blob/master/arbor/include/arbor/sampling.hpp#L43
3. Conclusion: a handle is to 1 `probe`, located at 1 position. Not to a `probeset`! All probes get a cell local id, regardless of being grouped together by way of being constructed from 1 `locset`. I confirm your observation. TODO: modify the image and text to reflect that a handle refers to a single `probe`.

The other direction:
5. `simulation.sample()` is called with a `arbor.cell_member` (or `tuple` for short), which betrays the flatness of the return value. `arbor.cell_member` only specifies `gid` and `index`, and `index` is over a "cell-local collection", in this case, over `probe_id`s, NOT `probeset_id`s. TODO: update documentation.
6. `sim.samples()` returns a `recorder`, a new word to add to our glossary. https://github.com/arbor-sim/arbor/blob/master/python/simulation.cpp#L38
7. A `recorder` is a struct holding `meta` (can be anything but afaik always a log/reg expr, the number of records, and the records (`sample_record`s), https://github.com/arbor-sim/arbor/blob/master/python/pyarb.hpp#L20
8. A `sample_record` holds a time and a `util::any_ptr`. Afaik they are right now always either scalar or vector values. https://github.com/arbor-sim/arbor/blob/master/arbor/include/arbor/sampling.hpp#L32 TODO: update rightmost column  in image: the expression is for the `loc` a `probe` is attached to, NOT `locset` a `probeset` is attached to. Also, indicate `value` can be scalar or vector.

 Now, the term `probeset_id` is unused. They don't have ids.  `probeset` still might be used to describe a thing like `arbor.cable_probe_membrane_voltage('"my_locset"')`. However, these objects don't return anything interesting.usable, so other than that we might say they describe a `probeset`, it is not a very important term to define. However, that might change, pending #1929 (for which I volunteer). For now, TODO remove term `probeset_id`.

Questions:

1. When you say `The point here is that rather than returning a [(meta, values_over_time)] a vector probe gives (list_of_locations, values_for_locations_over_time).`, you mean that is incorrect (see `probe(locset('(terminal)'), membrane_voltage)` example), or that it is what you expect/like to see? Is this what you suggest we consider a bug? In the bit above I describe how things are now, but there is logic to expecting to have and be able to use a `probeset_id`. Is that what you're saying?
2. Nowhere are "scalar probes" and "vector probes" defined. TODO: define them, and define for each probe whether they are vector or scalar. Question: which are the vector ones? Those about mech state?

thorstenhater:
There two levels to this:
1. For efficiency, the behviour in Python and C++ is substantially different
   - Python: Scalar probes give (by whatever mechanism) a record shaped like (this is not! the actual type) `(location, array[time, values])` where `time` and `values` are 1d arrays with lengths corresponding to time elapsed/sampling freq.
      Vector probes get `([location], array[time, values])` where `values` now has a second dimension corresponding to `locations` length.
    - C++: We get callbacks, but essentially the same data layout. 
2. Despite the intuitive idea of 'if probing a multi-location locset, we get a vector probe' we only get vectors when using the `_cell` kinded probes. _This_ is what I might consider a bug or at least violating good UX.

We need to document clearly both these points and think whether we want to change 2.

thorstenhater:
Another comment regarding https://github.com/arbor-sim/arbor/issues/1929: Spikes are _not_ recorded using the mechanism of probing.

brenthuisman:
Thanks for clarifying that `_cell` kinded probes are those that output vector data.

So, after checking, the nonobvious difference between a `probeset` and a `_cell` kinded probes (or vector probe) is this:

1. The points in a `probeset` translate to _items_ (corresponding to each point) in the list returned by `.samples()`. (contrary to earlier claims!)
    1. e.g.: a voltage probe placed on `"(join (location 0 0.5) (location 0 0.7))"` might return `[(array([[  0., -40.]]), (location 0 0.7)), (array([[  0., -40.]]), (location 0 0.5))]`. The handle points at a single index. So, the term `probeset` does make sense and should be left in.

2. The 'points' in a vector probe translate to _columns_ in the data of the first and only element in the list returned by `.samples()`.
    1. e.g.: `sample [(array([[  0., -40., -40., -40.]]), [(cable 0 0 0.333333), (cable 0 0.333333 0.666667), (cable 0 0.666667 1)])]`

We can call it a bug. I've update https://github.com/arbor-sim/arbor/issues/1929 to correct this.

brenthuisman:
Updated, see https://github.com/arbor-sim/arbor/pull/1898/commits/5b9bf7d328d8e356a184d14ba707448d7d8048b8 for the new diagram.

I think it's now clear what the difference is, and #1929 will remove that difference, as soon as I get back. Please do merge in the meantime :)

doc/python/probe_sample.rst
```diff
@@ -5,23 +5,66 @@ Cable cell probing and sampling
...
-the corresponding scalar- or vector-valued sample.
-
-Location expressions will be realised as zero or more specific sites on
-a cell; probe addresses defined over location expressions will describe zero,
+.. figure:: probe_sample-diag.svg
+    :width: 800
+    :align: center
+
+    A schematic view of how :term:`handles <handle>` let you access sampled data measured at a :term:`probeset`.
+    When setting a probe on a locset a probeset is created (a locset describes one or more locations).
```
Helveg, line 13: > When setting a probe on a locset a probeset is created (a locset describes one or more locations).

`probe` is already used in the schema (and everywhere else) to mean "part of a probeset", and here it is used as "thing that combined with a locset makes a probeset". We are trying to avoid this. Is it an option call the functions `probing devices`/`probing functions`/`probe objects` rather than `probes`? Alternatively we can call the thing inside of a probeset a `probeloc`?

doc/python/probe_sample.rst
```diff
@@ -5,23 +5,66 @@ Cable cell probing and sampling
...
+Sample data recorded by the Arbor simulation object is returned in the form 
+of a NumPy array, with the first column holding sample times, and subsequent 
+columns holding the corresponding scalar- or vector-valued sample.
+
+Probesets are defined over a location expression and will describe zero,
 one, or more probes, one per site. They are evaluated in the context of
 the cell on which the probe is attached.
 
+:term:`Vector probes <vector probe>` are a kind of probes that samples over a region, rather than a :term:`locset`.
+This means that the may output more than a single data point per timestamp. The layout of the outputs as returned
```
Helveg, line 36:
```suggestion
This means that they may output more than a single data point per timestamp. The layout of the outputs as returned
```

doc/python/probe_sample.rst
Helveg: Anyways, we could also leave it like this and rewrite / note it:

```
When setting a probe object on a locset, a probeset is created (a locset describes one or more locations).
Note that both the probe objects returned by probe functions, and the elements of a probeset may be referred to as probes.
```

Helveg: I also noticed below that we use `probeset functions`, maybe the solution is just to `probeset function` instead of `probe` here?

Helveg (changes requested):
Last duplicated concept: probe (`arbor.probe`) and `probe` (element of `probeset`)

thorstenhater:
Here's a script to demonstrate the difference.
```py
import arbor

tree = arbor.segment_tree()
p = tree.append(arbor.mnpos, arbor.mpoint(-3, 0, 0, 3), arbor.mpoint(3, 0, 0, 3), tag=1)
tree.append(p, arbor.mpoint(3, 0, 0, 3), arbor.mpoint(-3, 0, 0, 3), tag=2)
tree.append(p, arbor.mpoint(3, 0, 0, 3), arbor.mpoint(-3, 0, 0, 3), tag=2)

decor = (
    arbor.decor()
    .set_property(Vm=-40)
    .paint('"soma"', arbor.density("hh"))
    .place('"midpoint"', arbor.iclamp(10, 2, 0.8), "iclamp"))

cell = arbor.cable_cell(tree, labels, decor)

class single_recipe(arbor.recipe):
    def __init__(self):
        arbor.recipe.__init__(self)

    def num_cells(self):
        return 1

    def cell_kind(self, gid):
        return arbor.cell_kind.cable

    def cell_description(self, gid):
        return cell

    def probes(self, gid):
        return [arbor.cable_probe_membrane_voltage('(location 0 0.5)'),
                arbor.cable_probe_membrane_voltage_cell(),
                arbor.cable_probe_membrane_voltage('(join (location 0 0) (location 0 1))'),
                ]

    # (4.6) Override the global_properties method
    def global_properties(self, kind):
        return arbor.neuron_cable_properties()

recipe = single_recipe()
sim = arbor.simulation(recipe)
handles = [sim.sample((0, n), arbor.regular_schedule(0.1))
           for n in range(3) ]
sim.run(tfinal=1)

for hd in handles:
    print("Handle", hd)
    for d, m in sim.samples(hd):
        print(" * Meta:", m)
        print(" * Payload:", d.shape)
```

```
Handle 0
 * Meta: (location 0 0.5)
 * Payload: (10, 2)
Handle 1
 * Meta: [(cable 0 0 1), (cable 0 1 1), (cable 1 0 0), (cable 2 0 0), (cable 1 0 1), (cable 2 0 1)]
 * Payload: (10, 7)
Handle 2
 * Meta: (location 0 0)
 * Payload: (10, 2)
 * Meta: (location 0 1)
 * Payload: (10, 2)
"""
Thus
- 0: Single scalar probe.
- 1: Single vector probe.
- 2: Two scalar probes.

Conclusion: no bug present.

Might be worth to include this minimal example in the tutorials and/or discussion of probes.

doc/python/probe_sample.rst
brenthuisman: `probe` and `sample_recorder` (which is ultimately what arbor calls a thing that records samples) are not sufficiently distinguished I guess. 

When we place a probe(set) on a loc(set), we create a sample_recorder (no, not gonna define `sample_recorderset`!) Is this the same as a `sampler`? Probably, @thorstenhater can you confirm?

brenthuisman:
Thanks @thorstenhater, that's almost the script I have here :) I'll add it as an example, but I'll remove it at soon as the shape of record outputs is made consistent between probeset and vector probes.

doc/python/probe_sample.rst
brenthuisman: I can confirm.

PR #1899, opened 2022-05-18
remove test statements for move ctor

boeschf:
Remove checks for move ctor when it could be elided:

Newer compilers choose to elide the move constructor in certain cases. This leads to failure of 2 tests. As this behaviour is completely legal (and even desirable) it should not lead to failure. I therefore commented out the two statements in question and added a comment about this in the code. Note, that here we anyways only check the behaviour of `std::function<void()>` for which the `task` type is an alias at this point, so in my opinion the whole point of the test is a bit questionable.


bcumming (approved):
lgtm!

PR #1900, opened 2022-05-31
Tut/use avail threads

brenthuisman:
- Update Python many-cell examples to create contexts with `"avail_threads"`.

### DIFF
--- a/doc/tutorial/network_ring.rst
+++ b/doc/tutorial/network_ring.rst
@@ -116,8 +116,11 @@ To create a simulation, we must create an :class:`arbor.context` and :py:class:`
 
 Step **(12)** initalizes the ``threads`` parameter of :class:`arbor.context` with the ``avail_threads`` flag. By supplying
 this flag, a context is constructed that will use all locally available threads. On your local machine this will match the
-number of cpu-cores, times two if your cpu provides some form of simultaneous multithreading. Especially with large numbers
-of cells you will notice the speed-up. :func:`arbor.partition_load_balance` creates a default domain decomposition, which 
+number of CPU-cores, likely multiplied by two if your CPU provides some form of 
+`multithreading <https://en.wikipedia.org/wiki/Multithreading_(computer_architecture)>`_. Especially with large numbers
+of cells you will notice the speed-up. (You could instantiate the recipe with 5000 cells and observe the difference. Don't
+forget to turn of plotting if you do; it will take more time to generate the image then to run the actual simulation!)
+:func:`arbor.partition_load_balance` creates a default domain decomposition, which 
 for contexts initialized with ``threads=avail_threads`` distributes cells evenly over the available cores. You can print the
 objects to see what defaults they produce on your system.
 
--- a/doc/tutorial/network_ring.rst
+++ b/doc/tutorial/network_ring.rst
@@ -114,7 +114,7 @@ The execution
 
 To create a simulation, we must create an :class:`arbor.context` and :py:class:`arbor.domain_decomposition`.
 
-Step **(12)** creates a default execution context, and uses the :func:`arbor.partition_load_balance` to create a
+Step **(12)** creates an execution context that will use all locally available threads, and uses the :func:`arbor.partition_load_balance` to create a
 default domain decomposition. You can print the objects to see what defaults they produce on your system.
 
 Step **(13)** sets all spike generators to record using the :py:class:`arbor.spike_recording.all` policy.
--- a/doc/tutorial/network_ring.rst
+++ b/doc/tutorial/network_ring.rst
@@ -114,8 +114,12 @@ The execution
 
 To create a simulation, we must create an :class:`arbor.context` and :py:class:`arbor.domain_decomposition`.
 
-Step **(12)** creates an execution context that will use all locally available threads, and uses the :func:`arbor.partition_load_balance` to create a
-default domain decomposition. You can print the objects to see what defaults they produce on your system.
+Step **(12)** initalizes the ``threads`` parameter of :class:`arbor.context` with the ``avail_threads`` flag. By supplying
+this flag, a context is constructed that will use all locally available threads. On your local machine this will match the
+number of cpu-cores, times two if your cpu provides some form of simultaneous multithreading. Especially with large numbers
+of cells you will notice the speed-up. :func:`arbor.partition_load_balance` creates a default domain decomposition, which 
+for contexts initialized with ``threads=avail_threads`` distributes cells evenly over the available cores. You can print the
+objects to see what defaults they produce on your system.
 
 Step **(13)** sets all spike generators to record using the :py:class:`arbor.spike_recording.all` policy.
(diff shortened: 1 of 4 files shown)


python/example/network_ring.py
```diff
@@ -109,7 +109,7 @@ def global_properties(self, kind):
 recipe = ring_recipe(ncells)
 
 # (12) Create a default execution context, domain decomposition and simulation
```
thorstenhater, line 111: Comment out of sync.

doc/tutorial/network_ring.rst
```diff
@@ -114,7 +114,7 @@ The execution
 
 To create a simulation, we must create an :class:`arbor.context` and :py:class:`arbor.domain_decomposition`.
 
-Step **(12)** creates a default execution context, and uses the :func:`arbor.partition_load_balance` to create a
+Step **(12)** creates an execution context that will use all locally available threads, and uses the :func:`arbor.partition_load_balance` to create a
```
thorstenhater, line 117: In what sense:
1. Cores?
2. Hardware threads?
3. ...

Link to docs?

thorstenhater (changes requested):

doc/tutorial/network_ring.rst
```diff
@@ -114,8 +114,12 @@ The execution
 
 To create a simulation, we must create an :class:`arbor.context` and :py:class:`arbor.domain_decomposition`.
 
-Step **(12)** creates a default execution context, and uses the :func:`arbor.partition_load_balance` to create a
-default domain decomposition. You can print the objects to see what defaults they produce on your system.
+Step **(12)** initalizes the ``threads`` parameter of :class:`arbor.context` with the ``avail_threads`` flag. By supplying
+this flag, a context is constructed that will use all locally available threads. On your local machine this will match the
+number of cpu-cores, times two if your cpu provides some form of simultaneous multithreading. Especially with large numbers
+of cells you will notice the speed-up. :func:`arbor.partition_load_balance` creates a default domain decomposition, which
```
thorstenhater, line 120: This only is true if you request more cell groups, right? Both the 'distribute' and -- consequently -- the 'speed-up' parts.
See simulation.cpp:357, tasks push updates to whole _groups_.
```
> python3 brunel.py
<arbor.context: num_threads 2, has_gpu False, has_mpi False, num_ranks 1>
<arbor.domain_decomposition: domain_id 0, num_domains 1, num_local_cells 500, num_global_cells 500, groups 500>
```
You'll need to add something like this
```
    hint = arbor.partition_hint()
    hint.cpu_group_size = opt.group_size
    hints = {arbor.cell_kind.lif: hint}
    decomp = arbor.partition_load_balance(recipe, context, hints)
```
to get more groups on one MPI task _or_ partition explicitly.

doc/tutorial/network_ring.rst
```diff
@@ -114,8 +114,12 @@ The execution
 
 To create a simulation, we must create an :class:`arbor.context` and :py:class:`arbor.domain_decomposition`.
 
-Step **(12)** creates a default execution context, and uses the :func:`arbor.partition_load_balance` to create a
-default domain decomposition. You can print the objects to see what defaults they produce on your system.
+Step **(12)** initalizes the ``threads`` parameter of :class:`arbor.context` with the ``avail_threads`` flag. By supplying
+this flag, a context is constructed that will use all locally available threads. On your local machine this will match the
+number of cpu-cores, times two if your cpu provides some form of simultaneous multithreading. Especially with large numbers
```
thorstenhater, line 119: `times two` .... That wholly depends on the core arch. 
Better (more accurate)
> CPU cores times the number of SMT, as supported by your CPU. :footnote: Most Intel/AMD Chips have SMT=2 aka Hyperthreading. 

SMT should be written out once.

thorstenhater (changes requested):

doc/tutorial/network_ring.rst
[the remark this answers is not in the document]
brenthuisman: Old fashioned, I like it!

doc/tutorial/network_ring.rst
brenthuisman: By 'requesting cell groups', do you mean 'requesting ranks'?

What I mean to explain without introducing more jargon than required is that cells, also if they are the same kind/description, will be spread over the available execution threads (when `avail_threads` is set), and consequently a speed-up (compared to a default context) will be experienced.

doc/tutorial/network_ring.rst
brenthuisman: There are no consumer cpus with more than two, right?

Also, SMT is written out (it is never abbreviated).

thorstenhater: Depending on how wide you cast the (you _can_ technically buy a POWER9 workstation) there's SMT=8 consumer machines. Just say it's likely to be two, then no one will be surprised if it's not

SMT: Yes, but not in my example.

brenthuisman: POWER9 workstations are not mainstream!

I've added a likely, and explained in the MPI version that it is and should be not used.

doc/tutorial/network_ring.rst
thorstenhater: No, you can in fact request more groups using hints (like in the example I gave) or an explicit partitioning similar
to a list of lists of cell `gid`s. However, I realised that's not needed here, the default is one group per cell (here 500, see example) _and not_ one group per task. So, this is ok, but for optimisation one group per thread might be better (something better left out here, but keep in mind!)

doc/tutorial/network_ring.rst
```diff
@@ -116,8 +116,11 @@ To create a simulation, we must create an :class:`arbor.context` and :py:class:`
 
 Step **(12)** initalizes the ``threads`` parameter of :class:`arbor.context` with the ``avail_threads`` flag. By supplying
 this flag, a context is constructed that will use all locally available threads. On your local machine this will match the
-number of cpu-cores, times two if your cpu provides some form of simultaneous multithreading. Especially with large numbers
-of cells you will notice the speed-up. :func:`arbor.partition_load_balance` creates a default domain decomposition, which 
+number of CPU-cores, likely multiplied by two if your CPU provides some form of 
+`multithreading <https://en.wikipedia.org/wiki/Multithreading_(computer_architecture)>`_. Especially with large numbers
```
thorstenhater, line 120: Better link to here
https://en.wikipedia.org/wiki/Simultaneous_multithreading

thorstenhater (changes requested):

doc/tutorial/network_ring.rst
thorstenhater: Technically, you can order one now https://www.raptorcs.com/TALOSII/

doc/tutorial/network_ring.rst
brenthuisman: Say: logical cores.

thorstenhater (approved):

PR #1901, opened 2022-06-01
Code Quality: PVS Studio Finds

thorstenhater:
It's the season again when I put our code through the grinder and fix at least a few issues found.


arbor/memory/allocator.hpp
```diff
@@ -173,9 +173,9 @@ class allocator : public Policy {
     using rebind = allocator<U, Policy>;
 
 public:
-    allocator() {}
-    ~allocator() {}
-    allocator(allocator const&) {}
+    allocator() = default;
+    ~allocator() = default;
+    allocator(allocator const&) = default;
```
brenthuisman, line 178: That's different behavior, no?

arbor/benchmark_cell_group.cpp
```diff
@@ -39,7 +39,7 @@ benchmark_cell_group::benchmark_cell_group(const std::vector<cell_gid_type>& gid
         cg_targets.add_label(c.target, {0, 1});
     }
 
-    reset();
+    benchmark_cell_group::reset();
```
brenthuisman, line 42: Never knew this was ambiguous!

arbor/s_expr.cpp
```diff
@@ -281,7 +281,7 @@ class lexer {
         }
 
         // test if the symbol matches a keyword
-        auto it = keyword_to_tok.find(symbol.c_str());
+        auto it = keyword_to_tok.find(symbol);
```
brenthuisman, line 284: There are two kinds of people: those that append `.c_str()` gratuitously and those that really need it. How can PVSS tell which it is?

sup/include/sup/ioutil.hpp
```diff
@@ -32,7 +32,7 @@ class basic_null_streambuf: public std::basic_streambuf<charT, traitsT> {
     typedef typename streambuf_type::off_type off_type;
     typedef typename streambuf_type::traits_type traits_type;
 
-    virtual ~basic_null_streambuf() {}
+    virtual ~basic_null_streambuf() = default;
```
brenthuisman, line 35: Also not the same behavior.

brenthuisman:
I have the feeling a lot of the errors are related to the assumption by PVSS that `{}` and `= default` are equal.

arbor/benchmark_cell_group.cpp
thorstenhater: Oh, it's not. _Here_. But assume you derive from `benchmark_cell_group` overriding the `reset` method.
Now, you'll **still** get the base class' `reset`. This makes it explicit. There's no fix except making the class
`final`.

arbor/s_expr.cpp
thorstenhater: `Find` takes a `const std::string&` as `std::string` is the key type?!

sup/include/sup/ioutil.hpp
thorstenhater: As far as I can see that's not the case. The default deconstruction via `= default` will destroy all members (in reverse order), then the base classes. The same is true of `{}`. So, no difference, right?

arbor/memory/allocator.hpp
thorstenhater: The copy constructor might be arguable, however, `allocator has no members, so we are good?! Or am I missing something?

thorstenhater:
> I have the feeling a lot of the errors are related to the assumption by PVSS that `{}` and `= default` are equal.

But for the default ctor and dtor this is entirely correct, see nice answer here

https://stackoverflow.com/questions/20828907/the-new-syntax-default-in-c11?lq=1

The copy constructor is much more sketchy, but for empty classes it should be fine.

arbor/s_expr.cpp
brenthuisman: `.c_str()` isn't a `std::string`?

brenthuisman:
OK, so `{}` and `= default` are the same, apart from: "Giving a user-defined constructor, even though it does nothing, makes the type not an aggregate and also not trivial." The definitions of [trivial](https://www.geeksforgeeks.org/trivial-classes-c/) and [aggregate/POD](https://stackoverflow.com/questions/146452/what-are-pod-types-in-c) seem awfully close btw., but if that's important, why not _not_ declare the de-/constructor? Reading the above, `= default` gives you that behavior.

arbor/s_expr.cpp
thorstenhater: It's a `const char*`, like in C. If `std::string` where a `std::vector<char>`, `c_str` would be like `data`, returning the underlying buffer by pointer.

thorstenhater:
If you declare any of the 5 methods (copy ctor, move ctor, copy=, move=, non-default ctor), the rest will not be auto-generated. But you can get them back by `=default`

brenthuisman:
Ach so....

thorstenhater:
> > I have the feeling a lot of the errors are related to the assumption by PVSS that `{}` and `= default` are equal.
> 
> But for the default ctor and dtor this is entirely correct, see nice answer here
> 
> https://stackoverflow.com/questions/20828907/the-new-syntax-default-in-c11?lq=1
> 
> The copy constructor is much more sketchy, but for empty classes it should be fine.
> 
> The errors are caused by something different, and they do not occur (neither a warning) without the sanitizers. Fascinating.

So, the compile errors  were cause by the SIMD classes declaring `friend fma(SIMD, SIMD, SIMD)`, but **not** `friend fma(const SIMD&, ...)`. That's all.

arbor/morph/place_pwlin.cpp
```diff
@@ -220,9 +220,7 @@ std::pair<mlocation, double> place_pwlin::closest(double x, double y, double z)
             // v and w are the proximal and distal ends of the segment.
             const p3d v = seg.prox;
             const p3d w = seg.dist;
-
             const p3d vw = w-v;
-            const p3d vp = p-v;
```
brenthuisman, line 225: Strange to remove one but not both. Any idea why?

arbor/s_expr.cpp
brenthuisman: Yes, but that's not what I meant. I guess the answer to my question is: yes, in this case PVSS knows how `symbol` is used and `c_str()` is not required for that use.

brenthuisman (approved):
I went through the whole thing, looks reasonable to me. Have a few more questions, see above, but they are about PVSS's choices, not the code itself.

arbor/morph/place_pwlin.cpp
thorstenhater: `vp` is constructed inside a conditional and closer to its sole site of use. So, it's clearer and it might save the odd cycle.

arbor/fvm_layout.cpp
```diff
@@ -248,32 +248,34 @@ ARB_ARBOR_API fvm_cv_discretization fvm_cv_discretize(const cable_cell& cell, co
     double dflt_potential =   *(dflt.init_membrane_potential | global_dflt.init_membrane_potential);
     double dflt_temperature = *(dflt.temperature_K | global_dflt.temperature_K);
 
+    const auto& assignments = cell.region_assignments();
```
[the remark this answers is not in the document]
thorstenhater: They are use in multiple places, yielding longish expressions, so we name it.

arbor/include/arbor/s_expr.hpp
```diff
@@ -214,6 +214,7 @@ struct ARB_ARBOR_API s_expr {
     s_expr(s_expr l, s_expr r):
         state(pair_type(std::move(l), std::move(r)))
     {}
+    s_expr& operator=(const s_expr& s) { state = s.state; return *this; }
```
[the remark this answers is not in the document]
thorstenhater: No, not really. The rule is to implement `op=` if you implement copy and vice veras. The argument is one of correctness:
If you _need_ write one, it'll do something interesting. Therefore, you cannot use the default one.
You'll get a warning, but the result will be wrong, since the assignment operator will still be made for you see
https://gcc.godbolt.org/z/KfreGvxjY
Note especially the difference here
```
S a = S{};     // copy ctor, despite looking like assignment.
S a; a = S{}; // assignment
```

arbor/s_expr.cpp
thorstenhater: It knows since it can see this
```
std::map<K, V>::find(const K&)
```
and here `K=std::string`. We are calling
```
map.find(string.c_str())
```
so the temporary C string is converted back to a `std::string`.
That in turn is not possible without copying the bytes, so
1. not needed
2. wasteful

thorstenhater:
Most of them are reasonable things to admonish, although I ignored about 60% of the finds.
Next time, we can do some more. Here's a suggestion: Ask them for a license (free for FOSS),
run it and go over the results. Each has a key, which you can index into their exaplanation DB.
Or you can just ask (me).

PR #1904, opened 2022-06-09
Add some convenience to simulation creation.

### LINKED ISSUE #1862: Ergonomics
Working with Arbor at a high level requires quite a bit of redundant(?) typing. I propose to go through
our user facing APIs with a fine comb and think deeply on ergonomics.

Here's a list of examples I remember from the top of my head (mostly Python)

1. A dict should be accepted where a label_dict is expected. 
2. Mechanisms should not distinguish between GLOBAL and PARAMETER, ie `pas/e=42` becomes `'pas, e=42`.
3. SWC loading should give back the standard labels like NML and/or have them pre-defined somewhere.
5. Simplify construction of simulations
   ```py
   ctx = A.context()
   mdl = my_recipe()
   ddc = A.partition_load_balance(mdl, ctx)
   sim = A.simulation(mdl, ddc, ctx)
   ```
   there is no reason(?) to pass `ctx` and `ddc` multiple times. One idea could be this
   ```py
   ctx = A.context()
   mdl = my_recipe()
   ctx.partition_load_balance(mdl) # ctx now holds on to what ddc was above
   sim = A.simulation(mdl, ctx)
   ```
   or even
   ```py
   ctx = A.context()
   mdl = my_recipe()
   sim = ctx.partition_load_balance(mdl) # what else can you do w/ ddc?
   ```
   That doesn't mean the lower levels need to go away.
5. Related, probing is at least a 3 step process.
6. The Python module could use more sub-modules.

thorstenhater:
Encapsulates 80% of cases:
```cxx
 rec = recipe()
 ctx = make_context()
 dec = partition_load_balance(rec, ctx)
 sim = simulation(rec, dec, ctx)
```
is now written as
```cxx
 rec = recipe()
 sim = simulation(rec)
```
In python we use keyword args to allow
both to be specified separatly.

Partially fixes #1862

### DIFF
--- a/python/example/single_cell_cable.py
+++ b/python/example/single_cell_cable.py
@@ -208,7 +208,6 @@ if __name__ == "__main__":
     ]
     recipe = Cable(probes, **vars(args))
 
-
     # configure the simulation and handles for the probes
     sim = arbor.simulation(recipe)
     dt = 0.001


python/example/network_ring.py
```diff
@@ -109,9 +109,7 @@ def global_properties(self, kind):
 recipe = ring_recipe(ncells)
 
 # (12) Create an execution context using all locally available threads, domain decomposition and simulation
-context = arbor.context("avail_threads")
-decomp = arbor.partition_load_balance(recipe, context)
-sim = arbor.simulation(recipe, decomp, context)
+sim = arbor.simulation(recipe, context=arbor.context("avail_threads"))
```
Helveg, line 112:
```suggestion
context = arbor.context("avail_threads")
sim = arbor.simulation(recipe, context=context)
```

Helveg: Cleaner, more uniform.

Helveg (approved):
Nice! I checked only the Python impact

python/example/network_ring.py
thorstenhater: It's an extra line. Unacceptable.

example/lfp/lfp.cpp
```diff
@@ -193,7 +191,7 @@ int main(int argc, char** argv) {
     const double sample_dt = 0.1; // [ms]
     const double dt = 0.1;        // [ms]
 
-    arb::simulation sim(R, arb::partition_load_balance(R, context), context);
+    arb::simulation sim(R);
```
brenthuisman, line 194: `Recipe` instead of `R`, while we're at it.

python/example/network_ring.py
brenthuisman: What about making the load balancer the third arg, and optional? It's the least likely element that users will customize. We can then forego with the still somewhat verbose `context=arbor.context`, but say instead `sim = arbor.simulation(recipe, arbor.context("avail_threads"))`

python/example/single_cell_cable.py
```diff
@@ -176,12 +176,9 @@ def get_tmax(data):
         f'(location 0 {r})') for r in np.linspace(0, 1, 11)]
     recipe = Cable(probes, **vars(args))
 
-    # create a default execution context and a default domain decomposition
-    context = arbor.context()
-    domains = arbor.partition_load_balance(recipe, context)
 
     # configure the simulation and handles for the probes
-    sim = arbor.simulation(recipe, domains, context)
+    sim = arbor.simulation(recipe)
```
brenthuisman, line 181: `domains`; I never noticed :)

python/example/single_cell_recipe.py
```diff
@@ -59,15 +59,10 @@ def global_properties(self, kind):
...
-# (6) Create a default execution context and a default domain decomposition.
-
-context = arbor.context()
-domains = arbor.partition_load_balance(recipe, context)
-
 # (7) Create and run simulation and set up 10 kHz (every 0.1 ms) sampling on the probe.
 # The probe is located on cell 0, and is the 0th probe on that cell, thus has probe_id (0, 0).
 
-sim = arbor.simulation(recipe, domains, context)
+sim = arbor.simulation(recipe)
```
brenthuisman, line 65: These are examples. I'd like to show the verbose way at least once, and in this example in particular (it's where the `recipe` and `context` (and load balancers) are introduced.

To the accompanying tutorial page we can add how it could be shortened.

python/simulation.cpp
```diff
@@ -195,9 +195,13 @@ void register_simulation(pybind11::module& m, pyarb_global_ptr global_ptr) {
         // A custom constructor that wraps a python recipe with arb::py_recipe_shim
         // before forwarding it to the arb::recipe constructor.
         .def(pybind11::init(
-            [global_ptr](std::shared_ptr<py_recipe>& rec, const arb::domain_decomposition& decomp, const context_shim& ctx) {
+                 [global_ptr](std::shared_ptr<py_recipe>& rec,
+                              const std::optional<arb::domain_decomposition>& decomp,
```
brenthuisman, line 199: I guess here is where you should swap context and domdec.

arbor/include/arbor/load_balance.hpp
```diff
@@ -23,5 +23,5 @@ using partition_hint_map = std::unordered_map<cell_kind, partition_hint>;
 ARB_ARBOR_API domain_decomposition partition_load_balance(
     const recipe& rec,
     const context& ctx,
-    partition_hint_map hint_map = {});
+    const partition_hint_map& hint_map = {});
```
brenthuisman, line 26: Think of the performance savings :)

brenthuisman (changes requested):
Noice, a few requests.

python/example/single_cell_recipe.py
Helveg: Yes, introducing the long form is important!

python/example/network_ring.py
thorstenhater: You can currently set `context`, `domain_decomposition` or both. So, I opt for the parallel to our C++ API
and keep the order.

python/simulation.cpp
thorstenhater: See above.

python/example/network_ring.py
brenthuisman: Can we swap it there too? The argument applies there too: least likely last.

thorstenhater: That requires more thinking, breaking an API should be done carefully (even in 0.x.y). Note that this _here_ is 
not a breaking change, the old way still works. Consequently, I'd like to separate it into its own PR.
Finally, I am less than sure that the decomposition is the least likely.

example/lfp/lfp.cpp
thorstenhater: That expression seems a bit dated, yes.

python/example/network_ring.py
brenthuisman: I can't see how you'd be able to set a nonstandard domdec without supplying the context, so in that sense domdec is the least likely to be modified (any customization also requires a context, but not vice versa).

And you already add a new constructor to `arbor/include/simulation.hpp` with this different ordering, when is then used to update the cpp examples. You had to change the ordering there for nearly similar reasons: you can create a load balancer without having a context first.

So, I remain of the opinion to switch to the shape of that new constructor (because it is the most intuitive shape).

arbor/include/arbor/simulation.hpp
```diff
@@ -25,8 +27,13 @@ class simulation_state;
 
 class ARB_ARBOR_API simulation {
 public:
+
     simulation(const recipe& rec, const domain_decomposition& decomp, const context& ctx);
 
+    simulation(const recipe& rec,
```
brenthuisman, line 33: Sneakily swapped the order there, but! it's a much better order so please change all to be consistent.

example/dryrun/dryrun.cpp
```diff
@@ -174,10 +174,8 @@ int main(int argc, char** argv) {
                 params.num_ranks, params.cell, params.min_delay);
         arb::symmetric_recipe recipe(std::move(tile));
 
-        auto decomp = arb::partition_load_balance(recipe, ctx);
-
         // Construct the model.
-        arb::simulation sim(recipe, decomp, ctx);
+        arb::simulation sim(recipe, ctx);
```
brenthuisman, line 178: Evidence the load balancer is least likely to be manually provided ;)

example/generators/generators.cpp
```diff
@@ -132,11 +132,8 @@ int main() {
     // Create an instance of our recipe.
     generator_recipe recipe;
 
-    // Make the domain decomposition for the model
-    auto decomp = arb::partition_load_balance(recipe, context);
-
     // Construct the model.
-    arb::simulation sim(recipe, decomp, context);
+    arb::simulation sim(recipe, context);
```
brenthuisman, line 136: And more :P

python/example/network_ring.py
thorstenhater: Again, leave that discussion for a separate PR/Issue. Breaking changes should be done is isolation and this is big already. Good catch on the C++ side though, the order there should be reversed.

brenthuisman (changes requested):

example/brunel/brunel.cpp
```diff
@@ -244,11 +244,12 @@ int main(int argc, char** argv) {
 
         brunel_recipe recipe(nexc, ninh, next, in_degree_prop, w, d, rel_inh_strength, poiss_lambda, seed);
 
-        partition_hint_map hints;
-        hints[cell_kind::lif].cpu_group_size = group_size;
-        auto decomp = partition_load_balance(recipe, context, hints);
-
-        simulation sim(recipe, decomp, context);
+        simulation sim(recipe, context,
```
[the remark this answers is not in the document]
thorstenhater: That was just to show of the 'other' way of doing it, but there are no instances of the old way left.

example/generators/generators.cpp
thorstenhater: Bah, fine, you convinced me.

example/brunel/brunel.cpp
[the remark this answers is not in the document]
thorstenhater: I re-instantiated one example to the old ways.

Helveg (approved):
No objections 👍

example/lfp/lfp.cpp
```diff
@@ -79,8 +79,7 @@ struct lfp_demo_recipe: public arb::recipe {
         // Apical dendrite, length 490 μm, radius 1 μm, with SWC tag 4.
         tree.append(soma_apex, {0, 0, 10, 1},  {0, 0, 500, 1}, 4);
 
-        decor dec;
-
+        auto dec = arb::decor();
```
brenthuisman, line 82: Out of curiosity: why?

example/ring/ring.cpp
```diff
@@ -13,6 +13,7 @@
 
 #include <arborio/label_parse.hpp>
 
+#include <arbor/load_balance.hpp>
```
brenthuisman, line 16: Can be deleted.

example/ring/ring.cpp
```diff
@@ -160,10 +161,9 @@ int main(int argc, char** argv) {
         // Create an instance of our recipe.
         ring_recipe recipe(params.num_cells, params.cell, params.min_delay);
 
-        auto decomp = arb::partition_load_balance(recipe, context);
-
         // Construct the model.
-        arb::simulation sim(recipe, decomp, context);
+        auto decomposition = arb::partition_load_balance(recipe, context);
```
brenthuisman, line 165: Can be deleted.

python/example/network_ring.py
```diff
@@ -109,9 +109,8 @@ def global_properties(self, kind):
 recipe = ring_recipe(ncells)
 
 # (12) Create an execution context using all locally available threads, domain decomposition and simulation
-context = arbor.context("avail_threads")
-decomp = arbor.partition_load_balance(recipe, context)
-sim = arbor.simulation(recipe, decomp, context)
+ctx = arbor.context("avail_threads")
+sim = arbor.simulation(recipe, context=ctx)
```
brenthuisman, line 113:
```suggestion
sim = arbor.simulation(recipe, arbor.context("avail_threads"))
```

python/example/network_ring_mpi.py
```diff
@@ -118,8 +118,7 @@ def global_properties(self, kind):
 print(context)
 
 # (13) Create a default domain decomposition and simulation
-decomp = arbor.partition_load_balance(recipe, context)
-sim = arbor.simulation(recipe, decomp, context)
+sim = arbor.simulation(recipe, context=context)
```
brenthuisman, line 121:
```suggestion
sim = arbor.simulation(recipe, context)
```

python/example/single_cell_detailed_recipe.py
```diff
@@ -142,25 +142,19 @@ def global_properties(self, gid):
...
 
-# (4) Create an execution context
-context = arbor.context()
-
-# (5) Create a domain decomposition
-domains = arbor.partition_load_balance(recipe, context)
-
-# (6) Create a simulation
-sim = arbor.simulation(recipe, domains, context)
+# (7) Create a simulation
```
brenthuisman, line 145: Here and for other modified examples, the associated `rst` should be updated (don't forget the change in numbering here).

brenthuisman (changes requested):
Some minor things.

example/lfp/lfp.cpp
thorstenhater: Homogeneity with the other examples. There's no semantic difference.

python/example/single_cell_detailed_recipe.py
thorstenhater: Yes, that's in the queue with all the flake8 and black stuff ;)

python/example/network_ring.py
thorstenhater: Robin requested the exact opposite ;)

python/example/network_ring_mpi.py
thorstenhater: This works now :+1:

python/example/network_ring.py
brenthuisman: The main thing I care about is that as of The Swap(tm) `context=` can be left out. I'd have peace with
```suggestion
sim = arbor.simulation(recipe, ctx)
```

thorstenhater: So, I'll let you two battle that out, I'd chosen @brenthuisman 's suggestion and only change due to Robin's wishes.
Less typing is almost always to be preferred.

brenthuisman: What I meant is: no need to pass with the `context` key anymore. Second arg is expected to be a `arbor.context`.

brenthuisman (approved):

python/example/network_ring.py
Helveg: Fine with me :)

PR #1906, opened 2022-06-13
Adopt Black for Python.

thorstenhater:
Add black formatter


brenthuisman:
Building docs is not part of our tests, but relies on much of our Python code. Can you confirm it builds fine?

thorstenhater:
They do. All the literal include lines are in jeopardy though.

Helveg:
What about an [`--ignore-revs-file`](https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html)?

thorstenhater:
We can do that, but -- I assume -- not before merging this, since we need to commit IDs?

thorstenhater:
Then I can only say thank you for the hint, will do ;)

Helveg (approved):
I agree with black 👍 This makes the `'"label"'` syntax slightly weirder

.github/workflows/lint.yml
```diff
@@ -0,0 +1,24 @@
...
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    name: "Lint"
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+    steps:
```
brenthuisman, line 13: Maybe instead use their Action: https://black.readthedocs.io/en/stable/integrations/github_actions.html

Also, maybe make this a monthly job and have this Action check on each PR: https://github.com/marketplace/actions/python-black-formatter-only-on-edited-files
It only check changed files, so it won't put people on the hook for code unrelated to what they are changing.

doc/contrib/coding-style.rst
```diff
@@ -3,6 +3,26 @@
...
+
+.. code::
+
+   # Install the formatter if not present
+   pip install black
+   # Automatically apply style. If unsure what this does read on.
+   black --extend-exclude '/(ext|python/pybind11|doc/scripts/.*_theme)' . scripts/build-catalogue.in
+
+The formatter can also be run with ``--check`` to list offending files and
+``--diff`` to preview changes.
```
brenthuisman, line 21: Let's mention how people can set their editor to do this for them: https://black.readthedocs.io/en/stable/integrations/editors.html

python/__init__.py
```diff
@@ -9,12 +9,14 @@
...
+
     here = os.path.abspath(os.path.dirname(__file__))
-    with open(os.path.join(here, 'VERSION')) as version_file:
+    with open(os.path.join(here, "VERSION")) as version_file:
         return version_file.read().strip()
 
+
 __version__ = get_version()
-__config__  = config()
+__config__ = config()
```
brenthuisman, line 19: No more manual vertical aligning, I hope you can actually live with that ;)

brenthuisman (changes requested):

.github/workflows/lint.yml
brenthuisman: Even better/worse: have a bot autopush blacked files to a PR: https://github.com/psf/black/blob/2234639a31ed8aeca05e20aa6966005277e9ae07/.github/workflows/autoblack.yml

thorstenhater: Using their action is great though.

brenthuisman: You're right: by merging this PR there should be no way black will find errors in other areas but those touched in future PRs.

Having the bot autocorrect is neat though, and scary. What do you think?

doc/contrib/coding-style.rst
thorstenhater: 👍

PR #1908, opened 2022-06-15
Adopt flake8

thorstenhater:
This builds on top of the previous PRs regarding Python best practices and add the
flake8 linter.


brenthuisman:
flake8: good idea, but! readme says it's very tied to a Python minor version. You anointed Python 3.9 as standard, but since we want to keep our code compatible down to 3.6 (for now), should then the Action not run with 3.6?

Or is this more hassle than it's worth, considering the limited amount of Python?

.github/workflows/lint.yml
```diff
@@ -11,12 +11,23 @@ jobs:
     strategy:
       fail-fast: false
     steps:
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
```
brenthuisman, line 17:
```suggestion
          python-version: 3.6
```

python/example/network_ring_mpi.py
```diff
@@ -31,15 +31,15 @@ def make_cable_cell(gid):
     b0 = tree.append(s, arbor.mpoint(0, 0, 0, 2), arbor.mpoint(50, 0, 0, 2), tag=3)
 
     # Attach two dendrites (tag=3) of length 50 μm to the end of the first dendrite.
-    # Radius tapers from 2 to 0.5 μm over the length of the dendrite.
-    b1 = tree.append(
+    # (b1) Radius tapers from 2 to 0.5 μm over the length of the dendrite.
```
brenthuisman, line 34: No, that's not good. In an example, unused variables are OK, they are they to facilitate quick experimentation, not lessons of writing optimized code. Users way want to inspect the return value of `.append()`.

python/test/unit/test_catalogues.py
```diff
@@ -15,7 +15,7 @@ def __init__(self):
         self.props = arb.neuron_cable_properties()
         try:
             self.props.catalogue = arb.load_catalogue("dummy-catalogue.so")
-        except:
+        except Exception:
```
brenthuisman, line 18: Seems tautological, there's no possibility of `except` catching objects not inheriting from Exception.

brenthuisman (changes requested):

python/test/unit/test_catalogues.py
thorstenhater: https://stackoverflow.com/questions/54948548/what-is-wrong-with-using-a-bare-except

python/example/network_ring_mpi.py
thorstenhater: Well, this is not about performance. Flake8 complains as unused variables often indicate an oversight.
Example:
```py
b0 = tree.append(None, ...)
b1 = tree.append(b0, ...)
b2 = tree.append(b0, ...)
```
`b1` goes unused, so is this a typo or CopyPaste bug, and the user really meant
```py
b0 = tree.append(None, ...)
b1 = tree.append(b0, ...)
b2 = tree.append(b1, ...)
```

thorstenhater: > No, that's not good. In an example, unused variables are OK, they are they to facilitate quick experimentation, 

If that is inteded, we have the options of
1. Setting a file-wide igore for that warning
2. Suppressing it in this lines via `#noqu:XXX`

I'd just roll with it, since if experimentation is desired, adding a `x =` to the line shouldn't be to hard. We already
demonstrated that is possible above.

python/test/unit/test_catalogues.py
brenthuisman: TIL

thorstenhater: My pleasure

python/example/network_ring_mpi.py
brenthuisman: > Flake8 complains as unused variables often indicate an oversight.

In 'production' code, sure, but this isn't. (It isn't a copypaste bug either.)

Since it's my goto tutorial, I know that users only vaguely familiar with Python (i.e. scientific users ;)) can and are surprised by anything that's not systematic. 

Let's clarify why we don't assign then:

```suggestion
    # (b1) Radius tapers from 2 to 0.5 μm over the length of the dendrite. Because we're not going to reference this segment later on, we don't need to assign it to a variable.
```

.flake8
```diff
@@ -0,0 +1,24 @@
...
+               python/pybind11,
+               # auto-generated
+               doc/scripts/inputs.py
+               doc/scripts/make_images.py
+               # hidden
+               .*,
+               # artifacts
+               build,
+               # nah, don't care
+               spack/package.py
```
brenthuisman, line 24: Actually, they care! They have the same check running.

brenthuisman (changes requested):
Line numbers. I could take care of it in a follow up too, there's a few more tutorial mistakes left.

thorstenhater:
Line numbers to be fixed after all the convenience, Python formatting, and linting PR are in.

brenthuisman (approved):

PR #1909, opened 2022-06-15
Allow single quotes in Python s-exp 

thorstenhater:
Python uses `'` and `"` interchangibly, so allow this in Python s-exps.
```py
d.paint("'all'", ...)
d.paint("(region 'all')", ...)
```
and so on. 

NB. These expressions are therefore now equivalent
```
d.paint("\"fine'in's-exp\"")
d.paint("'fine\'in\'s-exp'")
```
when used from python.

### DIFF
--- a/arborio/label_parse.cpp
+++ b/arborio/label_parse.cpp
@@ -147,7 +147,7 @@ std::string eval_description(const char* name, const std::vector<std::any>& args
     };
 
     const auto nargs = args.size();
-    std::string msg = concat("'", name, "' with ", nargs, "argument", nargs!=1u?"s:" : ":");
+    std::string msg = concat("'", name, "' with ", nargs, " argument", nargs!=1u?"s:" : ":");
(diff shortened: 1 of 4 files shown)


Helveg:
In favor! I'd like to be able to use outer double quotes everywhere 👍 You're really racking up the formatting PRs here

thorstenhater:
This is a tiny bit more than pure formatting, but yes, it makes using black a bit easier on the eyes and thus ame from there.

brenthuisman:
This does not only allow it in Python, does it?

thorstenhater:
Only Python.

brenthuisman:
Hmm, could it be extended to C++ (and `ACC`)? It'd be surprising to learn that e.g. label expressions will work in Python but not in an `ACC` file.

thorstenhater:
Good arguments have been made against having this.

PR #1910, opened 2022-06-16
Temporarily disable A64FX CI

bcumming:
Disable a64fx ci due to to slurm issues on ault making it always fail.


bcumming:
bors try

brenthuisman (approved):

PR #1911, opened 2022-06-20
Predefine SWC Regions.

thorstenhater:
Add label_dict constructor with predefined labels for standard SWC regions

 - soma: (tag 1)
 - axon: (tag 2)
 - dend: (tag 3)
 - apic: (tag 4)


Helveg:
What's with  the huge diff?

thorstenhater:
It has been rebased on top of the flake8 PR.

brenthuisman:
Can you merge master again? Flake has been merged, so it'll be easier to review and hopefully easier to see why that test is failing.

thorstenhater:
I saw.

python/example/single_cell_detailed.py
```diff
@@ -21,15 +21,11 @@
 
 # (2) Create and populate the label dictionary.
 
-labels = arbor.label_dict()
+# Pre-defined SWC regions
+labels = arbor.label_dict(swc=True)
```
brenthuisman, line 25: Let's make it a bit clearer this is setting defaults:
```suggestion
labels = arbor.label_dict(default="swc")
```
or 

```suggestion
labels = arbor.label_dict(defaults_swc=True)
```
(there may be other defaults?)

python/example/single_cell_detailed.py
```diff
@@ -21,15 +21,11 @@
...
 
 # Regions:
```
brenthuisman, line 27: This is now a very empty section ;_;

`swc` being what it is, not all users may know what SWC defaults are. Let's add that to the associated tutorial page, with a link to our swc format page (if it isn't already there).

brenthuisman (changes requested):
To `doc/fileformat/swc.rst`, I'd like to mention these defaults explicitly, and that this is how you can start with a preloaded dict.

python/example/single_cell_detailed.py
thorstenhater: No, these are the only ones we have. Other _loaders_ return a structure containing labels and morphology, but for SWC it's essentially always the same. You're correct in that passing a boolean is not really the high point of design here. At the very least it should be kwargs-only, but I think it would be nice to have something to add SWC labels post-hoc, st
```py
ld = A.label_dict({....}).add_swc_labels()
```
will give you the labels you explicitly passed _plus_ our SWC defaults.

python/example/single_cell_detailed.py
thorstenhater: I want it _gone_, these things shouldn't be have to remembered let alone typed out again and again.

python/example/single_cell_detailed.py
thorstenhater: Done.

python/example/single_cell_detailed.py
thorstenhater: There's now way more text in the docs than we saved by this method ;)

Helveg (changes requested):
Link to the method

brenthuisman (approved):

PR #1912, opened 2022-06-21
Two context decomp swaps were forgotten

brenthuisman:
In the simulation object argument swap, two were forgotten.


AdhocMan (approved):
LGTM

Issue #1409, opened 2021-03-01
CMake install docs for Python environments


bcumming:
The process for python users who want to compile Arbor and its Python wrapper using CMake, then install the Python package can be surprising if they expect everything to get installed according to their local environment (e.g. in a venv).

This can be fixed by setting the `CMAKE_INSTALL_PREFIX` to the appropriate Python environment path, which will install all of Arbor in the expected location.

Add a note to the installation docs that explains this clearly here: https://arbor.readthedocs.io/en/latest/install/build_install.html#python-frontend

thorstenhater:
Escaped auto-close.

PR #1915, opened 2022-06-23
Reformulate Python build system in `pyproject.toml`, bump Python minver to 3.7

### LINKED ISSUE #1683: `setup.py` doesn't declare arbor package

brenthuisman:
* Failing Macos Python wheel builds fixed. (There is a weekly job, I just have not been checking it.)
* Macos Python wheels now come with dual-arch (x86-64 and arm64)
  * Testers wanted. I don't have Macs, let alone M1s.
* Moved (nearly) all Python build instructions to `pyproject.toml`
  * Enables 'build isolation', and need to specify build-deps only once, no need for users or CI scripts to pre-install them.
  * Enables editable `pip` installs (`pip install -e ./arbor`)
    * Testers wanted!
  * Compatible with 'build frontends' `pip` and `build`.
  * Passing CMake options actually got shorter
* Drop Python 3.6 support. This way of building does not work on that version anymore, but does 3.7 up to 3.11beta. I did my best to find the actual reason, or a formal declaration of cease of support, but couldn't find it. I assume something in this rather complex concoction of tooling is perhaps assuming 3.6 is in fact EOL.

Wheels available here: https://github.com/brenthuisman/arbor/actions/runs/2550193753

Closes #1683, #1414, #1409

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -20,5 +20,3 @@ extend-exclude =
                .*,
                # artifacts
                build,
-               # nah, don't care
-               spack/package.py
--- a/.github/workflows/ciwheel.yml
+++ b/.github/workflows/ciwheel.yml
@@ -48,13 +48,13 @@ jobs:
       - name: Update pip
         run: python3 -m pip install --upgrade pip
       - name: Get packages
-        run: python3 -m pip install numpy setuptools scikit-build ninja cmake
+        run: python3 -m pip install build #numpy setuptools scikit-build ninja cmake
       - uses: actions/checkout@v2
         with:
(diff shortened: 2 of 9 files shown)


brenthuisman:
For reasons unknown, Github is
1. missing the changes in `setup.py`. Git itself sees them?
2. Says reviews are not necesary for merging the PR. ???
3. Does not run any tests on the PR ??????

Made a new one, where all works as expected: #1916

PR #1916, opened 2022-06-24
Move Python build to `pyproject.toml`, bump Python minver to 3.7, fix macos wheel generation

### LINKED ISSUE #1683: `setup.py` doesn't declare arbor package

brenthuisman:
* Failing Macos Python wheel builds fixed. (There is a weekly job, I just have not been checking it.)
* Macos Python wheels now come with dual-arch (x86-64 and arm64)
  * Testers wanted. I don't have Macs, let alone M1s.
* Moved (nearly) all Python build instructions to `pyproject.toml`
  * Enables 'build isolation', and need to specify build-deps only once, no need for users or CI scripts to pre-install them.
  * Enables editable `pip` installs (`pip install -e ./arbor`)
    * Testers wanted!
  * Compatible with 'build frontends' `pip` and `build`.
  * Passing CMake options actually got shorter
* Drop Python 3.6 support. This way of building does not work on that version anymore, but does 3.7 up to 3.11beta. I did my best to find the actual reason, or a formal declaration of cease of support, but couldn't find it. I assume something in this rather complex concoction of tooling is perhaps assuming 3.6 is in fact EOL.

Wheels available here: https://github.com/brenthuisman/arbor/actions/runs/2550193753

Closes #1683, #1414, #1409

### DIFF
--- a/.github/workflows/ciwheel.yml
+++ b/.github/workflows/ciwheel.yml
@@ -24,38 +24,10 @@ jobs:
         with:
           fetch-depth: 0
           submodules: recursive
-      - name: Update pip
-        run: python -m pip install --upgrade pip
-
-      - name: Build wheels Linux
-        if: ${{ startsWith(matrix.os, 'ubuntu') }}
-        uses: pypa/cibuildwheel@v2.3.0
-        with:
-          output-dir: dist
-        env:
-          CIBW_BEFORE_ALL: yum -y install libxml2-devel
-          CIBW_BEFORE_BUILD: python -m pip install numpy setuptools scikit-build ninja cmake
-          CIBW_BUILD: "cp3*-manylinux_x86_64"
-          CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
-          CIBW_ARCHS_LINUX: x86_64
-          CIBW_REPAIR_WHEEL_COMMAND: 'auditwheel repair -w {dest_dir} {wheel} && python /project/scripts/patchwheel.py {dest_dir}'
-          CIBW_TEST_COMMAND: python -m unittest discover -v -s {project}/python
-
-      - name: Build wheels macos
-        if: ${{ startsWith(matrix.os, 'macos') }}
-        uses: pypa/cibuildwheel@v2.3.0
-        with:
-          output-dir: dist
-        env:
-          MACOSX_DEPLOYMENT_TARGET: "10.15" #needed to undo some CIBW settings
-          CIBW_BEFORE_BUILD: python -m pip install numpy setuptools scikit-build ninja cmake
-          CIBW_BUILD: "cp3*-macosx_x86_64"
-          CIBW_ARCHS_MACOS: x86_64 universal2
-          CIBW_TEST_COMMAND: python -m unittest discover -v -s {project}/python
-
-      # this action runs auditwheel automatically with the following args:
-      # https://cibuildwheel.readthedocs.io/en/stable/options/#repair-wheel-command
-
+      - name: Install cibuildwheel
+        run: python3 -m pip install cibuildwheel
+      - name: Build wheels
+        run: python3 -m cibuildwheel --output-dir dist
       - uses: actions/upload-artifact@v2
         with:
           name: dist
@@ -68,18 +40,16 @@ jobs:
     steps:
       - name: Set up Python
         uses: actions/setup-python@v2
-      - name: Update pip
-        run: python -m pip install --upgrade pip
       - name: Get packages
-        run: python -m pip install numpy setuptools scikit-build ninja cmake
+        run: python3 -m pip install build
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
           submodules: recursive
       - name: Make sdist
-        run:  python setup.py sdist
+        run: python -m build -s
       - name: Install sdist
-        run:  python -m pip install dist/arbor*.tar.gz
+        run:  python3 -m pip install dist/arbor*.tar.gz
       - name: Run Python tests
         run: python3 -m unittest discover -v -s python
       - name: Run Python examples
--- a/.github/workflows/ciwheel.yml
+++ b/.github/workflows/ciwheel.yml
@@ -47,7 +47,7 @@ jobs:
           fetch-depth: 0
           submodules: recursive
       - name: Make sdist
-        run: python -m build -s
+        run: python3 -m build -s
       - name: Install sdist
         run:  python3 -m pip install dist/arbor*.tar.gz
       - name: Run Python tests
--- a/doc/install/python.rst
+++ b/doc/install/python.rst
@@ -6,7 +6,7 @@ Python Installation
 Arbor's Python API will be the most convenient interface for most users.
 
 .. note::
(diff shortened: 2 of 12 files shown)


.github/workflows/ciwheel.yml
```diff
@@ -24,38 +24,10 @@ jobs:
...
-          CIBW_ARCHS_MACOS: x86_64 universal2
-          CIBW_TEST_COMMAND: python -m unittest discover -v -s {project}/python
-
-      # this action runs auditwheel automatically with the following args:
-      # https://cibuildwheel.readthedocs.io/en/stable/options/#repair-wheel-command
-
+      - name: Install cibuildwheel
+        run: python3 -m pip install cibuildwheel
+      - name: Build wheels
+        run: python3 -m cibuildwheel --output-dir dist
```
Helveg, line 30: We could switch to pypa's Github Action?

.github/workflows/ciwheel.yml
```diff
@@ -68,18 +40,16 @@ jobs:
...
       - name: Get packages
-        run: python -m pip install numpy setuptools scikit-build ninja cmake
+        run: python3 -m pip install build
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
           submodules: recursive
       - name: Make sdist
-        run:  python setup.py sdist
+        run: python -m build -s
```
Helveg, line 50:
```suggestion
        run: python3 -m build -s
```

doc/install/python.rst
```diff
@@ -77,22 +73,17 @@ Every time you make changes to the code, you'll have to repeat the second step.
...
-By default Arbor is installed with multi-threading enabled. To enable more
-advanced forms of parallelism and other features, Arbor comes with a few
-compilation options. These are of the form ``-D<KEY>=<VALUE>``, must be appended
-to the ``pip`` invocation via ``--install-option="-D<...>" --install-option="-D<...>" ...`` and can
-be used on both local (``pip3 install ./arbor``) and remote (``pip3 install
-arbor``) copies of Arbor. See the examples below.
+Arbor comes with a few compilation options, some of them related to advanced forms of parallelism and other features.
+The options and flags are the same :ref:`as documented for the CMAKE build <quickstart>`, but they are passed differently. To enable more
+They must be placed in the ``CMAKE_ARGS`` environment variable. The simplest way to do this is by prepending the ``pip``
+command with ``CMAKE_ARGS=""``, where you place the arguments separated by space inside the quotes.
```
Helveg, line 79:
```suggestion
Arbor comes with a few compilation options, some of them related to advanced forms of parallelism and other features.
The options and flags are the same :ref:`as documented for the CMAKE build <quickstart>`, but they are passed differently. To enable more
they must be placed in the ``CMAKE_ARGS`` environment variable. The simplest way to do this is by prepending the ``pip``
command with ``CMAKE_ARGS=""``, where you place the arguments separated by space inside the quotes.
```

Also, line lengths

doc/install/python.rst
```diff
@@ -117,6 +108,12 @@ The following flags can be used to configure the installation:
    and ``CMake`` under the hood, so all flags and options valid in ``CMake`` can
    be used in this fashion.
 
+   Allthough the `scikit-build documentation <https://scikit-build.readthedocs.io/en/latest/usage.html#environment-variable-configuration>`_
+   mentions that you can also pass the build options with ``--install-option=""``, this will cause ``pip`` to build
+   all dependencies, including all build-dependencies, instead of downloading them from PyPI. ``CMAKE_ARGS=""``
+   saves you the build time, and also downloading and setting up the dependencies they in turn require to be present.
+   Setting ``CMAKE_ARGS=""`` is in addition compatible with build front-ends like `build <https://pypa-build.readthedocs.io>`_.
+
```
Helveg, line 116: line length? There's usually a keyboard shortcut to linewrap a paragraph

Helveg (changes requested):

.github/workflows/ciwheel.yml
brenthuisman: The reasons for switching from their Action is that this way we're always running their latest. With the Action you need to pick a version... I prefer to catch their updates automatically (and get failure reports early).

.github/workflows/ciwheel.yml
brenthuisman: Nice catch!

doc/install/python.rst
brenthuisman: I'm opening up a new front in the war: https://sive.rs/1s

doc/install/python.rst
brenthuisman: It's consistent with some other line lengths in this file, but not all, correct. I'll break the again at `.`'s and `'`'s.

.github/workflows/ciwheel.yml
Helveg: 👍

Helveg (approved):
LGTM

spack/package.py
```diff
@@ -69,14 +69,14 @@ class Arbor(CMakePackage, CudaPackage):
 
     # python (bindings)
     extends("python", when="+python")
-    depends_on("python@3.6:", when="+python", type=("build", "run"))
+    depends_on("python@3.7:", when="+python", type=("build", "run"))
     depends_on("py-numpy", when="+python", type=("build", "run"))
     with when("+python"):
         depends_on("py-pybind11@2.6:", type=("build", "run"))
```
thorstenhater, line 75: This this weird. Do we depend on pybind11 2.6 or 2.8? Also, isn't pybind11 header-only?

thorstenhater (approved):
Single question, nothong blocking though.

spack/package.py
brenthuisman: 1. The second pybind line has the `when` option: there was a breaking change in pybind.
2. Yes, but that does not mean it's not a build-dep that Spack needs to supply? The submodules are not used for Spack builds. Perhaps you mean `'run'`? Yes, that could probably be removed!

spack/package.py
```diff
@@ -69,14 +69,14 @@ class Arbor(CMakePackage, CudaPackage):
...
     # python (bindings)
     extends("python", when="+python")
-    depends_on("python@3.7:", when="+python", type=("build", "run"))
+    depends_on("python@3.6:", when="+python", type=("build", "run"))
     depends_on("py-numpy", when="+python", type=("build", "run"))
     with when("+python"):
-        depends_on("py-pybind11@2.6:", type=("build", "run"))
-        depends_on("py-pybind11@2.8.1:", when="@0.5.3:", type=("build", "run"))
+        depends_on("py-pybind11@2.6:", type=("build"))
+        depends_on("py-pybind11@2.8.1:", when="@0.5.3:", type=("build"))
```
thorstenhater, line 76: Here's another question:
If we did not update this line **only** 0.5.3 would get 2.8.1? Meaning shouldn't it be
>=0.5.3 or Spack's equivalent? Or can we rely on CI to catch this?

brenthuisman: The trailing colon signifies 'and up'.

thorstenhater: That's a sneaky bit of syntax

PR #1903, opened 2022-06-09
Convenience: shorter code to build decor.

thorstenhater:
Add method chaining to `decor::{place, paint, set_default}`, so that we can now write
```c++
auto decor = arb::decor{}
  .paint("soma"_lab, arb::density("hh"))
  .paint("dend"_lab, arb::density("pas"))
  .set_default(arb::axial_resistivity{100}) // [Ω·cm]
```
to save some redundant typing. Adds the same to Python. Also modernise Python examples
where obvious.

Related to: #1862

PR #1917, opened 2022-06-24
Fix line numbers in tutorials and assorted doc corrections

brenthuisman:
- Fix line numbers in tutorials, fallout from #1906
- Mention arbor-contrib in a few relevant places
- Correct some fallout from #1904
- A new hardware and profiling tutorial, and covers things moved out of other examples in #1904
- Various other documentation fixes


doc/tutorial/single_cell_detailed_recipe.rst
```diff
@@ -90,38 +80,38 @@ what we did in the :ref:`previous example <tutorialsinglecellswc-gprop>`. One la
...
+Step **(5.8)** overrides the :meth:`arbor.recipe.event_generators` method. It takes one argument: ``gid``.
 Given the gid, this method returns *event generators* on that cell. These generators trigger events (or
 spikes) on specific *targets* on the cell. They can be used to simulate spikes from other cells, to kick-start
 a simulation for example. Our cell uses a current clamp as a stimulus, and has no targets, so we return
 an empty list.
 
-Step **(6.9)** overrides the :meth:`arbor.recipe.global_properties` method. It takes one argument: ``kind``.
+Step **(5.9)** overrides the :meth:`arbor.recipe.global_properties` method. It takes one argument: ``kind``.
 This method returns the default global properties of the model which apply to all cells in the network of
 that kind. We return ``self.the_props`` which we defined in step **(1)**.
```
thorstenhater, line 116: Add something like 

> We only use `cable` cells in this example (but there are more) and thus always return a `cable_cell_properties` object.

doc/tutorial/single_cell_detailed_recipe.rst
```diff
@@ -90,38 +80,38 @@ what we did in the :ref:`previous example <tutorialsinglecellswc-gprop>`. One la
...
 connections from other cell *sources* to specific *targets* on the cell with id ``gid``. Since we are
 simulating a single cell, and self-connections are not possible, we return an empty list.
 
-Step **(6.7)** overrides the :meth:`arbor.recipe.gap_junctions_on` method. It takes one argument: ``gid``.
+Step **(5.7)** overrides the :meth:`arbor.recipe.gap_junctions_on` method. It takes one argument: ``gid``.
 Given the gid, this method returns all the gap junctions on that cell. Gap junctions require 2 separate cells.
 Since we are simulating a single cell, we return an empty list.
 
-Step **(6.8)** overrides the :meth:`arbor.recipe.event_generators` method. It takes one argument: ``gid``.
+Step **(5.8)** overrides the :meth:`arbor.recipe.event_generators` method. It takes one argument: ``gid``.
```
thorstenhater, line 108: Here (and all similar)

We override these methods, but with their (empty) defaults. I think it's potentially better not to explain these methods in this early tutorial to reduce the amount of information to take in for the reader. Potentially though this should be part of a different PR. What's your idea on this?

doc/tutorial/single_cell_detailed_recipe.rst
```diff
@@ -59,18 +49,18 @@ examine the recipe in detail: how to create one, and why it is needed.
...
-Step **(6)** creates a ``single_recipe`` class that inherits from :class:`arbor.recipe`. The base recipe
+Step **(5)** creates a ``single_recipe`` class that inherits from :class:`arbor.recipe`. The base recipe
 implements all the methods defined above with default values except :meth:`arbor.recipe.num_cells`,
 :meth:`arbor.recipe.cell_kind` and :meth:`arbor.recipe.cell_description` which always have to be implemented
 by the user. The :meth:`arbor.recipe.global_properties` also needs to be implemented for
 :class:`arbor.cell_kind.cable` cells. The inherited recipe can implement any number of additional methods and
 have any number of instance or class variables.
 
-Step **(6.1)** defines the class constructor. In this case, we pass a ``cell`` and a set of ``probes`` as
+Step **(5.1)** defines the class constructor. In this case, we pass a ``cell`` and a set of ``probes`` as
```
thorstenhater, line 63: Note below as well, no need to send `probes` around. Do you know why this was done?

doc/tutorial/network_ring_gpu.rst
```diff
@@ -0,0 +1,126 @@
...
+***********
+
+The scientific results should be similar, other than number of cells, to those in :ref:`ring network <tutorialnetworkring>`,
+so we'll not discuss them here. Let's turn our attention to the :class:`~arbor.meter_manager`.
+
+.. literalinclude:: ../../python/example/network_ring_gpu.py
+   :language: python
+   :lines: 161-163
+
+Step **(18)** shows how :class:`arbor.meter_report` can be used to read out the :class:`~arbor.meter_manager`. It generates a table with the time between checkpoints. For example:
```
thorstenhater, line 110: The output below looks different for GPUs, no?

doc/tutorial/network_ring_gpu.rst
```diff
@@ -0,0 +1,126 @@
...
+A way to customize :class:`arbor.partition_load_balance` is by providing a :class:`arbor.partition_hint`. They let
+you configure how cells are distributed over the resources in the :class:`~arbor.context`, but without requiring you
+to know the precise configuration of a :class:`~arbor.context` up front. Whether you run your simulation on your 
+laptop CPU, desktop GPU, CPU cluster of GPU cluster, using :class:`partition hints<arbor.partition_hint>` you can
+just say: use GPUs, if available. You only have to change the :class:`~arbor.context` to actually define which 
+hardware Arbor will execute on.
+
+Step **(14)** creates a :class:`arbor.partition_hint`, and tells it to put 1000 cells in a groups allocated to GPUs, 
+and to prefer the utilization of the GPU if present. Lastly, a dictionary is created with which hints are assigned 
+to a particular :class:`arbor.cell_kind`. Different kinds may favor different execution, hence the option. In this
```
thorstenhater, line 71: > favour?

UK spelling used elsewhere.

doc/tutorial/network_ring_gpu.rst
```diff
@@ -0,0 +1,126 @@
...
+context, creates the :class:`arbor.domain_decomposition` object for us.
+
+A way to customize :class:`arbor.partition_load_balance` is by providing a :class:`arbor.partition_hint`. They let
+you configure how cells are distributed over the resources in the :class:`~arbor.context`, but without requiring you
+to know the precise configuration of a :class:`~arbor.context` up front. Whether you run your simulation on your 
+laptop CPU, desktop GPU, CPU cluster of GPU cluster, using :class:`partition hints<arbor.partition_hint>` you can
+just say: use GPUs, if available. You only have to change the :class:`~arbor.context` to actually define which 
+hardware Arbor will execute on.
+
+Step **(14)** creates a :class:`arbor.partition_hint`, and tells it to put 1000 cells in a groups allocated to GPUs,
```
thorstenhater, line 69: I _think_ GPU `p_l_b` prefers to maximise CellGroup sizes, but double check?

doc/tutorial/network_ring.rst
```diff
@@ -107,21 +107,23 @@ Step **(11)** instantiates the recipe with 4 cells.
...
 
 Step **(12)** initalizes the ``threads`` parameter of :class:`arbor.context` with the ``avail_threads`` flag. By supplying
 this flag, a context is constructed that will use all locally available threads. On your local machine this will match the
 number of logical cores in your system. Especially with large numbers
 of cells you will notice the speed-up. (You could instantiate the recipe with 5000 cells and observe the difference. Don't
 forget to turn of plotting if you do; it will take more time to generate the image then to run the actual simulation!)
-:func:`arbor.partition_load_balance` creates a default domain decomposition, which 
-for contexts initialized with ``threads=avail_threads`` distributes cells evenly over the available cores. You can print the
-objects to see what defaults they produce on your system.
+If no domain decomposition is specified, a default distributing work equally is created. This is sufficient for now.
```
thorstenhater, line 125: > default [one]?

doc/ecosystem/index.rst
```diff
@@ -3,7 +3,14 @@
 Ecosystem
 =========
 
-Arbor exists in a wider computational neuroscience software ecosystem. The Arbor simulator is also not the only component written by the Arbor developers.
+Arbor exists in a wider computational neuroscience software ecosystem. In addition to the Arbor simulator, the Arbor developepment team writes and maintains a set of useful tools to be used in conjuction with Arbor.
```
thorstenhater, line 6: > developepment

thorstenhater (changes requested):
Nice work! A few typos and suggestions.

General comments:
- Adding a new tutorial could have been done in its own PR
- Before merging, I'd suggest merging #1903 because it'll break the numbers again
- Maybe ditch the empty method explanations? It's a lot to take in for first timers and doesn't anything they care (yet) about.
- If we can standardise the marking of links between tutorial .py and .rst, we _could_ make a script to generate the line numbers.

python/example/single_cell_cable.py
```diff
@@ -72,6 +70,13 @@ def cell_kind(self, gid):
         assert gid == 0
         return arbor.cell_kind.cable
 
+    def probes(self, gid):
+        assert gid == 0
+        return self.the_probes
+
+    def global_properties(self, kind):
```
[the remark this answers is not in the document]
brenthuisman: Can we?

I think the asserts in this file might as well be removed, since I can't think of how `gid>0` can ever be interogated when `num_cells == 1`.

python/example/single_cell_detailed_recipe.py
```diff
@@ -105,44 +100,44 @@ def __init__(self, cell, probes):
         self.the_props.set_ion(ion="ca", int_con=5e-5, ext_con=2, rev_pot=132.5)
         self.the_props.catalogue.extend(arbor.allen_catalogue(), "")
 
-    # (6.2) Override the num_cells method
+    # (5.2) Override the num_cells method
     def num_cells(self):
         return 1
 
-    # (6.3) Override the num_targets method
+    # (5.3) Override the num_targets method
```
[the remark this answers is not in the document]
brenthuisman: I guess even Nora misses things ;)

python/example/single_cell_detailed_recipe.py
```diff
@@ -105,44 +100,44 @@ def __init__(self, cell, probes):
...
     def gap_junction_on(self, gid):
         return []
 
-    # (6.8) Override the event_generators method
+    # (5.8) Override the event_generators method
     def event_generators(self, gid):
         return []
 
-    # (6.9) Overrode the global_properties method
+    # (5.9) Overrode the global_properties method
```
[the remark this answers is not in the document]
brenthuisman: The code comments are not terribly useful, but the tutorial text refers to and explains each point. The tutorial is written as a first intro to the `recipe` class, so it makes sense to go into detail on all methods. And we always have a significantly reduced code comment w.r.t. tutorial text (hence the tutorial).

I think we should let it be right now. It's pretty consistent with the other tutorials.

python/example/single_cell_detailed_recipe.py
```diff
@@ -105,44 +100,44 @@ def __init__(self, cell, probes):
...
     def global_properties(self, gid):
         return self.the_props
 
 
 # Instantiate recipe
 # Pass the probe in a list because that it what single_recipe expects.
-recipe = single_recipe(cell, [probe])
+recipe = single_recipe(cell, [arbor.cable_probe_membrane_voltage('"custom_terminal"')])
```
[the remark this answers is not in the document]
brenthuisman: > Can't we move the probe list into the class

Not sure what you mean by that.

doc/tutorial/single_cell_recipe.rst
```diff
@@ -111,7 +107,7 @@ If we create the same analysis of the results we therefore expect the same resul
 
 .. literalinclude:: ../../python/example/single_cell_recipe.py
    :language: python
-   :lines: 75-92
+   :lines: 79-
```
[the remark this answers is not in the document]
brenthuisman: Nope. That means, include till end of file. One less line number to update ;)

doc/tutorial/single_cell_detailed_recipe.rst
brenthuisman: :+1:

doc/tutorial/single_cell_detailed_recipe.rst
brenthuisman: See other comment.

doc/tutorial/single_cell_detailed_recipe.rst
brenthuisman: I don't know why this was done. Maybe to be consistent with the pattern we use for the cell description? That happens in other tutorials this way too. It would be nicer perhaps to move the cell description into a function, or simply into the `cell_description` method. Is that what you mean for the probes as well?

doc/tutorial/network_ring_gpu.rst
brenthuisman: Right, the example output here is for one config, in this case my laptop CPU.

doc/tutorial/network_ring_gpu.rst
brenthuisman: I don't know what `p_l_b` is, but yes, the 1000 cells is superfluous and purely for showing off the option. I'll add a note.

doc/ecosystem/index.rst
brenthuisman: I guess there was a stray synapse encouraging me to write French, and the usual synapses generating noise f'ed it up.

doc/tutorial/network_ring_gpu.rst
thorstenhater: = `partition_load_balance`. Way to long to type out.

doc/tutorial/network_ring_gpu.rst
thorstenhater: Since we advertise GPU stuff in the example that is worth commenting.

python/example/single_cell_cable.py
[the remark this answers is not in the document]
thorstenhater: Jupp, that's fine, too. It also confuses people.

thorstenhater (changes requested):

python/example/single_cell_detailed_recipe.py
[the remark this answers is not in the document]
brenthuisman: We'd be going against the original idea of this tutorial: Nora and I thought it'd be a good idea to show this example and `single_cell_{model|recipe}.py` twice, with no other difference than the use of `recipe` vs `single_cell_model`. The reasoning behind doing it twice is to have each `single_cell_model` tutorial also available in `recipe` version.

Network is not the only thing you may want to switch to recipes for.

So, I think we should keep it. The tutorial begins with a link to the `s_c_m` version, so people who're bothered by useless ceremony know where to go. But those who aren't, might be discovering how to do more involved things this way. 

We're not optimizing for LOC, but for giving people handles for discovery. For reduction of cognitive load, there is already a version that is mentioned up front.

python/example/single_cell_detailed_recipe.py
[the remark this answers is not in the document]
brenthuisman: I'll move the probes, but the cell_description should probably stay broken out keep the flow of the code simpler: to keep it easy to separately discuss the cell and recipe. I'll remove the cell from the constructor.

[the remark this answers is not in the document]
thorstenhater: That's fine, but might warrant a comment (or sentence in the text) along the lines of:
We put it outside the class to make the cell only once (for efficiency) and up front (for explanation),
but this might be done just as well in `__init__` or in `get_cell_description`. The latter 
will be called multiple times (possibly > `num_cells` even).

python/example/single_cell_detailed_recipe.py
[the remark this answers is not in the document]
thorstenhater: My concern is not about the line count, but about the people reading these tutorial. They will have
a lot more things to process 

> We'd be going against the original idea of this tutorial: Nora and I thought it'd be a good idea to show this example and single_cell_{model|recipe}.py twice, with no other difference than the use of recipe vs single_cell_model. The reasoning behind doing it twice is to have each single_cell_model tutorial also available in recipe version.

Even more reason not to be showing these overloads? They aren't accessible from the single cell model version.

python/example/single_cell_detailed_recipe.py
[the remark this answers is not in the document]
brenthuisman: I've gathered up all the hard requirements to the `recipe` structure into an infobox, and made it more clear that other than meeting those requirements, the example is just an example.

doc/tutorial/single_cell_detailed_recipe.rst
```diff
@@ -49,71 +49,78 @@ examine the recipe in detail: how to create one, and why it is needed.
...
 
-Step **(5.1)** defines the class constructor. In this case, we pass a ``cell`` and a set of ``probes`` as
-arguments. These will be used to initialize the instance variables ``self.the_cell`` and ``self.the_probes``,
-which will be used in the overloaded ``cell_description`` and ``get_probes`` methods. Before variable
-initialization, we call the base C++ class constructor ``arbor.recipe.__init__(self)``. This ensures correct
-initialization of memory in the C++ class.
+.. note::
+   The inherited recipe can implement any number of additional methods and have any number of instance or class
+   variables. The structure shown in this example is a choice; it's up to you to structure your code the way you
+   see fit. What matters is that the methods required to be present return valid results of the correct type.
```
thorstenhater, line 65: This can be understood as that the overloads on all methods are required. Think of those less versed in the ways of OOP/Python.

doc/tutorial/single_cell_detailed_recipe.rst
```diff
@@ -49,71 +49,78 @@ examine the recipe in detail: how to create one, and why it is needed.
...
 :class:`arbor.cell_kind.cable`, so we simply return that.
 
-Step **(5.4)** overrides the :meth:`arbor.recipe.cell_description` method. It takes one argument: ``gid``.
+Step **(5.4)** overrides the :meth:`~arbor.recipe.cell_description` method. It takes one argument: ``gid``.
 Given the gid, this method returns the cell description which is the cell object passed to the constructor
-of the recipe. We return ``self.the_cell``.
+of the recipe. We return ``cell``, the cell created just above.
 
-Step **(5.5)** overrides the :meth:`arbor.recipe.get_probes` method. It takes one argument: ``gid``.
+Step **(5.5)** overrides the :meth:`~arbor.recipe.get_probes` method. It takes one argument: ``gid``.
```
thorstenhater, line 99: Why the `~`?

doc/tutorial/single_cell_detailed_recipe.rst
brenthuisman: TIL you don't need to override methods you don't use in your model. I'm not sure it'll bring much clarify to make that distinction here though. The way I read it now implies all that's overloaded in this example needs to be overloaded always, which is a fine message to pass on imho, even if not 100% correct.

python/example/single_cell_detailed_recipe.py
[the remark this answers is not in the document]
thorstenhater: Let's go the other direction: What do you think showing these overloads will teach people that they need to know here?
If it's the pure presence of these methods, add a note 'there's more, see later tutorials' and a link to the docs/tuts/...
Showing them here is honestly confusing, as there's no mention of 'this is just showing you these as options and we re-implement the defaults, you'd get the same result w/o any mention of these'. Users will needlessly be implementing these defaults over and over. The parallel between single cell model and this was good, stick to it ;)

python/example/single_cell_detailed_recipe.py
```diff
@@ -105,44 +99,43 @@ def __init__(self, cell, probes):
...
+        return [arbor.cable_probe_membrane_voltage('"custom_terminal"')]
 
-    # (6.6) Override the connections_on method
+    # (5.6) Override the connections_on method
     def connections_on(self, gid):
         return []
 
-    # (6.7) Override the gap_junction_on method
+    # (5.7) Override the gap_junction_on method
     def gap_junction_on(self, gid):
```
thorstenhater, line 123: This must absolutely be `gap_junctions_on`.

thorstenhater (changes requested):

doc/tutorial/single_cell_detailed_recipe.rst
brenthuisman: Tilde compression: to avoid printing `arbor.recipe.` all the time.

brenthuisman:
New tutorial category: `recipes`. Details about overloading methods and that otherwise the shape of a derived recipe is a matter of choice mentioned the first time, and not again.

doc/python/recipe.rst
```diff
@@ -18,11 +18,23 @@ Recipe
...
-
+    
     Recipes provide a cell-centric interface for describing a model.
     This means that model properties, such as connections, are queried using the global identifier ``gid`` of a cell.
     In the description below, the term ``gid`` is used as shorthand for the cell with global identifier.
 
+    **Required Constructor**
+
+    The constructor must be implemented and call the base class constructor, as at the moment there is no way
+    to instruct Python to do that automatically. This ensures correct initialization of memory in the C++ class.
```
thorstenhater, line 29: and the user asks: C++? This is Python?! What's going on?

In all seriousness, add a note here reiterating the point that pyarb is a thin wrapper around C++ Arbor.
Then: Due to the py<->cxx binding process omitting this leads to memory corruption.

doc/tutorial/single_cell_recipe.rst
```diff
@@ -38,9 +42,15 @@ and overrides and implements some of :class:`arbor.recipe` methods. Not all meth
 have to be overridden, but some will always have to be, such as :meth:`arbor.recipe.num_cells`.
 It returns `0` by default and models without cells are quite boring!
 
+.. note::
```
thorstenhater, line 45: I find this note  confusing. What's the argument being made? I assume it is this:
> A recipe must derive from `arbor.recipe` and must implement a number of methods (documented [there]).
In addition some methods may optionally be overridden. Beyond this, it is up to you, the user, to structure your code as you find convenient.

Also: `inherited` is not a word that can be used here. You can `inherit from` or have a `derived class`. In OOP that is.

doc/tutorial/single_cell_detailed_recipe.rst
```diff
@@ -3,127 +3,82 @@
...
+``arbor.recipe.__init__(self)`` to ensure correct initialization of memory in the C++ class.
 
-We also create the ``self.the_cat`` variable and set it to arbor's default mechanism catalogue. This will expose
-the *hh* and *pas* mechanisms but not the *Ih* mechanism, which is present in the allen catalogue. To be able
-to use *Ih*, we extend ``self.the_cat`` to include the allen catalogue.
+We also create an instance variable for the mechanism catalogue, the ``self.the_cat`` variable. This ensures
+its lifetime is guaranteed to extend to the entire duration of the simulation. Setting the mechanism catalogue
+in the constructor is a pattern you'll see return in these tutorials.
+
+We set ``self.the_cat`` to Arbor's default mechanism catalogue. This will expose the *hh* and *pas* mechanisms but not the *Ih* mechanism,
```
thorstenhater, line 52: This is no longer needed in the code. You can just write `self.the_props.catalogue.extend(...)`.
Also it has been already modernised, this is just outdated ;)

doc/python/recipe.rst
```diff
@@ -18,11 +18,23 @@ Recipe
...
     All recipes derive from this abstract base class.
-
+    
     Recipes provide a cell-centric interface for describing a model.
     This means that model properties, such as connections, are queried using the global identifier ``gid`` of a cell.
     In the description below, the term ``gid`` is used as shorthand for the cell with global identifier.
 
+    **Required Constructor**
+
+    The constructor must be implemented and call the base class constructor, as at the moment there is no way
```
thorstenhater, line 28: This is -- incidentally -- endemic to Python.
```
In [1]: class A:
   ...:     def __init__(self):
   ...:         print("A")
In [2]: class B(A):
   ...:     def __init__(self):
   ...:         print("B")
In [3]: a = A()
A
In [4]: b = B()
B
In [5]: class C(A):
   ...:     def __init__(self):
   ...:         super().__init__() # same a A.__init__() but more general
   ...:         print("C")
In [8]: c = C()
A
C
```

thorstenhater (changes requested):

doc/tutorial/single_cell_recipe.rst
brenthuisman: I reworded this.

doc/tutorial/single_cell_detailed_recipe.rst
brenthuisman: In future, someone writing new features might consider a pass by the examples then ;)

doc/python/recipe.rst
brenthuisman: I know, and it's unpleasant...

thorstenhater (approved):
Thanks, that was a ton of work. You might consider the one stylistic nit, but nothing blocking.

PR #1919, opened 2022-06-30
Mutable connection table

thorstenhater:
Add functionality, docs, and examples on editing the connection table.  This is a first, small PR
on the topic, further functionality will come as requested.

Teaser example
```cxx
    // This recipe -- omitted -- implements a mutable connection table 
    // and returns it via the `connections_on` callback. New entries are
    // added using `add_connection` which makes one link from a central 
    // `spike_source_cell` to the given `gid`
    auto rec = recipe();
    rec.add_connection(1);

    auto sim = arb::simulation(rec, ctx);

    // Record spikes
    std::vector<arb::spike> spikes;
    sim.set_local_spike_callback([&spikes](const auto& s) { spikes.insert(spikes.end(), s.begin(), s.end()); });

    // Run and print all spikes per gid
    sim.run(0.25, 0.025);
    show_spikes(spikes);

    // Add a new connection
    rec.add_connection(2);
    sim.update_connections(rec);

    // Run and print all spikes per gid
    sim.run(0.5, 0.025);
    show_spikes(spikes);
```

# Update July '22

- added re-wiring of Event Generators
- implemented a split between a recipe and a recipe-update (cheekily dubbed `topping` until further notice)
- `recipes` are `toppings` via inheritance


thorstenhater:
Good questions, I'll answer the easy one first (2)
I decided on using the recipe since it is a familiar interface to users. Regarding the edit command approach,
I wanted to reach for that first. However the code for building the actual table made it clear that all the
expensive work would have to be done again anyhow. I am referring to the sorting and partitioning of the 
actual backing tables. The rest is relatively cheap. Also, we'd need to keep around (or re-create) a set of
items that is used in builing the table, resolving the labels: `domain_decomposition`, `context` and the `label_resolver`s.
In general, performance is not really a consideration, since this is expected to occur every 1s-1min of biological time
or so (in vivo it's more like 10min-1h).

There's a additional wrinkle though: What if connectivity is random? Resetting the pRNG helps, but might not be sufficient
as other parts of the recipe/sim-building might call into it.

halfflat:
It would be all potential connections given the biological model; it might grow as O(N^2) for small N, where every neuron is 'close enough' to every other neuron, but it should still be O(N) for large models. Extra traffic would only arise in the cases where a source has all of its potential targets with zero weights (i.e. connected to nothing at all); not creating events with zero weight at the delivery side of things would be the optimization.

thorstenhater:
Hmm, about that for a compromise (fixing the perfectly valid ontological concerns): I factor out the rebuildable parts into 
their own components and recipes aggregate those components as do the 'update-recipes'. Just different subsets thereof.

python/example/plasticity.py
```diff
@@ -0,0 +1,74 @@
...
+        if gid in self.connected:
+            w, d = self.connected[gid]
+            return [A.connection((0, "source"), "synapse", w, d)]
+        return []
+
+    def add_connection(self, to):
+        self.connected[to] = (0.75, 0.1)
+
+
+A.mpi_init()
```
brenthuisman, line 47: Is it necessary to use MPI to make the change in connection table clear?

python/example/plasticity.py
```diff
@@ -0,0 +1,74 @@
...
+mpi = A.mpi_comm()
+ctx = A.context(mpi=mpi)
+rnk = ctx.rank
+csz = ctx.ranks
+assert csz == 3
```
brenthuisman, line 52: I don't see why this `context` would have 3 ranks. It does not know about the recipe yet.

example/plasticity/plasticity.cpp
```diff
@@ -0,0 +1,116 @@
...
+#ifdef ARB_MPI_ENABLED
+    arbenv::with_mpi guard(argc, argv, false);
+    auto ctx = arb::make_context({}, MPI_COMM_WORLD);
+#else
+    auto ctx = arb::make_context();
+#endif
+    auto rnk = arb::rank(ctx);
+    auto csz = arb::num_ranks(ctx);
+    auto rec = recipe(csz);
+    auto dec = arb::domain_decomposition(rec, ctx, {{rnk == 0 ? arb::cell_kind::spike_source : arb::cell_kind::cable, {rnk}, arb::backend_kind::multicore}});
```
brenthuisman, line 101: That's a _very_ terse partition hint!

example/plasticity/readme.md
```diff
@@ -0,0 +1,23 @@
+# Plasticity Example
+
+A miniapp that demonstrates how to use mutable connectivity. Should be run with
```
brenthuisman, line 3: It looks like the number of cable cells is a variable. You could mention this here, and maybe explain how that number is ultimately set (e.g. by passing a number of ranks to SLURM).

python/example/plasticity.py
```diff
@@ -0,0 +1,74 @@
+import arbor as A
```
brenthuisman, line 1: Can we add this file to `scripts/run_python_examples.sh`?

example/plasticity/plasticity.cpp
```diff
@@ -0,0 +1,116 @@
+#include <any>
```
brenthuisman, line 1: Can we add this file to `scripts/run_cpp_examples.sh`?

brenthuisman (changes requested):

brenthuisman:
One more thing: proper definitions for `partition`, `domain` and perhaps `cell_group` (this is sometimes, maybe, used as alias for domain, but it is not, I think!) are not present anywhere in the docs, yet relevant in the discussion of `arbor::communicator`. They would go well in the currently very brief `doc/concepts/domdec.rst`.

doc/dev/communication.rst
```diff
@@ -0,0 +1,251 @@
...
+   };
+
+will contain all connections in ``connections_`` partitioned by the
+domain of the source's ``gid`` in ``dom_dec``. Beginnings of the
+respective partitions are pointed to by the indices in
+``connection_part_``.
+
+The algorithm for building is slightly obscured by caching and the use
+of labels and resolving them via ``target_/source_resolution_map`` to
+local ids on the respective source and target cells.
```
lukasgd, line 248: These `label_resolution_map` were a bit confusing - are they documented somewhere?

arbor/simulation.cpp
```diff
@@ -261,6 +265,14 @@ simulation_state::simulation_state(
     epoch_.reset();
 }
 
+void simulation_state::update_connections(const recipe& rec) {
+    auto spikes = communicator_.num_spikes(); // TODO(TH) add those to the next communicator
+    communicator_ = arb::communicator(rec, ddc_, source_resolution_map_, target_resolution_map_, *ctx_);
+    // Use half minimum delay of the network for max integration interval.
+    t_interval_ = communicator_.min_delay()/2;
```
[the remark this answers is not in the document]
thorstenhater: Because we use the `min_delay` in our time stepping and it is defined as `min { c.delay | c in connections }`.
Thus an update might change the minimum.

arbor/communication/communicator.cpp
```diff
@@ -96,16 +96,17 @@ communicator::communicator(const recipe& rec,
     connections_.resize(n_cons);
     util::make_partition(connection_part_, src_counts);
     auto offsets = connection_part_;
-    std::size_t pos = 0;
+    auto src_domain = src_domains.begin();
```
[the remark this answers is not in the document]
thorstenhater: It's clean-up mostly (I wouldn't say 'cosmetic' though) which came up during designing the update functionality.

python/example/plasticity.py
thorstenhater: It was needed to ensure proper workings across multiple processes.

doc/dev/communication.rst
thorstenhater: Unfortunately not, they are an internal thing added during the switch 
```cxx
// Old
d = A.decor{};
d.place(...);
d.place(...);
do_something_to_item_two((0,1)); 

// New
d = A.decor{};
d.place(..., "item1");
d.place(..., "item2");
do_something_to_item_two("item2");
```
These maps translate textual labels, eg `"item"` into cell local addresses `(cell, item)`.

I can add the above explanation to a note/sidebar thing though.

doc/dev/communication.rst
```diff
@@ -0,0 +1,251 @@
...
+
+The ``connections`` vector is a list of connections partitioned by the
+domain (as in domain decomposition) of their source's ``gid``, while
+``connection_part`` stores the partioning indices.
+
+Building the Connection Table
+=============================
+
+The table of connections on the local rank is built during the
+connection of the ``communicator`` object
```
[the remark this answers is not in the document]
thorstenhater: Oh yes.

arbor/simulation.cpp
```diff
@@ -261,6 +265,14 @@ simulation_state::simulation_state(
     epoch_.reset();
 }
 
+void simulation_state::update_connections(const recipe& rec) {
+    auto spikes = communicator_.num_spikes(); // TODO(TH) add those to the next communicator
```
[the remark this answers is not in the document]
thorstenhater: Outdated, this was addressed and does what the comment suggested.

arbor/simulation.cpp
```diff
@@ -261,6 +265,14 @@ simulation_state::simulation_state(
...
+    communicator_ = arb::communicator(rec, ddc_, source_resolution_map_, target_resolution_map_, *ctx_);
```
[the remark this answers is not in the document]
thorstenhater: Also has changed, which caused the to be expected issues.

example/plasticity/plasticity.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+// Prints the locally collected spikes -- rank by rank -- and then clears the buffer
+void show_spikes(int ep, std::vector<arb::spike>& spikes, int rnk, int csz) {
+    sync();
+    if (!rnk) std::cout << "Epoch " << ep << '\n';
+    for (int ix = 0; ix < csz; ++ix) {
+        sync();
+        if (ix == rnk) {
+            std::cout << " * Rank " << rnk << '\n';
+            for(const auto& spike: spikes) std::cout << "   * " << rnk << '/' << csz << ": " << spike.source << "@" << spike.time << '\n';
```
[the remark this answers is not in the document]
thorstenhater: Installed

doc/dev/communication.rst
```diff
@@ -0,0 +1,251 @@
...
+       for (auto& d: displs) d /= traits::count();
+       return {std::move(buffer), std::move(displs)};
+   }
+
+   // From gathered_vector
+   template <typename T>
+   struct gathered_vector {
+       std::vector<T> values_;
+       std::vector<int> partition_;
+   };
```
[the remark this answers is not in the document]
thorstenhater: done

example/plasticity/plasticity.cpp
thorstenhater: outdated

doc/dev/communication.rst
```diff
@@ -0,0 +1,251 @@
+.. _communication:
```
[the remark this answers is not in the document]
thorstenhater: Always.

example/plasticity/readme.md
thorstenhater: The example has been overhauled.

doc/dev/communication.rst
```diff
@@ -0,0 +1,251 @@
...
+reduced to a simple time value. The connection is abstracted into a weight and a
+delay; modelling all axonal processes. While this may seem crude, it is a well
+supported model and commonly used in neuronal simulations.
+
+Connections are formed between sources (cable cell: spike detectors) and targets
+(cable cell: synapses). During runtime, events from all sources are concatenated
+on all MPI ranks using ``Allgatherv`` and targets are responsible for selecting
+events they have subscribed to. This is optimised for by sorting events locally
+(by source) and relying on the process layout to convert this into a *globally*
+sorted array.
```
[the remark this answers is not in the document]
thorstenhater: That's not really part of the communication. It's also explained in the dev section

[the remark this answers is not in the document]
thorstenhater: > Maybe it is helpful to clarify that a `arbor::communicator` holds this state (the concrete connection table, even if it's not a table) in Arbor, even when not using MPI (because the word communicator sounds MPI-ish).

That's explained in the second heading. The information here is starting with the goal (communicating spikes) and then working backwards from there (what's a communicator, how to build the tables, ...).

thorstenhater:
> One more thing: proper definitions for `partition`, `domain` and perhaps `cell_group` (this is sometimes, maybe, used as alias for domain, but it is not, I think!) are not present anywhere in the docs, yet relevant in the discussion of `arbor::communicator`. They would go well in the currently very brief `doc/concepts/domdec.rst`.

Note that `partition`'s use is overloaded. In this context it means either an ordered set of indices `I` such that
an array `A` of grouped items has groups with starting indices `I[i]` **or** it refers to the load balancing done by
`partition_load_balance`.

For the rest: I added some definitions.

thorstenhater:
Anything still amiss @brenthuisman @lukasgd

arbor/include/arbor/simulation.hpp
```diff
@@ -34,6 +34,8 @@ class ARB_ARBOR_API simulation {
                const context& ctx=make_context(),
```
boeschf, line 34: should we update `simulation`'s ctors to take `context` by value since it is a `shared_ptr` now?

thorstenhater: Good catch!

boeschf: I see you pass `context` by value now everywhere. I like it.

thorstenhater: Well, it's the proper thing to do.

python/example/plasticity.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class recipe(A.recipe):
+    def __init__(self, n):
+        super().__init__()
+        self.ccp = A.neuron_cable_properties()
+        self.connected = {}
```
brenthuisman, line 8: Upon re-reading I got confused again: _both_ connections are added through `.add_connection()`.

For readability's sake I propose that we instantiate the recipe with an explicit connection dict.

```suggestion
        self.conn_list = {x:[] for x in range(nc)}
```

python/example/plasticity.py
```diff
@@ -0,0 +1,62 @@
...
+        decor.paint("(all)", A.density("pas"))
+        decor.place("(location 0 0.5)", A.synapse("expsyn"), "synapse")
+        decor.place("(location 0 0.5)", A.spike_detector(-10.0), "detector")
+        return A.cable_cell(tree, A.label_dict(), decor)
+
+    def connections_on(self, gid):
+        if gid in self.connected:
+            w, d = self.connected[gid]
+            return [A.connection((0, "source"), "synapse", w, d)]
+        return []
```
brenthuisman, line 42:
```suggestion
        return self.conn_list[gid]
```

python/example/plasticity.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def add_connection(self, to):
+        self.connected[to] = (0.75, 0.1)
+
+
+rec = recipe(3)
+rec.add_connection(1)
```
brenthuisman, line 49:
```suggestion
rec.add_connection(1,A.connection((0, "source"), "synapse", 0.75, 0.1)) # Add connection between cell 0 and 1
```

python/example/plasticity.py
```diff
@@ -0,0 +1,62 @@
...
+        return A.cable_cell(tree, A.label_dict(), decor)
+
+    def connections_on(self, gid):
+        if gid in self.connected:
+            w, d = self.connected[gid]
+            return [A.connection((0, "source"), "synapse", w, d)]
+        return []
+
+    def add_connection(self, to):
+        self.connected[to] = (0.75, 0.1)
```
brenthuisman, line 45:
```suggestion
    def add_connection(self, to,what):
        self.connected[to].append(what)
```

python/example/plasticity.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+rec = recipe(3)
+rec.add_connection(1)
+sim = A.simulation(rec)
+sim.record(A.spike_recording.all)
+
+sim.run(0.25, 0.025)
+
+rec.add_connection(2)
```
brenthuisman, line 55:
```suggestion
rec.add_connection(2,A.connection((0, "source"), "synapse", 0.75, 0.1)) # Add connection between cell 0 and 2
```

doc/dev/communication.rst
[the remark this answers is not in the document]
brenthuisman: The second heading (Exchange of Spikes) shows an implementation for `communicator::exchange`, not a description of a `communicator`. The struct is shown much further down, line 205. I think it's easier to follow this text if people are aware of what it contains up front.

brenthuisman (changes requested):

python/example/plasticity.py
thorstenhater: > Upon re-reading I got confused again: both connections are added through .add_connection().

Yes, for homogeneity, the initial connection is just like others we might call into being.

> For readability's sake I propose that we instantiate the recipe with an explicit connection dict.

I am really not sure this is more readable. Also, it teaches bad habits: Doing it this way circumvents
the lazy construction we encourage. And for good reason, a recent experiment with the brunel.py example
that switched to an explicit connection list in Python blew up at ~4GB RAM used. The lazy version is fine,
though.

As written, we store the minimum data possible.

brenthuisman: The ingredient to `arbor.connection` are now spread out, so it's not immediately obvious what `.add_connection()` does. Alternatively, it can be renamed `.add_connection_to_cell_0()` or somesuch, to make it clear it creates very specific connections.

thorstenhater: Ok, that's verbose, but this _is_ an example after all.

thorstenhater: Done, the example also has way more comments now.

doc/dev/communication.rst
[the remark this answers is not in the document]
thorstenhater: I'd rather not do that upfront, it's way too much to take in. The reasoning is that `exchange` is what people
are looking for and `communicator`'s primary job. Thus the class is built around it and we explain that part 
first. The surrounding and supporting infrastructure and how it's set up comes later.

brenthuisman (approved):

PR #1924, opened 2022-07-19
remove spack ci cache in attempt to fix failing CI

bcumming:


brenthuisman:
Result: yes the problem is with the cache. Independently confirmed by Ebrains: Arbor v0.7 is live there though our package.py. 

Turns out Github many run into similar issues. Let's close this and see if it either solves with time, or try other solutions: #1926

PR #1931, opened 2022-08-01
Action must change `VERSION` except on version tag.

brenthuisman:
Trigger logic for wheel building is incorrect: VERSION must change in all cases except when a tag is pushed.

https://github.com/arbor-sim/arbor/actions/runs/2775920355


thorstenhater (approved):
👍

Issue #1845, opened 2022-02-22
Declaring both reading and writing an ion leads to modcc error 


schmitts:
With Arbor 0.6 having

`USEION p READ pi WRITE pi` in a mechanisms yields:

`the symbol defined pi can't be redeclared`

Omitting `READ` works.

PR #1936, opened 2022-08-04
modcc: Allow redundant, but correct READ declaration.

thorstenhater:
Fixes https://github.com/arbor-sim/arbor/issues/1845


boeschf (approved):
Looks good to me

PR #1937, opened 2022-08-04
Arborio reads from Stream objects now.

### LINKED ISSUE #1459: Allow reading from Buffer Object in PyArborio
Currently we have interfaces in ArborIO's Python wrapper that allows reading from a file like this
```Python
morph = arborio.load_XYZ('filename')
```
The more pythonic version is to read from a buffer/file descriptor-like object, which has a tiny bit
of boilerplate, but is more flexible
```Python
# read from file
with open('filename') as fd:
    morph = arborio.load_XYZ(fd)
# read from somewhere
my_string = foo() # HTTP? Redis?
morph = arborio.load_XYZ(StringIO(my_string))
```

thorstenhater:
Enable pyArborio input routines to consume Stream objects in addition to files by name.
- NML
- NeuroLucida ASC
- SWC

Closes #1459

### DIFF
--- a/python/cable_cell_io.cpp
+++ b/python/cable_cell_io.cpp
@@ -19,14 +19,10 @@ namespace pyarb {
 namespace py = pybind11;
 
 arborio::cable_cell_component load_component(py::object fn) {
-    const auto fname = util::to_path(fn);
-    std::ifstream fid{fname};
-    if (!fid.good()) {
-        throw arb::file_not_found_error(fname);
-    }
-    auto component = arborio::parse_component(fid);
+    auto contents  = util::read_file_or_buffer(fn);
+    auto component = arborio::parse_component(contents);
     if (!component) {
-        throw pyarb_error("Error while trying to load component from \"" + fname + "\": " + component.error().what());
+        throw pyarb_error(std::string{"Error while trying to load component: "} + component.error().what());
     }
     return component.value();
 };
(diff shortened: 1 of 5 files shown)


Helveg (commented):
I'm not up to speed with `arbor` testing, but it seems like you didn't add tests? I'd add a test for a string, a file object opened from filesystem, an in memory buffer, and something opened in an incorrect mode, for example truncated due to opening in write mode, and something opened in binary mode.

thorstenhater:
No tests so far, correct. Note that we _cannot_ read from a string. How would we distinguish a broken SWC in string from from a file name?

Helveg:
> I'd add a test for a string, a file object opened from filesystem, an in memory buffer, and something opened in an incorrect mode, for example truncated due to opening in write mode, and something opened in binary mode.

I'll be clearer:

* Pass a string `path` as argument, as a file path to a morphology file.
* Pass a file object from `open(path, "r")`
* Pass a file-like object from `io.StringIO(open(path, "r"))` to emulate the data coming from other buffers the `io` module supports.
* Pass a file object from `open(path, "w")` and `open(path, "rb")`

brenthuisman:
@Helveg You started this review, let me know when you OK it.

Helveg:
Well I don't think my requested changes have been implemented, right @thorstenhater?

thorstenhater:
They were ... though I forgot to add the relevant file.

thorstenhater:
Here you go.

Helveg (approved):
Appeased. Subject shall live to see another day. 👑

thorstenhater:
Oh ... I was referring to `black` ;).

brenthuisman (approved):

PR #1938, opened 2022-08-05
Excise fvm type aliases

thorstenhater:
- Remove the redundant `fvm_(value|size|index)_type` aliases and replace them with the raw ABI types.
- Simplify includes.


arbor/include/arbor/mechanism_abi.h
```diff
@@ -5,6 +5,9 @@
 
 #ifdef __cplusplus
 extern "C" {
+
+using namespace arb;
```
AdhocMan, line 9: I did a brief test test and the `using namespace` statement seems to be global if placed within an `extern` block, which does not limit its scope. So you might have to use the namespace prefix instead, since this is a header file.

AdhocMan: Or alternatively, I believe an enclosing `namespace arb { ... }` around the extern block should not affect any C symbol names inside and allow you to omit the prefix.

thorstenhater: That's an interesting interaction! I should have trusted my former self here.

thorstenhater: Reverted.

boeschf (approved):

Issue #1647, opened 2021-09-08
Newb-trap in spike recording docs


Helveg:
The [docs](https://docs.arbor-sim.org/en/latest/python/simulation.html#recording-spikes) seem to contain a nice self contained example of recording spikes that I can use on my own recipes in plug-and-play/copy-paste fashion and that it contains all necessary ingredients for recording spikes. It is also highly ranked in SEO concerning recording of spikes and no alternatives or additional docs seem to exist on the topic.

.

.

.

It does fail to mention that a `spike_detector` is required 🙃

PR #1939, opened 2022-08-05
add spike source docs.

thorstenhater:
Explain where spikes come from.

Closes https://github.com/arbor-sim/arbor/issues/1647


Helveg (commented):
Is there anything reasonable that can be done in the code example to show "oh hey, we have 3 spike detectors, so now we have spikes with id 0-2", that'll show the input output relationship?

thorstenhater:
Yes, @brenthuisman earmarked this for himself. But the gist is, we are going to attach a metadata field to the spike callback, eg location as s-exp string and label string, so you can engage in some filtering. At least in C++ that should work, for Python performance might be an issue.

thorstenhater:
Fixed: https://github.com/arbor-sim/arbor/issues/1941

doc/concepts/interconnectivity.rst
```diff
@@ -39,6 +39,17 @@ A recipe lets you define which sites are connected to which.
    action potential
       Spikes travel over :term:`connections <connection>`. In a synapse, they generate an event.
 
+   spike detector
```
brenthuisman, line 42: Have a `:ref:` to Python's `arbor.spike_detector`. It may be confusing to define the concept without referencing the class.

doc/concepts/interconnectivity.rst
```diff
@@ -39,6 +39,17 @@ A recipe lets you define which sites are connected to which.
...
+   Placed on a cell. Possible source of a connection. Detects crossing of a fixed threshold and generates
+   corresponding events.
+
+   spike source cell
```
brenthuisman, line 46: Same: have a `:ref:` to Python API `arbor.spike_source_cell`. It may be confusing to define the concept without referencing the class.

doc/concepts/interconnectivity.rst
```diff
@@ -39,6 +39,17 @@ A recipe lets you define which sites are connected to which.
...
+   Artificial cell to generate spikes on a given schedule.
+
+   recording
```
brenthuisman, line 49: Same-ish: a link to https://docs.arbor-sim.org/en/latest/python/simulation.html#recording-spikes

brenthuisman (changes requested):

doc/concepts/interconnectivity.rst
brenthuisman: Also, there is a stub in `/doc/concepts/decor.rst` for this. Maybe move this documentation there, since talking specifically about being placed (should that list of placeables be termified then?). A second (more abstract?) definition of a `spike detector` might be confusing.

thorstenhater:
> Also, there is a stub in /doc/concepts/decor.rst for this. Maybe move this documentation there, since talking specifically about being placed (should that list of placeables be termified then?). A second (more abstract?) definition of a spike detector might be confusing.

That's the core issue right there, they are both part of the data collection framework _and_ of the networking facilities. 
So, where to categorise them? I'd prefer to explain the detector as part of the interconnectivity as that is the topic where 
I consider them most important. And then -- of course -- link to data extraction (you know, you can actually write out
those spikes we use to network) and how to set them up (since that's how they come into being).

brenthuisman:
> That's the core issue right there, they are both part of the data collection framework _and_ of the networking facilities. So, where to categorise them?

=> they should be in both places. By adding a link I'll let you get away with adding text just once :) BUT, it would be confusing to have some of them in decor->placeables, and others under interconnectivity. Currently in the former there lives `1. Connection sites` and `3. Gap junction connection sites`. Could be then move that over too (and replace with links)? And maybe we can add a few lines (too decor->placeables) on the subject of spike detectors that explains their dual use and details their use for the case where people want to get spiking times/counts out?

doc/concepts/interconnectivity.rst
thorstenhater: It's now cross-referenced and cross-explained in simulation and interconnectivity, plus API and placing ... enough.

thorstenhater:
I likewise approve your additions ... it's still funny to me that we can circumevent review process this way.

PR #1940, opened 2022-08-05
Elaborate on mpi4py.

### LINKED ISSUE #1785: Passing `COMM_WORLD` to context constructors fails if `mpi4py` is installed after `arbor` is built
**Describe the bug**

A regression occurred where MPI communicators can no longer be passed to `context`:

```
>>> import arbor
>>> from mpi4py.MPI import COMM_WORLD
>>> arbor.context(mpi=COMM_WORLD)
RuntimeError: mpi must be None, or an MPI communicator
```

**Context**
I'm on PizDaint's `daint-gpu` stack with Spack built Python.

thorstenhater:
Explain that
- mpi4py might be needed when throwing MPI error.
- mpi4py needs to be baked into Arbor at build time.

Closes #1785

### DIFF
--- a/doc/install/python.rst
+++ b/doc/install/python.rst
@@ -91,8 +91,9 @@ The following flags can be used to configure the installation:
 
 * ``ARB_WITH_NEUROML=<ON|OFF>``: Enable support for NeuroML2 morphologies,
   requires ``libxml2`` library. Default ``OFF``
-* ``ARB_WITH_MPI=<ON|OFF>``: Enable MPI support, requires MPI library.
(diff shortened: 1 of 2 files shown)


Helveg (approved):
That's good!

boeschf (approved):
Looks like it gives a better error message, I think the issue can be resolved like this.

thorstenhater:
Explain that

    mpi4py might be needed when throwing MPI error.
    mpi4py needs to be baked into Arbor at build time.

PR #1943, opened 2022-08-06
Expose find_private_gpu to python

### LINKED ISSUE #1895: Allocating the same GPU on multiple processes leads to `RuntimeError: GPU memory` error
**Describe the bug**
When multiple Arbor processes target the same GPU, i.e. 2+ arbor processes on the same hardware unit create an `arbor.context` with an `arbor.proc_allocation` that targets the same `gpu_id`, the following error occurs:

```
  File "/users/bp000347/arbenv/bin/bsb", line 33, in <module>
    sys.exit(load_entry_point('bsb', 'console_scripts', 'bsb')())
  File "/users/bp000347/arbenv/bsb/bsb/cli.py", line 26, in scaffold_cli
    start_cli()
  File "/users/bp000347/arbenv/bsb/bsb/cli.py", line 255, in start_cli
    scaffoldInstance.run_simulation(cl_args.simulation, quit=True)
  File "/users/bp000347/arbenv/bsb/bsb/core.py", line 386, in run_simulation
    simulation, simulator = self.prepare_simulation(simulation_name)
  File "/users/bp000347/arbenv/bsb/bsb/core.py", line 419, in prepare_simulation
    simulator = simulation.prepare()
  File "/users/bp000347/arbenv/bsb/bsb/simulators/arbor/adapter.py", line 456, in prepare
    simulation = arbor.simulation(recipe, self.domain, context)
RuntimeError: GPU memory:: gpu_malloc unable to allocate 322812 bytes: all CUDA-capable devices are busy or unavailable
```

There are several problems:

* This seems like an entirely preventable error: Don't allow arbor processes to target the same GPU, you have all information in the `arbor.context` to figure out this will happen, and to prevent it. Some mention of this restriction in the docs on hardware allocation would also be good.
* The error does not contain enough information to point to a cause, who knows why the CUDA device is busy or unavailable, users will probably be unaware that they double-booked their GPU:
  * The call stack trace ends in `arbor.simulation`, no deeper CPP information is captured (see #1896)
  * The error message is a non-descript message, presumably raised by nvidia, uncaught, wrapped by pyarb and reraised
* Because the error is gobbled up by pyarb, there's no straightforward (or documented by Arbor) way to interrupt at the point where the error occurs. (see #1897)

thorstenhater:
Adds an `arbor.env` submodule to our bindings and exposes `find_private_gpu` through it.

Closes #1895

### DIFF
--- a/doc/python/hardware.rst
+++ b/doc/python/hardware.rst
@@ -68,6 +68,20 @@ Helper functions for checking cmake or environment variables, as well as configu
 
     Check if MPI is finalized.
 
+Env: Helper functions
+---------------------
+
+The ``arbor.env`` module collects helper functions for interacting with the environment.
+
+.. function:: find_private_gpu(comm)
+
+   Requires GPU and MPI. Will return an integer id of a GPU such that each GPU
+   is mapped to at most one MPI task (on the same node as the GPU). Raises an
+   exception if
+   - not built with GPU or MPI support
+   - unable to satisfy the constraints above
+   - handed an invalid or unknown MPI communicator object
+
 Prescribed resources
 ---------------------
 
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -38,6 +38,7 @@ set(pyarb_source
     schedule.cpp
     simulation.cpp
     single_cell_model.cpp
+    env.cpp
 )
 
 # compile the pyarb sources into an object library that will be
--- a/python/env.cpp
+++ b/python/env.cpp
@@ -1,0 +1,35 @@
+#include <pybind11/pybind11.h>
+
+#include <arborenv/gpu_env.hpp>
+
+#include "mpi.hpp"
+#include "error.hpp"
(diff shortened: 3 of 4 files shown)


thorstenhater:
bors try

thorstenhater:
bors try

boeschf:
should this new functionality be documented somewhere?

thorstenhater:
No! If we did, someone could use it!

No, of course, I just forgot.

boeschf (approved):
looks good, certainly useful to have this exposed. For future: may be worth to investigate if we can add some sane default GPU assignment if unspecified.

thorstenhater:
Re future: I assume @bcumming sensed a disturbance in the force right now. We could follow the ideas we implemented for the threading interface in Python and have 'none' (default), 'any' and 'private' tags to be passed to the context builder, in addition to ordinals. Even 'any' might be of disputable usefulness.

PR #1950, opened 2022-08-16
Make `mpoint`s hashable.

thorstenhater:
The GUI would like to have lookups from `arb::mpoint` to something, so please make it happen ;)


AdhocMan (approved):
lgtm

PR #1954, opened 2022-08-19
Remove abuse of arg_v.

### LINKED ISSUE #1754: Pybind seems to interpret some `decor.paint` docstrs as default values
```
arbor.decor().paint()
```

```
TypeError: paint(): incompatible function arguments. The following argument types are supported:
    1. (self: arbor._arbor.decor, region: str, mechanism: arbor._arbor.density) -> None
    2. (self: arbor._arbor.decor, region: str = 'the region label or description.', Vm: Optional[float] = initial membrane voltage [mV]., cm: Optional[float] = membrane capacitance [F/m²]., rL: Optional[float] = axial resistivity [Ω·cm]., tempK: Optional[float] = temperature [Kelvin].) -> None
    3. (self: arbor._arbor.decor, region: str, *, ion_name: str, int_con: Optional[float] = Initial internal concentration [mM], ext_con: Optional[float] = Initial external concentration [mM], rev_pot: Optional[float] = Initial reversal potential [mV]) -> None
```

Some of them look intended, as they don't have quotes around them, but `region: str = 'the region label or description.'` feels extra wrong. I find these hints, even if they'd be intended pretty cluttering. The user will pretty much always already have most of the signature right, and has given a wrong type, is forgetting one, giving one too many. The arg names and types are much more important than the docstr and the docstr clutters the print.

thorstenhater:
Do not abuse `pybind::arg_v("name", T default, "description")` as `pybind::arg_v("name", "description")`.
Use descriptive names and `pybind::arg` instead.

Closes #1754

### DIFF
--- a/python/cells.cpp
+++ b/python/cells.cpp
@@ -450,8 +450,8 @@ void register_cells(pybind11::module& m) {
                 s.scale(std::move(name), arborio::parse_iexpr_expression(ex).unwrap());
                 return s;
             },
-            pybind11::arg_v("name", "name of parameter to scale."),
-            pybind11::arg_v("ex", "iexpr for scaling"),
+            pybind11::arg("name"),
+            pybind11::arg("ex"),
             "Add a scaling expression to a parameter.")
         .def("__repr__",
             [](const arb::scaled_mechanism<arb::density>& d) {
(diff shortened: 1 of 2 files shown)


python/cells.cpp
```diff
@@ -774,13 +774,15 @@ void register_cells(pybind11::module& m) {
             "The group of spike detectors has the label 'label', used for forming connections between cells.")
         .def("discretization",
             [](arb::decor& dec, const arb::cv_policy& p) { return dec.set_default(p); },
-            pybind11::arg_v("policy", "A cv_policy used to discretise the cell into compartments for simulation"))
+            pybind11::arg("policy"),
+             "A cv_policy used to discretise the cell into compartments for simulation")
```
Helveg, line 778: These look like a different pattern, is this OK?

Helveg (commented):
So, were all the argument descriptions dropped?

thorstenhater:
`arg_v` is for use with default values, which these do not have. Consequently I shuffled the parameter description
into the function docstring where applicatble.

Helveg (approved):

brenthuisman (commented):
Most of them just removed the docstring, right? Not that they were terribly useful.

thorstenhater:
Yes, were redundant I removed them.

brenthuisman (approved):

PR #1957, opened 2022-08-23
Allow editing morphologies

### LINKED ISSUE #1589: Ignoring subsets of morphology upon loading (from .swc/.asc etc)
Hi. Discussed briefly with @halfflat whether or not it is possible to ignore/delete certain subsets like axons of morphologies in .swc file format (id tag "2") upon reading them in Arbor, and apparently it is not (except if done manually on the files themselves). 

Background: Reconstructions from sources like neuromorpho.org often contain the axon which is good of course for structural analysis, but perhaps less so for simulations focusing on dendritic and somatic responses, as well as recurrent networks. Hence, the elaborate axon geometries are replaced/represented by stylised axons with a few elements (e.g., Mainen & Sejnowski, 1996, https://doi.org/10.1038/382363a0) or an axon stub (Hay et al., 2011, https://doi.org/10.1371/journal.pcbi.1002107). The latter approach is used extensively in the BBP microcircuit neuron models (https://bbp.epfl.ch/nmc-portal/welcome.html).

So basically what I'm looking for is something that would allow me to 
- load a .morphology file
- delete certain parts of the loaded geometry
- add new cable elements with geometry of my own choice that is attached to some site on the loaded morphology

### LINKED ISSUE #1777: Translate Morphology
Move a morphology/segment tree in space. 
We already have some support for this using `pw_lin` and `isometry`, it would be useful to have the option to modify the morphology itself.

thorstenhater:
Add methods over `segment_trees`
- join_at: splice two trees together at a given id
- split_at: split a tree into two tree at a given id
- equivalence: two trees represent the same morphology
- equality: two trees are identical.
- apply isometry

Allow loading raw segment trees in arborio _except_ NML.

Closes #1589 
Closes #1777


arbor/morph/segment_tree.cpp
```diff
@@ -12,6 +12,108 @@
 
 namespace arb {
 
+template<typename P>
+segment_tree fold(const segment_tree& tree,
+                  P predicate,
+                  std::pair<msize_t, msize_t> start,
+                  const segment_tree& acc={}) {
```
boeschf, line 19: I am a bit confused about the naming of the function arguments. If I understand correctly, `predicate` is actually a "sentinel" functor, i.e. the folding stops once it evaluates to true. Also I am not sure why the last argument is named `acc` - usually this would be called something like "initval". Anyways, would be useful to describe what this function is actually doing.

thorstenhater: Sure, I can add some more explanation to the function. It's an internal helper that appeared as a common pattern.

thorstenhater: Also, while it is kind of a fold, copy_if seems more to the point.

boeschf: Thanks for adding the documentation. I still think the description for the predicate is unclear, though. Otherwise I like it.

arbor/morph/segment_tree.cpp
```diff
@@ -12,6 +12,117 @@
...
+    if(a.size() != b.size()) return false;
+
+    // invert parent <*> child relation
+    std::map<arb::msize_t, std::vector<arb::msize_t>>
+        a_children_of,
+        b_children_of;
+    for (auto ix = 0; ix < a.size(); ++ix) {
+        a_children_of[a.parents()[ix]].push_back(ix);
+        b_children_of[b.parents()[ix]].push_back(ix);
+    }
```
boeschf, line 94: factor out in helper function?

arborio/include/arborio/swcio.hpp
```diff
@@ -136,5 +137,15 @@ ARB_ARBORIO_API arb::morphology load_swc_arbor(const swc_data& data);
...
+ARB_ARBORIO_API arb::segment_tree load_swc_neuron_raw(const swc_data& data);
+
+// Convert a valid, ordered sequence of SWC records into a morphology.
+//
+// Note that 'one-point soma' SWC files are explicitly not supported.
+//
+// The segments of the generated morphology  will be contiguous. There will be
+// one segment for each SWC record after the first: this record defines the tag
+// and distal point of the segment, while the proximal point is taken from the
+// parent record.
```
boeschf, line 149: which function(s) does this comment refer to?

boeschf (commented):
c++ docs are missing. Otherwise looks good.

boeschf (approved):
looks good!

arbor/morph/segment_tree.cpp
thorstenhater: Jupp.

boeschf (commented):
looks good, thanks!

boeschf (approved):

Issue #1242, opened 2020-11-16
Docs wishlist


brenthuisman:
This Issue keeps track of open but identified TODOs for the documentation. A docs wishlist! Please keep issue current (edit this text) so we keep track of things we know need to be corrected but are too small for a separate issue and/or cannot be handled right now.

- [x] Cell Mechanism concepts page: include listing of default mechanisms, their name/tag and description, enumeration of parameters and their role/meaning/explanation wherever not absolutely trivial.
    - e.g. `gnabar` == maximum possible sodium conductance
- [x] A concepts page that describes probes and samplers. (See #474)
- [x] C++ API documentation for recipe is out of date #1080
- [x] remove `em` from custom css, and change all *terms* to :term:`terms`
- [ ] C++ API for `mechanism_catalogue`
- [x] C++ API for `morphology`
- [x] C++ API for `segment_tree`
- [x] C++ API for `regions` and `locsets`
- [x] C++ `cable_cell` page split and reorganize
- [x] C++ `event_generator` docs
- [x] Python `event_generator` docs
- [x] Python API for `cable_cell`
- [x] Python API for `morphology`
- [x] Python API for `gap_junctions` (incomplete)
- [x] Python API for `iclamp`
- [x] Python API for `spike_detector`
- [x] Python API for `initial_ion_data`
- [x] Python API for `cable_global_properties`
- [x] Concepts/cable_cell/decorations/place: fill the gaps
- [x] Contribution: finish docs howto
- [x] Contribution: add example howto?
- [x] Contribution: add test howto
- [x] **Developer** docs, Arbor internals.

## Out of scope for this issue

* New tutorials, moved here: https://github.com/arbor-sim/arbor/issues/1369
* Use docstrings on user facing API: makes them guaranteed up-to-date, testable. #1259, #1240

## Edits
- @thorstenhater add `event_generator`

thorstenhater:
Hi @brenthuisman ,

could you update the progress here?

brenthuisman:
Updated.

PR #1962, opened 2022-08-26
Remove explicit generator

### LINKED ISSUE #1488: Refactor `explicit_generator`
Right now, it accepts a `pse_vector` so a single generator can connect to multiple targets on the cell.
I propose that we make it more similar to the other generators, which work on a single target at a time.

```
arb::explicit_generator(arb::pse_vector{{0, 1.0, weight_}, {0, 3.0, weight_}, {1, 2.9, weight_}});
```
would become:
```
{arb::explicit_generator(0, weight, {1.0, 3.0}), 
 arb::explicit_generator(1, weight, {2.9})};
```

Meaning that the explicit generator would have the following signature: 

```
event_generator explicit_generator(const cell_lid_type& target, float weight, std::vector<time_type> times);
```

thorstenhater:
- Remove the old, multi-target `event_generator` class in favour of `schedule_generator(tgt, weight, explicit_schedule)`
- Much simplification ensues, `event_generator` is no longer a type-erasing container, but just what 
   `schedule_generator` was before
- Make the label resolution in generators a bit more eager, no longer at simulation time, 
   but now during setup (bit give a wee bit of perf as well)

Closes #1488

### DIFF
--- a/arbor/include/arbor/event_generator.hpp
+++ b/arbor/include/arbor/event_generator.hpp
@@ -5,6 +5,7 @@
 #include <memory>
 #include <random>
 #include <type_traits>
+#include <optional>
 
 #include <arbor/assert.hpp>
 #include <arbor/common_types.hpp>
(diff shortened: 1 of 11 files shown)


arbor/include/arbor/event_generator.hpp
```diff
@@ -49,130 +50,38 @@ namespace arb {
...
-};
-
-// Convenience routines for making schedule_generator:
 
 // Generate events with a fixed target and weight according to
 // a provided time schedule.
 
-struct schedule_generator {
-    schedule_generator(cell_local_label_type target, float weight, schedule sched):
+struct event_generator {
```
brenthuisman, line 62: In the PR message, you wrote the opposite: `event_generator` was suppressed in favor of a simplified `schedule_generator`. Accidental swap or am I missing sth?

brenthuisman: Grepping thought the docs finds mentions of `event_generator` that might need updates in:
1. `./cpp/hardware.rst`
2. `./cpp/recipe.rst`

test/unit/test_recipe.cpp
```diff
@@ -201,20 +201,21 @@ TEST(recipe, event_generators) {
...
+        std::vector<arb::event_generator>
+            gens_0 = {arb::explicit_generator({"synapse0"}, 0.1, std::vector<arb::time_type>{1.0}),
+                      arb::explicit_generator({"synapse1"}, 0.1, std::vector<arb::time_type>{2.0})},
+            gens_1 = {arb::explicit_generator({"synapse0"}, 0.1, std::vector<arb::time_type>{1.0})};
 
         auto recipe_0 = custom_recipe({cell_0, cell_1}, {{}, {}}, {{}, {}},  {gens_0, gens_1});
         auto decomp_0 = partition_load_balance(recipe_0, context);
 
-        EXPECT_NO_THROW(simulation(recipe_0, context, decomp_0));
+        EXPECT_NO_THROW(simulation(recipe_0, context, decomp_0).run(1, 0.1));
```
brenthuisman, line 213: Why does it need to run now?

brenthuisman (changes requested):
Simplification: I like it. Some questions:

arbor/include/arbor/event_generator.hpp
thorstenhater: It's been a small dance
- rm event_generator (because that's the point)
- mv schedule_generator event_generator (since it's now the only kind of generator)
- So, now we still have event_generator, but it's different.

good point about the doc.

thorstenhater: Update the docs are fine; just callbacks returning an `event_generator`.

test/unit/test_recipe.cpp
thorstenhater: Actually it always had to. See, the old `event_generator` did up-front target checking and threw an error if any
of them (there could be multiple in one, remember) did not exist. The schedule based one **did not** and did 
defer that to runtime **per event**. 

So, an `event_generator` would have been fine if construction passed, but `schedule_generator` only if simulation
runs. This PR makes the check on schedule eager, removes the need to store the resolver and the runtime checking.
Thus, it's simpler, faster, and safer now.

TL;DR: This requirement for `run` is just paranoia. ;)

thorstenhater:
See above: it's simpler, faster, and safer.

thorstenhater:
Yes, these sections are still empty. Is that a requirement for this PR, though? It does not add anything in this regard nor
change any of the related methods.

I added this to #1242 to remind ourselves.

brenthuisman (approved):
Agree that adding related documentation would be ideal; no need to spin that out to separate PRs for that.

Issue #1757, opened 2021-11-12
Scheduler API additions


Helveg:
**Describe the feature you need**
I would like some extra Python features for schedulers:

* Default "all events" `.events()` equal to `.events(0, float("inf"))`
* join operator (pref `|`, `+` might be confused with a translation of existing spikes)

If there are any internal advantages of using `regular_schedule` over an `explicit_schedule` with the same spikes times, it would be nice if these advantages could be maintained.

**Explain what it is supposed to enable**
Fire-and-forget composition of input signals on spike source cells from external formats like SONATA, tools or user code.

**Additional context**
I'm writing an interface between arbor and my framework's network format, and I have definitions of `devices` that might impinge multiple schedules on the same spike source cell or event generator. It can be solved with full Python code like this, but the requested additions obviate the need for this logic:

```
    def get_schedule(self, gid):
        schedule = arbor.explicit_schedule([])
        for device in self._devices_on[gid]:
            pattern = device.get_pattern(gid)
            merged = pattern + schedule.events(0, float("inf"))
            schedule = arbor.explicit_schedule(merged)
        return schedule
```

thorstenhater:
Hi @Helveg,

let's go over your suggestions 

1. Sure, that's easy enough.
2. A join operator `<>` on schedules would yield the time-ordered union, right? This is done internally when you define multiple source (via `generators_on` or `spike_source` connections), so there's no need for this in Arbor and would just duplicated internally functionality.
3. Sure, but why? We are allowing at most R/O access.
4. I am not sure what this is supposed mean.

Helveg:
3. That's for the fire-and-forget part, so that I can read out what the final schedule looks like
4. That they can be serialized to something and sent across a wire in a distributed environment. It's already relatively simple to extract the state and initialize a new one at the other side of the wire, but it would be nicer if it's supported from your side.

thorstenhater:
2. No, it's deep within the communications code where these events are interleaved with the incoming spikes. It's nothing we can make accessible.
3. Sure, I got that before, but why?
4. You could just serialise a type tag and the associated parameters? We do not support sending this over the wire, since Arbor's model of distributed computation is the `recipe` we currently have no plans on changing that.

thorstenhater:
Just a heads up, but #1962 might make your life a bit easier here.

thorstenhater:
Stepping back a bit, your request for multiple interleaved schedules on a source cell
sounds much more reasonable.

[...]

PR #1963, opened 2022-08-26
Allow multiple schedules per source_cell.

thorstenhater:
Partially addresses #1757, `spike_source_cell` can have multiple interleaved schedules now.


test/unit/test_spike_source.cpp
```diff
@@ -115,3 +115,66 @@ TEST(spike_source, exhaust)
     test_seq(regular_schedule(0, 1, 5));
     test_seq(explicit_schedule({0.3, 2.3, 4.7}));
 }
+
+TEST(spike_source, multiple)
+{
+    // This test assumes that seq will exhaust itself before t=10 ms.
+    auto test_seq = [](std::vector<schedule> seq) {
```
boeschf, line 122: you could probably simplify and combine the two tests by having a variadic lambda using universal references (you could then also test for r-value references in the same test)
```
auto test_seq = [](auto&&... seqs) {
    // ...
    spike_source_cell("src", static_cast<decltype(seqs)&&>(seqs)...);
    // ...
}
```

boeschf (changes requested):

arbor/include/arbor/spike_source_cell.hpp
```diff
@@ -11,10 +11,12 @@ namespace arb {
 
 struct ARB_SYMBOL_VISIBLE spike_source_cell {
     cell_tag_type source; // Label of source.
-    schedule seq;
+    std::vector<schedule> seqs;
 
     spike_source_cell() = delete;
-    spike_source_cell(cell_tag_type source, schedule seq): source(std::move(source)), seq(std::move(seq)) {};
+    template<typename... Seqs>
+    spike_source_cell(cell_tag_type source, Seqs... seqs): source(std::move(source)), seqs{seqs...} {};
```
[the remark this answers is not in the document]
thorstenhater: As written `seqs` is passed by value, thus copied into the argument list and then moved into the class
(actually copied again, but let's trust in the compiler for a sec).

Since we are in C++17 this copy is elided, see various notes on the topic, here's one I
could summon quickly https://andreasfertig.blog/2022/02/why-you-should-use-stdmove-only-rarely/.

Quoting the relevant bit

> Due to the mandatory copy elision in C++17, even with <code>-fno-elide-constructors</code>, the compiler must now elide this nonsense construction. <em>However</em>, if we apply <code>std::move</code> to the temporary copy elision doesn't apply anymore, and we're back in seeing a move construction. 

Am I missing something?

test/unit/test_spike_source.cpp
thorstenhater: Good point, the split was more based on my internal organisation of matters than anything else!

thorstenhater: On second thought though
1. constructing the `expected` value inside the lambda is kind of awkward now
   ```cxx
           std::vector<schedule> schedules{seqs...};
        for (auto& schedule: schedules) {
            auto ts = as_vector(schedule.events(0, 10));
            actual.insert(actual.end(),
                          ts.begin(), ts.end());
        }
        ```
        It works though since once we invoke the copy ctr and the other time it's the init list, but
        I find that kind of confusing to read.
2. We now have to explicitly cast to a std::vector<...> when calling `test_seq`
3. `static_cast<decltype(seqs)&&>` this is not needed.

test/unit/test_spike_source.cpp
```diff
@@ -115,3 +115,41 @@ TEST(spike_source, exhaust)
...
+
+        // epoch ending at 10ms
+        epoch ep(0, 0., 10.);
+        group.advance(ep, 1, {});
+
+        auto expected = spike_times(group.spikes());
+        std::sort(expected.begin(), expected.end());
+
+        auto actual = std::vector<time_type>{};
+        std::vector<schedule> schedules{seqs...};
```
boeschf, line 135: this should probably in the beginning of the lambda, such that it copies `seqs` before a potential move. It could be correct, if the (potential) move does not actually move anything and leaves the `seqs` in a good state. Haven't checked though.

thorstenhater: In the way of things it's fine, we keep one more copy alive then we should, but it's really not relevant, especially not in a test. I'd even argue it would be fine anywhere else.

boeschf: The advantage of perfect forwarding in the lambda is that you can stick in either l-value or r-value references and, thus, test both variants. If you feel this is unnecessary, we can leave it as is. I would prefer testing both variants by adding another call to the lambda:
```
auto req_sched = regular_schedule(0, 1, 5)
test_seq(req_sched, explicit_schedule({0.3, 2.3, 4.7}));
```

test/unit/test_spike_source.cpp
```diff
@@ -115,3 +115,41 @@ TEST(spike_source, exhaust)
     test_seq(regular_schedule(0, 1, 5));
     test_seq(explicit_schedule({0.3, 2.3, 4.7}));
 }
+
+TEST(spike_source, multiple)
+{
+    // This test assumes that seq will exhaust itself before t=10 ms.
+    auto test_seq = [](auto&&... seqs) {
+        ss_recipe rec(1u, spike_source_cell("src", seqs...));
```
boeschf, line 123:
```suggestion
        std::vector<schedule> schedules{seqs...};
        ss_recipe rec(1u, spike_source_cell("src", static_cast<decltype(seqs)&&>(seqs)...));
```

test/unit/test_spike_source.cpp
```diff
@@ -115,3 +115,41 @@ TEST(spike_source, exhaust)
...
+
+        // epoch ending at 10ms
+        epoch ep(0, 0., 10.);
+        group.advance(ep, 1, {});
+
+        auto expected = spike_times(group.spikes());
+        std::sort(expected.begin(), expected.end());
+
+        auto actual = std::vector<time_type>{};
+        std::vector<schedule> schedules{ seqs...};
```
boeschf, line 135:
```suggestion
```

test/unit/test_spike_source.cpp
```diff
@@ -115,3 +115,41 @@ TEST(spike_source, exhaust)
...
+    };
+
+    auto seqs = std::vector<schedule>{regular_schedule(0, 1, 5),
+                                      explicit_schedule({0.3, 2.3, 4.7})};
+    test_seq(seqs);
+    test_seq(std::vector<schedule>{regular_schedule(0, 1, 5),
+                                   explicit_schedule({0.3, 2.3, 4.7})});
+    test_seq(regular_schedule(0, 1, 5),
+             explicit_schedule({0.3, 2.3, 4.7}));
+}
```
boeschf, line 155:
```suggestion
    regular_schedule reg_sched(0, 1, 5);
    test_seq(reg_sched, explicit_schedule({0.3, 2.3, 4.7}));
}
```

boeschf (commented):
added the suggestions for clarifying my comment

test/unit/test_spike_source.cpp
thorstenhater: I am not opposed to adding this, but what are we exactly testing? The removal of the copy ctor? We cannot observe
whether that occurs or not, since we do not instrument the ctors.

boeschf: We would be using the move constructor, instead. But you are right, there is not much to test there. If you pass by value to the `spike_source_cell` constructor, you can also remove one of the tests where you pass an l-value ref of a vector.

thorstenhater: I like the perfect forwarding constructor and the extra test doesn't hurt.

boeschf (approved):

PR #1965, opened 2022-08-29
Add spike counts to pre-commit tests.

thorstenhater:
- Enhance `run_cpp_examples.sh` to do some very basic checking of spikes counts
- Add a script to check all releases
- The former must succeed for all commits to master

Here's the expected spike counts as of 0.7
- brunel 6998
- bench 972
- ring 94 (19 pre-v0.7)
- gap_junctions 30

### DIFF
--- a/doc/contrib/release.rst
+++ b/doc/contrib/release.rst
@@ -75,6 +75,8 @@ Release
    
 #. Change ``VERSION``. Make sure does not end with ``-rc`` or ``-dev``.
 
+#. Update ``scripts/check-all-tags.sh`` to check the current tag.
+
 #. Tag
 
    - commit and open a PR for the above changes.
--- a/scripts/check-all-tags.sh
+++ b/scripts/check-all-tags.sh
@@ -62,7 +62,12 @@ do
         echo " * SIMD=$simd"
         check brunel 6998
         check bench 972
-        check ring 19
+        if [[ "$tag" < "v0.7" ]]
+        then
+            check ring 19
+        else
+            check ring 94
+        fi
         check gap_junctions 30
     done
 done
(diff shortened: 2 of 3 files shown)


scripts/check-all-tags.sh
```diff
@@ -0,0 +1,74 @@
...
+    if [ "$expected" == "$actual" ]
+    then
+        echo "   - $prog: OK"
+    else
+        echo "   - $prog: ERROR wrong number of spikes: $expected ./. $actual"
+        ok=1
+    fi
+}
+
+for tag in "v0.4-79855b66" "v0.5.2-51e35898" "v0.6-930c23eb" "v0.7-d0e424b4"
```
brenthuisman, line 57: This is not just the tag but a tag-hash tuple. Is that necessary? Loop above seems to do well w/o.

doc/contrib/release.rst
```diff
@@ -75,6 +75,8 @@ Release
    
 #. Change ``VERSION``. Make sure does not end with ``-rc`` or ``-dev``.
 
+#. Update ``scripts/check-all-tags.sh`` to check the current tag.
```
brenthuisman, line 78: Automated testing will have caught any changes by now, so all that's left is to add nextver to the loops, right?

brenthuisman (changes requested):
Noice.

doc/contrib/release.rst
thorstenhater: Yes, just add `tag-hash`. The check-all script is for our manual regression testing, it's not used in an automated workflow.

scripts/check-all-tags.sh
thorstenhater: The tags are just for human convenience and/or double checking.

brenthuisman (approved):

PR #1967, opened 2022-09-05
🦑  Excise `time`

### LINKED ISSUE #1252: Proposal: remove time parameter from mechanisms.
Time was added as a visible value to NMODL mechanisms in order to incorporate time-dependent current sources. But apart from the impact of #1251, there are a couple of reasons to seek an alternative approach:

1. The integration can't see any discontinuities in the input current source, which prevents us from adjusting integration steps to match any such discontinuities.
2. It's insufficient for representing arbitrary time-dependent input currents, without also providing facilities for loading into the mechanism arbitrarily large representations of these current values.

I'd prefer that we extend the existing stimulus interface to encompass this use case:
* Provide stimulus options that correspond to common experimental stimulus protocols.
* Provide a stimulus current class that can represent e.g. arbitrary piecewise-linear current curves.

thorstenhater:
Excises `time` from NMODL as per our policies not to let users peek behind the curtain.

Closes #1252

### DIFF
--- a/arbor/backends/event.hpp
+++ b/arbor/backends/event.hpp
@@ -13,7 +13,7 @@ namespace arb {
 struct target_handle {
     cell_local_size_type mech_id;    // mechanism type identifier (per cell group).
     cell_local_size_type mech_index; // instance of the mechanism
-    cell_size_type intdom_index;     // which integration domain (acts as index into e.g. vec_t)
+    cell_size_type intdom_index;     // which integration domain (acts as index into arrays)
 
     target_handle() = default;
     target_handle(cell_local_size_type mech_id, cell_local_size_type mech_index, cell_size_type intdom_index):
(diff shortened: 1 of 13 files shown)


thorstenhater:
bors try

boeschf:
This will potentially break some code where users rely on this feature. What's the our policy with respect to deprecation warnings/breaking changes?

thorstenhater:
That we are in `0.x.y` land and will politely inform our users ;) `Time` has been in NMODL mainly for implementing I_Clamps, which we do differently now. None of our internal catalogues uses it.

In general we try to minimise our ABI surface and try to keep our users away from the soft underbelly of Arbor.
Using `time` someone could try to implement their own solvers/integrator, which we decidedly do not like.
If desparately needed, adding a trivial ODE (`t' = 1`) restores it, but makes our dislike obvious.
This also saves an array in shared state, (plus one on the GPU, if present).

boeschf (approved):

Issue #1951, opened 2022-08-17
error: 'make_value_iterator' is not a member of 'pybind11'


espenhgn:
**Describe the bug**

Hi! 
While building Arbor on the JSC JUSUF HPC resource, I encounter an issue with an undefined function in `pybind11`: 
```bash
...
[100%] Building CXX object python/CMakeFiles/pyarb_obj.dir/cells.cpp.o
[100%] Linking CXX executable ../bin/lmorpho
[100%] Built target lmorpho
/p/home/jusers/hagen2/jusuf/sources/arbor/python/cells.cpp: In lambda function:
/p/home/jusers/hagen2/jusuf/sources/arbor/python/cells.cpp:290:35: error: 'make_value_iterator' is not a member of 'pybind11'; did you mean 'make_key_iterator'?
  290 |                  return pybind11::make_value_iterator(ld.cache.begin(), ld.cache.end());
      |                                   ^~~~~~~~~~~~~~~~~~~
      |                                   make_key_iterator
make[2]: *** [python/CMakeFiles/pyarb_obj.dir/build.make:104: python/CMakeFiles/pyarb_obj.dir/cells.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:2017: python/CMakeFiles/pyarb_obj.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

To me, the issue appears to be that `pybind11-v2.8.1` in the git submodule should be used (CMake finds it), but the system provided `pybind11-v2.7.1` is still used during compilation. Seems like the required version was bumped via https://github.com/arbor-sim/arbor/pull/1742. 

I could get it to compile by replacing `make_value_iterator` by `make_key_iterator` in `cells.cpp`, but I don't know what sort of problems that may have introduced. 

Installation script:
```bash
module load Python/3.6.9 SciPy-Stack GCC/11.2.0 ParaStationMPI/5.5.0-1 mpi4py/3.1.3 h5py/3.5.0 GSL/2.7 git/2.33.1-nodocs CMake/3.21.1 Doxygen/1.9.1 Autotools/20210726 Boost/1.78.0 PyYAML/5.4.1

cd arbor
git checkout master
git pull
git submodule update --init --recursive

mkdir build
cd build
rm -rf *

export CC=`which gcc`
export CXX=`which g++`
[...]

PR #1968, opened 2022-09-05
Move pybind11 to `/ext`

brenthuisman:
- Ticks off a box in #1731.
- May address an error described in #1951


thorstenhater (approved):
👍

espenhgn:
Thanks! I tested now this PR on the cluster, however, it does not fix #1951 in case an older pybind11 version is available.

PR #1975, opened 2022-09-09
generate-catalogue is gone

thorstenhater:
This is prep for meson in addition to simplifying (ie reducing) the amount of scripting
we have. In the long run, I'd like to remove `BuildModules.cmake` too, but this might
be superseded by meson and this PR is self-contained.

# Changes
- `modcc` accepts now a list of NMODL files
- likewise, `modcc` is now able to spit out a catalogue.cpp file
- remove the `generate-catalogue` script, simplify a-b-c
- remove some options from `modcc` we never use
- No longer allow external modcc

# TODO
- [x] Re-enable raw CXX files when building catalogues.


brenthuisman:
TODO looks done, correct?

thorstenhater:
✅

brenthuisman (changes requested):
I think `doc/dev/extending_catalogues.rst` can use an update. Looks like it never mentioned this script (only in a comment).

thorstenhater:
Done, and the overall 'API' is much simpler _again_

brenthuisman (approved):

PR #1977, opened 2022-09-13
load_asc Python fix 

lukasgd:
The `load_asc` in Python failed as the C++ function it expects a filename, not a string like `contents` added in #1937.

### DIFF
--- a/arborio/neurolucida.cpp
+++ b/arborio/neurolucida.cpp
@@ -610,7 +610,7 @@ parse_hopefully<sub_tree> parse_sub_tree(asc::lexer& L) {
 
 
 // Perform the parsing of the input as a string.
-asc_morphology parse_asc_string(const char* input) {
+ARB_ARBORIO_API arb::segment_tree parse_asc_string_raw(const char* input) {
     asc::lexer lexer(input);
 
     std::vector<sub_tree> sub_trees;
@@ -768,6 +768,14 @@ asc_morphology parse_asc_string(const char* input) {
         }
     }
 
+    return stree;
+}
+
+
+ARB_ARBORIO_API asc_morphology parse_asc_string(const char* input) {
+    // Parse segment tree
+    arb::segment_tree stree = parse_asc_string_raw(input);
+
     // Construct the morphology.
     arb::morphology morphology(stree);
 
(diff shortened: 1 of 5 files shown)


thorstenhater (changes requested):
Hey, thanks for spotting & fixing this. It's actually quite worrying that this is never exercised in a test.

Please do add a test to ensure this never comes up again; you can simply reuse the code from the SWC loaders.

lukasgd:
94b060b should address this,
- Python tests for SWC and ASC
- raw loader added for ASC for consistency (following up on #1957)
- fixed some compiler warnings

arborio/neurolucida.cpp
```diff
@@ -781,7 +789,8 @@ asc_morphology parse_asc_string(const char* input) {
     return {stree, std::move(morphology), std::move(labels)};
 }
 
-ARB_ARBORIO_API asc_morphology load_asc(std::string filename) {
+
+inline std::string read_file(std::string filename) {
```
thorstenhater, line 793: Hey, I have the lingering feeling that this method is probably repeated all over the code base. 
If do not feel it's too much to pile on here, you could hunt these places down and move this here to something like `util::read_file_to_string`. Otherwise just leave a note, and we'll take care of it later.

python/morphology.cpp
```diff
@@ -356,18 +356,25 @@ void register_morphology(py::module& m) {
...
             }
             catch (std::exception& e) {
                 // Try to produce helpful error messages for SWC parsing errors.
                 throw pyarb_error(util::pprintf("error loading neurolucida asc file: {}", e.what()));
             }
         },
-        "filename"_a, "Load a morphology and meta data from a Neurolucida ASCII .asc file.");
+        "filename_or_stream"_a,
+        pybind11::arg_v("raw", false, "Return a segment tree instead of a fully formed morphology"),
+        "Load a morphology and meta data from a Neurolucida ASCII .asc file.");
```
thorstenhater, line 377:
```suggestion
        "Load a morphology or segment_tree and meta data from a Neurolucida ASCII .asc file.");
```

thorstenhater (changes requested):

python/morphology.cpp
lukasgd: Committed - sorry, saw too late that I could just approve it from the website here.

arborio/neurolucida.cpp
thorstenhater: A quick scan of the codebase turned up very few instances. Ok!

thorstenhater (approved):

PR #1979, opened 2022-09-15
Python/concurrency defaults

brenthuisman:
- [x] Rename Python's `create_context` to `make_context_shim`, to align with `arb::make_context`
- [x] A special no-arg `make_context_shim` initializes a context with no MPI, and a `arbenv::default_allocation()`. This is a `proc_allocation` with `arbenv::thread_concurrency()` threads and a `arbenv::default_gpu()`. This no-arg constructor is not exposed and used only by the no-arg `arbor.context()` constructor and the single-arg (=recipe) `arbor.simulation()` constructor.
  - This removes any effort for default (== local) use, but keeps proper defaults in the case users request MPI.
- [x] Binds a few more helper functions from `arbenv` in `arbor.env`, e.g. `arbor.env.thread_concurrency()` binds `arbenv::thread_concurrency()`
- [x] Update tests and tutorial texts.
- ~~[x] Align Pybind minver in `python/CMakeLists.txt` with elsewhere.~~
- [x] Docs updated/markup corrected to reflect changes in #2001
  - [x] `doc/cpp/hardware.rst`
  - [x] `doc/concept/interconnectivity.rst`
- [x] Update `proc_allocation` printer for #2094
- [x] Expand arbor.config() with timestamp and python_lib_path.


thorstenhater:
Hi @brenthuisman,

thanks that might save a few people a question on the help board. However, the tests run into a timeout.
I'll restart the job for now, but if it persists, please check what's going on.

brenthuisman:
I restarted the job a few times now, it's the Python examples running on the Pip source install that does not terminate. Logs are not very helpful (we need to print script filenames to see which is actually not terminating)! I'll try to take a look today or tomorrow.

thorstenhater:
Interesting ... can you figure out _where_ exactly it spins?

thorstenhater:
Ok, this appears exactly in this constellation
- Python
- threads > 1
- calling `sim.update` after `sim.__init__`

and results in an uninterruptible hang. I kneejerk blame the threading implementation, but
it's a weird one.

thorstenhater:
Ok, found the perpetrator. See #1984.

thorstenhater:
This should be unblocked now.

boeschf:
seems to hang in python examples agian

brenthuisman:
Runs OK now locally!

doc/python/hardware.rst
```diff
@@ -152,9 +152,8 @@ The Python wrapper provides an API for:
 
         .. attribute:: threads
 
-            The number of threads available locally for execution. Must be set to 1 at minimum. 1 by default.
-            Passing ``"avail_threads"`` (as string) will query and use the maximum number of threads the system makes available.
-
+            The number of threads available locally for execution. Defaults to the maximum number of threads the system makes available if gpu_id and mpi are not set, else defaults to 1.
```
thorstenhater, line 155: A quick word/investigation into whether this is honouring the CGroups/ExecMask settings would be nice, but not a blocker.

doc/tutorial/network_ring.rst
```diff
@@ -117,13 +117,10 @@ and :py:class:`arbor.domain_decomposition`. The first lets Arbor know what hardw
...
-this flag, a context is constructed that will use all locally available threads. On your local machine this will match the
-number of logical cores in your system. Especially with large numbers
-of cells you will notice the speed-up. (You could instantiate the recipe with 5000 cells and observe the difference. Don't
-forget to turn of plotting if you do; it will take more time to generate the image then to run the actual simulation!)
-If no domain decomposition is specified, a default one distributing work equally is created. This is sufficient for now.
-You can print the objects to see what defaults they produce on your system.
+Step **(12)** creates a simulation object from the recipe. Optionally, the :py:class:`~arbor.simulation` constructor takes two more
+parameters: a :class:`arbor.context` and a :class:`arbor.domain_decomposition`. In a followup of this tutorial that will be demonstrated.
+For now, it is enough to know that for simulations that don't require customized execution those arguments can be left out. Arbor
+will without any further arguments use all locally available threads.
```
thorstenhater, line 123:
```suggestion
Without further arguments Arbor will use all locally available threads.
```
Grammar ;)
also, due to the suggestion quirks, remove the hanging 'Arbor' a line up.

doc/tutorial/network_ring_mpi.rst
```diff
@@ -28,14 +28,14 @@ The hardware context
...
-context can be customized by the user. We may specify the number of threads in the thread pool; determine the
-id of the GPU to be used; or create our own MPI communicator.
-
-The configuration of the context will need to be changed to reflect the change in hardware.
-First of all, we scrap setting `threads="avail_threads"` and instead use 
-`MPI <https://en.wikipedia.org/wiki/Message_Passing_Interface#Overview>`_ to distribute the work over nodes, cores and threads.
+and the MPI communicator for distributed simulations. In some Arbor tutorials, the :class:`arbor.single_cell_model` or
+:class:`arbor.simulation` objects created the execution context :class:`arbor.context` behind the scenes. The details
+of the execution context can be customized by the user. We may specify the number of threads in the thread pool;
+determine the id of the GPU to be used; or create our own MPI communicator.
```
thorstenhater, line 34:
```suggestion
determine the id of the GPU to be used; or pass our own MPI communicator.
```
You might also hint at the arbenv get_private_gpu facilities.

python/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ if(ARB_USE_BUNDLED_PYBIND11)
   # instead of using find_package.
   add_subdirectory(${pb11_src_dir} pybind11)
 else()
-  find_package(pybind11 2.8.1 REQUIRED)
+  find_package(pybind11 2.10.0 REQUIRED)
```
thorstenhater, line 19: Sneaking version bump?
We agreed to separate these out into their own PRs.

python/context.cpp
```diff
@@ -125,27 +125,15 @@ void register_contexts(pybind11::module& m) {
...
-                    return create_context(arbenv::thread_concurrency(), gpu, mpi);
-                }
-                throw pyarb_error(
-                        util::pprintf("Attempt to set threads to {}. The only valid thread options are a positive integer greater than 0, or 'avial_threads'.", threads));
-
-            }),
-            "threads"_a, "gpu_id"_a=pybind11::none(), "mpi"_a=pybind11::none(),
-            "Construct a distributed context with arguments:\n"
-            "  threads: A string option describing the number of threads. Currently, only \"avail_threads\" is supported.\n"
+            "  threads: The number of threads available locally for execution. Must be set to 1 at minimum. Defaults to the maximum number of threads the system makes available if gpu_id and mpi are not set, else defaults to 1.\n"
```
thorstenhater, line 136:
```suggestion
            "  threads: The number of threads to be used. Must be set to 1 at minimum. Defaults to the maximum number of threads the system makes available if gpu_id and mpi are not set, else defaults to 1.\n"
```
This argument is prescriptive rather than descriptive. Arbor will not decide to use less threads than the number you pass here.
Also, technically, it's not only execution, but also construction.

python/context.cpp
```diff
@@ -67,7 +67,7 @@ struct proc_allocation_shim {
     }
 };
 
-context_shim create_context(unsigned threads, pybind11::object gpu, pybind11::object mpi) {
+context_shim make_context(unsigned threads, pybind11::object gpu, pybind11::object mpi) {
```
thorstenhater, line 70: Nice!

python/simulation.cpp
```diff
@@ -210,7 +212,7 @@ void register_simulation(pybind11::module& m, pyarb_global_ptr global_ptr) {
                               const std::optional<arb::domain_decomposition>& decomp,
                               std::uint64_t seed) {
                 try {
-                    auto ctx = ctx_ ? ctx_ : std::make_shared<context_shim>(arb::make_context());
+                    auto ctx = ctx_ ? ctx_ : std::make_shared<context_shim>(arb::make_context({arbenv::thread_concurrency(),-1}));
```
thorstenhater, line 215: So, under the idea that Python and C++ APIs should co-evolve I am against this. Keep it at the default here and change the default in C++. But still keep the option to ask for 'avail_threads' explicitly.

python/example/network_ring_gpu.py
```diff
@@ -123,7 +123,7 @@ def global_properties(self, kind):
 # (11) Set up the hardware context
 # gpu_id set to None will not use a GPU.
 # gpu_id=0 instructs Arbor to the first GPU present in your system
-context = arbor.context(threads="avail_threads", gpu_id=None)
+context = arbor.context(1, gpu_id=None)
```
thorstenhater, line 126: Why exactly do we opt to use one thread here?

doc/python/hardware.rst
thorstenhater, line 155: Considering the rest of the change set, how about digging in a bit deeper and come up with sensible defaults for _all_ cases? It's not a requirement to merge, but if we are going to put it in eventually, why not in 0.8?
I don't know whether arbenv has the proper calls, but
- MPI: figure out how many threads a task is allowed to use (using hwloc maybe) and set that when handed 'avail_threads'. Other than that there's OS-specific calls (which hwloc uses), we would need to consider MacOS and Linux.
- GPU: can just use 'avail_threads', it speeds up construction a bit. Independent of the change above. If 'avail_threads' receives super-charged capabilities wrt MPI, it'll benefit, if not, then not.
- MPI+GPU: do whatever MPI does. Likely we are using one task per GPU.

thorstenhater (changes requested):
Nice! A few changes I think we should have before merging
- Please pull out the pybind version bump.
- Make C++ behave like we'd like Python to 
- Have 'avail_threads' even in presence of GPU

If you like you can also make this even more  fancy by make it
play nice with MPI, but that's not neccessary to merge.

python/CMakeLists.txt
brenthuisman: A PR was merged in my absence with an outdated version number here. Pybind11 was bumped over a month ago to 2.10.0

brenthuisman:
C++ was left out, since I assumed (some people in the past certainly argued so) that there should be no magic. I'm OK with keeping them aligned, just want to make sure this has changed.

thorstenhater:
For me alignment beats the no_magic paradigm, so if there must be magic, it should be working the same always.
One might argue that under this requirement it is not magic anymore.

python/CMakeLists.txt
thorstenhater: Ah!

doc/python/hardware.rst
brenthuisman: It does (as per arbenv::thread_concurrency()).

thorstenhater: I know, but it would be nice to have it written in the docs.

python/example/network_ring_gpu.py
brenthuisman: Since an integer number must be provided when the nondefault constructor is used (`"avail_threads"` has been removed), and 1 is as good as any.

thorstenhater: Oh dear, I missed the removal. Please keep it in. Reasoning: We shouldn't be doing something the user cannot ask for.

doc/python/hardware.rst
brenthuisman: OK, it can be added.

brenthuisman: My initial idea was to only touch the default constructor, to provide the best experience for novice users.

If you'd like to touch more than the default constructor: we could initialize `threads` to `thread_concurrency`. It's straightforward to understand for the user, but in particular when MPI is used, may result in surprising behavior.

doc/tutorial/network_ring_mpi.rst
brenthuisman: will do!

python/example/network_ring_gpu.py
brenthuisman: Before I do that: it requires a combinatorial explosion of constructors, especially if we want to be able to set it in any combination with the other options (maybe we should switch to kwargs only?) due to the different type. I considered `py::any()`, but that didn't make it any more brief.

python/simulation.cpp
brenthuisman: So, also in cpp make `avail_threads` an option you mean?

python/example/network_ring_gpu.py
thorstenhater: Those who dabble in dynamically typed languages are damned to type `type_of(...)` into all eternity.
kw_args is fine in Python, I think (I can never remember the ordering anyhow). C++ should be ok with
taking an integral type for the thread count (which can be defaulted to `arbenv::thread_concurrency`)
and the GPU id.

doc/python/hardware.rst
brenthuisman: MPI config is it's own world of expertise, and I wager normal users stick to 1 thread per rank (I have never seen or done differently, unless it was you trying to squeeze out every last drop).

So, my pref is then to default `threads` to `thread_concurrency` unless `mpi` is set, in which case it stays `1`. If cleverer people than I come up with a thought out proposal for MPI then I'm happy to consider it.

python/simulation.cpp
thorstenhater: Well, currently it is, right, via `thread_concurrency`?

brenthuisman: Ah, sure, but it isn't a default (yet). I thought that was your request?

I was thinking that the best place to set this default (once) is in `proc_allocation`, but no, because it does not know about MPI. So, the default `context` constructors should set `num_threads` to `arbenv::thread_concurrency()`, if MPI is not enabled.

In the interest of keeping things similar, should in Python the variable be called `thread_concurrency`, and be a member of `arbor`? e.g. `arbor.context(arbor.thread_concurrency(), None)`? I like it!

thorstenhater: Should be in `arbenv`, yes, but that's the vision.

doc/python/hardware.rst
thorstenhater: As I said, I'd be fine with merging this part as is, but consider the extra effort for making it extra-nice for the user interesting. I think the common patterns are
1. No MPI, 1 thread/(logical)core 
  - with GPU 
  - without GPU
3. MPI, still 1 thread per (logical)core but now each task gets handed `1/n`th. N is the number of tasks according to
  - 1 task per socket
  - 1 task GPU

So, the approach is to compute 
- the (logical)core count: `arbenv::cores(logical=true|false)`
- the number of tasks: handed to you via `Comm_size`, defaults to 1

and `context` or `thread_concurrency` could be instructed whether to distribute threads over 
tasks or not.

python/simulation.cpp
brenthuisman: Unlike in Python, there is no explicit constructor in C++, it all goes through `proc_allocation`: `execution_context(const proc_allocation& resources = proc_allocation{})`. Also, having this in C++ would make (all of) Arbor dependent on `arbenv`.

Since there is no way to currently do sth like `auto ctx =  arbor::context{gpu_id=1}`, we could set `proc_allocation.num_threads` by changing the non-mpi constructor to `execution_context(const proc_allocation& resources = proc_allocation{arbenv::thread_concurrency(),-1});` To the extent that `execution_context` construction maps to Python, it sets `num_threads` under the same circumstances.

thorstenhater: > Unlike in Python, there is no explicit constructor in C++, it all goes through proc_allocation: execution_context(const proc_allocation& resources = proc_allocation{}). 

You could call proc_allocation a set of kwargs. Once we have achieved C++20 in its full glory this becomes 
a thing
```cxx
execution_context({.threads=4, .comm=MPI_COMM_WORLD})
```

> Also, having this in C++ would make (all of) Arbor dependent on arbenv.

That's fine, to a certain extent, `arbenv` is a library made for Arbor to interact with the environment.
As long we do not pull it into the deepest layers.

Also, nothing stopping us from making new, purpose-built c'tors, eg
```cxx
context from_mpi_one_thread_per_logical_core(MPI_Comm& comm, int gpu_id=-1);
context from_mpi_one_thread_per_physical_core(MPI_Comm& comm, int gpu_id=-1);
// ...
```

Also, if proc_allocation is really too much of a hassle to have around, feel free to remove it.

brenthuisman: I've updated the code as per my earlier suggestion in the interested of merging this now.

- Removal of proc_allocation is fine by me, but I'm not a user. The C++ tests do, so that's why I have not suggested removal.
- MPI cleverness: to me that can be seen as a separate issues (the PR was initially written to get a better default for local use). I'll ask in the channel if users are missing something.

doc/python/hardware.rst
brenthuisman: See above's note on MPI.

thorstenhater: As I said, it'd been nice to have rather than critical. I am personally fine with typing out these, but as we were sanding off some rougher edges it offered an easy win.

brenthuisman:
Hmm, I see a CI build error pop up that I didn't see locally:

```
CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
  "arbor" of type SHARED_LIBRARY
    depends on "arborenv" (weak)
  "arborenv" of type SHARED_LIBRARY
    depends on "arbor" (weak)
At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.
```
So, how do we handle this for the shared variant? Do we need that variant for anything (I don't recall noticing it before)?
The `Python build test` does not show the error.

doc/python/hardware.rst
brenthuisman: I'd want to test this out, but for this week I don't think I'll have the time. Making the release will almost certainly eat up the rest of my hours, and I prefer not to postpone it further. Maybe a small discussion in the team meet about removing proc_alloc and have these new constructors?

thorstenhater: As said, optional, and only if `proc_alloc` blocks you from achieving more.

brenthuisman:
Updated, changed. PR message reflects the major changes.

doc/concepts/interconnectivity.rst
thorstenhater, line 1: This is largely unrelated formatting changes? Fine, but confusing to review.

test/unit/test_spikes.cpp
```diff
@@ -214,6 +214,7 @@ TEST(SPIKES_TEST_CLASS, threshold_watcher_interpolation) {
     dict.set("mid", arb::ls::on_branches(0.5));
 
     arb::proc_allocation resources;
+    resources.num_threads = 1;
```
thorstenhater, line 217: Why one?

test/unit/test_domain_decomposition.cpp
```diff
@@ -152,6 +152,7 @@ namespace {
 TEST(domain_decomposition, homogenous_population)
 {
     proc_allocation resources;
+    resources.num_threads = 1;
```
thorstenhater, line 155: Same as below.

test/unit/test_domain_decomposition.cpp
```diff
@@ -219,6 +220,7 @@ TEST(domain_decomposition, homogenous_population)
 TEST(domain_decomposition, heterogenous_population)
 {
     proc_allocation resources;
+    resources.num_threads = 1;
```
thorstenhater, line 223: Same as below.

python/example/network_ring.py
```diff
@@ -121,9 +121,8 @@ def global_properties(self, kind):
 ncells = 4
 recipe = ring_recipe(ncells)
 
-# (12) Create an execution context using all locally available threads and simulation
-ctx = arbor.context(threads="avail_threads")
-sim = arbor.simulation(recipe, ctx)
+# (12) Create a simulation using all locally available threads
```
thorstenhater, line 124: Yeah, it's short, yet less clear what's going on. 
```suggestion
# (12) Create a simulation using the default settings:
# - Use all threads available
# - Use round-robin distribution of cells across groups with one cell per group
# - No GPU
# - No MPI
# - Empty label dictionary
# Other constructors of simulation can be used to change all of these.
```

python/conversion.hpp
```diff
@@ -66,6 +66,17 @@ std::optional<T> py2optional(pybind11::object o, const char* msg) {
     return o.is_none()? std::nullopt: std::optional<T>(std::move(value));
 }
 
+template <typename T, typename F>
```
thorstenhater, line 69:
```cxx
template <typename T, typename F>
std::optional<T> optional_when(T&& o, F&& pred) {
    if (pred(std::forward(T)) {
      return {std::forward(T)};
    } else {
      return {};
    }
}
```
- A slightly more descriptive name
- Do not construct the optional if not needed
- Play nicely with rvalue refs, lvalue ref, and values.

python/context.cpp
```diff
@@ -136,8 +124,10 @@ void register_contexts(pybind11::module& m) {
              "bind_procs"_a=false,
              "bind_threads"_a=false,
              "Construct an allocation with arguments:\n"
-             "  threads: The number of threads available locally for execution. Must be set to 1 at minimum. 1 by default.\n"
-             "  gpu_id:  The identifier of the GPU to use, None by default.\n")
+             "  threads:      The number of threads available locally for execution. Must be set to 1 at minimum. 1 by default.\n"
```
thorstenhater, line 127: Might this be a symbolic constant or string? Such as 'avail_threads'?

python/context.cpp
```diff
@@ -33,94 +34,81 @@ std::ostream& operator<<(std::ostream& o, const context_shim& ctx) {
...
+void proc_allocation_shim::set_gpu_id(pybind11::object gpu) {
+    auto gpu_id = py2optional<int>(gpu, "gpu_id must be None, or a non-negative integer", is_nonneg());
+    proc_allocation.gpu_id = gpu_id.value_or(-1);
+};
 
-    // helper function to use arb::make_context(arb::proc_allocation)
-    arb::proc_allocation allocation() const {
-        return arb::proc_allocation(num_threads, gpu_id.value_or(-1), bind_procs, bind_threads);
+void proc_allocation_shim::set_num_threads(unsigned threads) {
+    if (0==threads) {
```
thorstenhater, line 59: Don't we throw this from C++?

python/CMakeLists.txt
```diff
@@ -43,11 +43,12 @@ set(pyarb_source
     env.cpp
 )
 
-set_property(SOURCE config.cpp PROPERTY COMPILE_DEFINITIONS ARB_BINARY="${CMAKE_INSTALL_BINDIR}"       APPEND)
-set_property(SOURCE config.cpp PROPERTY COMPILE_DEFINITIONS ARB_LIB="${CMAKE_INSTALL_LIBDIR}"          APPEND)
-set_property(SOURCE config.cpp PROPERTY COMPILE_DEFINITIONS ARB_DATA="${CMAKE_INSTALL_DATAROOTDIR}"    APPEND)
-set_property(SOURCE config.cpp PROPERTY COMPILE_DEFINITIONS ARB_CXX_COMPILER="${CMAKE_CXX_COMPILER}"   APPEND)
-set_property(SOURCE config.cpp PROPERTY COMPILE_DEFINITIONS ARB_PREFIX="${CMAKE_INSTALL_PREFIX}"       APPEND)
+set_property(SOURCE config.cpp PROPERTY COMPILE_DEFINITIONS ARB_BINARY="${CMAKE_INSTALL_BINDIR}"         APPEND)
```
thorstenhater, line 46: What changed here?

arbor/execution_context.hpp
```diff
@@ -25,7 +25,7 @@ struct ARB_ARBOR_API execution_context {
     task_system_handle thread_pool;
     gpu_context_handle gpu;
 
-    execution_context(const proc_allocation& resources = proc_allocation{});
+    execution_context(const proc_allocation& resources = proc_allocation{1,-1});
```
thorstenhater, line 28: I'd like a symbolic constant here, ie adding
```cxx
constexpr static int gpu_nil_id = -1;
```
to `context.hpp`

thorstenhater (changes requested):
Shaping up to a nice addition, some comments.

doc/concepts/interconnectivity.rst
brenthuisman: True-ish, I had to touch `hardware.rst` and discovered some things neglected/incorrect in https://github.com/arbor-sim/arbor/pull/2001.

test/unit/test_spikes.cpp
brenthuisman: Unless testing the defaults, I like tests be explicit.

python/example/network_ring.py
brenthuisman: Good point. Slightly modified in upcoming commit. Label dict left out, it's set elsewhere.

python/context.cpp
brenthuisman: `arb:::proc_allocation` does not to any checking. Nor does anything else for that matter. Only the pyobject ever did.

python/context.cpp
brenthuisman: Available as `arbor.env.default_concurrency()`.

python/CMakeLists.txt
brenthuisman: Forgot to list in PR message (added): `arbor.config()` now returns `timestamp` and `python_lib_path`. The prefix need not be (and for pip installs, isn't) where CMake installed the package.

python/example/network_ring.py
```diff
@@ -121,7 +121,12 @@ def global_properties(self, kind):
 ncells = 4
 recipe = ring_recipe(ncells)
 
-# (12) Create a simulation using all locally available threads
+# (12) Create a simulation using the default settings:
+# - Use all threads available
+# - Use round-robin distribution of cells across groups with one cell per group
+# - Use GPU if present
```
thorstenhater, line 127: Do we? Boy we are smart ;)

thorstenhater (approved):
One weirdness ... Other than that good.

PR #1984, opened 2022-09-22
Fix connectivity update concurrency bug.

thorstenhater:
1. Fix Python bindings for `recipe::update`
   - *drop* the GIL before handing off to C++
   - tighten exception safety
2. Run plasticity examples with threads; both C++ and Python.
   - C++: Guard against I/O interleaving.
   - Py: Drop spikes from source, prettify reporting.
   - C++: use decor chaining.
3. Modernise PYBIND11_OVERLOAD -> *RIDE (advised since 2.6) in simulation.
4. No longer initialise connectivity twice.


thorstenhater:
The PR accidentally ate another one.

PR #1985, opened 2022-09-22
Clean up plasticity

thorstenhater:
1. Fix Python bindings for `recipe::update`
   - *drop* the GIL before handing off to C++
   - tighten exception safety
2. Run plasticity examples with threads; both C++ and Python.
   - C++: Guard against I/O interleaving.
   - Py: Drop spikes from source, prettify reporting.
   - C++: use decor chaining.
3. Modernise PYBIND11_OVERLOAD -> *RIDE (advised since 2.6) in simulation.
4. No longer do we initialise connectivity twice.
   - Simplify communicator construction.
   - Fix unit tests that needed to two-phase init communicator.

### DIFF
--- a/example/plasticity/plasticity.cpp
+++ b/example/plasticity/plasticity.cpp
@@ -62,7 +62,7 @@ struct recipe: public arb::recipe {
             .place(center, arb::synapse("expsyn"), syn)
             .place(center, arb::threshold_detector{-10.0}, det)
             .set_default(arb::cv_policy_every_segment());
-        return arb::cable_cell({tree}, decor);
+        return arb::cable_cell({tree}, {}, decor);
     }
 };
 
--- a/example/plasticity/plasticity.cpp
+++ b/example/plasticity/plasticity.cpp
@@ -57,29 +57,38 @@ struct recipe: public arb::recipe {
         if (gid == 0) return arb::spike_source_cell{src, arb::regular_schedule(f_spike)};
         // all others are receiving cable cells; single CV w/ HH
         arb::segment_tree tree; tree.append(arb::mnpos, {-r_soma, 0, 0, r_soma}, {r_soma, 0, 0, r_soma}, 1);
-        auto decor = arb::decor{};
-        decor.paint(all, arb::density("hh", {{"gl", 5}}));
-        decor.place(center, arb::synapse("expsyn"), syn);
-        decor.place(center, arb::threshold_detector{-10.0}, det);
-        decor.set_default(arb::cv_policy_every_segment());
-        return arb::cable_cell({tree}, {}, decor);
+        auto decor = arb::decor{}
+            .paint(all, arb::density("hh", {{"gl", 5}}))
+            .place(center, arb::synapse("expsyn"), syn)
+            .place(center, arb::threshold_detector{-10.0}, det)
+            .set_default(arb::cv_policy_every_segment());
+        return arb::cable_cell({tree}, decor);
     }
 };
+
+// For demonstration: Avoid interleaving std::cout in multi-threaded scenarios.
+// NEVER do this in HPC!!!
+std::mutex mtx;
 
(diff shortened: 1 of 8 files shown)


example/plasticity/plasticity.cpp
```diff
@@ -57,29 +62,39 @@ struct recipe: public arb::recipe {
...
+            .paint(all, arb::density("hh", {{"gl", 5}}))
+            .place(center, arb::synapse("expsyn"), syn)
+            .place(center, arb::threshold_detector{-10.0}, det)
+            .set_default(arb::cv_policy_every_segment());
         return arb::cable_cell({tree}, {}, decor);
     }
 };
 
+// For demonstration: Avoid interleaving std::cout in multi-threaded scenarios.
+// NEVER do this in HPC!!!
```
boeschf, line 75: :)

example/plasticity/plasticity.cpp
```diff
@@ -57,29 +62,39 @@ struct recipe: public arb::recipe {
...
 void sampler(arb::probe_metadata pm, std::size_t n, const arb::sample_record* samples) {
     auto* loc = arb::util::any_cast<const arb::mlocation*>(pm.meta);
-    std::cout << std::fixed << std::setprecision(4);
+
     for (std::size_t i = 0; i<n; ++i) {
+        std::lock_guard<std::mutex> lock{mtx};
         auto* value = arb::util::any_cast<const double*>(samples[i].data);
-        std::cout << "|  " << samples[i].time << " |      " << loc->pos << " | " << *value << " |\n";
+        std::cout << std::fixed << std::setprecision(4)
+                  << "|  " << samples[i].time << " |      " << loc->pos << " | " << *value << " |\n";
```
boeschf, line 85: would be good practice to reset the `cout` state to what it was before, after this line (I know it is not important here, though).

test/unit-distributed/test_communicator.cpp
```diff
@@ -531,8 +531,8 @@ TEST(communicator, ring)
     auto global_sources = g_context->distributed->gather_cell_labels_and_gids(local_sources);
 
     // construct the communicator
-    auto C = communicator(R, D, label_resolution_map(global_sources), label_resolution_map(local_targets), *g_context);
-
+    auto C = communicator(R, D, *g_context);
+    C.update_connections(R, D, label_resolution_map(global_sources), label_resolution_map(local_targets));
```
boeschf, line 535: this looks like a common pattern - would it make sense to add an overload à la
```
void communicator::update_connections(
    const connectivity& rec,
    const domain_decomposition& dom_dec,
    const cell_labels_and_gids&  local_sources);
```

example/plasticity/plasticity.cpp
thorstenhater: It sadly brings lots of hassle, but I'll add something ;)

test/unit-distributed/test_communicator.cpp
thorstenhater: Actually, why not allow automatic conversion of sources to label map?

boeschf: yes, that too. but also default conversion from local to global?

example/plasticity/plasticity.cpp
boeschf: not strictly needed for this PR

example/plasticity/plasticity.cpp
thorstenhater: I thought a fair warning was needed ;)

test/unit-distributed/test_communicator.cpp
thorstenhater: Hmm, upon surveying the situation, it's only ever used in the tests, and there only thrice.

boeschf (approved):

PR #1989, opened 2022-09-27
Heaviside step

llandsmeer:
This PR adds the Heaviside step function to Arbor's iexpr, needed for some inhomogeneous density mechanism like the ones in [this model](https://raw.githubusercontent.com/OpenSourceBrain/L5bPyrCellHayEtAl2011/master/neuroConstruct/generatedNeuroML2/L5PC.cell.nml). Arguable, one could also use segment groups but this makes automated translation of an existing model using NeuroML's `H()` much much easier.

The implementation returns

```python
def step(x):
    if x > 0:
        return 1
    else:
        return 0
```

which makes sense to me, but does not have to be the semantics that we follow.

For example the [NeuroML/LEMS implementation](https://github.com/LEMS/jLEMS/blob/554803b3ebfc444cfc7b044fe71d69fc933b6b13/src/main/java/org/lemsml/jlems/core/expression/FunctionNode.java#L83) uses the half-maximum convention

```java
ret = 0.5*(Math.signum(arg)+1.);
```

which has `H(0.0) == 0.5`.

### DIFF
--- a/arbor/iexpr.cpp
+++ b/arbor/iexpr.cpp
@@ -285,7 +285,7 @@ struct step: public iexpr_interface {
     step(iexpr_ptr v): value(std::move(v)) {}
 
     double eval(const mprovider& p, const mcable& c) const override {
-        return value->eval(p, c) >= 0.0 ? 1.0 : 0.0;
+        return value->eval(p, c) > 0.0 ? 1.0 : 0.0;
     }
 
     iexpr_ptr value;
--- a/arbor/iexpr.cpp
+++ b/arbor/iexpr.cpp
@@ -596,6 +596,10 @@ std::ostream& operator<<(std::ostream& o, const iexpr& e) {
         o << "exp " << std::get<0>(std::any_cast<const std::tuple<iexpr>&>(e.args()));
         break;
     }
+    case iexpr_type::step: {
+        o << "step " << std::get<0>(std::any_cast<const std::tuple<iexpr>&>(e.args()));
+        break;
+    }
     case iexpr_type::log: {
         o << "log " << std::get<0>(std::any_cast<const std::tuple<iexpr>&>(e.args()));
         break;
--- a/arbor/iexpr.cpp
+++ b/arbor/iexpr.cpp
@@ -281,6 +281,16 @@ struct exp: public iexpr_interface {
     iexpr_ptr value;
 };
 
+struct step: public iexpr_interface {
+    step(iexpr_ptr v): value(std::move(v)) {}
+
+    double eval(const mprovider& p, const mcable& c) const override {
+        return value->eval(p, c) >= 0.0 ? 1.0 : 0.0;
+    }
+
+    iexpr_ptr value;
+};
+
 struct log: public iexpr_interface {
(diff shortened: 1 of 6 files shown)


thorstenhater (changes requested):
Nice work, thanks for making NML2 a bit easier. A question on the design, though:
If this is specifically for the support of NML2 semantics, why not implement it the
same way (or add an alternative say `hstep`, which has that (admittedly weird) behaviour)?

llandsmeer:
Because I wasn't sure if the Math.signum implementation was 'the standard' or just the easiest way to get a step function in java. I couldn't find the official definition. The python LEMS implementation also has the half-maximum convention explicitly spelled out though, which makes it more official I think?

https://github.com/LEMS/pylems/blob/1180c4fa58a8e2c1b23a284fd7088c3b2e398f00/lems/sim/build.py#L773

```python
        elif func == "H":

            def heaviside_step(x):
                if x < 0:
                    return 0
                elif x > 0:
                    return 1
                else:
                    return 0.5

            return "heaviside_step"
```

The LEMS paper also mentions

> The LEMS language was developed in parallel to the initial reference implementation jLEMS (source code at https://github.com/LEMS/jLEMS), and the serialization of LEMS in XML followed closely the internal classes used in jLEMS for specifying the model. Documentation of all of the elements in LEMS can be found here: http://lems.github.io/LEMS/elements.html. This is automatically generated from documentation in the jLEMS source code.

Which I guess makes the Java version the official standard..

I'll change it to the half-maxmimum convention

thorstenhater:
The problem is that IEEE745 float/doubles are _signed_, so we actually have ±0. Java might be different. Let's just define
`hw-step` or similar to delineate from a proper heaviside.

arbor/iexpr.cpp
```diff
@@ -281,6 +281,23 @@ struct exp: public iexpr_interface {
...
+
+    double eval(const mprovider& p, const mcable& c) const override {
+        double x = value->eval(p, c);
+        if (x > 0) {
+            return 1.0;
+        } else if (x < 0) {
+            return 0.0;
+        } else {
+            return 0.5;
+        }
```
boeschf, line 295: replacing branches with operations, very obscure
```suggestion
        // x <  0:  0
        // x == 0:  0.5
        // x >  0:  1
        return 0.5*((0. < x) - (x < 0.) + 1);
```

llandsmeer: This is also more in line with the 'official' NeuroML java (sign) implementation as opposed to the Python (if statements)

llandsmeer:
> The problem is that IEEE745 float/doubles are _signed_, so we actually have ±0. Java might be different. Let's just define `hw-step` or similar to delineate from a proper heaviside.

How would half-maximum heaviside deviate here from binary heaviside w.r.t. `-0`?

Something like this:

```
hm_step(-0) = 0.5
hm_step(+0) = 0.5

bin_step(-0) = 0
bin_step(+0) = 1
```

or

```
hm_step(-0) = 0
hm_step(+0) = 0.5

bin_step(-0) = 1
bin_step(+0) = 1
```

?

Not that I hope that any production code every relies on this behaviour :)

PR #1991, opened 2022-09-28
:atom: Meson

thorstenhater:
Trial use of meson to build Arbor. There are some known problems/missing feature, but you can try 
this preview today. Depending on the feedback, the mesonic lodge™ will polish this branch and make
it the default. 

⚠️ Please be quick, since this will bitrot quite fast otherwise.

# Quick start
```
# in arbor
pip install meson ninja
mkdir mbuild
cd mbuild
meson -Darb_with_bundled_libs=true -Darb_with_python=true
# runs all tests
ninja test
```

# Missing/Todo
- [ ] port CI
- [ ] port pip install
- [ ] a-b-c uses meson


brenthuisman:
Excellent!

But. Python is going to make merging this difficult. The Meson support status of the various components in the Python build are:

- [Scikit-build](https://github.com/scikit-build/scikit-build/issues/432): wont implement.
- [Meson Python build backend (which could be used in Pyproject.toml)](https://github.com/mesonbuild/meson/issues/7863): not implemented yet.
- [Third party Python build backend](https://github.com/FFY00/meson-python): we'd depend on a hobby project it looks like, and have to accept [a small explosion in dependencies](https://github.com/mesonbuild/meson/issues/7863#issuecomment-1178564193).

The latter is the only option as far as I can establish, but I think that may simply be too premature.

brenthuisman:
Small update as of today: `meson-python` seems to work, but will not be merged into meson proper in current form: https://github.com/mesonbuild/meson/issues/7863#issuecomment-1634336523

thorstenhater:
This has basically rotted to the point of being better rewritten from scratch.

PR #1992, opened 2022-09-30
Optimize CPU-side solvers

thorstenhater:
# Description
* remove storage of RHS and face-conductance from cable solver
* remove RHS from diffusion solver
* elide copy RHS->U and RHS->Xd respectively
* solve will now directly mangle its inputs U / Xd
* assembly is now part of solve
* fix tests accordingly
* add `__restrict__` in the assembly part to encourage auto-vectorization

# Analysis
* these tricks do _not_ work for GPU, as we pack memory there
* further manual vectorisation found no speed-up
  - likely this means the explicit vectorisation in shared state is likewise redundant; might be subject to removal soon
  - experiments done on AVX2
* we save quite a bit of memory
  - cable_solver: `rhs + face_conductance = 2*#CV doubles`
  - diffusion_solver: `rhs = #CV doubles` per diffusive species
* and some copies: one for the cable solver and one per diffusive species
  - these are redundant, since we only ever call into `solver::solve<T>(T& to)`

### DIFF
--- a/arbor/backends/multicore/cable_solver.hpp
+++ b/arbor/backends/multicore/cable_solver.hpp
@@ -19,19 +19,13 @@ struct cable_solver {
 
     iarray parent_index;
     iarray cell_cv_divs;
-
-    array d;     // [μS]
-    array u;     // [μS]
-    array rhs;   // [nA]
-
-    array cv_capacitance;      // [pF]
-    array face_conductance;    // [μS]
-    array cv_area;             // [μm^2]
-
     iarray cell_to_intdom;
 
-    // the invariant part of the matrix diagonal
-    array invariant_d;         // [μS]
+    array d;              // [μS]
+    array u;              // [μS]
+    array cv_capacitance; // [pF]
+    array cv_area;        // [μm^2]
+    array invariant_d;    // [μS] invariant part of matrix diagonal
 
     cable_solver() = default;
     cable_solver(const cable_solver&) = default;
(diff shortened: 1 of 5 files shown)


arbor/backends/multicore/cable_solver.hpp
```diff
@@ -74,71 +66,82 @@ struct cable_solver {
...
     }
 
-    const_view solution() const {
-        // In this back end the solution is a simple view of the rhs, which
-        // contains the solution after the matrix_solve is performed.
-        return rhs;
-    }
+    // Setup and solve the cable equation
+    // * expects the voltage from its first argument
+    // * will likewise overwrite the first argument with the soluction
```
boeschf, line 71:
```suggestion
    // * will likewise overwrite the first argument with the solution
```

arbor/backends/multicore/cable_solver.hpp
```diff
@@ -74,71 +66,82 @@ struct cable_solver {
...
+    template<typename T>
+    void solve(T& rhs, const_view dt_intdom, const_view current, const_view conductivity) {
+        value_type * const __restrict__ d_ = d.data();
```
boeschf, line 74: `__restrict__` is not a c++ keyword, it is implemented by gcc and clang, though. For visual c++ it would be `__declspec(restrict)`. Maybe we should put the `restrict` keyword behind an arbor macro à la `ARB_RESTRICT`?

arbor/backends/multicore/cable_solver.hpp
```diff
@@ -74,71 +66,82 @@ struct cable_solver {
...
-    const_view solution() const {
-        // In this back end the solution is a simple view of the rhs, which
-        // contains the solution after the matrix_solve is performed.
-        return rhs;
-    }
+    // Setup and solve the cable equation
+    // * expects the voltage from its first argument
+    // * will likewise overwrite the first argument with the soluction
+    template<typename T>
+    void solve(T& rhs, const_view dt_intdom, const_view current, const_view conductivity) {
```
boeschf, line 73: Do you know what the reasoning was behind the original split into assemble/solve?

thorstenhater: It's much more useful on the GPU side, where the memory is shuffled into place to accomodate the 
kernel's thread layout. So, I presume it was done for homogeneity.

arbor/backends/multicore/cable_solver.hpp
thorstenhater: NVCC and XLC support it as well; at least. We do not target MSVC. 
That said, it's a good idea regardless. It also plan to employ it in the ABI
soon.

boeschf (approved):

PR #1994, opened 2022-10-04
missing std namespace

boeschf:
tiny fix: use std variants of `sqrt` and `move` (removes warnings, and is safer)


thorstenhater (approved):
👍

PR #1996, opened 2022-10-04
Do not restrict SWC record identifier

schmitts:
All integers do occur in the wild according to:

http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html

Standardized swc files (www.neuromorpho.org) -
0 - undefined
1 - soma
2 - axon
3 - (basal) dendrite
4 - apical dendrite
5+ - custom

A lot of data does not conform exactly to this standard however e.g.

CNIC data -
0 - undefined
1 - soma
2 - axon
3 - (basal) dendrite
4 - apical dendrite
5 - fork point
6 - end point
7 - custom

VNED data -  seems to be standard, but uses
10 - related to soma ?

Gulyas data - each number represents structure with same diameter.

Other data has been observed with
-1 -  also possibly related to soma ?

### DIFF
--- a/arborio/swcio.cpp
+++ b/arborio/swcio.cpp
@@ -46,7 +46,7 @@ swc_mismatched_tags::swc_mismatched_tags(int record_id):
 {}
 
 swc_unsupported_tag::swc_unsupported_tag(int record_id):
-    swc_error("Only SWC record identifiers of 1, 2, 3 or 4 are supported.", record_id)
+    swc_error("Unsupported SWC record identifier.", record_id)
 {}
 
 // Record I/O:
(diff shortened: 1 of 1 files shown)


thorstenhater (changes requested):
Since Arbor does not imbue these tags with special meaning I see no reason to restrict them.
Before I approve, have you checked the docs/tests are ok?

schmitts:
"All checks have passed" and I couldn't find any explicit mention of allowed SWC identifier in the Arbor docs. Maybe @brenthuisman knows a potential spot?

thorstenhater:
👍 

Could you add the table you quoted (and its source!) to the docs at
https://docs.arbor-sim.org/en/latest/fileformat/swc.html

brenthuisman:
`fileformat/swc.rst` is indeed the place where current conventions are documented. An update to it would be appreciated!

schmitts:
@brenthuisman Yes, I'll update the PR and add some links and docs.

schmitts:
Going through https://docs.arbor-sim.org/en/latest/fileformat/swc.html I don't see a need to add anything.

thorstenhater (approved):

PR #2001, opened 2022-10-17
Feature: External Connectivity

### LINKED ISSUE #1913: Coupling to NEST
This is one stepping stone towards a CoSim workflow as described by HBP:SGA3 T5.5.

# Minimalist Steering
- Receive KILL command
  -  hard kill: UNIX signal 
  -  soft kill: command via, will stop after current call to `simulation::run` finishes
- Send heartbeat 
  - 1st implementation via stdio
  -  can leverage #1870 
- Output minimal delay
  - via stdio
  - maybe JSON

# Coupling
- Open a connection to another spiking simulator (or its proxies)
  -  `MPI_Comm_connect` (https://www.mpich.org/static/docs/v3.2/www3/MPI_Comm_connect.html)
  - Metadata for connect must be transferred via stdio or similar measures
- Receive remote spikes in our native format over said intercommunicator
  - _likely_ Intercomm `Allgather` works, but these are weird, so do some reading
  - one or two rounds of exchange? So, Arbor -> Arbor and then Arbor -> NEST, or do all in one
- Allow 'remote' connections during setup.
  - Will need an extra tag and/or method on recipe

# Notes
- There's prior work by @apeyser in https://github.com/apeyser/arbor/tree/feature/nestio
  - Ideas look sane, but not mergeable anymore
- As said Intercomm collectives are weird, but here they do what we want?!
  - After the AllgatherV both Arbor and the coupled application have all slices of spikes
  - _However_ the semantics imply that we need two rounds anyhow, so maybe AllgatherV(intra) and Broadcast(inter) is better?
Screenshot 2022-06-22 at 09 40 17

thorstenhater:
# Introduction

- Define a protocol for interacting with Arbor through MPI from outside
- Add documentation accordingly
- Implement a Proof of Concept example

# TODO
- [x] Python bindings
- [x] Update Docs
- ~~[X] Optional: Add a NEST + Arbor example/showcase.~~ Handled externally by CoSim team

# ⚠️ Breaking Changes
- Due to the large number of arguments, context creation is now kwargs-only.
   This means `arb.context("avail_threads", 1, mpi)` now needs to be written as
    ```py
    arb.context(threads="avail_threads", gpu=1, mpi=mpi)
    ```
- Ditto for `proc_allocation` must now be
   ```py
    arb.proc_allocation(threads="avail_threads", gpu=1)
    ```
- The mixed form is now
    ```py
    arb.context(arb.proc_allocation(threads="avail_threads", gpu=1),  # positional
                        mpi=mpi) # kw-only.
    ```
- Note that this is likely the common usage anyhow, I needed to fix four calls in our examples and tests; out of dozens.
# Related Issues
Closes #1913


sontheimer:
Successful installation and test on the Cosim VM (Vagrant and Virtualbox):
- https://github.com/multiscale-cosim/ARBOR-NEST-usecase1

No errors and greenlight on the following commands:
- ```cmake /home/vagrant/multiscale-cosim/ARBOR-NEST-usecase1/arbor -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_C_COMPILER=$CC -DARB_USE_BUNDLED_LIBS=ON -DARB_WITH_PYTHON=ON -DPYTHON_EXECUTABLE=`which python3` -DARB_WITH_MPI=ON```
- `make install`
- `make -j 4 examples`
- `mpirun -n 2 ./bin/remote`
- `pip3 install /home/vagrant/multiscale-cosim/ARBOR-NEST-usecase1/arbor/`

doc/cpp/recipe.rst
```diff
@@ -80,6 +80,17 @@ Recipe
 
         By default returns an empty list.
 
+    .. cpp:function:: virtual std::vector<connection> external_connections_on(cell_gid_type gid) const
```
boeschf, line 83: shouldn't this return a `vector<ext_cell_connection>`?

doc/cpp/recipe.rst
```diff
@@ -80,6 +80,17 @@ Recipe
...
+
+        Returns a list of all the **incoming** connections for `gid` from a
+        remote simulation. Each connection ``con`` should have a valid synapse
+        label ``con.dest`` on the post-synaptic target `gid`, and a valid source
+        label ``con.source.lid`` on the pre-synaptic source
+        ``con.source.gid``. See :cpp:type:`connection`.
```
boeschf, line 89: same here? also, I couldn't find the place this is pointing to

doc/concepts/interconnectivity.rst
```diff
@@ -74,6 +74,142 @@ full recipe.
...
+                                    32),     # tag
+                                    "tgt",
+                                    weight,
+                                    delay)]]
+
+Note that Arbor now recognizes two sets of ``GID``: An external and an internal
+set. This allows both Arbor and the coupled simulation to keep their own
+numbering schemes. However, internally Arbor will tag external cells and spikes
+by setting their ``GID``s' most significant bit. This _halves_ the effecively
+available ``GID``s.
```
boeschf, line 114: not clear to me what this means in practice - if this imposes a restrictions on the number of cells/maximum gid (internal or external), then we probably should provide a factory to generate the labels which does a check on the bounds.

doc/dev/index.rst
```diff
@@ -17,6 +17,7 @@ Here we document internal components of Arbor. These pages can be useful if you'
    cable_cell
    cell_groups
    communication
+   external
```
boeschf, line 20: where is this file?

boeschf (commented):
some initial comments on documentation

doc/concepts/interconnectivity.rst
thorstenhater: Yes, a model building error will be thrown if such a connection is found.

doc/cpp/recipe.rst
thorstenhater: Should!

doc/cpp/recipe.rst
thorstenhater: My mistake `ext_cell_connection` is isomorphic to `connection` ;)

doc/dev/index.rst
thorstenhater: Currently, this is all covered in the _user_ docs, since I expect _external_ developers to come to look at this. Thus, I ended up not adding this.

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,182 @@
...
+        default: throw unexpected_message{};
+    }
+    return result;
+}
+
+// Exchange control message. NOTE: This makes clever (?) use of MPI_Allreduce
+// to achieve simulatenous duplex mode broadcast by preparing zeroed buffers
+// on all ranks except the root and then adding up tons of zeros with a single
+// payload.
+inline
```
boeschf, line 137: please explain why this is better than a regular broadcast

boeschf (commented):
some comments about remote comm

example/remote/remote.cpp
```diff
@@ -0,0 +1,148 @@
...
+        // * We never stop until the other group kills the process ;)
+        std::cerr << "[EXT]" << mpi.rank << " " << mpi.local_rank << '\n';
+        for (int ep = 0;; ++ep) {
+            std::cerr << "[EXT] Epoch " << ep << '\n';
+            arb::cell_gid_type src = mpi.local_rank;
+            auto time = ep*2.0*dt + mpi.local_rank*dt + 0.05;
+            std::vector<arb::spike> spikes(10, {{src, 0}, time});
+            std::cerr << "[EXT] Waiting for control message from Arbor\n";
+            auto msg = arb::remote::exchange_ctrl(mpi.inter, arb::remote::msg_epoch{});
+            if(!std::holds_alternative<arb::remote::msg_epoch>(msg)) break;
```
boeschf, line 91: can't you do something with `msg`? Like determine the spike times?

arbor/include/arbor/communication/remote.hpp
thorstenhater: It's two broadcasts in duplex, ie assume group 0 has data `[42, 0, 0]` and group `[23, 0]` (42 and 23 are stand-ins for our payloads while the 0s shall represent our zeroed buffers). After the `Allreduce(..., inter)` we'll get `[23, 23, 23]` and `[42, 42]` in groups 0 and 1. If we were using broadcasts, we'd need to take care with ordering, ie who goes first group=0:rank=0 or group=1:rank=0. Thus we'd have to offer two different functions, explain who uses what when, ... This solution is a bit trippy, but it's one call, no ordering :D

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+inline
+void mpi_checked(int rc, const std::string& where) {
+    if (rc != MPI_SUCCESS) throw mpi_error{where, rc};
+}
+
+// Prepare byte buffer for sending.
+inline
+std::vector<char> make_send_buffer(const ctrl_message& msg, std::vector<char>& send) {
+    std::size_t ptr = 0;
```
boeschf, line 82: check for size of `send` is missing

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,182 @@
...
+        memcpy(send.data() + ptr, &m, sizeof(m));
+    };
+    std::visit(visitor, msg);
+    return send;
+}
+
+// Consume receive buffer and generate a message
+inline
+ctrl_message make_ctrl_message(const std::vector<char>& recv) {
+    std::size_t ptr = 0;
```
boeschf, line 99: check for size of `recv` is missing

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,182 @@
...
+// Magic protocol tag
+constexpr std::uint8_t ARB_REMOTE_MAGIC = 0xAB;
+
+// Remote protocol version
+constexpr std::uint8_t ARB_REMOTE_VERSION_MAJOR = 0x00;
+constexpr std::uint8_t ARB_REMOTE_VERSION_MINOR = 0x01;
+constexpr std::uint8_t ARB_REMOTE_VERSION_PATCH = 0x00;
+
+// Message buffer length
+constexpr std::size_t ARB_REMOTE_MESSAGE_LENGTH = 1024;
```
boeschf, line 24: can't this be inferred from the variant type?

doc/concepts/interconnectivity.rst
```diff
@@ -74,6 +74,142 @@ full recipe.
...
+ scenarios where both sides are launched as a single job (eg via ``SLURM``), but
+ might do so where unrelated jobs are used.
+
+Tying It All Together
+---------------------
+
+We recommend to make use of the facilities offered in
+``arbor/communication/remote.hpp``, as does Arbor internally. Refer to the
+``remote.cpp`` example on how they are used.
+
```
boeschf, line 209: remote.hpp should be explained in more detail imo.

arbor/include/arbor/communication/remote.hpp
boeschf: ah I wasn't aware that the intercomm broadcast is *not* duplex, and therefore has asymmetric function arguments. I see, very clever. But more expensive. Could  we do a sendrecv root to root and then local broadcast instead?

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,182 @@
...
+};
+
+inline
+void mpi_checked(int rc, const std::string& where) {
+    if (rc != MPI_SUCCESS) throw mpi_error{where, rc};
+}
+
+// Prepare byte buffer for sending.
+inline
+std::vector<char> make_send_buffer(const ctrl_message& msg, std::vector<char>& send) {
```
boeschf, line 81: I think this function should not return anything (output is provided by l-value reference argument).

arbor/include/arbor/communication/remote.hpp
thorstenhater: Could be done, true. Idea here is to potentially interact with less sophisticated languages, think C or Python, that only have byte buffers.  Whomever also might elect not to use this header.
Having a fixed-size buffer with some headroom seemed sensible under this light.

arbor/include/arbor/communication/remote.hpp
thorstenhater: True. I tend to inline it ... it's a simpler helper and shouldn't be used on its own.

arbor/include/arbor/communication/remote.hpp
thorstenhater: See below.

arbor/include/arbor/communication/remote.hpp
thorstenhater: This is an internal helper only. We sadly cannot hide these, so I'll inline them.

arbor/include/arbor/communication/remote.hpp
thorstenhater: Yes, but if that's more efficient?! Allreduce usually is one of the core functions to optimise, it's part of the acceptance test for most major installations :D

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,182 @@
...
+    if ((mag != ARB_REMOTE_MAGIC) ||
+        (maj != ARB_REMOTE_VERSION_MAJOR) ||
+        (min != ARB_REMOTE_VERSION_MINOR) ||
+        (pat != ARB_REMOTE_VERSION_PATCH)) throw unexpected_message{};
+    std::uint8_t tag = recv[ptr++];
+    auto payload = recv.data() + ptr;
+    ctrl_message result = msg_null{};
+    switch(tag) {
+        case msg_null::tag: {
+            result = *reinterpret_cast<const msg_null*>(payload);
```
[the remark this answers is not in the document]
thorstenhater: Good point.

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,182 @@
...
+};
+
+struct unexpected_message: remote_error {
+    unexpected_message(): remote_error{"Arbor remote: Received unknown tag."} {}
+};
+
+struct mpi_error: remote_error {
+    mpi_error(const std::string& where, int rc):
+        remote_error{"MPI failed in " + where + " with error=" + std::to_string(rc)} {}
+};
```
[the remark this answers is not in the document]
thorstenhater: useful to have, yes.

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,219 @@
...
+#include <cstring>
+#include <cstdint>
+#include <memory>
+#include <exception>
+#include <variant>
+#include <vector>
+#include <string>
+
+#include <mpi.h>
+
```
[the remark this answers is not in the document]
thorstenhater: Good idea, it'll also point to documentation and licensing!

example/remote/remote.cpp
boeschf: what are `msg_epoch::t_start` and `msg_epoch::t_end` used for then?

thorstenhater: For now we expect both parts (=arbor + remote) to run on the same epoch. The example could hold itself to a higher standard, true.

arbor/communication/communicator.cpp
```diff
@@ -68,16 +79,25 @@ void communicator::update_connections(const connectivity& rec,
...
             auto gid = gids[i];
-            gid_infos[i] = gid_info(gid, i, rec.connections_on(gid));
+            gid_infos[i] = {gid, i, rec.connections_on(gid), rec.external_connections_on(gid)};
         });
-    cell_local_size_type n_cons =
-        util::sum_by(gid_infos, [](const gid_info& g){ return g.conns.size(); });
+
+    cell_local_size_type
+        n_cons = util::sum_by(gid_infos, [](const auto& g){ return g.conns.size(); }),
+        n_ext_cons = util::sum_by(gid_infos, [](const auto& g){ return g.ext_conns.size(); });
```
boeschf, line 100: what's the harm in repeating the type? Please don't abuse the comma operator too much.

thorstenhater: Anything typed is too much. Also, I never considered this an operator use, for me this is just like `f(a, b, c)`.
If this can be overloaded I understand your aversion.

arbor/communication/mpi_context.cpp
```diff
@@ -98,11 +104,64 @@ struct mpi_context_impl {
...
 };
 
 template <>
 std::shared_ptr<distributed_context> make_mpi_context(MPI_Comm comm, bool bind) {
     return std::make_shared<distributed_context>(mpi_context_impl(comm, bind));
 }
 
+struct remote_context_impl {
+    mpi_context_impl mpi_;
+    MPI_Comm portal_ = MPI_COMM_NULL;
```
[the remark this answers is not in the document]
thorstenhater: Personal preference, but not by much.

arbor/communication/communicator.cpp
```diff
@@ -31,6 +31,28 @@ communicator::communicator(const recipe& rec,
...
+    auto msb = 1 << (sizeof(cell_gid_type)*8 - 1);
+    // set the MSB
+    return {(c.rid | msb) ^ msb, c.index};
+}
+
+inline
+cell_member_type global_cell_of(const cell_member_type& c) {
+    auto msb = 1 << (sizeof(cell_gid_type)*8 - 1);
+    // set the MSB
+    return {(c.gid | msb) ^ msb, c.index};
```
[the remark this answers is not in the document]
thorstenhater: Yes. The comment should be 'clear', not 'set'

arbor/communication/communicator.hpp
```diff
@@ -44,7 +44,8 @@ class ARB_ARBOR_API communicator {
     ///
     /// Takes as input the list of local_spikes that were generated on the calling domain.
     /// Returns the full global set of vectors, along with meta data about their partition
-    gathered_vector<spike> exchange(std::vector<spike> local_spikes);
+    std::pair<gathered_vector<spike>,
+              std::vector<spike>> exchange(std::vector<spike> local_spikes);
```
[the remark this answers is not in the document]
thorstenhater: Probably I should make a type for this, now that this becoming serious.

doc/concepts/interconnectivity.rst
```diff
@@ -74,6 +74,163 @@ full recipe.
...
+We package this in the suplemental header ``arbor/communication/remote.hpp`` as
+``gather_spikes``. This function will accept a ``std::vector<arb_spike>`` where
+``arb_spike`` is a binary compatible version of Arbor's internal spike type that
+is to be sent from the local rank of the coupled packaged, eg ``b1`` from above.
+After the operation Arbor has received the concatenation of all such vectors and
+the routine will return the concatenation of all spikes produced and exported by
+Arbor on all ranks of the participating package.
+
+Please refer to our developer's documentation for more details the actual spike
+exchange process. Due to the way MPI define intercommunicators, the exchange is
```
boeschf, line 170:
```suggestion
exchange process. Due to the way MPI defines intercommunicators, the exchange is
```

doc/concepts/interconnectivity.rst
```diff
@@ -74,6 +74,163 @@ full recipe.
...
+
+Control Plane and Epochs
+------------------------
+
+Before initiating the actual simulation, Arbor sets the ``epoch`` length to half
+the minimal delay in the global network. The minimal delay can be queried using
+``simulation::min_delay``.
+
+Before the start of each ``epoch``, a control message must be exchanged between
+the root -- ie rank 0 -- process of the Arbor process and that of coupled
```
boeschf, line 181:
```suggestion
Arbor's root -- ie rank 0 -- process and that of the coupled
```

doc/concepts/interconnectivity.rst
```diff
@@ -74,6 +74,163 @@ full recipe.
...
+simulation. The control message is transferred by ``MPI_Sendrecv(12)`` of a byte
+buffer of length 1024. The payload comprises
```
boeschf, line 183:
```suggestion
simulation. The control message is a byte buffer of length 1024 and is 
transferred by ``MPI_Sendrecv(12)``. The payload comprises
```

arbor/include/arbor/simulation.hpp
```diff
@@ -51,6 +53,8 @@ class ARB_ARBOR_API simulation {
 
     time_type run(time_type tfinal, time_type dt);
 
+    time_type min_delay();
```
boeschf, line 56: would it be useful to add something like `max_epoch_length`, too?

doc/concepts/interconnectivity.rst
```diff
@@ -74,6 +74,163 @@ full recipe.
...
+outstanding communication. An exception will be raised. Note that Arbor might
+terminate even without sending or receiving an ``abort`` message in exceptional
+circumstances.
+
+On ``epoch`` Arbor will commence the next epoch. Note that Arbor may expect the
+last epoch to be shortened, ie when the total runtime is not a multiple of the
+epoch length.
+
+``Done`` signals the sending side is finished with the current simulation
+period. *May* cause the receiving side to quit.
```
boeschf, line 203: what do you mean by period here? epoch or simulation time overall?

example/remote/Makefile
boeschf, line 1: why to we need a Makefile? Seems to be here by accident, considering your paths are in there...

example/remote/remote.cpp
```diff
@@ -0,0 +1,149 @@
...
+    else if (mpi.group == 0) {
+        // Simulate external spikes
+        // * We send from a fictious cell with gid=local rank
+        // * The time is a function of rank and epoch.
+        // * We never stop until the other group kills the process ;)
+        std::cerr << "[EXT]" << mpi.rank << " " << mpi.local_rank << '\n';
+        for (int ep = 0;; ++ep) {
+            std::cerr << "[EXT] Epoch " << ep << '\n';
+            arb::cell_gid_type src = mpi.local_rank;
+            auto time = ep*2.0*dt + mpi.local_rank*dt + 0.05;
```
boeschf, line 87: Can we calculate the time based on the received message to demonstrate the value of the payload?

scripts/run_cpp_examples.sh
```diff
@@ -36,6 +36,20 @@ do
     cd -
 done
 
+# MPI only examples
+if [[ $PREFIX = mpirun* ]]
```
boeschf, line 40: wouldn't it be nicer to add `remote` to the list of examples above in this case?

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,238 @@
...
+}
+
+// Exchange control message. NOTE: This makes clever (?) use of MPI_Allreduce
+// to achieve simulatenous duplex mode broadcast by preparing zeroed buffers
+// on all ranks except the root and then adding up tons of zeros with a single
+// payload.
+inline
+ctrl_message exchange_ctrl(const ctrl_message& msg, MPI_Comm comm) {
+    int is_inter = 0;
+    mpi_checked(MPI_Comm_test_inter(comm, &is_inter), "exchange ctrl block: comm type");
```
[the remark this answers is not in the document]
thorstenhater: True

scripts/run_cpp_examples.sh
thorstenhater: That I do not understand?

example/remote/Makefile
thorstenhater: Yes, accident.

arbor/include/arbor/communication/remote.hpp
```diff
@@ -0,0 +1,238 @@
...
+
+struct arb_cell_id {
+    std::uint32_t gid;
+    std::uint32_t lid;
+};
+
+struct arb_spike {
+    arb_cell_id source;
+    arb_time_type time;
+};
```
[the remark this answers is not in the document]
thorstenhater: What do you propose? `Meta` isn't in C++ yet ;)

example/remote/remote.cpp
thorstenhater: This is computing fake spikes to send for the next epoch, so yes, we could use the payload, kind of, awkwardly.
For now I just print the bounds.

arbor/include/arbor/communication/remote.hpp
thorstenhater: You mean the rough equivalent to `static_assert(sizeof(std::variant<...> < 1024)`?

arbor/include/arbor/simulation.hpp
```diff
@@ -17,6 +18,7 @@
 #include <arbor/schedule.hpp>
 #include <arbor/spike.hpp>
 #include <arbor/util/handle_set.hpp>
+#include <arbor/util/unique_any.hpp>
```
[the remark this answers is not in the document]
thorstenhater: Jupp.

arbor/include/arbor/spike.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include <ostream>
 #include <type_traits>
+#include <vector>
```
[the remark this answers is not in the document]
thorstenhater: I like vectors ;)

scripts/run_cpp_examples.sh
boeschf: Lines 29 and following: same loop body, so lines 41ff are code repetition. Just a detail.

arbor/include/arbor/communication/remote.hpp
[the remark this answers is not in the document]
thorstenhater: I really do not expect these two to change at all, maybe once if we ever need to completely overhaul our comms strategy.

arbor/include/arbor/simulation.hpp
thorstenhater: True.

scripts/run_cpp_examples.sh
thorstenhater: Yeah. I decided not to go down this rabbit hole
https://stackoverflow.com/questions/8063228/check-if-a-variable-exists-in-a-list-in-bash
and instead copy paste the loop. Sue me ;)

arbor/include/arbor/communication/remote.hpp
boeschf: yes, something like this, including the meta bits

example/remote/readme.md
```diff
@@ -0,0 +1,11 @@
...
+
+This example shows
+1. setting up a simulation that uses an external tool to source spikes.
+2. the in-band control and spike exchange protocols. 
+3. use of the supplemental tools exposed by Arbor to handle these connections.
+
+Please run it with at least two MPI processes.
+
+Also included is a cross language client (Python)/ server (C++) example for
+demonstration. (Not yet tested due to OpenMPI fickleness)
```
boeschf, line 11: can you add a line on how you'd use the server/client example?

arbor/include/arbor/simulation.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <array>
+#include <any>
```
boeschf, line 4: any keeps creeping in...

boeschf (commented):
some minor details, otherwise looks very good

example/remote/readme.md
thorstenhater: Sure!

example/remote/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+add_executable(remote EXCLUDE_FROM_ALL remote.cpp)
+add_executable(server EXCLUDE_FROM_ALL server.cpp)
+add_executable(client EXCLUDE_FROM_ALL client.cpp)
```
boeschf, line 3: there is no such file, only the python variant

boeschf (approved):
thanks, looks good to me!

PR #2003, opened 2022-10-17
use fetch_content for gtest

boeschf:
investigate whether this helps with our test cleanup problems


ext/CMakeLists.txt
```diff
@@ -22,38 +22,87 @@ endif()
 add_library(ext-tinyopt INTERFACE)
 target_include_directories(ext-tinyopt INTERFACE tinyopt/include)
 
+
```
thorstenhater, line 25: A few questions:
- is `add_sub_directory` insufficient?
- if so, why?
- should libgtest be installed? I'd prefer not to.
- do we need to re-set CXX and ilk?

thorstenhater (changes requested):
Only dubious part is CMake, but that might just be CMake, honestly.

ext/CMakeLists.txt
boeschf: - The benchmark library uses the external project technique as well - by not using `add_sub_directory` we can limit the spill of the submodule content into our project. 
- libgtest (and google-benchmark for that matter) do not get installed
- haven't tried not setting it, but it shouldn't hurt

thorstenhater: ok, convinced me ;)

thorstenhater (approved):
👍

PR #2007, opened 2022-10-19
cleanup documentation

boeschf:
fixes a bunch of sphinx warnings

### DIFF
--- a/doc/concepts/labels.rst
+++ b/doc/concepts/labels.rst
@@ -870,4 +870,3 @@ API
 ---
 
 * :ref:`Python <pylabels>`
-* :ref:`C++ <cpplabels>`
--- a/doc/concepts/labels.rst
+++ b/doc/concepts/labels.rst
@@ -870,3 +870,4 @@ API
 ---
 
 * :ref:`Python <pylabels>`
+* :ref:`C++ <cpplabels>`
--- a/doc/concepts/lif_cell.rst
+++ b/doc/concepts/lif_cell.rst
@@ -23,3 +23,4 @@ API
 ---
 
 * :ref:`Python <pylifcell>`
+* :ref:`C++ <cpplabels>`
--- a/doc/concepts/lif_cell.rst
+++ b/doc/concepts/lif_cell.rst
@@ -23,4 +23,4 @@ API
 ---
 
 * :ref:`Python <pylifcell>`
-* :ref:`C++ <cpplabels>`
+* :ref:`C++ <cpplifcell>`
--- a/doc/concepts/lif_cell.rst
+++ b/doc/concepts/lif_cell.rst
@@ -23,4 +23,3 @@ API
 ---
 
 * :ref:`Python <pylifcell>`
-* :ref:`C++ <cpplifcell>`
--- a/doc/concepts/spike_source_cell.rst
+++ b/doc/concepts/spike_source_cell.rst
@@ -22,4 +22,3 @@ API
 ---
 
 * :ref:`Python <pyspikecell>`
-* :ref:`C++ <cppspikecell>`
--- a/doc/concepts/spike_source_cell.rst
+++ b/doc/concepts/spike_source_cell.rst
@@ -22,4 +22,4 @@ API
 ---
 
 * :ref:`Python <pyspikecell>`
-* :ref:`C++ <cpplabels>`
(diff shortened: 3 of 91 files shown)


doc/concepts/labels.rst
```diff
@@ -870,4 +870,3 @@ API
 ---
 
 * :ref:`Python <pylabels>`
-* :ref:`C++ <cpplabels>`
```
brenthuisman, line 873: I really want to keep these in, because users should know that stuff can be done in both lagnuages. Them being missing is something I don't object to showing, and otherwise we should track separately the missing bits which in the past proved to go out of date too quickly.

doc/concepts/lif_cell.rst
```diff
@@ -23,4 +23,3 @@ API
 ---
 
 * :ref:`Python <pylifcell>`
-* :ref:`C++ <cpplifcell>`
```
brenthuisman, line 26: idem

doc/concepts/spike_source_cell.rst
```diff
@@ -22,4 +22,3 @@ API
 ---
 
 * :ref:`Python <pyspikecell>`
-* :ref:`C++ <cppspikecell>`
```
brenthuisman, line 25: idem

doc/dev/cable_cell.rst
```diff
@@ -80,7 +80,7 @@ Discretisation splits the segments described by the morphology into *control
 volumes* (CV; sometimes called *compartments*) according to a ``cv_policy``.
 This allows us to construct a system of linear equations, the Hines matrix, to
 describe the evolution of the CV voltages according to the cable equation. Refer
-to :ref:`Discretisation <discretisation>` and :ref:`Cable equation
+to the :ref:`decor's discretization method <pycablecell-decor>` and :ref:`Cable equation
```
brenthuisman, line 83: Don't let Sam see this ;)

doc/python/mechanisms.rst
```diff
@@ -126,14 +126,14 @@ Cable cell mechanisms
...
     .. method:: density(name)
-        :noindex:
 
         constructs :attr:`mech` with *name* and default parameters.
 
         :param name: name of mechanism.
         :type name: str
 
     .. method:: density(name, params)
+        :noindex:
```
brenthuisman, line 136: Why is it here anyway?

doc/python/probe_sample.rst
```diff
@@ -4,6 +4,7 @@ Cable cell probing and sampling
 ===============================
 
 .. module:: arbor
+   :noindex:
```
brenthuisman, line 7: Not sure this should be here? There are currently definitely links to these elements.

brenthuisman (changes requested):
Thanks! Few small points:

doc/python/mechanisms.rst
boeschf: I moved the `noindex` attribute from the first method overload to the second just to be consistent.

doc/python/probe_sample.rst
boeschf: The module is already indexed in `doc/python/cell.rst`

doc/dev/cable_cell.rst
boeschf: I can leave the dead link as I did with the above links - is the chapter "Discretization" something we intend to write down at some point?

brenthuisman: A minor war was fought over discreti**s**ation vs discreti**z**ation. Don't ask :) No request for change meant.

doc/python/probe_sample.rst
brenthuisman: I thought this directive also excludes all that follows. Did you check that it does not break links to stuff in this file?

doc/python/mechanisms.rst
brenthuisman: Sure, I guess for crossref within the docs that's good enough.

doc/python/probe_sample.rst
boeschf: I don't think that this is the case, I do not get warnings about missing links, and the html doc still shows the option to make permalinks for the following entries/sections

brenthuisman (approved):

PR #2008, opened 2022-10-19
fix cmake: ninja + ExternalProject

boeschf:
Uses a solution as described here:
https://stackoverflow.com/questions/50400592/using-an-externalproject-download-step-with-ninja


thorstenhater:
Nice! I remember Ninja being a bit more picky

thorstenhater (approved):

PR #2009, opened 2022-10-20
Use HTTPS access instead of ssh/git for gtest.

thorstenhater:
Cause certain constrained/fortified systems to fail to build/clone arbor.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -21,3 +21,5 @@
 	path = ext/googletest
 	url = git@github.com:google/googletest.git
 	branch = release-1.12.1
+[submodule "ext/googletest/"]
+	url = https://github.com/google/googletest.git


.gitmodules
```diff
@@ -21,3 +21,5 @@
 	path = ext/googletest
 	url = git@github.com:google/googletest.git
 	branch = release-1.12.1
+[submodule "ext/googletest/"]
+	url = https://github.com/google/googletest.git
```
boeschf, line 25: please remove ssh/git address and replace with https instead of adding another submodule

thorstenhater: Interesting ... I was relying on
```
 git submodule set-url ext/googletest/ https://github.com/google/googletest.git
 ```
doing the right thing here.

boeschf (approved):

PR #2012, opened 2022-10-21
Workaround for malloc failure in CI

AdhocMan:
This replaces the default malloc implementation with tcmalloc as a potential workaround for our failures in CI.
I haven't been able to track down the exact cause for the failures, but I've run the CI pipeline multiple times with these changes and so far it seems to fix the issue (fingers crossed).

### DIFF
--- a/.github/workflows/test-everything.yml
+++ b/.github/workflows/test-everything.yml
@@ -92,12 +92,14 @@ jobs:
         # This is a workaround for the unfortunate interaction of MacOS and OpenMPI 4
         # See https://github.com/open-mpi/ompi/issues/6518
         OMPI_MCA_btl: "self,tcp"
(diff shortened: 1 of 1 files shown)


thorstenhater (changes requested):
Ok, if this works that's amazing. However, I am a worried that we are plastering over an issue we do not understand. Also, this adds compleixty with no indication as to when we can get 
rid of it.

That leaves me with some questions
1. Is there any potential explenation as to why this works and why the vanilla version fails?
2. Please document (two sentences is plenty) that we are doing this and why in the dev section.
3. Is there any motivation for trying this particular route? Is there a bug in one of the underlying layers? libc-malloc?

AdhocMan:
Since #2013 seems to fix the issue, I'm closing this PR for now.

PR #2014, opened 2022-10-24
Include ubenches into CI

boeschf:
- add benchmark workflow
- update benchmarks to new cable_cell API

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -1,0 +1,40 @@
+name: Benchmarks
+
+on:
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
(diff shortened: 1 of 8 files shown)


brenthuisman (dismissed):
lgtm, hope they run w/o error.

thorstenhater (approved):
Thanks, these should always build, no matter what. In the future we might consider adding them to a graph to have an idea if/how our performance changes over time. More benchmarks would be needed, too, but these are a nice first step.

thorstenhater:
Hey @boeschf 

I am using this PR as another canary to see whether my attempt to rid us of the CI
failures is helpful.

thorstenhater (approved):
👍

PR #2017, opened 2022-10-25
Bump versions

### LINKED ISSUE #1815: Maintenance: Bump versions across the board
We will remove support for the following in ~0.7~ 0.8
- [x] python 3.6. EOL reached
- [x] CUDA 10
- [x] GCC 8
- [x] Clang 8
- [x] Apple 9

and bump these by one each.

thorstenhater:
Bump versions as described in #1815 

- GCC: goes up to 9 from 8, needed by CUDA 11
- CUDA: goes up to 11 from 10
- Clang: 10 up from 8
- Python: stays at 3.7

Our minimally supported Ubuntu is 20.04 LTS as GH Actions deprecated 18.04. MacOS is minimally
at v11 by the same constraints. I changed the compiler versions to reflect this while staying within 
our policies: all of them were released in early 2020. I was tempted to raise Python to 3.8 (2019) or
even 3.9 (2020) by the same policy; but decided to be a bit more conservative there. That said, Python
will be bumped next.

Closes #1815

### DIFF
--- a/.github/workflows/test-everything.yml
+++ b/.github/workflows/test-everything.yml
@@ -112,32 +112,6 @@ jobs:
           python-version: ${{ matrix.config.py }}
       - name: Update pip
         run: python -m pip install --upgrade pip
-      # - name: OpenMPI cache
-        # uses: actions/cache@v3
-        # id:   cache-ompi
-        # with:
-          # path: ~/openmpi-4.1.4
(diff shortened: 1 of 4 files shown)


brenthuisman (approved):
Python will stay at 3.7 likely for a loooooong time. Thanks Google Colab :)

PR #2019, opened 2022-10-25
Speed up CI

boeschf:
- use `Cmake`'s `Ninja` backend for building `Arbor`
- use `ccache` for storing build artifacts
  - unique cache key: needs to reflect os, compiler, shared/static variant and CPU vector extensions (avx, avx2, etc)


brenthuisman:
Hurray: https://github.com/arbor-sim/arbor/actions/runs/3321531860/jobs/5490065274

boeschf:
well, the ninja stuff runs
the error came because I tried to additionally use `ccache` - and I am pretty sure it is due to the cache `key` not including the CPU architecture / SIMD vector length

brenthuisman:
The key might need to be specific for each config in the matrix. Reading the [docs](https://docs.github.com/en/github-ae@latest/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_id), I think `job_id` might be that unique string.

thorstenhater:
Nice. I am going to merge this and after the CI is through fold this into #1958 to check whether we get the weird behaviour.

thorstenhater (approved):
Nice job. We should -- for the future -- investigate whether we can factor out some part of the .yml.

PR #2022, opened 2022-10-27
update cray documentation

boeschf:
- updated documentation about cray systems
- added comments about cuda target

### DIFF
--- a/doc/install/build_install.rst
+++ b/doc/install/build_install.rst
@@ -747,7 +747,7 @@ environment variable to ``dynamic``.
 
 .. note::
 
-    This step is no longer required beginning with the Cray PE 10.06 release, where
+    This step is no longer required beginning with the Cray PE 19.06 release, where
     the default linking mode is dynamic.
 
     On older systems, if ``CRAYPE_LINK_TYPE`` isn't set, there will be warnings like
@@ -772,7 +772,7 @@ Putting it all together, a typical workflow to build Arbor on a Cray system is:
 
 .. code-block:: bash
 
-    export CRAYPE_LINK_TYPE=dynamic    # may no longer be required
+    export CRAYPE_LINK_TYPE=dynamic    # only required if Cray PE version < 19.06
     
     # For GPU setup
     module load daint-gpu/21.09        # system specific
--- a/doc/install/build_install.rst
+++ b/doc/install/build_install.rst
@@ -666,6 +666,22 @@ using the supplied MPI compiler wrappers in preference.
         $ CC --version
         g++ (GCC) 6.2.0 20160822 (Cray Inc.)
 
+
+Heterogeneous systems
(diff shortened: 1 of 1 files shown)


doc/install/build_install.rst
```diff
@@ -706,27 +721,37 @@ then choose GCC 7.1.0
...
 .. code-block:: bash
 
-    export CRAYPE_LINK_TYPE=dynamic
+    $ export CRAYPE_LINK_TYPE=dynamic
 
-Putting it all together, a typical workflow to build Arbor on a Cray system is:
+.. note::
 
-.. code-block:: bash
+    This step is no longer required beginning with the Cray PE 10.06 release, where
```
brenthuisman, line 750: If no longer required, then remove.

doc/install/build_install.rst
```diff
@@ -736,6 +761,42 @@ Putting it all together, a typical workflow to build Arbor on a Cray system is:
     Often the library or executable will work, however if a different glibc is loaded,
     Arbor will crash at runtime with obscure errors that are very difficult to debug.
 
+For building Arbor's python interface, it may be necessary to load a more recent
```
brenthuisman, line 764: May be, or is?

brenthuisman (changes requested):
Minor remarks.

doc/install/build_install.rst
boeschf: Haven't tested all systems, so cannot tell you exactly. It is required on Piz Daint, though. This isn't supposed to be an exhaustive system documentation, and people are expected to look up their system specs / best practices / ask sysadmin etc.

doc/install/build_install.rst
boeschf: Depends on the Cray PE version. What do you suggest instead?

doc/install/build_install.rst
brenthuisman: I guess 'knows about the Cray system' is the operating clause, not 'more recent Python'? Maybe:

```suggestion
For building Arbor's Python interface, it may be necessary to load a Python that knows about the target system; in the case of the CSCS Cray system:
```

doc/install/build_install.rst
brenthuisman: "Cray PE" is not defined, so it may not be obvious that it refers to our how a user controls it. I image it's an env with a set of GCC builds etc. one can load? Or is it done system wide and has the user no choice?

boeschf: Cray PE= Cray Programming Environment. Yes it comes with a bunch of tools: compilers, MPI, scientific libraries, but it also sets a bunch of environment variables etc. It usually the minimum you have on a system provided by Cray (HPE).

brenthuisman (approved):

Issue #1987, opened 2022-09-26
Add tutorials and examples for SDE


thorstenhater:
#1884 is about to get merged, we should add a few basic examples at least to 
show how to use it from Python and C++. Also, some standard mechanisms could
be included in a 'stochastic' catalogue.

### TODO

- [x] add a simple example
  - [x] C++
  - [x] Python
- [x] add stochastic catalogue
  - [x] O-U process point source
- [ ] Optional: add tutorial

schmitts:
I can contribute the Graupner-Brunel example.

thorstenhater:
That'd be cool.

PR #2030, opened 2022-11-03
SDE examples

boeschf:
Addresses #1987 
- reproducibility test for random number generator
- C++ example demonstrating Ornstein-Uhlenbeck process
- Python example featuring stochastic mechanism
- stochastic mechanism catalogue

The Python example is modeled after @jlubo's work here: https://github.com/jlubo/arbor_ou_lif_example

## TODO
- [x] reproducibility test
- [x] C++ example
- [x] Python example
- [x] Catalogue


brenthuisman (changes requested):
Looks good, thanks! Some small points.

python/example/ou_lif/Readme.md
```diff
@@ -0,0 +1,29 @@
...
+### References: [1] Ricciardi, LM. 1977. _Diffusion Processes and Related Topics in Biology_.
+Springer, Berlin, Germany.
+
+[2] Moreno-Bote, R; Parga, N. 2010. Response of integrate-and-fire neurons to noisy inputs filtered
+by synapses with arbitrary timescales: Firing rate and correlations. _Neural Comput._ 22.
+
+[3] Destexhe, A; Rudolph, M; Paré, D. 2003. The high-conductance state of neocortical neurons in
+vivo. _Nat. Rev. Neurosci._ 4.
+
+[4] Dayan, P; Abbott, LF. 2001. _Theoretical Neuroscience_. MIT Press, Cambridge/MA, USA.
```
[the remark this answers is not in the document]
boeschf: added a paragraph

python/example/ou_lif/ou_lif.py
```diff
@@ -0,0 +1,305 @@
...
+            tmp = np.zeros(len(data[:, 0]))
+        data_curr.append(tmp)
+
+    # get spikes
+    spike_times = sim.spikes()['time']
+
+    # assemble, store, and plot data
+    # ==============================
+
+    data_header = "Time, V, I_stim, I_bg"
```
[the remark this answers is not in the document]
boeschf: I'd appreciate if you could do it later, thanks

mechanisms/stochastic/noisy_expsyn_curr.mod
```diff
@@ -0,0 +1,87 @@
+: Exponential current-based synapse
+: Input current modeled by an Ornstein-Uhlenbeck process
```
jlubo, line 2:
```suggestion
: Ornstein-Uhlenbeck input
: Can be used to model the input current from a population of exponential current-based synapses.
```
I'm suggesting this because from a neuroscience perspective, it is not one single synapse being modeled here, but a population. Accordingly, I think the filename should also be changed, e.g. to `ornstein_uhlenbeck.mod`.

python/example/ou_lif/ou_lif.py
```diff
@@ -0,0 +1,342 @@
...
+
+        # protocol for stimulation
+        # "time_start": starting time, "scheme": protocol type
+        self.bg_prot = {"time_start": 0, "scheme": "FULL", "label": "ou_bg"}
+        self.stim1_prot = {"time_start": 10000, "scheme": "TRIPLET", "label": "ou_stim"}
+        self.stim2_prot = {
+            "time_start": 15000,
+            "scheme": "ONEPULSE",
+            "label": "ou_stim",
+        }
```
jlubo, line 142:
```suggestion
        self.stim2_prot = {"time_start": 15000, "scheme": "ONEPULSE", "label": "ou_stim"}
```
Formatting adapted

python/example/ou_lif/Readme.md
```diff
@@ -0,0 +1,37 @@
...
+synaptic input from a population of neurons.  This is mathematically based on the diffusion
+approximation, which analytically describes the input current that arises from presynaptic Poisson
+spiking if the spiking activity is sufficiently high and the evoked postsynaptic potentials are
+sufficiently small compared to the threshold potential [1,2].  Moreover, the exponentially decaying
+autocorrelation function of the OU process resembles that of currents evoked by the spiking activity
+of many neocortical neurons [3].
+
+Here, we use the Ornstein-Uhlenbeck process in the following form, `dI/dt = -(1/τ) * (I - μ) +
+√(2/τ) * σ * W`.  Assuming homogeneous Poisson spiking for a population of *N* neurons in the
+putative population, the mean current is `μ = N * f * w_out`, and the volatility (standard
```
jlubo, line 17:
```suggestion
√(2*τ) * σ * W`.  Assuming homogeneous Poisson spiking at frequency *f* for a putative population of *N* neurons, 
the mean current is `μ = N * f * w_out`, and the volatility (standard
```
The multiplication is just a suggestion: in my opinion, one can follow the equations more easily this way (also see the other related comments).

python/example/ou_lif/Readme.md
```diff
@@ -0,0 +1,37 @@
...
+
+This example is taken from [Jannik Luboeinski's repo](https://github.com/jlubo/arbor_ou_lif_example)
+and adapted for inclusion into Arbor.
+
+The simulation consists of a single compartment cable cell with simple leaky integrate-and-fire (LIF)
+dynamics with Ornstein-Uhlenbeck (OU) input current.  The OU input current can approximate the
+synaptic input from a population of neurons.  This is mathematically based on the diffusion
+approximation, which analytically describes the input current that arises from presynaptic Poisson
+spiking if the spiking activity is sufficiently high and the evoked postsynaptic potentials are
+sufficiently small compared to the threshold potential [1,2].  Moreover, the exponentially decaying
```
jlubo, line 11:
```suggestion
dynamics with Ornstein-Uhlenbeck (OU) input current. The OU input current can approximate the
synaptic input from a population of neurons. This is mathematically based on the diffusion
approximation, which analytically describes the input current that arises from presynaptic Poisson
spiking if the spiking activity is sufficiently high and the evoked postsynaptic potentials are
sufficiently small compared to the threshold potential [1,2]. Moreover, the exponentially decaying
```
Spaces removed

doc/concepts/mechanisms.rst
```diff
@@ -77,6 +77,13 @@ Two catalogues are provided that collect mechanisms associated with specific pro
 * ``bbp_catalogue`` For models published by the Blue Brain Project (BBP).
 * ``allen_catalogue`` For models published on the Allen Brain Atlas Database.
 
+A fourth catalogue ``stochastic_catalogue`` provides mechanisms expressed as stochastic differential
+equations:
+
+* *noisy_expsyn_curr* Synapse mechanism similar to *expsyn_curr*, where the exponential decay is
```
jlubo, line 83:
```suggestion
* *ornstein_uhlenbeck* Synapse mechanism that can stochastically account for a population of *expsyn_curr* synapses. The mechanism is mathematically equal to *expsyn_curr* but with the exponential decay being
```
Relating to comment below.

mechanisms/stochastic/noisy_expsyn_curr.mod
```diff
@@ -0,0 +1,87 @@
...
+: statistic properties:
+:   E[I_ou] = μ - (μ - I_ou_0) * e^(-t/τ)
+: Var[I_ou] = σ^2 * (1 - e^(-2 * t/τ))
+:
+: in the limit t->∞:
+:   E[I_ou] = μ
+: Var[I_ou] = σ^2
+
+NEURON {
+    POINT_PROCESS noisy_expsyn_curr
```
jlubo, line 23:
```suggestion
    POINT_PROCESS ornstein_uhlenbeck
```
See comment above.

python/example/ou_lif/ou_lif.py
```diff
@@ -0,0 +1,342 @@
...
+    # =====================================================
+
+    # synaptic time constant in ms
+    tau_syn = 5.0
+    # mean in nA
+    mu_bg = 0.15
+    # volatility in nA
+    sigma_bg = 0.5
+    # instantiate mechanism
+    ou_bg = arb.mechanism("noisy_expsyn_curr")
```
jlubo, line 84:
```suggestion
    ou_bg = arb.mechanism("ornstein_uhlenbeck")
```
See comment above.

python/example/ou_lif/ou_lif.py
```diff
@@ -0,0 +1,342 @@
...
+    # firing rate of the putative neurons in Hz
+    f = 100
+    # synaptic weight in nC
+    w_out = h_0 / R_leak
+    # mean in nA
+    mu_stim = N * f * w_out
+    # volatility in nA
+    sigma_stim = np.sqrt((1000.0 * N * f) / (2 * tau_syn)) * w_out
+    # instantiate mechanism
+    ou_stim = arb.mechanism("noisy_expsyn_curr")
```
jlubo, line 103:
```suggestion
    ou_stim = arb.mechanism("ornstein_uhlenbeck")
```
See comment above.

mechanisms/CMakeLists.txt
```diff
@@ -21,6 +21,12 @@ make_catalogue(
   VERBOSE ${ARB_CAT_VERBOSE}
   ADD_DEPS ON)
 
+make_catalogue(
+  NAME stochastic
+  MOD noisy_expsyn_curr
```
jlubo, line 26:
```suggestion
  MOD ornstein_uhlenbeck
```
See comment below.

python/example/ou_lif/ou_lif.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def num_cells(self):
+        return 1
+
+    def probes(self, gid):
+        # probe membrane potential, total current, and external input currents
+        return [
+            arb.cable_probe_membrane_voltage('"center"'),
+            arb.cable_probe_total_ion_current_cell(),
+            arb.cable_probe_point_state_cell("noisy_expsyn_curr", "I_ou"),
```
jlubo, line 161:
```suggestion
            arb.cable_probe_point_state_cell("ornstein_uhlenbeck", "I_ou"),
```
See comment above.

jlubo (changes requested):

python/example/ou_lif/ou_lif.py
```diff
@@ -0,0 +1,305 @@
...
+
+        # initialize catalogue and cell properties
+        # ========================================
+
+        self.props = arb.neuron_cable_properties()
+        cat = arb.load_catalogue("./ou_lif-catalogue.so")
+        cat.extend(arb.stochastic_catalogue(), "")
+        self.props.catalogue = cat
+
+        # cell morphology
```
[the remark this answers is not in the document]
brenthuisman: `def make_cell()` is what I meant, thanks!

python/example/ou_lif/ou_lif.py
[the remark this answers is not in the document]
brenthuisman: OK

python/example/ou_lif/lif.mod
```diff
@@ -0,0 +1,66 @@
+: Leaky-integrate and fire neuron
+: author Yannik Luboeinski
+: based on the LIF mechanism written by Sebastian Schmitt (https://github.com/tetzlab/FIPPA/blob/main/STDP/mechanisms/lif.mod)
```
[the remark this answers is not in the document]
brenthuisman: Would is also be OK to link here: https://github.com/arbor-contrib/FIPPA/blob/main/STDP/mechanisms/lif.mod ?
Just in case the Tetzlab changes their mind at any point ;)

mechanisms/stochastic/noisy_expsyn_curr.mod
boeschf: ok, thanks, that makes sense. However, Ornstein-Uhlenbeck is a particular mathematical equation which can be used to model many things. Can we come up with a more descriptive name which reflects the neuroscience?

boeschf:
@jlubo Let me elaborate why I chose the particular variables in the formulation of the SDE:
- it reflects the physical process: the deterministic term is scaled by the inverse of a relaxation time $\frac{1}{\tau}$ (it is a rate of change) and the stochastic term is scaled by $\frac{1}{\sqrt{\tau}}$ (change with respect to the  Browinan motion)
- the stochastic term is also scaled by $\sqrt{2}$ such that the input parameter $\sigma$ reflects the long term standard deviation

I understand that in the literature people may use a different formulation. I am open to changing the variables, however, I would prefer a more vanilla formulation in this case:
$$X' = \theta (\mu - X) + \sigma W,$$
where the parameters $\theta$, $\mu$ and $\sigma$ are specified in the python code.

Let me know what you think.

python/example/ou_lif/ou_lif.py
boeschf: Our linter complains when changing this to a one-liner, so I'll leave it as is.

python/example/ou_lif/lif.mod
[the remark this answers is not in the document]
jlubo: I guess so, @schmitts please comment if you object! Should the repo [arbor_ou_lif_example](https://github.com/jlubo/arbor_ou_lif_example) then also be forked by [arbor-contrib](https://github.com/arbor-contrib)?

mechanisms/stochastic/noisy_expsyn_curr.mod
jlubo: What about `ornstein_uhlenbeck_input` or `ou_input`? I guess that would account for the neuroscience aspect of it, modeling synaptic input.

jlubo:
Sure, that looks like a valid reasoning. The only thing that's not clear to me is how in this equation σ is supposed to represent the long-term standard deviation, even though is defined by σ_{Gauss}/√(2*τ), or what you mean by long-term standard deviation. 

Anyway, I think we can keep the mathematical description the way you have suggested it. Just as far as I can see σ then can't be interpreted as standard deviation of the OU process.

python/example/ou_lif/lif.mod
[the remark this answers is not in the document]
schmitts: @brenthuisman is the expert on these kind of questions.

boeschf:
Regarding the standard deviation: please have a look at https://github.com/boeschf/arbor/blob/sde_test/doc/concepts/mechanisms.rst#stochastic-processes:
In essence, once you wait long enough ( $t \rightarrow \infty$ ), we get
$$E[I_{ou}] \rightarrow \mu,$$ 
$$Var[I_{ou}] \rightarrow \sigma.$$

brenthuisman:
It's what makes much of particle physics much simpler than it should be ;)

brenthuisman (approved):
let's go!

jlubo:
And @boeschf you were right, I just did the maths and both formulations yield the correct long-term variance :)

PR #2034, opened 2022-11-09
fix some compiler warning

boeschf:
- minor changes


boeschf:
yeah, forgot those few places earlier

PR #2035, opened 2022-11-14
Additional builtin functions to Arbor's NMODL dialect

boeschf:
Added common math functions which I came across in the wild quite frequently. Implementations may be more optimized than hand-rolled versions.

- sqrt: usual square root
- heaviside_right: right-continuous heaviside step function
- heaviside_left: left-continuous heaviside step function
- signum: sign bit function


test/unit-modcc/test_printers.cpp
```diff
@@ -58,25 +58,29 @@ TEST(scalar_printer, constants) {
...
+        {"y=heaviside_right(c-theta)", "y=((arb_value_type)((c-theta)>=0.))"},
+        {"y=heaviside_left(c-theta)",  "y=((arb_value_type)((c-theta)>0.))"},
     };
 
     // create a scope that contains the symbols used in the tests
     Scope<Symbol>::symbol_map globals;
     auto scope = std::make_shared<Scope<Symbol>>(globals);
 
-    for (auto var: {"x", "y", "z", "a", "b", "c"}) {
+    for (auto var: {"x", "y", "z", "a", "b", "c", "c", "theta"}) {
```
thorstenhater, line 83: 2x `c`?

thorstenhater (commented):
Thanks, `sqrt` has been sorely missed! I have seen less need for the step functionals, but ok.
A few general comments:
- inhomogeneous parameters have a `step` function already, consider renaming this or here for consistency
- would `sqrt` benefit from an approximative treatment like `exp`?
- I find the derivatives of `step` a bit iffy. But I guess currently we accept if-then-else in `DERIVATIVE`s as well.

A side note:
I am not very confident about the SVE backend, as we have no CI there, which makes me 
quite uncomfortable to continue supporting it. Also, no users.

test/unit-modcc/test_printers.cpp
boeschf: typo. thanks for catching it

brenthuisman:
Are any ARM64 clusters (expected to be) available for neuroscientists? AFAIK nobody is planning to put one up for general availability, which means a big 'at your own peril' is enough?

For reference, running the testsuite under QEMU emulating an SVE chip takes 24-36 hours.

thorstenhater:
About the derivative:
By definition there is none, so 'correct' would be to bail. However, it may well be that we need to provide one.

The question approximating `sqrt` was more of a general thought, but it's not worth it.

thorstenhater (approved):

PR #2038, opened 2022-11-17
Update __ballot -> __ballot_sync in reduce_by_key.

thorstenhater:
Fails to compile on current NVHPC. `ballot` is deprecated since CUDA 9.


brenthuisman:
'it' being the deprecation or sth else?

thorstenhater:
The fix is non-sense and the problem occurs due to a misconfiguration of the ou example within CMake.

PR #2039, opened 2022-11-17
BUGFIX: add ARB_CUDA flag to example catalogue

boeschf:
- adding this flag manually follows the same strategy as the dummy catalogue in unit tests

### DIFF
--- a/example/ornstein_uhlenbeck/ou.cpp
+++ b/example/ornstein_uhlenbeck/ou.cpp
@@ -20,8 +20,6 @@ public:
         // build catalogue with stochastic mechanism
         cell_gprop_.catalogue = global_default_catalogue();
         cell_gprop_.catalogue.import(arb::global_ornstein_uhlenbeck_catalogue(), "");
-        //cell_gprop_.catalogue.import(build_catalogue(), "");
-        //cell_gprop_.catalogue = make_catalogue(global_default_catalogue());
         cell_gprop_.default_parameters = neuron_parameter_defaults;
         
         // paint the process on the whole cell


example/ornstein_uhlenbeck/CMakeLists.txt
```diff
@@ -6,6 +6,10 @@ make_catalogue_standalone(
     CXX_FLAGS_TARGET ${ARB_CXX_FLAGS_TARGET_FULL}
     VERBOSE ON)
 
+if(ARB_WITH_NVCC)
```
AdhocMan, line 9: There is also `ARB_WITH_CUDA_CLANG` and `ARB_WITH_HIP_CLANG` to consider.

boeschf: maybe, but then the tests should also handle this, and they do not - what is the difference?

AdhocMan: For HIP, we need to define `ARB_HIP` and `ARB_CUDA` for the other two. See https://github.com/arbor-sim/arbor/blob/master/CMakeLists.txt#L369

boeschf: yes, you are right. Changes:
- use `make_catalogue` instead of `make_catalogue_standalone`
- link to `arbor-private-deps`
- mark all files as c++ sources when using `cuda-clang` or `hip-clang`

example/ornstein_uhlenbeck/ou.cpp
```diff
@@ -20,7 +19,9 @@ class recipe: public arb::recipe {
 
         // build catalogue with stochastic mechanism
         cell_gprop_.catalogue = global_default_catalogue();
-        cell_gprop_.catalogue.import(build_catalogue(), "");
+        cell_gprop_.catalogue.import(arb::global_ornstein_uhlenbeck_catalogue(), "");
+        //cell_gprop_.catalogue.import(build_catalogue(), "");
```
AdhocMan, line 23: Some leftover code here.

AdhocMan (changes requested):
Looks good to me apart from that one comment. Maybe also use bors to run tests including GPU, if that's still working.

example/ornstein_uhlenbeck/ou.cpp
boeschf: thanks for spotting

boeschf:
bors try

AdhocMan (approved):
Not sure how to check if bors is still running, but this lgtm otherwise.

PR #2040, opened 2022-11-17
Throw better errors when we cannot look up ion diffusivity

thorstenhater:
- Stop throwing `std::bad_optional_access` if we cannot look up the (default) diffusivity of an ion.
- Add more tests to nail down the errors related to diffusivty.

### DIFF
--- a/arbor/fvm_layout.cpp
+++ b/arbor/fvm_layout.cpp
@@ -16,8 +16,6 @@
 #include <arbor/morph/mprovider.hpp>
 #include <arbor/morph/morphology.hpp>
 
-#include <iostream>
-
 #include "fvm_layout.hpp"
 #include "threading/threading.hpp"
 #include "util/maputil.hpp"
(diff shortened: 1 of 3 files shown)


boeschf:
could it be that you haven't updated the submodules - not sure why there would be a diff otherwise?

arbor/fvm_layout.cpp
```diff
@@ -315,11 +313,18 @@ ARB_ARBOR_API fvm_cv_discretization fvm_cv_discretize(const cable_cell& cell, co
...
             }
         }
-        // Fetch defaults, either global or per cell
-        auto gd = *(value_by_key(dflt.ion_data,        ion).value().diffusivity
-                  | value_by_key(global_dflt.ion_data, ion).value().diffusivity);
-        if (gd <= 0.0) {
-            throw cable_cell_error{util::pprintf("Illegal global diffusivity '{}' for ion '{}'.", gd, ion)};
+        arb_value_type def = 0.0;
+        if (auto data = value_by_key(global_dflt.ion_data, ion);
+            data && data->diffusivity) {
```
boeschf, line 318: the if statement with initializer is is a c++17 feature I was not aware of! cool

test/unit/test_diffusion.cpp
```diff
@@ -288,3 +288,64 @@ TEST(diffusion, decay_by_event) {
...
+                tree.append(0, { -1, 0, 0, 3}, {1, 0, 0, 3}, 2);
+                // Utilise diffusive ions
+                dec.place("(location 0 0.5)"_ls, synapse("inject/x=bla", {{"alpha", 200.0}}), "Zap");
+                dec.paint("(all)"_reg, density("decay/x=bla"));
+            }
+
+            cell_size_type num_cells()                                   const override { return 1; }
+            cell_kind get_cell_kind(cell_gid_type)                       const override { return cell_kind::cable; }
+            std::any get_global_properties(cell_kind)                    const override { return gprop; }
+            util::unique_any get_cell_description(cell_gid_type)         const override { return cable_cell({tree}, dec); }
```
boeschf, line 308:
```suggestion
            cell_size_type num_cells()                           const override { return 1; }
            cell_kind get_cell_kind(cell_gid_type)               const override { return cell_kind::cable; }
            std::any get_global_properties(cell_kind)            const override { return gprop; }
            util::unique_any get_cell_description(cell_gid_type) const override { return cable_cell({tree}, dec); }
```

arbor/fvm_layout.cpp
thorstenhater: It's extremely handy.

boeschf (approved):

PR #2042, opened 2022-11-17
simplify make catalogue

boeschf:
Current state:
- unit tests and an example build their own mechanism catalogue
- repeated logic for detecting GPUs, passing compile time flags

Proposal:
- we should leverage the cmake variables and targets which are already there since these catalogues are internal to arbor
=> build a static library with mechanisms, which links to `arbor-private-deps`


thorstenhater (approved):
I like, simplification and a bit of isolation!

PR #2043, opened 2022-11-17
spack gpu option

boeschf:
- add conditional variant for cuda builds to enable GPU-based random number generation


thorstenhater (approved):
👍

PR #2045, opened 2022-11-21
BluePyOpt API updates

lukasgd:
This is a small PR to track updates in the BluePyOpt API (cf. https://github.com/BlueBrain/BluePyOpt/pull/393) compared to the v0.8 release.


brenthuisman (approved):
https://github.com/BlueBrain/BluePyOpt/pull/393 merged, let's go!

PR #2048, opened 2022-11-25
Use pugixml instead of libxml2

thorstenhater:
- Simplify XML parsing infrastructure
- Use a vendored XML library instead of a runtime dependency
- Should enable NML2 morph on WASM for @llandsmeer


test/unit/test_strprintf.cpp
```diff
@@ -51,12 +51,12 @@ TEST(strprintf, wrappers) {
     char buf[200];
 
     auto uptr = std::unique_ptr<int>{new int(17)};
-    sprintf(buf, "uptr %p", uptr.get());
+    snprintf(buf, 200, "uptr %p", uptr.get());
```
brenthuisman, line 54: good!

brenthuisman (changes requested):
Great! Can't find any problem, other than you sneaking in version bumps for three other deps in `/ext`. If you remove those, I'll merge.

doc/dependencies.csv
```diff
@@ -8,7 +8,6 @@ Build option/target,Tool name,Minimum version,Notes,Update criteria
 --,Apple Clang,11,Apple LLVM version 11.0.0 (clang-1100.0.33.8),
 ARB_GPU,Hip Clang,ROCm 3.9,HIP support is currently experimental.,
 ARB_GPU,CUDA,11.0,,"* It is available on all of the target HPC systems (Piz Daint, Juwels Booster)"
```
brenthuisman, line 10: Can you add pugi here?

thorstenhater: Sure.

thorstenhater:
These version bumps are an effect of my flailing at `git submodule` to remove a change in another PR. 
I am more and more confused by these.

thorstenhater:
> Also, I see conflicts("+shared", when="~pic") there. Should I start twitching?

No clue ... does this imply that they won't build if PIC=OFF, but SHARED=ON?

brenthuisman:
gitmodules are unfortunately part of the repo, not any object (commit, branch, tag) in it (although you can change them through one?). If you set a module to a certain hash, it'll be set to that for any commit you check out.

Fortunately (but perhaps confusingly) the fact that Github forks are different repos, means that you can have different versions of submodules that way between your PR and the main repo. So, update a module in your repo, review an update in another's repo, merge with main repo: three sets of module hashes at play. I try to limit working on them one by one for that reason.

I agree, this is Not Great.

brenthuisman (changes requested):
Tiny change.

thorstenhater:
I know all that, I have been a victim for quite some time now. It's just very annoying and errorprone. 
See #2049

spack/package.py
```diff
@@ -70,9 +69,9 @@ class Arbor(CMakePackage, CudaPackage):
     # misc dependencies
     depends_on("fmt@7.1:", when="@0.5.3:")  # required by the modcc compiler
     depends_on("fmt@9.1:", when="@0.7.1:")
+    depends_on("pugixml@1.11:", when="@0.7.1:")
```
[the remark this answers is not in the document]
thorstenhater: Yes and if I read the docs correctly (and pugi adheres to semver), this should be compatible and OK.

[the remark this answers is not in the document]
brenthuisman: Unless I made mistakes, I fixed the submodules to the commithash of the version stated in the spack file. Which ones are not according to you?

brenthuisman:
One never knows *all* about Git ;)

Also, maybe add a pugi mention to https://github.com/arbor-sim/arbor/blob/master/ATTRIBUTIONS.md

spack/package.py
```diff
@@ -70,9 +69,9 @@ class Arbor(CMakePackage, CudaPackage):
...
     depends_on("nlohmann-json")
```
brenthuisman, line 73: Good that you point out there's no versions for some other deps here either.

brenthuisman (approved):
Let's go.

spack/package.py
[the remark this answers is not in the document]
thorstenhater: No, that's not the point, rather that `fmt:9.1` is not giving the patch version. And I do not think it should; thus I didn't give one for pugixml

PR #2049, opened 2022-11-28
Hammer down submodules.

thorstenhater:
Fix submodule versions inplace. git submodule update should now checkout the correct versions.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,25 +1,25 @@
 [submodule "google-benchmark"]
 	path = ext/google-benchmark
 	url = https://github.com/google/benchmark.git
-	branch = main
+	branch = v1.7.1
 [submodule "ext/fmt"]
 	path = ext/fmt
 	url = https://github.com/fmtlib/fmt.git
-	branch = master
+	branch = 9.1.0	
 [submodule "ext/random123"]
 	path = ext/random123
 	url = https://github.com/DEShawResearch/random123.git
-	branch = main
+	branch = v1.14.0
 [submodule "ext/json"]
 	path = ext/json
 	url = https://github.com/nlohmann/json.git
-	branch = master
+	branch = v3.11.2
 [submodule "ext/pybind11"]
 	path = ext/pybind11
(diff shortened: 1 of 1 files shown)


brenthuisman (changes requested):
Sorry, this does not do what you think it does. I didn't go into it for [v0.8](https://github.com/arbor-sim/arbor/commit/8e82ec1947a3a6d84452969d1d628b2292654319#diff-fe7afb5c9c916e521401d3fcfb4277d5071798c3baf83baf11d6071742823584)'s changes, but I reverted Fabian's attempt. Reasons:

1. It's not formally supported [Git Submodules Docs](https://www.git-scm.com/docs/gitmodules). I don't know if it's wise to depend on an undocumented feature.
2. Even though it seems to work with upstream Git (which is not the only client out there, I have not tried them all), it does not hammer down anything: only upon fresh clone (or sync) are the repos checked out to the associated commit hash. Changing the value in `.gitmodules` doesn't do anything for existing clones/copies. Similarly, updating the commit hash the usual way does not update the value in `.gitmodules`. I.e. to know if what's in `.gitmodules` is actually checked out, you still need to verify manually. In summary, in my opinion this does not creaty the clarity that is sought for.
3. Minor point: you lose tracking of a branch by entering a non-branch value. Not a biggy because we manually bump to known released versions anyway, but our dep checker [wouldn't work anymore](https://github.com/arbor-sim/arbor/blob/master/.github/workflows/check-submodules.yml).
4. Some projects work around it by creating a branch for every tag (version), but only some of ours deps do, and I thought it best to stick to a known cumbersome config than an inconsistent one.

thorstenhater:
So, I tested this locally with the explicit versions
1. Checkout _some_ commit. 
2. Run `git submodule update`
3. Observe the correct version coming back.

which is pretty much what I expect. `git submodule sync`
can be used like this to achieve something similar
1. Garble your submodules 
2. `git submodule sync`
3. `git submodule update`

That's enough for me, now that I know.

However, having _one_ source for versions in machine
readable (and actable) form would be better.

PR #2057, opened 2022-12-06
consistent mechanism ids

boeschf:
Mechanism ids are assigned by iterating through an unordered map within `fvm_mechanism_data`. This is sub-optimal as the ids depend on implementation defined behaviour. In particular, the ids are not consistent between `libc++` and `libstdc++`. The ids, however, are part of the key used to produce unique random values for solving SDEs, and thus, we get different random values with different OSs/different standard libraries even when using the same global seed value. In order to achieve reproducible results, I propose to change the container in question to a ordered `std::map`.


thorstenhater:
Yes, please. But, we are using a map of unsigned to mechanism which is built linearly. How about using a vector?
Then, we should have this layout:
```cxx
std::vector<mechanism> mechanisms_;
std::vector<std::size_t> revpot_mechs_;
// more index maps for different kinds.
```
That should enable patterns like 
- initialising _all_ mechanisms
- compute current for a single kind
- etc

What do you think.

boeschf:
that sounds interesting. I'll experiment and may extend the PR.

brenthuisman (approved):

PR #2058, opened 2022-12-07
Set Python to known version for all CI workflows

brenthuisman:
See the same jobs fail, which turn out to execute using different Python versions:
- https://github.com/arbor-sim/arbor/actions/runs/3618204857/jobs/6097794385
- https://github.com/arbor-sim/arbor/actions/runs/3630819640/jobs/6141257441

### DIFF
--- a/.github/workflows/spack.yml
+++ b/.github/workflows/spack.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04, ubuntu-22.04]
+        os: [ubuntu-20.04]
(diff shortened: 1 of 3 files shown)


thorstenhater (dismissed):
This is ok for me, but you add two workflow runs. CI is still taking time and costing _someone_. Do we really need that bit extra?

brenthuisman:
Update: Github have a bit of GCC 12 missing in their image, thanks @schmitts: https://github.com/actions/runner-images/issues/6724

Meanwhile we can stick to 20.04 for Spack workflows.

schmitts (approved):
LGTM

llandsmeer (approved):

PR #2062, opened 2022-12-12
GitHub action to build an arbor wheel for pyodide and emscripten

llandsmeer:
This workflow creates a .whl file given a specified Pyodide version in the workflow file

Somewhere in the latest few arbor commits some kind of regression got introduced that prevent the .so file from actually getting imported, so until that is fixed this workflow needs to be run manually. The build procedure should be correct however.


.github/workflows/build-pyodide-package.yml
```diff
@@ -0,0 +1,108 @@
...
+        cd arbor_src/build_arborlib_em
+        cmake .. \
+            --toolchain "${ABS_TOOLCHAIN}" \
+            -DCMAKE_INSTALL_PREFIX=$(realpath prefix) \
+            -DARB_USE_BUNDLED_LIBS=ON \
+            -DARB_MODCC="${ABS_MODCC}" \
+            -DARB_ARCH=generic
+        make -j 8
+        make install
+    - name: build-arbor-python-so-emscripten
```
brenthuisman, line 58: So this is a second build of libarbor? Why is that necessary?

.github/workflows/build-pyodide-package.yml
```diff
@@ -0,0 +1,108 @@
...
+        PY_VERSION_FLAT=${{ env.PYMAJOR }}${{ env.PYMINOR }}
+        ARBOR_PYLIB_OUTPUT=arbor/_arbor.cpython-${PY_VERSION_FLAT}-wasm32-emscripten.so
+        ABS_ARBOR_PYLIB_OUTPUT=$(pwd)/$ARBOR_PYLIB_OUTPUT
+        source emsdk-main/emsdk_env.sh
+        mkdir -p arbor arbor.dist-info
+        mkdir -p arbor_src/build_python_lib_em
+        cd arbor_src/build_python_lib_em
+        time em++ \
+            ../python/*.cpp \
+            -I ../ext/pybind11/include/ \
```
brenthuisman, line 71: Are all these flags required?

.github/workflows/build-pyodide-package.yml
```diff
@@ -0,0 +1,108 @@
...
+            -s FORCE_FILESYSTEM=1 \
+            -s LINKABLE=1 \
+            -s USE_SDL=0 \
+            -s SIDE_MODULE=1 \
+            -s MODULARIZE=1 \
+            -s WASM_BIGINT=1 \
+            '-DARB_PREFIX="/dev/null"' \
+            '-DARB_BINARY="/dev/null"' \
+            '-DARB_LIB="/dev/null"' \
+            '-DARB_CXX_COMPILER="/dev/null"' \
```
brenthuisman, line 88: Same question. I guess these are ignored, but then I'd like to know why ;)

.github/workflows/build-pyodide-package.yml
```diff
@@ -0,0 +1,108 @@
...
+            '-DARB_DATA="/dev/null"' \
+            -g0 \
+            -O2
+    - name: build-python-wheel-emscripten
+      run: |
+        cp -v arbor_src/VERSION arbor
+        echo IyBUaGUgUHl0aG9uIHdyYXBwZXIgZ2VuZXJhdGVkIHVzaW5nIHB5YmluZDExIGlzIGEgY29tcGlsZWQgZHluYW1pYyBsaWJyYXJ5LAojIHdpdGggYSBuYW1lIGxpa2UgX2FyYm9yLmNweXRob24tMzgteDg2XzY0LWxpbnV4LWdudS5zbwojCiMgVGhlIGxpYnJhcnkgd2lsbCBiZSBpbnN0YWxsZWQgaW4gdGhlIHNhbWUgcGF0aCBhcyB0aGlzIGZpbGUsIHdoaWNoIHdpbGwgaW1wb3J0cwojIHRoZSBjb21waWxlZCBwYXJ0IG9mIHRoZSB3cmFwcGVyIGZyb20gdGhlIF9hcmJvciBuYW1lc3BhY2UuCgpmcm9tIC5fYXJib3IgaW1wb3J0ICogICMgbm9xYTogRjQwMwoKCiMgUGFyc2UgVkVSU0lPTiBmaWxlIGZvciB0aGUgQXJib3IgdmVyc2lvbiBzdHJpbmcuCmRlZiBnZXRfdmVyc2lvbigpOgogICAgaW1wb3J0IG9zCgogICAgaGVyZSA9IG9zLnBhdGguYWJzcGF0aChvcy5wYXRoLmRpcm5hbWUoX19maWxlX18pKQogICAgd2l0aCBvcGVuKG9zLnBhdGguam9pbihoZXJlLCAiVkVSU0lPTiIpKSBhcyB2ZXJzaW9uX2ZpbGU6CiAgICAgICAgcmV0dXJuIHZlcnNpb25fZmlsZS5yZWFkKCkuc3RyaXAoKQoKCl9fdmVyc2lvbl9fID0gZ2V0X3ZlcnNpb24oKQpfX2NvbmZpZ19fID0gY29uZmlnKCkgICMgbm9xYTpGNDA1CgojIFJlbW92ZSBnZXRfdmVyc2lvbiBmcm9tIGFyYm9yIG1vZHVsZS4KZGVsIGdldF92ZXJzaW9uCg== \
```
brenthuisman, line 95: Clever, but hard to verify and understand. I would either accept these files as files in our git repo, or maybe a separate repo (for now), so that reproducing the wheel does not require the workflow file. We could then hook deployment to a website (eg `arbor-sim.org/online` to that repo) as well. What do you think?

.github/workflows/build-pyodide-package.yml
```diff
@@ -0,0 +1,108 @@
...
+      run: |
+        mkdir -p arbor_src/build_modcc_host
+        cd arbor_src/build_modcc_host
+        cmake .. -DARB_USE_BUNDLED_LIBS=ON
+        make modcc -j 8
+    - name: build-libarbor-emscripten
+      run: |
+        ABS_TOOLCHAIN=$(realpath emsdk-main/upstream/emscripten/cmake/Modules/Platform/Emscripten.cmake)
+        ABS_MODCC=$(realpath arbor_src/build_modcc_host/bin/modcc)
+        sed -i.bak -E \
```
brenthuisman, line 45: Why is this needed?

.github/workflows/build-pyodide-package.yml
```diff
@@ -0,0 +1,108 @@
...
+    - name: setup-emsdk
+      run: |
+        ./emsdk-main/emsdk install ${{env.PYODIDE_EMSCRIPTEN_VERSION}}
+        ./emsdk-main/emsdk activate ${{env.PYODIDE_EMSCRIPTEN_VERSION}}
+    - name: build-modcc-host
+      run: |
+        mkdir -p arbor_src/build_modcc_host
+        cd arbor_src/build_modcc_host
+        cmake .. -DARB_USE_BUNDLED_LIBS=ON
+        make modcc -j 8
```
brenthuisman, line 40: Github runners only give 2 cores afaik.

.github/workflows/build-pyodide-package.yml
```diff
@@ -0,0 +1,108 @@
+name: Build pyodide arbor
+
+on:
```
brenthuisman, line 3: Would a weekly build be an idea?

brenthuisman (changes requested):
Great work, thanks! A few remarks. Also: if a bug was introduced in the main tree, we should fix it. Can you describe the bug?

.github/workflows/build-pyodide-package.yml
llandsmeer: Good one, comments comments to explain would be a good idea. The point is that these are all required to be given to the compiler or the compile will fail - yet these make no sense here as we're not installing in a prefix or a site-packages directory

.github/workflows/build-pyodide-package.yml
brenthuisman: Sure!

.github/workflows/build-pyodide-package.yml
llandsmeer: Sadly yes. Usually these are produced by CMake or the actual build tool, but I could not convince CMake to produce an pyodide/emscripten .so file. About half the flags are required by arbor and make sure it find the right python header files, the other half is needed to make the .so file `import`-able by pyodide

.github/workflows/build-pyodide-package.yml
llandsmeer: We're not building libarbor in the third step - we are linking the libarbor* files together with the pybind binding files to a single .so file that can be loaded by pyodide. Maybe could use some more explanation in the comments

.github/workflows/build-pyodide-package.yml
llandsmeer: Because Random123 refuses to build on any non-standard platform sadly

This is the code that we delete:

```c++
#if !defined(__x86_64__) && !defined(__i386__) && !defined(__powerpc__) && !defined(__arm__) && !defined(__aarch64__) && !defined(__s390x__)
#  error "This code has only been tested on x86, powerpc and a few arm platforms."
#include <including_a_nonexistent_file_will_stop_some_compilers_from_continuing_with_a_hopeless_task>
{ /* maybe an unbalanced brace will terminate the compilation */
 /* Feel free to try the Random123 library on other architectures by changing
 the conditions that reach this error, but you should consider it a
 porting exercise and expect to encounter bugs and deficiencies.
 Please let the authors know of any successes (or failures). */
#endif
```

So technically we're doing something very wrong, but the code seems to work fine for now.. Maybe we should test the random numerics on wasm :)

brenthuisman:
Building and deploying these wheels will be moved out into a separate repo: https://github.com/arbor-sim/playground

I'll let @llandsmeer do the honors of closing the PR :)

PR #2065, opened 2022-12-14
Allow __ in profiler names.

thorstenhater:
Allow `__` in profiler names, now that we split on `:`.


brenthuisman (approved):
Can't complain about anything, but what might have been the reason for this?

thorstenhater:
For not having `__`? C++ functions may not contain these.

PR #2067, opened 2022-12-14
🐙 Optimise PPACK

thorstenhater:
- pointers in `PPACK_IFACE_BLOCK` will be `__restrict__`
  - this is expected to produce better binary code for the _scalar_ case
  - for GPU we expect this to better utilise data caches as per CUDA docs 
- `ved_di` is not used anymore, gone and ABI bumped
- GPU code never used the indexing structs, so remove from `PPACK_IFACE_BLOCK`

NB. This builds on earlier work to test performance metrics; to be deleted and cherrypicked.


llandsmeer:
Can confirm this works on CPU SIMD, with a 5% speedup

boeschf:
could you merge master? easier to see what's new that way

thorstenhater:
Sure!

modcc/printer/cprinter.cpp
```diff
@@ -249,24 +249,30 @@ ARB_LIBMODCC_API std::string emit_cpp_source(const Module& module_, const printe
...
+                                   "[[maybe_unused]] arb_size_type {0}index_constraints_n_constant                 = pp->index_constraints.n_constant;\\\n"
+                                   "[[maybe_unused]] arb_size_type {0}index_constraints_n_independent              = pp->index_constraints.n_independent;\\\n"
+                                   "[[maybe_unused]] arb_size_type {0}index_constraints_n_none                     = pp->index_constraints.n_none;\\\n"
+                                   "[[maybe_unused]] arb_index_type* __restrict__ {0}index_constraints_contiguous  = pp->index_constraints.contiguous;\\\n"
+                                   "[[maybe_unused]] arb_index_type* __restrict__ {0}index_constraints_constant    = pp->index_constraints.constant;\\\n"
+                                   "[[maybe_unused]] arb_index_type* __restrict__ {0}index_constraints_independent = pp->index_constraints.independent;\\\n"
+                                   "[[maybe_unused]] arb_index_type* __restrict__ {0}index_constraints_none        = pp->index_constraints.none;\\\n"),
```
boeschf, line 275: are these index constraints and their representation documented somewhere?

boeschf:
I like the explicit types instead of `auto` in the generated code.

modcc/printer/cprinter.cpp
thorstenhater: It's an internal indicator to basically multi-version the SIMDised loops.
- contiguous: this bundle of `N` indices is `{k, k+1, ..., k+N-1}`
- constant: `{k, k, ..., k}`
- independent and none: seem to be the same, we know nothing about the set

These are then used to drive the ld/st operations from/to SIMD types, so that the 
actual loop can operator on full-width SIMD types.

I do not know whether this has been documented, it was already here when I came ;)

thorstenhater:
Ideally one would hope for a warning about implicit conversions as well.

boeschf (approved):

Issue #2068, opened 2022-12-15
Pressing Ctrl-C during Arbor simulation in a Python process will not stop the process


llandsmeer:
To reproduce:

 - start a long running arbor simulation run()
 - press Ctrl-C
 - nothing will happen, until the simulation is done

To fix, we should call Py_CheckSignals() in the mainthread and then propagate the error in pybind, which will then result in a KeyboardInterrupt error

 - [Python documentation on CheckSignals](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_CheckSignals)
 - [PyBind11 documentation on propagating exceptions](https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html#handling-exceptions-from-python-in-c)

Eg

```c++
if (PyErr_CheckSignals() != 0)
    throw py::error_already_set();
```

In practice, this might be a bit more complicated. We could for example replace the current

```c++
arb::time_type run(arb::time_type tfinal, arb::time_type dt) {
    return sim_->run(tfinal, dt);
}
```

with

```c++
arb::time_type run(arb::time_type tfinal, arb::time_type dt) {
    arb::time_type tepoch = 0;
    while (true) {
        if (PyErr_CheckSignals() != 0) {
            throw py::error_already_set();
        }
        tepoch += dt * 100;
        arb::time_type res = sim_->run(std::min(tepoch, tfinal), dt);
        if (tepoch >= tfinal) {
            return res;
        }
    }
}
```

ie checking every 100 timesteps for a signal. 

Alternatively, we could check every timestep or decide we don't want this.

I'll happily write the PR but I think we should first decide how we go about this.

schmitts:
What about sending SIGQUIT with Ctrl-\ instead of SIGINT with Ctrl-C?

llandsmeer:
Ctrl-\ doesn't seem to work, but indeed usually killing the process works. Still very annoying in a jupyter notebook, which will then also delete all stored variables

llandsmeer:
Copying the note from #2070:

> So we decided that the ^C check needs to happen in the epoch callback. Problem is that the epoch callback only gets called in simulations with spikes (also affecting the progressbar.. ). So that doesnt solve the problem. No idea how to proceed, maybe close for now

So first #2089 needs to be solved before tackling this issue

thorstenhater:
This an artifact of our binding from Python to C++ and cannot be fixed, even with a custom
[...]

PR #2070, opened 2022-12-15
Catch ^C in run()

llandsmeer:
See #2068 - Allow simulation to stop on Ctrl-C press in python

Implements https://pybind11.readthedocs.io/en/stable/faq.html#how-can-i-properly-handle-ctrl-c-in-long-running-functions

We now cycle through a lot of GIL locks. At 1000 timesteps/checks I couldn't see any runtime impact (the check if acquire GIL, check a memory location, release GIL)

### DIFF
--- a/python/simulation.cpp
+++ b/python/simulation.cpp
@@ -97,7 +97,21 @@ public:
     }
 
     arb::time_type run(arb::time_type tfinal, arb::time_type dt) {
-        return sim_->run(tfinal, dt);
+        arb::time_type tepoch = 0;
+        while (true) {
+            if (PyErr_CheckSignals() != 0) {
+                throw py::error_already_set();
+            }
+            tepoch += dt * 1000;
+            arb::time_type res;
+            {
+                pybind11::gil_scoped_release guard;
+                res = sim_->run(std::min(tepoch, tfinal), dt);
+            }
+            if (tepoch >= tfinal) {
+                return res;
+            }
+        }
     }
 
     void set_binning_policy(arb::binning_kind policy, arb::time_type bin_interval) {
@@ -239,7 +253,6 @@ void register_simulation(pybind11::module& m, pyarb_global_ptr global_ptr) {
              pybind11::call_guard<pybind11::gil_scoped_release>(),
              "Clearing spike and sample information. restoring memory")
         .def("run", &simulation_shim::run,
-            pybind11::call_guard<pybind11::gil_scoped_release>(),
             "Run the simulation from current simulation time to tfinal [ms], with maximum time step size dt [ms].",
             "tfinal"_a, "dt"_a=0.025)
         .def("set_binning_policy", &simulation_shim::set_binning_policy,


llandsmeer:
Please check the implementation - maybe you don't like that I add up a lot of floating numbers

brenthuisman (commented):
Nice! I don't think precision is required for `tepoch`, so adding many tiny floats isn't a problem.

Only question I have is about impact on large simulations. What simulation did you test on?

llandsmeer:
The floating point addition thing will only be problematic if `dt * 1000` is smaller than the accuracy of float64 at tfinal. Eg if you simulate for `1e15` seconds at a dt of `0.001`. You might have different problems then.

In what way do you think this could be impacting large simulations? Essentially we check for ^C every 1000 timesteps, so as long as the cost of the check is cheap in comparison to a single timestep (which is always the case?) then we should be good. Indeed I only tested this up to ~100 cells for now.

brenthuisman:
It wouldn't be the first time an infrequent check turns out to be costly at scale. I would prefer it if you can confirm I worry for nothing with a larger simulation, is that possible?

thorstenhater:
Could you benchmark this again, but 
1. Make the actual workload trivial: two LIF cells, one simple synapse with delay `d`. The epoch is now `d`.
2. Run the simulation with a small `dt` for a long time `T`.
3. Vary ` dt <= d <= T` and graph the simulation wallclock time (1 Thread!) over `d`

python/simulation.cpp
```diff
@@ -97,7 +97,21 @@ class simulation_shim {
     }
 
     arb::time_type run(arb::time_type tfinal, arb::time_type dt) {
-        return sim_->run(tfinal, dt);
+        arb::time_type tepoch = 0;
```
thorstenhater, line 100: `epoch` is an internal moniker for the timespan `t_min_delay/2`, for the sake of clarity, could you find another name?

python/simulation.cpp
```diff
@@ -97,7 +97,21 @@ class simulation_shim {
...
+        while (true) {
+            if (PyErr_CheckSignals() != 0) {
+                throw py::error_already_set();
+            }
+            tepoch += dt * 1000;
```
thorstenhater, line 105: Depending on the model 1000 $\Delta t$ can already be an appreciable delay. Not sure we should use a magic number here.

thorstenhater (changes requested):
Before you delve in addressing my whole two comments, take a moment to consider whether you can leverage the `epoch_callback` functionality here. Also helps with not having diverging Python/C++ code bases. That said it probably requires having the callback a `continue=yes/no` value **and** layering these callbacks.

python/simulation.cpp
llandsmeer: Alternatively, we (different options):

 - call the check function each dt
 - check the elapsed time every 1dt and call the check function if that's more than 100ms
 - record the time it takes to execute the first dt, extrapolate how many steps it would take to execute (>) 100ms, and take that many steps in between.

llandsmeer:
The callback is a good idea, much cleaner.. the layering with `        sim_->set_epoch_callback(arb::epoch_progress_bar());` is going to be interesting

llandsmeer:
So we decided that the ^C check needs to happen in the epoch callback. Problem is that the epoch callback only gets called in simulations with spikes (also affecting the progressbar..). So that doesnt solve the problem. No idea how to proceed, maybe close for now

brenthuisman:
OK, I'll close. Can you copy that comment to the issue, so it's not overlooked by the next attempt? Thanks!

PR #2072, opened 2022-12-20
v0.8.1

brenthuisman:


CHANGELOG.md
```diff
@@ -1,3 +1,29 @@
+# v0.8.1
+
+** 2022 12 20 **
+
+A 🎄 holiday release! Not much has changes in a month, but we'd like to share it all the same. Notably, the [Arbor GUI](https://github.com/arbor-sim/gui/) [is co-released](https://github.com/arbor-sim/gui/releases/tag/v0.8) as of Arbor v0.8, and v0.8.1 will be no different.
```
boeschf, line 5:
```suggestion
A 🎄 holiday release! Not much has changed in a month, but we'd like to share it all the same. Notably, the [Arbor GUI](https://github.com/arbor-sim/gui/) [is co-released](https://github.com/arbor-sim/gui/releases/tag/v0.8) as of Arbor v0.8, and v0.8.1 will be no different.
```

CHANGELOG.md
```diff
@@ -1,3 +1,29 @@
...
+
+## Bug fixed
+
+- Fix ornstein_uhlenbeck example on gpu (#2039)
+- Setting ARB_MODCC was broken and nunfunctional. Fixed. (#2029)
+- The `--cxx` flag in `arbor-build-catalogue` is now properly used; falls back to `c++`. (#2051)
+
+## Full commit log
+
+...
```
boeschf, line 25: no commit log?

python/example/example_requirements.txt
```diff
@@ -1,4 +1,5 @@
 # pip requirements file
+numpy!=1.24.0 # See https://github.com/numpy/numpy/issues/22826
```
boeschf, line 2: should we add this restriction also to the spack package?

CHANGELOG.md
brenthuisman: They are generated when I push the tag, so they can't be yet. I add them in the post-release PR. I'll add a note about that.

boeschf: ok, thanks

python/example/example_requirements.txt
brenthuisman: The restriction is only required for the examples (a single one in fact), which isn't a target in the spack package. So I don't think it's necessary.

boeschf: still, numpy is a dependency in the spack package when building with python, so would be nice to have imo

brenthuisman: This dependency is because Arbor exposes some data as numpy arrays (spike array). This won't trigger the big. The user would need to write code themselves that runs into it.

boeschf: ok let's leave it then.

boeschf (approved):
lgtm, thank you!

Issue #1421, opened 2021-03-09
Reporting or calculating model memory usage


halfflat:
It is currently unclear how much memory one could expect any given model to require, and this becomes especially pertinent when the model is gap-junction based, requiring that all such cells lie in the same single cell group.

This information could be provided by an offline calculator/documentation reflecting the known memory costs based on (for cable cells) number of CVs, mechanism state, ionic extent etc., or provided through the simulator interface. If the latter, it could constitute part of the information returned by an implementation of #940.

brenthuisman:
What seems ergonomic to me is to be able to query a cell description for the estimate. What else is needed is the mech (catalog) and cv policy, correct?

halfflat:
We'd need to know which cell group will be handling it, and with a view to run-time random variables, we'd also need the `mprovider` at hand to perform any estimate.

That said though, performing this query outside of the context of the `simulation` object and realized cell groups is a bit thorny for at least a couple of reasons:
* It ties the estimate to the implementation details within a cell group, and estimation is likely to be almost as hard as just building the cell group around the cell regardless.
* Estimates on a cell-by-cell basis will miss any overhead (or optimizations) that arise at the level of a cell group, where multiple cells are bundled together.

For these reasons, I think the simulator object based approach would be the most faithful and also the least disruptive. A simulation comprising one or more cell groups can be set up, the resource usage queried, and then the simulation can be discarded.

brenthuisman:
It's an estimate, so it does not need to be a precise number. Do you think the per cell desc estimate can be off more than say a factor of 2 from actual memory consumption? Or is it really going to be wildly dependent on hardware context?

halfflat:
It's hard to estimate how bad the estimate will be :) The more work we do, the closer we will be, but also the closer we will be to making something that looks like a cell group and asking it.

brenthuisman:
A rule of thumb times two can go a long way to being useful, without requiring that much effort ;)

PR #2081, opened 2023-01-23
Checkpoint/Resume

### LINKED ISSUE #560: Checkpoint support
We have no checkpointing support, but it sure would be good to have.

Nothing concrete here on this issue: open to collecting comments regarding use cases, design ideas, etc.

thorstenhater:
# What is this?

Implement (partial!) dumping of simulation state  through a serialization object that can be use to reset the simulation
to an earlier state. The produced snapshot includes just enough information to perform that reset,
not all bits.

For a successful resume, the following constraints apply:
- `context` and `domain_decomposition` must be identical to those used during `serialize`
- the `simulation` we call `resume` on must be 
  - valid
  - constructed using the same recipe

## Not included
- pointer data that forms internal references.
- internal states of generators, samplers, and probes. Might be needed to change in between snapshots.

# ⚠️ Breaking changes

I needed to bump CUDA language level to 17 (up from 14). That'll require CUDA 11+; released in 2020.

# TODO
- [x] GPU.
- [x] ~At least partially check the constraints~ It's largely impossible. See docs. Suggestions welcome, though.
- [x] Python bindings.
- [X] Docs

# Issues

Closes #560


arbor/include/arbor/schedule.hpp
```diff
@@ -230,6 +152,130 @@ class poisson_schedule_impl {
...
+        virtual void reset() override { wrapped.reset(); }
+        virtual iface_ptr clone() override { return std::make_unique<wrap<Impl>>(wrapped); }
+        virtual void serialize(serializer& s, const std::string& k) const override { ::arb::serialize(s, k, wrapped); }
+        virtual void deserialize(serializer& s, const std::string& k) override { ::arb::deserialize(s, k, wrapped); }
+
+        Impl wrapped;
+    };
+};
+
+// Costructors
```
brenthuisman, line 257:
```suggestion
// Constructors
```

brenthuisman (changes requested):
Very nice! Some questions and comments below:

arbor/backends/multicore/shared_state.hpp
```diff
@@ -123,22 +135,53 @@ struct ARB_ARBOR_API istim_state {
...
-        std::vector<arb_value_type*> parameters_;
-        std::vector<arb_value_type*> state_vars_;
-        std::vector<arb_ion_state>   ion_states_;
-
-        std::array<std::vector<arb_value_type*>, cbprng::cache_size()> random_numbers_;
-        std::vector<arb_size_type> gid_;
-        std::vector<arb_size_type> idx_;
-        cbprng::counter_type random_number_update_counter_ = 0u;
-    };
+struct mech_storage {
```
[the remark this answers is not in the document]
thorstenhater: We do have this once per backend with partially differing implemetations, eg what's a vector here is an GPU array over there.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,243 @@
...
+
+    // Construct a simulation.
+    auto model = recipe{};
+    auto ctx = arb::make_context();
+    auto ddc = arb::partition_load_balance(model, ctx);
+    auto simulation = arb::simulation{model, ddc, ctx};
+
+    // Run forwards, snapshot state.
+    simulation.run(T, dt);
+    serialize(serializer, simulation);
```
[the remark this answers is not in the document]
thorstenhater: It's going to be overwritten, similar to a map, so
```
std::map<int, int> m;
m[42] = 23;
m[42] = 0;
```
will result in `{{42, 0}}`.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,243 @@
...
+
+In Python, currently only support for (de)serialization from/to JSON strings is offered.
+
+  .. code:: python
+
+    import arbor as A
+
+    rec = my_recipe()
+    sim = A.simulation(rec)
+    jsn = sim.serialize()
```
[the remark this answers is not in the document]
thorstenhater: So, the idea that we have an open design and we can new serializers to _any_ type, even if we didn't write it in the first place. Example
```cxx
std::map<...> m = {...};
// How to define anything that allows us this?
m.serialize(...);
// Only option:
void serialize(const std::map<...>&) {...}
serialize(m); // OK
```
as Python doesn't have overloading, is more to the OO side of the spectrum
AND allows monkeypatching, attaching methods to objects feels more idiomatic.
Alternative:
```py
def serialize(o: Object):
   if isinstance(o, ...):
      ...
```
but this is toxic.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,243 @@
...
+    sim.deserialize(jsn)
+
+
+Intermediate: Writing your own Storage Engine in C++
+----------------------------------------------------
+
+A storage engine is used to construct the serializer and is responsible for
```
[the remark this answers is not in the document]
thorstenhater: ?

python/simulation.cpp
```diff
@@ -232,6 +249,13 @@ void register_simulation(pybind11::module& m, pyarb_global_ptr global_ptr) {
              "Rebuild the connection table from recipe::connections_on and the event"
              "generators based on recipe::event_generators.",
              "recipe"_a)
+        .def("deserialize", &simulation_shim::deserialize,
```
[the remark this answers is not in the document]
thorstenhater: Cute indeed, but no clue if it'll work, since this is currently legal (and needed)
```cxx
std::unique_ptr<double> pd = std::make_unique(42.0);
serialize(sd, pd);
*pd = 23;
deserialize(sd, pd);
```

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,243 @@
...
+   all ranks.
+3. The hardware configuration must be identical: SIMD, GPU, and Scalar CPU code
+   have different binary layouts which makes serializing across hardware settings
+   invalid.
+
+Thus, the recommended approach is to treat a simulation 'script' (``.cxx`` /
+``.py``), the parallel-runtime parameters (think ``mpirun``) and the associated
+snapshots as a single unit.
+
+In Python, currently only support for (de)serialization from/to JSON strings is offered.
```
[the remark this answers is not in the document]
thorstenhater: Fine, it started out as a more general idea, but this is what we have for now. Maybe it's more accurately described as SerDes MVP ;) 
I see some possibilities here: 
1. We use this as is and if there's demand, we'll extend it.
  - I need to add a format meta data block and a version
  - Also, I decided to lock down the missing bits, ie you won't be able to swap MPI ranks, even if equivalent.
3. We try to go all the way now, but needs some more work.

doc/fileformat/serdes.rst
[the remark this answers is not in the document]
brenthuisman: Whoops, C++ only. You can't provide your own storage engine in Python.

python/simulation.cpp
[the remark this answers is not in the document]
brenthuisman: But that's not more of an issue when we call it 'Pickle', is it?

(Could even pass the json to ZipFile)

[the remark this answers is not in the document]
thorstenhater: Pickle is a known, well-defined term with its own protocol. I don't think we can overload that name.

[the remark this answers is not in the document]
brenthuisman: From the linked docs, I understood this is true for Python objects, but not for C++-Pybind ones, and that that's the reason you can implement these methods at will.

Looking at the [Python pickle docs](https://docs.python.org/3/library/pickle.html), the warning is clear.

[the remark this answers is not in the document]
thorstenhater: I am not sure which warning you're referring to, but Pickle defines a protocol ie
https://docs.python.org/3/library/pickle.html#data-stream-format
that we do not adhere to. My comment about pointers was referring to the fact
that this PR relies the structure being in place already, so a more accurate counter
example would be this
```cxx
int* pi = nulltpr;
deserialize(sd, pid); // BOOM, tries to deref pi and assign to its pointee
pi = new int;
deserialize(sd, pid); // Fine, we now have a value to write to.
```

doc/fileformat/serdes.rst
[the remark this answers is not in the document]
brenthuisman: Let's go with 1. But, I'm still in favor of calling it snapshotting, I think that better describes the use-case.

doc/fileformat/serdes.rst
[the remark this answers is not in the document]
thorstenhater: Thus venomous?

doc/fileformat/serdes.rst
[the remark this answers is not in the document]
thorstenhater: True.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,243 @@
...
+but instead a reset the simulation state to this point in time. The saved data
+is not a complete image of the state, but strives to be a minimal subset from
+which to restore.
+
+Usage
+-----
+
+SerDes in itself is independent of the file format used, it allows for multiple
+backends. The general usage is show in the examples below; we assume familiarity
+with the C++ and Python recipe interfaces:
```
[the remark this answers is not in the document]
thorstenhater: This is just for the _recipe_, not for checkpointing.

arbor/backends/multicore/shared_state.hpp
```diff
@@ -191,10 +234,10 @@ struct ARB_ARBOR_API shared_state: shared_state_base<shared_state, array, ion_st
                  const std::vector<arb_value_type>& init_membrane_potential,
                  const std::vector<arb_value_type>& temperature_K,
                  const std::vector<arb_value_type>& diam,
-                 const std::vector<arb_index_type>& src_to_spike,
-                 const fvm_detector_info& detector_info,
+                 const std::vector<arb_index_type>& src_to_spike_,
+                 const fvm_detector_info& detector,
                  unsigned align,
-                 arb_seed_type cbprng_seed_ = 0u);
+                 arb_seed_type cbprng_seed_=0);
```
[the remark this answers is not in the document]
thorstenhater: Yes, but that was unintentional.

arbor/simulation.cpp
```diff
@@ -121,6 +121,58 @@ class simulation_state {
     label_resolution_map source_resolution_map_;
     label_resolution_map target_resolution_map_;
 
+    // We do not serialize:
```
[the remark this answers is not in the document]
thorstenhater: In the docs there is the general advice/discussion of this:
- serialize stores only what is needed for restore
- not for data extraction
- see also implementation advice.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,243 @@
+.. _formatserdes:
+
+Serialization/Deserialization (SerDes)
+======================================
+
+While not a fileformat on its own SerDes allows users to dump a snapshot of the
+simulation state for a later time. This is **not** intended for data extraction,
+but instead a reset the simulation state to this point in time. The saved data
+is not a complete image of the state, but strives to be a minimal subset from
```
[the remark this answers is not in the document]
thorstenhater: Added more detail.

arbor/simulation.cpp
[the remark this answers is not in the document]
thorstenhater: Extended said section.

arbor/backends/multicore/shared_state.hpp
```diff
@@ -191,10 +234,10 @@ struct ARB_ARBOR_API shared_state: shared_state_base<shared_state, array, ion_st
                  const std::vector<arb_value_type>& init_membrane_potential,
                  const std::vector<arb_value_type>& temperature_K,
                  const std::vector<arb_value_type>& diam,
-                 const std::vector<arb_index_type>& src_to_spike,
-                 const fvm_detector_info& detector_info,
+                 const std::vector<arb_index_type>& src_to_spike_,
+                 const fvm_detector_info& detector,
```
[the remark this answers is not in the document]
thorstenhater: `*_info` is so darn generic that you might as well leave it out ;) That said, this item isn't a detector on its own, so yeah.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,243 @@
+.. _formatserdes:
+
+Serialization/Deserialization (SerDes)
+======================================
+
+While not a fileformat on its own SerDes allows users to dump a snapshot of the
+simulation state for a later time. This is **not** intended for data extraction,
```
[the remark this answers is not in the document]
thorstenhater: reformulated.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,243 @@
+.. _formatserdes:
+
+Serialization/Deserialization (SerDes)
```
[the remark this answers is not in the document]
thorstenhater: 1. Ok
2. Renamed.
3. No, this is intractable. Depends on mechs in use, backend, morphology. Feel free to check out https://github.com/arbor-sim/arbor/issues/1421

doc/fileformat/index.rst
```diff
@@ -20,3 +20,4 @@ File formats
    neuroml
    nmodl
    swc
+   serdes
```
brenthuisman, line 23: An additional entry must be added to `doc/index.rst`.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,262 @@
...
+          variable out-of-band, we strongly advise against it.
+
+Usage
+-----
+
+Checkpoints are independent of the storage engine used, it allows for multiple
+backends. The general usage is show in the examples below; we assume familiarity
+with the C++ and Python recipe interfaces:
+
+  .. code:: c++
```
brenthuisman, line 40: The extra indent before `.. code::` creates a blockquote. Dedent pls. Same for all following code blocks.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,262 @@
...
+=============
+
+While not a fileformat on its own, checkpoints allow users to dump a snapshot of
+the simulation state for a later time. This is **not** intended for data
+extraction (see :ref:`probesample` for this functionality), but instead a reset
+the simulation state to this point in time. The saved data is not a complete
+image of the state, but strives to be a minimal subset from which to restore
+(see :ref:`below <impl guide>`).
+
+Checkpoints are currently _not_ portable; i.e. the following constraints apply:
```
brenthuisman, line 13:
```suggestion
Checkpoints are currently **not** portable; i.e. the following constraints apply:
```

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,262 @@
...
+
+Note that we left the definition of ``io`` open, as ``serializer`` uses it
+through a well-defined interface (see next section). Thus, one can simply add
+new implementations. Arbor currently ships with ``arborio::json_serdes`` that
+produces JSON output, more might be added over time.
+
+In Python, currently only support for (de)serialization from/to JSON strings is
+offered.
+
+  .. code:: python
```
brenthuisman, line 75: I would move the Python demo up-top of the usage section, from which you can naturally proceed to explain C++ offers the option of custom storage engines.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,262 @@
...
+bracket a write of an array value and announce that the following keyes are to
+be interpreted as integer indices. Analogous for the ``map`` counterparts and
+the associated ``begin_read`` and ``end_read`` methods. Finally, ``next_key`` is
+used during reading of containers to retrieve an optional next key and advanced
+the internal iterator. If empty, the container is exhausted, else the contained
+key can be used to retrieve the associated value. See examples below and the JSON
+interface in ``arborio``.
+
+
+Adding Snapshotting to new Objects (C++ only)
```
brenthuisman, line 133: Swap this section with `Writing your own Storage Engine`. This is more relevant to most users.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,262 @@
...
+             virtual void end_read_map() = 0;
+             virtual void begin_read_array(const key_type&) = 0;
+             virtual void end_read_array() = 0;
+
+             virtual ~interface() = default;
+         };
+
+The ``read`` and ``write`` methods are responsible for inserting and extracting
+the relevant items. The ``begin_write_array`` and ``end_write_array`` methods
+bracket a write of an array value and announce that the following keyes are to
```
brenthuisman, line 124:
```suggestion
bracket a write of an array value and announce that the following keys are to
```

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,262 @@
...
+the internal iterator. If empty, the container is exhausted, else the contained
+key can be used to retrieve the associated value. See examples below and the JSON
+interface in ``arborio``.
+
+
+Adding Snapshotting to new Objects (C++ only)
+---------------------------------------------
+
+This is not available at the Python interface, due to a mismatch in features at
+the level of languages and binings generation.
```
brenthuisman, line 137:
```suggestion
the level of languages and bindings generation.
```

brenthuisman: Any user-written C++ objects in a simulation modified thus, will already be serialized with a call to `A.simulation().serialize()`, right? If so, worth mentioning here.

doc/fileformat/serdes.rst
```diff
@@ -0,0 +1,262 @@
...
+              typename V>
+    void serialize(serializer& ser, const K& k, const std::map<Q, V>& v) {
+        ser.begin_write_map(to_key(k));
+        for (const auto& [q, w]: v) serialize(ser, q, w);
+        ser.end_write_map();
+    }
+
+Reading data is a bit more involved, as writing data might be partial and work
+only in conjunction with proper setup beforehand. Thus, one needs to take care
+when overwriting values. The sotrage is polled for the next key using
```
brenthuisman, line 183:
```suggestion
when overwriting values. The storage is polled for the next key using
```

brenthuisman (changes requested):

doc/fileformat/serdes.rst
thorstenhater: Why favour **not** over _not_ here?

doc/fileformat/serdes.rst
brenthuisman: Why couldn't a user add new serializable objects to a simulation?

doc/fileformat/serdes.rst
thorstenhater: Still, one is bold, the other italic?

doc/fileformat/serdes.rst
thorstenhater: To the `simulation` _type_ inside libarbor?

doc/fileformat/serdes.rst
thorstenhater: > user-written C++ objects in a simulation

I am not sure I understand, but see above?

doc/fileformat/serdes.rst
brenthuisman: Bold is to be consistent with the previous emphasized not.

doc/fileformat/serdes.rst
brenthuisman: Serializing a simulation object is where things begin, right? I assume any object in it, will be serialized if they have implemented `read`/`write` or use of of the `ARB_SERDES_*` macros. Should somebody add a new type like that, and have a recipe using the type, then it'll be included in the serialization of the simulation object, right?

thorstenhater: No, `recipe` is by definition not serializable. We can -- in general -- not serialize a function except dumping its source code / assembly, which we cannot access from inside C++ (not practically, that is). In that sense `recipe`
_is_ its own serialization.

Thus the comment in the docs about 'treating recipe + main + serialize as one bundle'. This and only this is a 
complete representation.

A user cannot add an arbitrary object to a simulation at C++ level. This checkpoint relies on us to identify and
annotate the parts to serialize. See here on what we would need for making such functionality:
https://herbsutter.com/2017/07/26/metaclasses-thoughts-on-generative-c/

Python is a different story, but we'll not scan our object tree to find user data and speculatively dump it for them.

brenthuisman: Right. Discussed over VC.

doc/fileformat/serdes.rst
brenthuisman: Seen above.

doc/fileformat/serdes.rst
thorstenhater: Done

arbor/include/arbor/serdes.hpp
```diff
@@ -0,0 +1,476 @@
...
+    ser.write(to_key(k), static_cast<unsigned long long>(v));
+}
+
+template<typename K>
+ARB_ARBOR_API void serialize(serializer& ser, const K& k, unsigned long v) {
+    ser.write(to_key(k), static_cast<unsigned long long>(v));
+}
+
+template<typename K>
+ARB_ARBOR_API void serialize(serializer& ser, const K& k, const float v) {
```
boeschf, line 208:
```suggestion
ARB_ARBOR_API void serialize(serializer& ser, const K& k, float v) {
```

arbor/include/arbor/serdes.hpp
```diff
@@ -0,0 +1,476 @@
...
+    ser.write(to_key(k), double{v});
+}
+
+template<typename K>
+ARB_ARBOR_API void serialize(serializer& ser, const K& k, const double v) {
```
boeschf, line 213:
```suggestion
ARB_ARBOR_API void serialize(serializer& ser, const K& k, double v) {
```

arbor/include/arbor/serdes.hpp
```diff
@@ -0,0 +1,476 @@
...
+    ser.write(to_key(k), v);
+}
+
+template<typename K>
+ARB_ARBOR_API void serialize(serializer& ser, const K& k, const bool v) {
```
boeschf, line 218:
```suggestion
ARB_ARBOR_API void serialize(serializer& ser, const K& k, bool v) {
```

arbor/include/arbor/serdes.hpp
```diff
@@ -0,0 +1,476 @@
...
+    ser.write(to_key(k), static_cast<long long>(v));
+}
+
+template<typename K>
+ARB_ARBOR_API void serialize(serializer& ser, const K& k, const unsigned long long v) {
```
boeschf, line 223:
```suggestion
ARB_ARBOR_API void serialize(serializer& ser, const K& k, unsigned long long v) {
```

arbor/include/arbor/schedule.hpp
```diff
@@ -230,6 +152,130 @@ class poisson_schedule_impl {
...
+// and `events(t2, t3)` are made without an intervening call to `reset()`,
+// then 0 ≤ _t0_ ≤ _t1_ ≤ _t2_ ≤ _t3_.
+
+// Forward decls ... friend is weird.
+template<typename K>
+void serialize(serializer&, const K&, const explicit_schedule_impl&);
+template<typename K>
+void serialize(serializer&, const K&, const regular_schedule_impl&);
+template<typename K>
+void serialize(serializer&, const K&, const empty_schedule&);
```
boeschf, line 167: needed?

arbor/include/arbor/schedule.hpp
```diff
@@ -230,6 +152,130 @@ class poisson_schedule_impl {
...
+    ARB_SERDES_WRITE(tstop_);
+    ser.end_read_map();
+}
+
+template<typename K>
+void deserialize(serializer&, const K&, explicit_schedule_impl&);
+template<typename K>
+void deserialize(serializer&, const K&, regular_schedule_impl&);
+template<typename K>
+void deserialize(serializer&, const K&, empty_schedule&);
```
boeschf, line 182: needed?

arbor/memory/wrappers.hpp
```diff
@@ -151,5 +153,63 @@ auto on_gpu(const C& c) -> device_vector<typename C::value_type> {
...
+void serialize(::arb::serializer ser, const K& k, const device_vector<T>& vs) {
+    auto hvs = on_host(vs);
+    ser.begin_write_array(to_key(k));
+    for (int ix = 0; ix < hvs.size(); ++ix) serialize(ser, ix, hvs[ix]);
+    ser.end_write_array();
+}
+
+template<typename K,
+         typename V>
+void serialize(::arb::serializer ser, const K& k, host_vector<V>& hvs) {
```
boeschf, line 175:
```suggestion
void deserialize(::arb::serializer ser, const K& k, host_vector<V>& hvs) {
```

arbor/backends/gpu/shared_state.hpp
```diff
@@ -40,6 +40,17 @@ struct ARB_ARBOR_API ion_state {
     using solver_type = arb::gpu::diffusion_state<arb_value_type, arb_index_type>;
     using solver_ptr  = std::unique_ptr<solver_type>;
 
+    // ARB_SERDES_ENABLE(ion_state,
+                      // node_index_,
+                      // write_Xi_, write_eX_, write_Xo_,
+                      // iX_, eX_, Xi_, Xd_, Xo_, gX_,
+                      // init_Xi_, init_Xo_, init_eX_,
+                      // reset_Xi_, reset_Xo_,
+                      // charge);
```
boeschf, line 49: remove?

arbor/backends/gpu/shared_state.hpp
```diff
@@ -118,7 +129,34 @@ struct ARB_ARBOR_API istim_state {
...
+    std::vector<arb_value_type>  globals_;
+    std::vector<arb_value_type*> parameters_;
+    std::vector<arb_value_type*> state_vars_;
+    std::vector<arb_ion_state>   ion_states_;
+    memory::device_vector<arb_value_type*> parameters_d_;
+    memory::device_vector<arb_value_type*> state_vars_d_;
+    memory::device_vector<arb_ion_state>   ion_states_d_;
+    random_numbers random_numbers_;
+    ARB_SERDES_ENABLE(mech_storage, data_, random_numbers_);
+};
```
boeschf, line 144: this is repeated within the struct `shared_state`, please remove either.

thorstenhater: Yupp, merge played a trick on me!

arbor/backends/gpu/shared_state.hpp
thorstenhater: Yes, there are/were alternatives that allow inspecting a more complete picture of the overall state.

arbor/include/arbor/serdes.hpp
```diff
@@ -0,0 +1,476 @@
...
+#include <vector>
+#include <memory>
+#include <map>
+#include <unordered_map>
+#include <array>
+#include <stdexcept>
+
+#include <arbor/export.hpp>
+
+namespace arb {
```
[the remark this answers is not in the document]
thorstenhater: We'll have to go to the latter, since the rules for inline (= in a struct) `friend` are really weird wrt namespaces.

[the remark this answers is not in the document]
thorstenhater: I think `arb::serialize`/`arb::deserialize` are specific enough, I can see that `arb::to_key` and friends might not be.
So, either I'll put those in `arb::serdes_utils` or call them `arb::to_serdes_key`. Preferences?

arbor/include/arbor/serdes.hpp
```diff
@@ -0,0 +1,476 @@
...
+        return std::string{key};
+    }
+    else {
+        return std::to_string(key);
+    }
+}
+
+template <typename K>
+void from_key(K& key, const key_type& k) {
+    using T = std::decay_t<std::remove_cv_t<std::remove_reference_t<K>>>;
```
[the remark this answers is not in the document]
thorstenhater: See above.

arbor/include/arbor/serdes.hpp
```diff
@@ -0,0 +1,476 @@
...
+    if constexpr (std::is_same_v<std::string, T>) {
+        key = k;
+    }
+    else if constexpr (std::is_same_v<char*, T>) {
+        key = std::string{k};
```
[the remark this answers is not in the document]
thorstenhater: If `K` is a stringy type, it might be a (possibly  `const`) `char[]`, `char*`, or a `std::string`, 
or a reference to any of those.
After `decay` and stripping `&`, `const`, and `volatile` we have `char*` left (or `std::string`), no?

arbor/include/arbor/schedule.hpp
thorstenhater: Nope.

arbor/include/arbor/schedule.hpp
thorstenhater: Nope.

arbor/include/arbor/serdes.hpp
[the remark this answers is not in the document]
thorstenhater: Went with `from/to_serdes_key`.

arbor/include/arbor/serdes.hpp
```diff
@@ -0,0 +1,476 @@
...
+struct ARB_SYMBOL_VISIBLE illegal_key_type: serdes_error {
+    illegal_key_type(): serdes_error{"SerDes keys must be an integral or string-like type."} {}
+};
+
+// Handling keys
+using key_type = std::string;
+
+template <typename K>
+key_type to_key(K&& key) {
+    using T = std::decay_t<std::remove_cv_t<std::remove_reference_t<K>>>;
```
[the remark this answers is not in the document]
boeschf: strange :man_shrugging:

arbor/include/arbor/serdes.hpp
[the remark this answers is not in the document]
boeschf: agreed. My question was about the assignment. The lhs seems to be a `char*`. I was expecting something like this: `key = k.c_str();`. Do I have it backwards?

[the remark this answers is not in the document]
thorstenhater: Ah, good catch! Although, that would be incorrect, since assume the original type is a `char[5]` and we try
to write `foobarbaz` into it. Not only won't this compile, it would be really Bad if it did, which it will for `char*` where it will potentially overwrite unowned memory. BAD. So, I'd conclude we don't allow converting to `char*`/`char[]`, due to security issues. Conversvion _from_ can be more lenient.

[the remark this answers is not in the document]
thorstenhater: Other options will run into problems, too. 
- Who will have deallocate the `char*`?
- We could not deallocate, but that leaks
- Also, we could have a pointer into nothingness since the original string might disappear. 

Thus, we can only copy out _values_.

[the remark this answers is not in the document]
boeschf: I concur

boeschf:
can you merge and resolve conflicts? I see some changes which seem to be already merged.

brenthuisman (dismissed):
lgtm

arbor/memory/wrappers.hpp
```diff
@@ -151,5 +153,63 @@ auto on_gpu(const C& c) -> device_vector<typename C::value_type> {
     return device_vector<T>(make_const_view(c));
 }
 
+template<typename K,
+         typename T>
+void serialize(::arb::serializer ser, const K& k, const host_vector<T>& hvs) {
+    ser.begin_write_array(to_key(k));
```
boeschf, line 159: shouldn't this be `to_serdes_key(k)` now?

brenthuisman:
@boeschf doublechecks CUDA 11 avail on piz daint.

arbor/memory/wrappers.hpp
```diff
@@ -152,64 +152,45 @@ auto on_gpu(const C& c) -> device_vector<typename C::value_type> {
     using T = typename C::value_type;
     return device_vector<T>(make_const_view(c));
 }
-
+} // namespace memory
+  //
```
boeschf, line 156:
```suggestion

```

arbor/include/arbor/serdes.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <optional>
-#include <string_view>
+// #include <string_view>
```
boeschf, line 4:
```suggestion
```

arbor/include/arbor/serdes.hpp
```diff
@@ -143,205 +147,219 @@ struct serializer {
...
-ARB_ARBOR_API void serialize(serializer& ser, const K& k, std::string_view v) {
-    ser.write(to_serdes_key(k), std::string{v});
+ARB_ARBOR_API void serialize(arb::serializer& ser, const K& k, const std::string& v) {
+    ser.write(arb::to_serdes_key(k), v);
 }
 
+// template<typename K>
+// ARB_ARBOR_API void serialize(serializer& ser, const K& k, std::string_view v) {
+    // ser.write(to_serdes_key(k), std::string{v});
+// }
```
boeschf, line 166: remove?

arbor/backends/gpu/event_stream.hpp
```diff
@@ -84,6 +89,49 @@ class event_stream : public event_stream_base<Event, typename memory::device_vec
         arb_assert(num_events == base::ev_data_.size());
     }
 
+    friend void serialize(serializer& ser, const std::string& k, const event_stream<Event>& t) {
```
boeschf, line 92: Wouldn't it be less error-prone if the base class template `event_stream_base` had its own serialization hooks which could be called from here?

arbor/include/arbor/serdes.hpp
```diff
@@ -143,205 +147,219 @@ struct serializer {
...
+ARB_ARBOR_API void serialize(::arb::serializer& ser, const K& k, const std::vector<V, A>& vs) {
+    ser.begin_write_array(arb::to_serdes_key(k));
+    for (std::size_t ix = 0; ix < vs.size(); ++ix) serialize(ser, ix, vs[ix]);
+    ser.end_write_array();
+}
+
+template <typename K,
+          typename V>
+ARB_ARBOR_API void serialize(::arb::serializer& ser, const K& k, const std::vector<V>& vs) {
+    ser.begin_write_array(arb::to_serdes_key(k));
```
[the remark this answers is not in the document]
thorstenhater: Yeah, the second one can go. We don't want to serialize allocators since upon deserialization we want to use the current state, not the serialized one. Assume
```
auto vec = vec_w_alloc();
serialize(ser, vec); // Writes out allocator state
auto vec_two = vec_w_alloc(); // Change allocator state
deserialize(ser, vec); // Restores global(!!!) allocator state
// Now what happens?
auto vec_three = vec_w_alloc(); // Change allocator state
```

arbor/include/arbor/serdes.hpp
thorstenhater: yes

arbor/backends/gpu/event_stream.hpp
thorstenhater: Hmm. It's a tradeoff. In the current way I can use the `ENABLE` macro in the `multicore` subclass
and need a custom serializer in the `gpu` case. Your proposal is slightly more code to write:
1. `ENABLE` in base
2. custom one-liner in `multicore`
3. custom serializer in `gpu` minus the base members plus one line for serializing `base`

Seen that way: Yes, probably better ;)

boeschf:
cscs-ci run daint-gpu

brenthuisman:
cscs-ci run daint-gpu

brenthuisman:
cscs-ci run daint-gpu

brenthuisman:
Still an error in the distributed GPU job:

```
gap_junctions: srun: error: nid05416: task 1: Exited with exit code 1
srun: launch/slurm: _step_signal: Terminating StepId=48211655.1
slurmstepd: error: *** STEP 48211655.1 ON nid05415 CANCELLED AT 2023-08-07T17:55:22 ***
   - gap_junctions: srun: error: nid05415: task 0: Exited with exit code 143
```

thorstenhater:
JUWELS disagrees
```
$ srun --account=XXXX --partition=XXXX bin/gap_junctions
gpu:      yes
threads:  1
mpi:      yes
ranks:    1

Using default parameters.
running simulation

30 spikes generated at rate of 3.33333 ms between spikes

meter                         time(s)      memory(MB)
model-init                      0.016           1.654
model-run                       1.098           1.409
meter-total                     1.114           3.062
```

boeschf (approved):

PR #2093, opened 2023-03-02
Take less locks for shorter times when sampling.

thorstenhater:
Title say almost all, locks to the sampler maps are taken even if said map is empty.
So, just don't do that. A sampling heavy experiment shows this delta; we sample
93 cells 10 times per `dt`.  Release build + Profiling + SIMD on a single thread.
## Before
```
Running simulation for 1s...
Simulation done, took: 3.344s
REGION               CALLS     WALL    THREAD        %
root                     -    3.185     3.185    100.0
  advance                -    3.116     3.116     97.8
    samplesetup       8000    2.147     2.147     67.4
    sampledeliver     8000    0.494     0.494     15.5
    integrate            -    0.457     0.457     14.3
      setup           8000    0.282     0.282      8.8
```
## After
```
Running simulation for 1s...
Simulation done, took: 1.878s
REGION               CALLS     WALL    THREAD        %
root                     -    1.725     1.725    100.0
  advance                -    1.656     1.656     96.0
    samplesetup       8000    0.938     0.938     54.4
    sampledeliver     8000    0.407     0.407     23.6
    integrate            -    0.293     0.293     17.0
```

### DIFF
--- a/arbor/arbexcept.cpp
+++ b/arbor/arbexcept.cpp
@@ -26,7 +26,7 @@ arbor_internal_error::arbor_internal_error(const std::string& what):
 domain_error::domain_error(const std::string& w): arbor_exception(w) {}
 
 bad_cell_probe::bad_cell_probe(cell_kind kind, cell_gid_type gid):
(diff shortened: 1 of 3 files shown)


brenthuisman (approved):
lgtm

PR #2094, opened 2023-03-06
📊 Pin threads

thorstenhater:
Allow thread and process affinity to request process and thread affinity via `proc_allocation`.
Affinity improves resource utilisation and thus performance, see experiments below, mainly
of CPU-only simulations. All other scenarios will not pay for it, so this is a straight win.

This makes `hwloc` an optional dependency. If found, we offer affinity, else we don't. It's
commonly deployed on HPC clusters, though, so we have good chances of finding it.
Some care is taken to interact nicely with MPI and SLURM.

Benchmarks on JUWELS-CPU (48 cores x 2 SMT) with a small-ish network of ~100 simple
cells, (release, SIMD).
thread-scaling-GCL-juwels


cmake/FindHWLOC.cmake
```diff
@@ -0,0 +1,297 @@
...
+#=============================================================================
+# Copyright 2012-2013 Inria
+# Copyright 2012-2013 Emmanuel Agullo
+# Copyright 2012-2013 Mathieu Faverge
+# Copyright 2012      Cedric Castagnede
+# Copyright 2013      Florent Pruvost
+#
+# Distributed under the OSI-approved BSD License (the "License");
+# see accompanying file MORSE-Copyright.txt for details.
+#
```
boeschf, line 39: what about the license - do we need to provide the file?

doc/cpp/hardware.rst
```diff
@@ -249,6 +249,18 @@ The core Arbor library *libarbor* provides an API for:
 
         The number of CPU threads available.
 
+    .. cpp:member:: bool bind_procs
+
+        Try to generate a binding mask for all processes on a node.
+        Do not set to true if process binding is handled externally,
+        eg by SLURM or OpenMPI.
+
```
boeschf, line 257: ok, but how is it useful then? I assume it would widen the mask if possible - is that what is intended?

arbor/affinity.hpp
```diff
@@ -0,0 +1,73 @@
...
+    // Allocate one set per item
+    auto cpusets = std::vector<hwloc_cpuset_t>(count, {});
+    // Distribute items over topology, giving each of them as much private cache
+    // as possible and keeping them locally in number order.
+    hwloc(hwloc_distrib(topology,
+                        &root, 1,                        // single root for the full machine
+                        cpusets.data(), cpusets.size(),  // one cpuset for each thread
+                        INT_MAX,                         // maximum available level = Logical Cores
+                        0),                              // No flags
+          "Distribute");
```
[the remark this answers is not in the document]
thorstenhater: `distrib` produces a distribution that maximizes the caches available to each item in the spread.
Yes, hyperthreads are in included by setting INT_MAX/PU/logical cores here. The distribution will
spread out as far as possible, regardless of the numbering scheme presented by the HW/OS, which
is why HWLOC.
If anything has set us a mask, we restrict our topology to that mask and distribute within, which
is why the comment above advises against binding in that scenario.

doc/cpp/hardware.rst
thorstenhater: Nope, if SLURM made a mask, we'll live inside, thus wasting HW, so by default off and this
warning. If you want Arbor to manage this, turn off SLURMs attempt `--cpu-bind=none` and use
this. Provided for completeness.

cmake/FindHWLOC.cmake
thorstenhater: This file came to us from the Eigen Library on the Android project ;)

arbor/threading/threading.cpp
```diff
@@ -3,9 +3,16 @@
 
 #include <arbor/assert.hpp>
 #include <arbor/util/scope_exit.hpp>
+#include <arbor/arbexcept.hpp>
 
 #include "threading/threading.hpp"
 
+#include <iostream>
```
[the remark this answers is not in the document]
thorstenhater: Direly.

arbor/CMakeLists.txt
```diff
@@ -134,6 +134,17 @@ endif()
...
 add_library(arbor ${arbor_sources} ${arbor-builtin-mechanisms})
+
+if(HWLOC_FOUND)
+    target_include_directories(ext-hwloc INTERFACE ${HWLOC_INCLUDE_DIRS})
+    message("hwloc ${HWLOC_LIBRARY_DIRS} ${HWLOC_LIBRARIES}")
+    target_link_directories(ext-hwloc INTERFACE ${HWLOC_LIBRARY_DIRS})
+    target_link_libraries(ext-hwloc INTERFACE ${HWLOC_LIBRARIES})
+    target_compile_definitions(arbor-private-deps INTERFACE ARB_HAVE_HWLOC)
+    target_link_libraries(arbor INTERFACE ext-hwloc)
+endif()
```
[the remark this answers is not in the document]
thorstenhater: Pasted in the wrong place and escaped the purge.

arbor/threading/threading.cpp
```diff
@@ -132,7 +175,10 @@ thread_local unsigned task_system::current_task_queue_ = -1;
 // Default construct with one thread.
 task_system::task_system(): task_system(1) {}
 
-task_system::task_system(int nthreads): count_(nthreads), q_(nthreads) {
+task_system::task_system(int nthreads, bool bind):
+    count_(nthreads),
+    bind_(bind),
+    q_(nthreads) {
```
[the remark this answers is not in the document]
thorstenhater: On `q_`? That wasn't touched, but allocates one queue per worker (incl the main thread).

arbor/communication/mpi_context.cpp
```diff
@@ -24,9 +26,18 @@ struct mpi_context_impl {
...
+    explicit mpi_context_impl(MPI_Comm comm, bool bind=false): comm_(comm) {
         size_ = mpi::size(comm_);
         rank_ = mpi::rank(comm_);
+        if (bind) {
+            MPI_Comm local;
+            MPI_Comm_split_type(comm, MPI_COMM_TYPE_SHARED, rank_, MPI_INFO_NULL, &local);
+            auto mpi_guard = util::on_scope_exit([&] { MPI_Comm_free(&local); });
+            int local_rank = -1, local_size = -1;
+            MPI_Comm_rank(local, &local_rank);
+            MPI_Comm_size(local, &local_size);
```
AdhocMan, line 38: Two points here:
- Maybe use `MPI_Comm_free(&local)` to clean up the communicator afterwards
- Why not use the mpi functions from `mpi.hpp`? These would include error checks.

CMakeLists.txt
```diff
@@ -238,8 +238,20 @@ if(NOT ARB_USE_BUNDLED_RANDOM123)
     target_include_directories(ext-random123 INTERFACE ${RANDOM123_INCLUDE_DIR})
 endif()
 
-add_subdirectory(ext)
+# in the event we can find hwloc, just add it
+find_package(HWLOC QUIET)
```
AdhocMan, line 242: It would be nice to have an option to control the inclusion of HWLOC. Especially for package managers like spack, where ideally it should fail if HWLOC is not found despite being requested.

arbor/threading/threading.cpp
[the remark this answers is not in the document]
boeschf: right, sorry I didn't see it being there before.

arbor/communication/mpi_context.cpp
thorstenhater: - See l 35. It's also exception safe ;)
- Good point.

CMakeLists.txt
thorstenhater: I can add an advanced option to force it, yes.

arbor/communication/mpi_context.cpp
boeschf: C++ bindings are more or less deprecated I believe

arbor/threading/threading.cpp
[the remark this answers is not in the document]
thorstenhater: I changed the formatting, as the line got too long for my tastes

arbor/communication/mpi_context.cpp
thorstenhater: @AdhocMan talks about our internal wrappers.

thorstenhater: And yes, MPIC++ as been deprecated since (pre?) 3.0, I think.

cmake/FindHWLOC.cmake
thorstenhater: But I guess we do. Would be way easier if `hwloc` had their own or we could rely on delegating to `pkg-config`.

thorstenhater: That was quite an adventure. MORSE doesn't exist anymore, but their code is all over the place. Then again, the license is weird. So, now we have a simple wrapper around pkg-config, as OpenMPI/hwloc provide `.pc` files.

CMakeLists.txt
thorstenhater: Done.

python/context.cpp
```diff
@@ -107,13 +118,20 @@ void register_contexts(pybind11::module& m) {
...
+             "gpu_id"_a=pybind11::none(),
+             "bind_procs"_a=false,
+             "bind_threads"_a=false,
+             "Construct an allocation with arguments:\n"
+             "  threads: The number of threads available locally for execution. Must be set to 1 at minimum. 1 by default.\n"
+             "  gpu_id:  The identifier of the GPU to use, None by default.\n")
         .def_property("threads", &proc_allocation_shim::get_num_threads, &proc_allocation_shim::set_num_threads,
             "The number of threads available locally for execution.")
+        .def_property("bind_procs", &proc_allocation_shim::get_bind_procs, &proc_allocation_shim::set_bind_procs,
+            "Try to bind MPI procs?")
```
brenthuisman, line 132: Specifically MPI proc? This is not what's in the API docs, which even advise against when using MPI.

doc/concepts/hardware.rst
```diff
@@ -3,31 +3,39 @@
 Hardware context
```
brenthuisman, line 3: Combining reformatting with changes makes spotting the change difficult. Rather do this in separate changesets.

doc/cpp/hardware.rst
```diff
@@ -249,6 +249,18 @@ The core Arbor library *libarbor* provides an API for:
 
         The number of CPU threads available.
 
+    .. cpp:member:: bool bind_procs
+
+        Try to generate a binding mask for all processes on a node.
```
brenthuisman, line 254: Document/link what a binding mask is, and how it is generated. (Also on Python page)

brenthuisman (changes requested):

python/context.cpp
thorstenhater: Read that again. The docs have -- as does the API -- _two_ levels of binding.
1. Proc: If managed externally, please don't try this, since it'll be redundant at best and likely detrimental.
2. Thread: See below.

So, if you want to go the full distance:
- Disable SLURM binding attempts (these are proc-level only!) `--cpu-bind=none`
- If using OpenMPI and your site is misconfigured, you might need to disable their's, too.
- Set `bind_procs` and `bind_threads` in `proc_allocation`.

Otherwise you'll want `procs=false`, `threads=true`.

arbor/threading/threading.hpp
```diff
@@ -161,7 +162,8 @@ class ARB_ARBOR_API task_system {
     task_system();
 
     // Create nthreads-1 new std::threads running run_tasks_loop(tid)
-    task_system(int nthreads);
+    task_system(int nthreads, bool bind_threads=false);
```
boeschf, line 165: I wonder weather the bind option should be `true` by default? Is there a common use case where this may not be the best option?

thorstenhater: I had an internal debate about precisely this. I went with off by default since most users likely don't 
care on their laptop w/ 2 cores, this would change a default, and might raise an error.

arbor/affinity.hpp
```diff
@@ -0,0 +1,73 @@
...
+  process
+};
+
+
+#ifdef ARB_HAVE_HWLOC
+
+inline
+void hwloc(int exp, const std::string& msg) {
+    if (-1 == exp) {
+        throw arbor_internal_error(std::string{"HWLOC Thread failed at: "} + msg);
```
brenthuisman, line 24: It's a mysterious error with a mysterious error message. I would prefer to take responsibility for it: what is the consequence of the error? Maybe sth like: "Arbor cannot proceed, due to an error in HWLOC: %msg. Consider doing X/changing Y."

thorstenhater: It's a mystery wrapped in an enigma.

arbor/threading/threading.cpp
```diff
@@ -87,23 +92,59 @@ void task_system::run(priority_task ptsk) {
...
 
-void task_system::run_tasks_loop(int i) {
-    auto guard = util::on_scope_exit([] { current_task_queue_ = -1; });
-    current_task_queue_ = i;
+#define HWLOC(exp, msg) if (-1 == exp) throw arbor_internal_error(std::string{"HWLOC Thread failed at: "} + msg);
+
+// If we have found hwloc, pin our thread to a single physical core else noop
+void bind_my_thread(int index, int n_threads) {
+#ifdef ARB_HAVE_HWLOC
+    // Create the topology and ensure we don't leak it
```
boeschf, line 100: I am confused - why not using `affinity.hpp`?

CMakeLists.txt
```diff
@@ -238,8 +243,25 @@ if(NOT ARB_USE_BUNDLED_RANDOM123)
...
+    target_link_directories(ext-hwloc INTERFACE ${hwloc_LIBRARY_DIRS})
+    target_link_libraries(ext-hwloc INTERFACE ${hwloc_LIBRARY})
+    target_include_directories(ext-hwloc INTERFACE ${hwloc_INCLUDE_DIR})
+    target_compile_definitions(ext-hwloc INTERFACE ARB_HAVE_HWLOC)
+    target_link_libraries(arbor-private-deps INTERFACE ext-hwloc)
+else()
+    if(ARB_USE_HWLOC)
+        message(SEND_ERROR "Requested support for hwloc, but CMake couldn't find it.")
+    endif()
+endif()
```
boeschf, line 261: should we move this stuff into `ext/CMakeLists.txt`?

arbor/execution_context.cpp
```diff
@@ -16,7 +16,7 @@ namespace arb {
 
 execution_context::execution_context(const proc_allocation& resources):
     distributed(make_local_context()),
```
boeschf, line 18: I guess it doesn't really make sense to set process affinity for the local context?

thorstenhater: We could, but is this really for performance-conscious users?

arbor/threading/threading.cpp
thorstenhater: I am confused as well. Maybe a bad interaction between git and my editor...

CMakeLists.txt
thorstenhater: Hmm, I am a bit torn here: `ext` is where we store 3rd party projects. This is a system dependency.

arbor/threading/threading.cpp
```diff
@@ -92,46 +88,10 @@ void task_system::run(priority_task ptsk) {
...
-#endif
-}
-
-#undef HWLOC
-
 void task_system::run_tasks_loop(int index) {
     auto guard = util::on_scope_exit([] { current_task_queue_ = -1; });
     current_task_queue_ = index;
-    if (bind_) bind_my_thread(index, count_);
+    if (bind_) set_affinity(index, count_, affinity_kind::thread);
```
boeschf, line 94: what happens if count is greater than number hardware resources?

arbor/execution_context.cpp
boeschf: I'd prefer to have the same behaviour for a program on single process regardless of whether it was compiled with MPI, ymmv.

arbor/threading/threading.cpp
thorstenhater: You can experiment by using the `hwloc-distrib` tool. But it does the sensible thing:
```
❯ hwloc-distrib --verbose --from machine --to PU --single 16
distributing 16
0x00000001
0x00000001
0x00000002
0x00000002
0x00000004
0x00000004
0x00000008
0x00000008
0x00000010
0x00000010
0x00000020
0x00000020
0x00000040
0x00000040
0x00000080
0x00000080
```

arbor/execution_context.cpp
thorstenhater: Yeah, if someone made the effort to pass `bind_proc=true`, we should honour their wishes, whatever we might think.

arbor/threading/threading.cpp
boeschf: thanks, looks sensible indeed

arbor/execution_context.cpp
thorstenhater: _but_, consider what will happen: we will distribute 1 item across the available mask _without_ singlification (that's done only for threads). That's a no-op.

boeschf: true, it doesn't seem that useful

arbor/affinity.hpp
brenthuisman: Merci.

python/context.cpp
brenthuisman: I can't retrace my steps. Docs are clear.

brenthuisman (approved):
A request, but not a showstopper: add hwloc to `spack/package.py`. There is no such thing as an optional dependency, so it could be a variant.

brenthuisman:
That was fast!

thorstenhater:
I need this to build upon. Since we don't have stacked PRs (maybe we should?) this is the way.

PR #2095, opened 2023-03-07
BUGFIX: gpu mechanism init

boeschf:
Fix GPU code generation when initializing mechanism without arrays by not launching multiply kernel. Would otherwise lead to runtime error.


thorstenhater (approved):
Stupid oversight! :+1: Thanks for fixing.

Out of curiosity, how did you find this and how did it slip under the radar for so long?

boeschf:
I ran the unit tests on Piz daint. Unfortunately, we don't run anything on GPU in the github tests, and bors is borked still.

PR #2096, opened 2023-03-07
BUGFIX: add missing functions to sve backend

boeschf:
Some auxiliary functions were missing. Noticed while trying Arbor on AWS Graviton based nodes.


thorstenhater (approved):
lgtm

PR #2098, opened 2023-03-16
Enable Spack tests

brenthuisman:
[Spack can run tests](https://spack.readthedocs.io/en/latest/packaging_guide.html#id24), and relies on [ancient GNU standards for Make targets](https://www.gnu.org/software/make/manual/make.html#Standard-Targets), so these are hereby added. This is a request from Ebrains.


test/CMakeLists.txt
```diff
@@ -19,3 +19,7 @@ add_subdirectory(unit-modcc)
 # Microbenchmarks.
 # Builds: one executable per microbenchmark under ubench/.
 add_subdirectory(ubench)
+
+# 'check' target executes tests.
+add_custom_target(check COMMAND unit)
```
thorstenhater, line 24: 1. Comment seems to repeat the code. 
2. Comment seems to miss the important bit: We _need_ this to be called `check` since someone external relies on it, right?

python/config.cpp
thorstenhater, line 1: Just a though, not necessarily for this PR: We should add the `config` dict to our docs, ask people to add its output to bug report, and add a related field to the issue temple.

thorstenhater (changes requested):
A detail and a spin-off.

python/config.cpp
brenthuisman: Good idea, how bout I add it to the issue template?

Issue #2101, opened 2023-03-24
Issues when building for SVE


adegomme:
version : all > 0.6

**Describe the bug**

When building (or cross building) with SVE enabled (for benchmarking purposes), we encountered a few issues in the generated portions of code, with various compilers (arm/gcc/llvm).

Process is to build modcc for the native system and use it to build the SVE version, as documentation says, so:

```
mkdir -p $BUILDDIR
cd $BUILDDIR
git clone --depth=1 --recursive https://github.com/arbor-sim/arbor.git arbor
cd $BUILDDIR/arbor
mkdir build_modcc
cd build_modcc
cmake ..
make -j modcc 
mkdir -p $BUILDDIR/arbor/build
cd $BUILDDIR/arbor/build
cmake -DCMAKE_INSTALL_PREFIX=/usr/local/arbor -D ARB_WITH_MPI=ON -D ARB_WITH_PYTHON=ON -D ARB_VECTORIZE=ON -D CMAKE_CXX_FLAGS="-O3 -mcpu=neoverse-v1" -D CMAKE_C_FLAGS="-O3 -mcpu=neoverse-v1 " -D ARB_MODCC=$BUILDDIR/arbor/build_modcc/bin/modcc -D ARB_ARCH=armv8.4-a+sve $BUILDDIR/arbor 
cmake --build $BUILDDIR/arbor/build --target all -- -j$(nproc) 
```

The first issue encountered was

```
/var/tmp/arbor/build/mechanisms/generated/bbp/K_Tst_cpu.cpp:37:27: error: constexpr variable 'min_align_' must be initialized by a constant expression
static constexpr unsigned min_align_ = std::max(S::min_align(simd_value{}), S::min_align(simd_index{}));
/var/tmp/arbor/build/mechanisms/generated/bbp/K_Tst_cpu.cpp:37:72: note: non-literal type 'arb::bbp_catalogue::kernel_K_Tst::simd_value' (aka '__SVFloat64_t') cannot be used in a constant expression
static constexpr unsigned min_align_ = std::max(S::min_align(simd_value{}), S::min_align(simd_index{}));
```

Removing the "constexpr" in the cprinter is enough to go past it, but I don't know if performance may be affected here.

`sed -i 's/static constexpr unsigned min_align_/static unsigned min_align_/g' modcc/printer/cprinter.cpp`

The second issue is

```
[...]

PR #2104, opened 2023-04-03
bugfix: sve code generation

boeschf:
solves issue #2101


test/unit/test_simd.cpp
```diff
@@ -1889,9 +1889,11 @@ typedef ::testing::Types<
                   simd_t<     simd<int,    0, simd_abi::sve>, int,    4>,
                   simd_t<simd_mask<double, 0, simd_abi::sve>, bool,   4>>,
 
+#if ARB_SVE_WIDTH >= 8
```
thorstenhater, line 1892: So, if we cap the SVE length at 8 in the tests, I'd assume we should do so in all of our SIMD. Or we shouldn't cap it, ever. 
What's the reasoning here?

thorstenhater (changes requested):
Hey,

looking good, except two minor things:
- In the tests, you ap SVE_LENGTH at 8. What's the idea there? See also comment on file.
- A question due to my ignorance of SVE:
   ```cxx
        // sve.hpp:304
       static  svint64_t pow(const svint64_t& x, const svint64_t& y, const svbool_t& mask = svptrue_b64()) {
         auto len = svlen_s64(x);
         int32 a[len], b[len], r[len];
         copy_to_masked(x, a, mask);
         copy_to_masked(y, b, mask);

        for (unsigned i = 0; i<len; ++i) {
            r[i] = std::pow(a[i], b[i]);
        }
        return copy_from_masked(r, mask);
    }
   ```
   Note the use of VLA here as `svlen_*` is not a constexpr. I'd really like to get rid of it, since it's non-standard and if we fix the length to 8, it should be possible?!

PR #2106, opened 2023-04-17
Add LTO and M1 arch to CMake

thorstenhater:
Tweak CMake to set up Link Time Optimisations (helps pybind11) and set arch/tune for Arm Macs (M1 for now).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 # Have LTO where possible, ie add -flto
 check_ipo_supported(RESULT HAVE_LTO OUTPUT ERR_LTO)
-if(HAVE_LTO)
+if(HAVE_LTO AND NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
   message (STATUS "LTO support found, enabling")
   set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
 else()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.19)
 include(CMakeDependentOption)
+include(CheckIPOSupported)
 
 # Make CUDA support throw errors if architectures remain unclear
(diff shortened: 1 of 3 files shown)


thorstenhater:
> This seems to work fine on the M2 as well.

Good to hear, thanks for checking.

arbor/backends/multicore/threshold_watcher.hpp
```diff
@@ -58,6 +58,7 @@ class threshold_watcher {
     /// Crossing events are recorded for each threshold that
     /// is crossed since the last call to test
     void test(array& time_since_spike, const arb_value_type& t_before, const arb_value_type& t_after) {
+        if (cv_index_.empty() || n_detectors_ == 0) return;
```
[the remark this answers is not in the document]
thorstenhater: So, this has always an issue, I guess, **but** LTO has exposed it. Example 
```cxx
std::vector<T> v;
assert(v.size() == 0); // OK, always passes.
assert(v.data() == nullptr); // MIGHT pass, but not always
assert(v.data() != nullptr); // MIGHT pass, but not always
```
See also here
https://stackoverflow.com/questions/25419851/what-should-stdvectordata-return-if-the-vector-is-empty

CMakeLists.txt
```diff
@@ -16,6 +17,14 @@ include(GNUInstallDirs)
 # Effectively adds '-fpic' flag to CXX_FLAGS. Needed for dynamic catalogues.
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
+# Have LTO where possible, ie add -flto
+check_ipo_supported(RESULT HAVE_LTO OUTPUT ERR_LTO)
+if(HAVE_LTO)
```
AdhocMan, line 22:
```suggestion
if(HAVE_LTO AND NOT DEFINED CMAKE_INTERPROCEDURAL_OPTIMIZATION)
```
It could be useful to still be able to override this. Setting it as a cache variable would affect any other project that includes arbor, so just checking if it is already set might be a better solution. Alternatively there there could be an `ARB_LTO` option.

thorstenhater: Good idea!

AdhocMan (approved):
lgtm

PR #2107, opened 2023-04-17
Add zero'ing

thorstenhater:
# Add specialised zero method to rangeutils.

Zeroing current arrays can take substantial amounts of time in simple models. Thus
add zero leveraging memset which is slightly faster and communicates intentions
a bit better.

### DIFF
--- a/arbor/util/rangeutil.hpp
+++ b/arbor/util/rangeutil.hpp
@@ -86,7 +86,9 @@ void fill(Seq&& seq, const V& value) {
 // Zero a container, specialised for contiguous sequences
 // i.e.: Array, Vector, String.
 
-template <typename Seq, typename = std::enable_if_t<sequence_traits<Seq&&>::is_contiguous>>
+template <typename Seq,
+          typename = std::enable_if_t<sequence_traits<Seq&&>::is_contiguous>,
+          typename = std::is_trivially_copyable<typename sequence_traits<Seq&&>::value_type>>
 void zero(Seq& vs) {
     // NOTE: All contiguous containers have `data` and `size` methods.
     using T = typename sequence_traits<Seq&&>::value_type;
--- a/arbor/util/rangeutil.hpp
+++ b/arbor/util/rangeutil.hpp
@@ -9,6 +9,7 @@
 #include <iterator>
 #include <ostream>
 #include <numeric>
+#include <cstring>
 
 #include "util/meta.hpp"
 #include "util/range.hpp"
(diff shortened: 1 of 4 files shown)


arbor/util/rangeutil.hpp
```diff
@@ -9,6 +9,7 @@
 #include <iterator>
 #include <ostream>
 #include <numeric>
+#include <cstring>
```
brenthuisman, line 12: Are you sure? I don't see anything using this.

brenthuisman (approved):

arbor/util/rangeutil.hpp
thorstenhater: std::memset?

brenthuisman: :eyes:

arbor/util/rangeutil.hpp
```diff
@@ -82,6 +83,16 @@ void fill(Seq&& seq, const V& value) {
     std::fill(canon.begin(), canon.end(), value);
 }
 
+// Zero a container, specialised for contiguous sequences
+// i.e.: Array, Vector, String.
+
+template <typename Seq, typename = std::enable_if_t<sequence_traits<Seq&&>::is_contiguous>>
```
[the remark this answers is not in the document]
thorstenhater: Oh how I wish for concepts 😢

arbor/util/rangeutil.hpp
```diff
@@ -82,6 +83,18 @@ void fill(Seq&& seq, const V& value) {
     std::fill(canon.begin(), canon.end(), value);
 }
 
+// Zero a container, specialised for contiguous sequences
+// i.e.: Array, Vector, String.
+
+template <typename Seq,
+          typename = std::enable_if_t<sequence_traits<Seq&&>::is_contiguous>,
+          typename = std::is_trivially_copyable<typename sequence_traits<Seq&&>::value_type>>
```
AdhocMan, line 91:
```suggestion
          typename = std::enable_if_t<
              sequence_traits<Seq &&>::is_contiguous &&
              std::is_trivially_copyable_v<typename sequence_traits<Seq &&>>>>
```
The `value_type` of `is_trivially_copyable` is just a type that always exists, so using it this way has no useful effect. You'll want to use `value` or the `_v` definition.
And yes, concepts would be nice!

arbor/util/rangeutil.hpp
AdhocMan, line 12:
```suggestion
#include <cstring>
#include <type_traits>
```

AdhocMan (changes requested):

AdhocMan (approved):
lgtm

PR #2111, opened 2023-04-19
Python/pip strikes again: execute `pip install` in venvs

brenthuisman:
A recent release of `pip` merged a PR addressing PEP 668. Distros  of Python may now mark `site-packages` directories as 'managed', which means `pip` is not allowed to write to them. This means Arbors CMake is now, due to assumptions about prefixes, is not able to correctly install the Python package anymore: `pip` is a hard requirement for that. We already document that `pip` is the preferred way, so that's OK, but what's new is that `pip` doesn't seem to default to a user install anymore, if the system dir is not writable (or 'managed'), ergo: install does not work. Moreover, on at least Debian 12, there's even no longer such a thing as user installs, so even the `--user` flag does not help.

`venv`s seem therefore the only place to install Python packages these days, so let's.

### DIFF
--- a/.github/workflows/test-everything.yml
+++ b/.github/workflows/test-everything.yml
@@ -111,7 +111,7 @@ jobs:
         with:
           python-version: ${{ matrix.config.py }}
       - name: Update pip and setup venv
-        run: python -m pip install --upgrade pip && python -m venv env && . env/bin/activate
+        run: python -m pip install --upgrade pip
       - name: Install Python packages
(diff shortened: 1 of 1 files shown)


thorstenhater:
Forcing users to do the right thing? Oh dear, what did we come to?

thorstenhater (approved):

thorstenhater (changes requested):

thorstenhater (approved):
🎉

PR #2112, opened 2023-04-19
Fix crash on probing non-existent mechanism state.

### LINKED ISSUE #2066: Arbor crashes on GPU if a mechanism is probed that is not placed
Arbor crashes with `IndexError: _Map_base::at` on GPU if a mechanism is probed that is not placed. This is different from the behaviour on CPU where Arbor runs and, as expected, samples are empty when retrieved after the simulation. The attached script reproduces the bug on `27db9a7f700bb05680a811ba3445e734d61aacf6`. 

The line that triggers is:

https://github.com/arbor-sim/arbor/blob/b63cca1c50a2a4d0d7d24f1b8de1fdd5581806cf/arbor/fvm_lowered_cell_impl.hpp#L1062

script:

[single_cell_recipe.py.gz](https://github.com/arbor-sim/arbor/files/10226552/single_cell_recipe.py.gz)

stack trace:

[gdb.txt.gz](https://github.com/arbor-sim/arbor/files/10226580/gdb.txt.gz)

thorstenhater:
Closes #2066

### DIFF
--- a/arbor/fvm_lowered_cell_impl.hpp
+++ b/arbor/fvm_lowered_cell_impl.hpp
@@ -897,62 +897,82 @@ void resolve_probe(const cable_probe_density_state_cell& p, probe_resolution_dat
     R.result.push_back(std::move(r));
 }
 
+inline
+auto point_info_of(cell_lid_type target,
+                   int mech_index,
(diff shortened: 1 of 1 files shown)


boeschf (approved):
no complaints

PR #2115, opened 2023-04-25
Fix crash in FindId Visitor.

### LINKED ISSUE #2114: NMODL parsing issue with `WHITE_NOISE` block (segmentation fault when compiling)
The following NMODL code only compiles if either the empty `else` block is uncommented or the `WHITE_NOISE` block is removed:

```
NEURON {
	POINT_PROCESS example
}

STATE {
	x
}

INITIAL {
	x = 10
}

BREAKPOINT {
	SOLVE state METHOD stochastic
}

WHITE_NOISE {
	xi
}

DERIVATIVE state {
	if (x > 2) {

	}
	else if (x > 1) {

	}
	:else {
	:
	:}
}
```

Exits with the following output:
```
Building catalogue 'custom' from mechanisms in /home/<user>/Desktop/Bug else case in mechanism/mechanisms
 * NMODL
   * example
Build log:
[ 25%] modcc generating: /tmp/tmp3z13mtrr/build/generated/custom/custom_catalogue.cpp;/tmp/tmp3z13mtrr/build/generated/custom/example_cpu.cpp
  File "/home/<user>/arbor_v0.8.2-dev/bin/arbor-build-catalogue", line 315, in <module>
    sp.run(make_cmd, shell=True, check=True, stdout=out, stderr=err)
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'make' returned non-zero exit status 2.
 Error:

Segmentation fault (core dumped)
make[2]: *** [CMakeFiles/custom-catalogue.dir/build.make:74: generated/custom/custom_catalogue.cpp] Error 139
make[1]: *** [CMakeFiles/Makefile2:82: CMakeFiles/custom-catalogue.dir/all] Error 2
make: *** [Makefile:91: all] Error 2
```

Occurs for Arbor v0.8.2-dev. The return of `arbor.config()` is:

```
{'mpi': False, 'mpi4py': False, 'gpu': None, 'vectorize': True, 'profiling': True, 'neuroml': True, 'bundled': True, 'version': '0.8.2-dev', 'source': '2023-04-19T20:12:31+0
[Bug_parsing_NMODL_with_WHITE_NOISE_2023-04-24.zip](https://github.com/arbor-sim/arbor/files/11313609/Bug_parsing_NMODL_with_WHITE_NOISE_2023-04-24.zip)
2:00 443031de77af7a86d9e941b8acebe2a635f3c81e', 'build_config': 'RELEASE', 'arch': 'native', 'prefix': '/home/<user>/arbor_v0.8.2-dev', 'binary_path': 'bin', 'lib_path': 'lib', 'data_path': 'share', 'CXX': '/usr/bin/c++', 'pybind-version': '2.10.1'}
```

Scripts to reproduce: [Bug_parsing_NMODL_with_WHITE_NOISE_2023-04-24.zip](https://github.com/arbor-sim/arbor/files/11313610/Bug_parsing_NMODL_with_WHITE_NOISE_2023-04-24.zip)

thorstenhater:
Closes #2114

### DIFF
--- a/modcc/symdiff.cpp
+++ b/modcc/symdiff.cpp
@@ -79,7 +79,7 @@ public:
     void visit(IfExpression* e) override {
         if (!found()) e->condition()->accept(this);
         if (!found()) e->true_branch()->accept(this);
-        if (!found()) e->false_branch()->accept(this);
+        if (!found() && e->false_branch()) e->false_branch()->accept(this);
     }
 
 private:
--- a/test/unit-modcc/mod_files/white_noise_under_conditional.mod
+++ b/test/unit-modcc/mod_files/white_noise_under_conditional.mod
@@ -1,0 +1,28 @@
+NEURON {
+  POINT_PROCESS example
+}
+
+STATE {
+  x
+}
+
+INITIAL {
+  x = 10
+}
+
+BREAKPOINT {
+  SOLVE state METHOD stochastic
+}
+
+WHITE_NOISE {
+  xi
+}
+
+DERIVATIVE state {
+  if (x > 2) {
+
+  }
+  else if (x > 1) {
+
+  }
+}
--- a/test/unit-modcc/test_module.cpp
+++ b/test/unit-modcc/test_module.cpp
@@ -188,3 +188,21 @@ TEST(Module, stochastic_density) {
     EXPECT_EQ(wnb.parameters.size(), 2u);
     EXPECT_EQ(wnb.used.size(), 2u);
 }
+
+TEST(Module, stochastic_conditional) {
+    Module m(io::read_all(DATADIR "/mod_files/white_noise_under_conditional.mod"), "wn_else.mod");
+    EXPECT_NE(m.buffer().size(), 0u);
+
+    Parser p(m, false);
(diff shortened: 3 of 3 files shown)


modcc/symdiff.cpp
```diff
@@ -79,7 +79,7 @@ class FindIdentifierVisitor: public Visitor {
     void visit(IfExpression* e) override {
         if (!found()) e->condition()->accept(this);
```
AdhocMan, line 80: Is there any way `condition()` or `true_branch()` could ever be null as well?

thorstenhater: Semantically? No, these represent `if (cond) {then} {else}`, where only the `else` part is optional. If we technically can produce such an expression, I do not know, but it'd constitute a bug in the transformations. A user can never, though.

AdhocMan (approved):
lgtm

modcc/symdiff.cpp
AdhocMan: Ok, makes sense.

PR #2121, opened 2023-05-12
Scikit build core

### LINKED ISSUE #2109: `pip install arbor_src_dir` pollutes `venv`s.
When installing Arbor with a `venv` active (created `python -m venv some_path`), a copy of the Arbor source tree is installed to `some_path`. This only happens when installing from source, not when installing from PyPI.

When `ninja install`ing, this does not happen. So it's either `pip` or `scikit-build` doing. A cursory scan of their issues and release notes does not point at an obvious change.

Possibly related to failures like https://github.com/arbor-sim/arbor/actions/runs/4719569758/jobs/8370545628?pr=2107

brenthuisman:
- Replaces `scikit-build` with `scikit-build-core`, which is beta-ish.
- Delete stale `setup.py`
- Create `scripts/test_executables.sh`, with a `modcc` and `a-b-c` test. Call script in all workflows (CMake|Pip|Spack).
- Suppress setting `march` and `mtune` to `native` on versions of Apple Clang <15. Only explicit targets are allowed, but since Apple Clang v15 `native` _is_ valid, and we only really use Apple Clang for the universal Python wheels, this seems an acceptably simple solution.

Downsides
1. https://scikit-build-core.readthedocs.io/en/latest/configuration.html#dynamic-metadata-wip
Looks like we cannot use `VERSION` to populate the `pyroject.toml` version string, so it is duplicated. 
2. Writing outside of the Python package dir, therefore not copy a-b-c and modcc to /bin. Workaround is to make them callable through a function, and bind the function to a script. There is a way to write "scripts" (i.e. executables) outside of the package dir, but then headers and such are still not copied. We could deduce the correct paths for all outputs (libraries, documentation) from the env var, but that seems no less brittle than the proposed solution, used by the CMake project itself. See https://github.com/scikit-build/scikit-build-core/issues/194#issuecomment-1426408196

Fixes #2109


brenthuisman:
In principle, I think this is a good idea, `scikit-build-core`. If you can live with having to set VERSION twice, please merge.

python/__init__.py
```diff
@@ -16,6 +16,33 @@ def get_version():
...
+    import subprocess
+
+    sys.exit(
+        subprocess.call(
+            [os.path.join(os.path.dirname(__file__), "bin", "modcc"), *sys.argv[1:]]
+        )
+    )
+
+
+def build_catalogue():
```
thorstenhater, line 31: Also ... since we call `sys.exit` here, does this mean we terminate our simulation by calling the harmless function `build_catalogue`?

python/__init__.py
```diff
@@ -16,6 +16,33 @@ def get_version():
         return version_file.read().strip()
 
 
+def modcc():
```
thorstenhater, line 19: Apart from the error handling: What's the motivation?
Shelling out is usually frowned upon

python/__init__.py
```diff
@@ -16,6 +16,33 @@ def get_version():
...
+    sys.exit(
+        subprocess.call(
+            [
+                os.path.join(os.path.dirname(__file__), "bin", "arbor-build-catalogue"),
+                *sys.argv[1:],
+            ]
+        )
+    )
+
+
```
thorstenhater, line 45: Finally: Documentation :D

setup.py
thorstenhater, line 1: Nice!

python/__init__.py
thorstenhater, line 45: PS: I lied; one more. If we start adding Python-only functions to Arbor (all for it), let's make a separate module and just use `__init__.py` for importing all of the sub-modules.

pyproject.toml
```diff
@@ -75,3 +79,4 @@ repair-wheel-command = "auditwheel repair -w {dest_dir} {wheel} && python3 /proj
 
 [[tool.cibuildwheel.overrides]]
 select = "*-musllinux*"
+#TBD
```
thorstenhater, line 82: Elaborate, please.

thorstenhater (changes requested):
A few comments, I like the cleaner looks though.

Can we have an introduction for skbuild-core, though?

pyproject.toml
brenthuisman: As you can see, it isn't new. It's a placeholder for musllinux specific options. I fixed musl builds a while back, but not passing the tests. If you want to remove the placeholder, we can.

python/__init__.py
```diff
@@ -16,6 +16,33 @@ def get_version():
         return version_file.read().strip()
 
 
+def modcc():
+    import os
+    import sys
+    import subprocess
+
+    sys.exit(
+        subprocess.call(
```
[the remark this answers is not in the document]
brenthuisman: Does the pip version need work arounds? One can still use their own builds of these tools (because then you know where you've installed them).

The functions are there mainly to have something to bind a Python scripts to (see `[project.scripts]` in `pyproject.toml`). They're not meant to be used. We could prefix them with underscores to indicate that better?

python/__init__.py
brenthuisman: Yes. More reason to indicate these functions are not meant to be called by users.

python/__init__.py
brenthuisman: Since this is a workaround for functionality not meant for direct use, I don't think this is necessary yet.

python/__init__.py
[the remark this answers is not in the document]
thorstenhater: Something like this, or shove them into an `internal_111_donot_use` namespace.

brenthuisman:
Todo:

- Add a `modcc` and `a-b-c` test to `run_python_examples.sh`, and execute `run_python_examples.sh` in CMake|Spack|Pip CI tests.

doc/install/python.rst
```diff
@@ -106,7 +106,7 @@ The following flags can be used to configure the installation:
 
 .. note::
 
-   There are more, advanced flags that can be set. We are using ``scikit-build``
+   There are more, advanced flags that can be set. We are using ``scikit-build-cire``
```
thorstenhater, line 109:
```suggestion
   There are more, advanced flags that can be set. We are using ``scikit-build-core``
```

cmake/CompilerOptions.cmake
```diff
@@ -111,12 +111,17 @@ function(set_arch_target optvar arch)
         string(REGEX REPLACE "-.*" "" target_model "${target}")
 
         # Use -mcpu for all supported targets _except_ for x86 and Apple arm64, where it should be -march.
-        if("${target}" MATCHES "aarch64-apple-darwin" OR "${target}" MATCHES "arm64-apple-darwin")
-            set(arch_opt "-march=${arch} -mtune=${arch}")
-        elseif(target_model MATCHES "x86|i[3456]86" OR target_model MATCHES "amd64" OR target_model MATCHES "aarch64")
-            set(arch_opt "-march=${arch} -mtune=${arch}")
+        
+        if (CMAKE_CXX_COMPILER_ID MATCHES "AppleClang" AND CMAKE_CXX_COMPILER_VERSION LESS 15)
+            set(arch_opt "")
```
thorstenhater, line 116: Ok, that'll _work_ but is certainly not ideal. You can try to detect being on
AArch64 and then use `apple-m1`. However, that also isn't complete, there's an M2 already...

thorstenhater (changes requested):

cmake/CompilerOptions.cmake
brenthuisman: 1. I can't detect for arch, because building a fat binary happens on x86_64 and also runs into this wall. Afaik there's no CMake builtin way to detect universal builds, in which case only horrible hacks will save us.
2. Indeed, we won't know what the right arch is. I don't think having our own little matrix here is a good idea.

Since it's problem with old AppleClang, with time the people hitting this, which was already near 0, will go down. Only reason really is that I can't upgrade AppleClang in Github runners, so until they move to 15, this needs to be handled. Anyone that cares about optimized builds can always do it themselves.

thorstenhater (changes requested):

cmake/CompilerOptions.cmake
thorstenhater: Ok, accepted.

thorstenhater (approved):

henryiii:
FYI, we now have a built-in regex plugin so you can read version from a local file, if that's helpful.

PR #2127, opened 2023-05-30
Add an error on NET_RECEIVE w/ > 1 args.

thorstenhater:
NEURON allows NET_RECEIVE to take more than one argument (`weight` being always the first).
These arguments can then be set when connecting to the POINT_PROCESS (per _connection_ not
target synapse); similar to parameters.

Arbor doesn't allow this, but `modcc` compiles 
```
NET_RECEIVE(w, foo) {
     g = g + foo
}
```
just fine. Except it will use an undefined variable `foo` in the generated C++ leading to a compilation
error when building the native object.

This PR fixes this and generates an error instead
```
❯  modcc net_receive.mod
error trace:
  * net_receive.mod:(line 1,col 1) NET_RECEIVE does take at most one argument (Arbor limitation!)
```

### DIFF
--- a/modcc/module.cpp
+++ b/modcc/module.cpp
@@ -571,6 +571,10 @@ bool Module::semantic() {
 
     if (has_symbol("net_receive", symbolKind::procedure)) {
         auto net_rec_api = make_empty_api_method("net_rec_api", "net_receive");
+        // handle Arbor specifics
+        if (net_rec_api.second->args().size() > 1) {
+            error(pprintf("NET_RECEIVE does take at most one argument (Arbor limitation!)"));
+        }
         net_rec_api.first->body(net_rec_api.second->body()->clone());
         if (net_rec_api.second) {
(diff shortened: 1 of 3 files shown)


modcc/module.cpp
```diff
@@ -571,6 +571,10 @@ bool Module::semantic() {
 
     if (has_symbol("net_receive", symbolKind::procedure)) {
         auto net_rec_api = make_empty_api_method("net_rec_api", "net_receive");
+        // handle Arbor specifics
+        if (net_rec_api.second->args().size() > 1) {
+            error(pprintf("NET_RECEIVE does take at most one argument (Arbor limitation!)"), net_rec_api.first->location());
```
Helveg, line 576:
```suggestion
            error(pprintf("NET_RECEIVE takes at most one argument (Arbor limitation!)"), net_rec_api.first->location());
```

Better or nah?

brenthuisman (approved):

PR #2142, opened 2023-06-22
spack ci uses a stable spack branch instead of develop

bcumming:
Fix the failing spack github action.


thorstenhater:
Hey @bcumming,

spack seems still unhappy with that try?

bcumming:
Not happy at all.

brenthuisman:
It unfortunately doesn't fix the problem, so closing.

PR #2143, opened 2023-06-26
Fix GPU/NVC problems

thorstenhater:
- Fix GPU compilation of `ion-state`
- Rename `ls::restrict` -> `ls::restrict_to` as _some_ compilers might take offense
- Fix some(!) of the warnings issued by `nvc`

⚠️ Breaking ⚠️ 

`restrict` is now called `restrict-to`.


modcc/expression.hpp
```diff
@@ -294,7 +294,7 @@ class ARB_LIBMODCC_API IdentifierExpression : public Expression {
     }
 
     expression_ptr clone() const override;
-
+    using Expression::semantic;
```
boeschf, line 297: Seems strange that this would be needed - do you have any explanation why only for this specific override the base class' method needs to be pulled into scope?

modcc/perfvisitor.hpp
```diff
@@ -41,6 +41,7 @@ static std::ostream& operator << (std::ostream& os, FlopAccumulator const& f) {
 
 class FlopVisitor : public Visitor {
 public:
+    using Visitor::visit;
```
boeschf, line 44: again same thing: is this a bug in nvc++, or what do I overlook?

test/unit/test_morph_expr.cpp
```diff
@@ -664,39 +664,39 @@ TEST(region, thingify_moderate_morphologies) {
         EXPECT_TRUE(region_eq(mp, radius_gt(reg_c_, 2), cl{{0,0.55,0.7},{2,0,0.5},{3,0.1,0.375},{3,0.9,1}}));
         EXPECT_TRUE(region_eq(mp, radius_gt(reg_d_, 2), cl{{0,0.55,0.7},{2,0,0.5},{3,0.1,0.375},{3,0.75,0.9}}));
 
-        // Test restriction
+        // Test restrict_toion
```
boeschf, line 667:
```suggestion
        // Test restrict_to
```

modcc/expression.hpp
thorstenhater: No, this is an entirely correct complaint. The problem is that we have this situation
```cxx
struct Base {
  virtual void foo() {}
  virtual void foo(int) {}
};

struct Derived {
  void foo() override {}
};
```
which is overriding only one overload. The `using` declares we want to use the `Base::foo(int)`
but overload `Base::foo()`.

modcc/perfvisitor.hpp
thorstenhater: See above.

boeschf (approved):

thorstenhater:
Something weird happened here ... this PR branch no longer tacks my branch, but the corresponding one in Arbor.
Which I cannot push to..

boeschf (approved):

PR #2144, opened 2023-06-26
Update busyring to generate rings

bcumming:
The busyring benchmark was not generating perpetual spike chains, i.e. after generating one round of spikes on each ring, no more spikes were generated.

The reason was the `event-freq` parameter, which injected events at the start of each sub-ring at a fixed frequency, which was keeping the first cell in each ring in a state where the event from the last cell in the ring wouldn't trigger a spike. 

While root causing the issue, I refactored the cell generation code to simplify it.


example/busyring/parameters.hpp
```diff
@@ -33,13 +33,12 @@ struct ring_params {
     ring_params() = default;
 
     std::string name = "default";
-    unsigned num_cells = 10;
-    unsigned ring_size = 10;
-    double min_delay = 10;
-    double duration = 100;
+    unsigned num_cells = 5000;
```
thorstenhater, line 36: How long would you expect this to run on a reasnably modern desktop/laptop? 
An ideal runtime for people trying would -- for me -- be around 10s.

thorstenhater (commented):
Looks good, question is whether we should lower the default settings

example/busyring/parameters.hpp
bcumming: I want something that does an interesting amount of computation, and this is a pretty good sweet spot that won't take too long on the CI/CD

thorstenhater: If it doesn't bog down CI too much I am good

thorstenhater (approved):
I was convinced :D

PR #2148, opened 2023-07-04
SVE backend

boeschf:
This PR attempts to fix the bugs reported in issue #2101 and introduces a new fixed-sized (VLS) sve backend. This backend does away with the need for variable sized arrays and runtime calls to figure out the simd width. Later we may want to retire the original sizeless (VLA) sve backend in favor of the new one.

Changes:
- new sve backend with abi `vls_sve`: fixed-sized types, can be used wherever avx types and co can be used
- configure time detection of sve register size
- added pertinent compiler flags when sve is in use
- modcc now generates correct code for all backends, and generates a `constexpr simd_witdth_` (was previously not working with sve)
- fixed small bug in sve `exp` function
- updated tests


brenthuisman:
What is the significance of the cscs tests failing? Should they be passing?

boeschf:
No, ignore the cscs tests here - they fail because PR #2149 isn't merged yet, and I have the webhook enabled to run these tests for for me always (other users need to trigger them using a comment). Thus, it tries to run test scripts which aren't available yet.

thorstenhater (approved):
As far as I can tell this looks sane and good catch on the misused intrinsic in `exp`

Issue #2136, opened 2023-06-13
Strange looking doc section


Helveg:
Something seems off here: https://docs.arbor-sim.org/en/stable/concepts/simulation.html#configuring-data-extraction

Helveg:
Also the signature of `set_ion` looks funky: https://docs.arbor-sim.org/en/latest/python/cable_cell.html#arbor.cable_cell_global_properties.ion_reversal_potential

And the top right logo is broken for that page as well.

boeschf:
Thanks @Helveg for reporting. Should be fixed now.

PR #2152, opened 2023-07-17
Fix documentation 

boeschf:
Addresses #2136 and fixes some more warnings


doc/python/hardware.rst
```diff
@@ -193,6 +193,7 @@ The Python wrapper provides an API for:
         using :py:func:`~arbor.env.default_allocation()`.
 
     .. method:: context(threads=threads, gpu_id=gpu_id, mpi=mpi_comm, inter=mpi_comm)
+        :noindex:
```
brenthuisman, line 196: Why?

brenthuisman (approved):
1 Q:

doc/python/hardware.rst
boeschf: The method above has the same name (overload) and sphinx issues a warning that the method is already indexed.

Issue #2147, opened 2023-07-04
Track dependency versions based on Spack stable branch.


brenthuisman:
We could have our dependencies track the Spack stable branch. Since Spack's a major way Arbor is rolled out (CSCS, JSC, Ebrains), it has an obvious benefit.

brenthuisman:
All our submodules are hosted on github. Which means the following is an idea for automation:

- install latest Spack release
- spack info $packagename
- E.g. pugixml:

```
Preferred version:  
    1.13      https://github.com/zeux/pugixml/releases/download/v1.13/pugixml-1.13.tar.gz
```
- From this you can derive the https://github.com/zeux/pugixml/releases/v1.13 URL, which contains a commit id.
- Can we do this without parsing HTML?

brenthuisman:
For reference: `spack list --format=version_json` doesn't give all the info that we need.

brenthuisman:
Related: https://github.com/arbor-sim/arbor/issues/1731

brenthuisman:
CI now tests deps provided by both Spack stable and develop: https://github.com/arbor-sim/arbor/blob/master/.github/workflows/test-spack.yml

PR #2153, opened 2023-07-17
Dependencies: should we update submodules? When to bump versions?

boeschf:
Appease issue #2146

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,14 +7,15 @@ on:
 jobs:
   build:
     name: Lint
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
     steps:
+      - uses: actions/checkout@v3
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: 3.7
+          python-version: '3.10'
       - name: Get packages
         run: |
           python -m pip install --upgrade pip
--- a/doc/concepts/interconnectivity.rst
+++ b/doc/concepts/interconnectivity.rst
@@ -115,11 +115,11 @@ similarly
                                     weight,
                                     delay)]]
 
-Note that Arbor now recognizes two sets of ``GID``: An external and an internal
+Note that Arbor now recognizes two sets of ``GID``\: An external and an internal
 set. This allows both Arbor and the coupled simulation to keep their own
 numbering schemes. However, internally Arbor will tag external cells and spikes
-by setting their ``GID``s'  most significant bit. This _halves_ the effecively
-available ``GID``s.
+by setting their ``GID``\s'  most significant bit. This _halves_ the effecively
+available ``GID``\s.
 
 To consume external spike events, a specialised ``context`` must be created by
 calling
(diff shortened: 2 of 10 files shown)


brenthuisman:
Have you checked if these versions are available on Spack?

brenthuisman (changes requested):
- The commit ID for pybind11 does not match the version in Spacks `package.py`.
- atm, some versions here are surpassing the latest available in the latest Spack release I think. Even if there's no automated testing/updating, let's add this rule to `doc/contrib/dependency-management.rst`
- google-benchmark (in Spack: `benchmark`) isn't specified in `package.py`. Should we?
- `doc/dependencies.csv` wants some attention :)

boeschf:
> * The commit ID for pybind11 does not match the version in Spacks `package.py`.

done

>     * atm, some versions here are surpassing the latest available in the latest Spack release I think. Even if there's no automated testing/updating, let's add this rule to `doc/contrib/dependency-management.rst`

I checked with latest spack release, should be ok
the rule is already present:  `#. ``doc/dependencies.csv``, git submodules and ``spack/package.py`` shall be in sync.`

>     * google-benchmark (in Spack: `benchmark`) isn't specified in `package.py`. Should we?

It is always built from source/submodule right now. Only needed for target `ubenches` which is meant for devs only

>     * `doc/dependencies.csv` wants some attention :)

Not sure what to put there - I can fill in the current versions but those are for sure not the minimum requirements

boeschf:
strictly speaking those module updates are not necessary for Arbor to work. Thinking about closing this PR as not needed. What are your thoughts?

brenthuisman:
We are doing two different things at once with the Spack job, which should probably be separated:

1. We would like to ensure we build with certain (recent) versions to detect breakage early.
2. We want to build with the same version as the submodules, so as to not be surprised by one job breaking and the other not. 
3. We want to provide a `package.py` that's no stricter than necessary on the deps. E.g. some deps don't have version requirements, which, if true, is better for people building specs.

For reference: we have run into (pybind11) errors not showing up in our CI, because we had different version between a submodule and the Spack minver. That took some time to discover. But, you are right,`package.py` is supposed to be as loose as possibly to make life of specs builders easy, and not really the place to do this sort of testing.

Proposal:
1. Spack prefers building the latest version, so divergence between CI using the submodule and Spack already exists. Spack reuses what's in cache, which we enabled, but we can force fetching most recent with `--fresh`. So let's do that.
2. We bump the submodule and Spack dep in tandem, and never past Spack latest release, and only if there's demonstrated breaking or changes that we require.
3. Some Spack deps are unversioned, and but the Git submodules are. We leave it as-is, until there is reason to bump, and then they'll be brought in line.

This results in us testing the lowest known working versions of the submodules in most of our CI, and the Spack CI tests with a reasonably current version (as per Spack releases), so we have some notification of breakage over different versions.

Actions:
1. Reject updates in this PR.
2. Disable submodule updates bot.
3. Update `doc/contrib/dependency-management.rst` with this proposal.
4. Add `--fresh` flag to Spack job.
5. Close #2147 and #1731

I can fold 3 and 4 in my Spack updates branch.

brenthuisman:
Another idea:

Force submodule-use for header-only deps (which is all but random123?). This isolates us and other builders/integrators from having to worry about version incompatibilities.

brenthuisman:
I've moved this discussed to the relevant issue, and I'll close this PR pending a decision.

https://github.com/arbor-sim/arbor/issues/1731

PR #2154, opened 2023-07-17
fix ODR violation

boeschf:
While linking unit tests, examples, pyarb I get the following warnings about ODR violations:
```
arbor/arbor/util/maputil.hpp:21:6: warning: ‘keys’ violates the C++ One Definition Rule [-Wodr]
   21 | auto keys(Seq&& m) {
      |      ^
arbor/arbor/util/maputil.hpp:21:6: note: return value type mismatch
   21 | auto keys(Seq&& m) {
      |      ^
arbor/arbor/util/range.hpp:40:8: note: type ‘struct range’ defined in anonymous namespace cannot match across the translation unit boundary
   40 | struct range {
      |        ^
arbor/arbor/util/maputil.hpp:21:6: note: ‘keys’ was previously declared here
   21 | auto keys(Seq&& m) {
      |      ^
arbor/arbor/util/maputil.hpp:21:6: note: code may be misoptimized unless ‘-fno-strict-aliasing’ is used
```
This seems related to the return type deduction of two function templates with `auto` return type which return a `range` template instantiation. The type of the latter is deduced based on a `constexpr if` expression. I am not entirely sure where the ODR violation happens, but by using trailing return types for the former functions the problem is avoided.


thorstenhater (changes requested):
I am seriously confused by the fix. We ask the compiler to deduce via `auto` which
can only work on the returned value, which seemingly breaks ODR. But now asking
it to deduce the _same type_ in the _same fashion_ just with _other words_ is fine?

I suggest adding a comment about this nonsensical behaviour with the request to check
again in the future.

boeschf:
Yes it is quite nonsensical. I checked this again today and it turns out that the warning is only emitted  when I compile Arbor with `-DBUILD_SHARED_LIBS=ON` and have `LTO` enabled.

thorstenhater:
That's even more hilarious.

PR #2156, opened 2023-07-19
Spack-solvable versions, add tests to Spack workflow

brenthuisman:
## Changes

1. Spack interprets version info consisting of 2-tuples as ranges, leading to v0.8 being uninstallable if there has been a point release (v0.8.1).
2. Change Spack CI to use latest release _and_ develop branch, because the former is how Arbor is rolled out in various centers, and thus should be verified to work, and the latter lets us know of upcoming issues with dependencies.
3. Changed Spack CI to not use `build_spack_package.sh`. This script does everything in `/tmp`, and Github Actions appear to be wiping `/tmp` between steps, which means breaking up the run into steps is difficult. Breaking up is desirable, because then one can see whether building or testing breaks.
4. Run Python tests after building the Spack package.
5. Also use Spack-provided deps when Spack is instructed to build master, not submodules.
6. Tiny fix for the `ciwheelbuild` failure.

## Future issues

Out of scope for now, but identified as problems:

1. Currently, Spack cannot execute the unit tests, because the Google-Test package isn't built and Arbor doesn't look for it. @boeschf is looking at the latter: making it possible to build with an external Google Test.
    - Will be solved by https://github.com/arbor-sim/arbor/issues/1731 
2. The former is another problem: Spack does not or not always (it's not clear) concretize test dependencies. This is an open issue with low priority: https://github.com/spack/spack/issues/29447
  This mean that despite oru `package.py` properly including `depends_on("googletest@1.12.1", type="test", when="@0.7.1:")`, Spack does not build it, at least not in the dev-build workflow as we currently do. And it isn't clear what an always working workaround is, see issue. (Spack can't display the deps for the `test` variant: https://github.com/spack/spack/issues/23586 `spack spec -I arbor` doesn't show Google Test, and neither does `spack dependencies --deptype test arbor`.)
    - Blocked by upstream.


.github/workflows/spack.yml
```diff
@@ -25,18 +27,49 @@ jobs:
...
+          key: spack-cache-${{ matrix.os }}-${{ matrix.spack-version }}-${{ github.run_id }}
+          restore-keys: spack-cache-${{ matrix.os }}-${{ matrix.spack-version }}-
+      - name: Get Spack (develop)
+        if: ${{ matrix.spack-version == 'develop' }}
+        run: |
+          git clone -c feature.manyFiles=true --depth 1 https://github.com/spack/spack.git
+      - name: Get Spack (latest_release)
+        if: ${{ matrix.spack-version == 'latest_release' }}
+        run: |
+          wget "$(curl -sH "Accept: application/vnd.github.v3+json" https://api.github.com/repos/spack/spack/releases/latest | grep browser_download_url |  cut -d '"' -f 4)"
```
thorstenhater, line 39: wuhu.

doc/contrib/release.rst
```diff
@@ -126,7 +126,7 @@ Post Release
 
 #. Update Zenodo with authors and changelog created in previous step and submit.
 
-#. Make a new PR setting ``VERSION`` to the next with a trailing ``-dev``. E.g. if you just release ``3.14``, change ``VERSION`` to ``3.15-dev``
+#. Make a new PR setting ``VERSION`` to the next with a trailing ``-dev``. E.g. if you just released ``3.14.15``, change ``VERSION`` to ``3.15.16-dev``. Make the number portion always consists of a 3-tuple. Shorter versions are uninstallable by Spack (``spack install arbor@0.8`` will install v0.8.1, due to anything short than a 3-tuple being interpreted as a version range).
```
thorstenhater, line 129:
```suggestion
#. Make a new PR setting ``VERSION`` to the next with a trailing ``-dev``. E.g. if you just released ``3.14.15``, change ``VERSION`` to ``3.15.16-dev``. Make the number portion always consists of a triplr. Shorter versions are uninstallable by Spack (``spack install arbor@0.8`` will install v0.8.1, due to anything shorter than a triple being interpreted as a version range). 
```

thorstenhater (changes requested):
Great, some wording though (and I am scared of `curl | bash`)

.github/workflows/spack.yml
brenthuisman: 1. I understand, BUT:
2. We were already doing this (see `build_spack_package.sh`)
3. It is done through a GitHub API, not some random website. It's the recommended way to get release artifacts.

doc/contrib/release.rst
```diff
@@ -126,7 +126,7 @@ Post Release
 
 #. Update Zenodo with authors and changelog created in previous step and submit.
 
-#. Make a new PR setting ``VERSION`` to the next with a trailing ``-dev``. E.g. if you just release ``3.14``, change ``VERSION`` to ``3.15-dev``
+#. Make a new PR setting ``VERSION`` to the next with a trailing ``-dev``. E.g. if you just released ``3.14.15``, change ``VERSION`` to ``3.15.16-dev``. Make the number portion always consists of a triple. Shorter versions are uninstallable by Spack (``spack install arbor@0.8`` will install v0.8.1, due to anything shorter than a triple being interpreted as a version range).
```
thorstenhater, line 129:
```suggestion
#. Make a new PR setting ``VERSION`` to the next with a trailing ``-dev``. E.g. if you just released ``3.14.15``, change ``VERSION`` to ``3.15.16-dev``. Make sure the number portion always consists of a triple. Shorter versions are uninstallable by Spack (``spack install arbor@0.8`` will install v0.8.1, due to anything shorter than a triple being interpreted as a version range). 
```
🤓

thorstenhater (approved):

thorstenhater (approved):
🎉

thorstenhater (approved):
👍

PR #2157, opened 2023-07-20
Python3.12: Replace `distutils.version.LooseVersion`

penguinpee:
Python3.12 dropped distutils.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ changelog = "https://github.com/arbor-sim/arbor/releases"
 [build-system]
 requires = [
     "scikit-build-core",
+    "looseversion",
 ]
 build-backend = "scikit_build_core.build"
 
--- a/scripts/tsplot
+++ b/scripts/tsplot
@@ -11,7 +11,7 @@ import matplotlib as M
 import matplotlib.pyplot as P
 import numbers
 from functools import reduce
-from distutils.version import LooseVersion
+from looseversion import LooseVersion
 from itertools import chain, islice, cycle
 
 # Run-time check for matplot lib version for line style functionality.


penguinpee:
This is the first step in an attempt to update arbor to 0.8.1 in Fedora. Fedora recently introduced Python3.12 in rawhide. For the current release 0.8.1 the patch is slightly different:

```patch
diff --git a/pyproject.toml b/pyproject.toml
index ec34cf9a..671f10ef 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,6 +53,7 @@ requires = [
     "scikit-build",
     "cmake>=3.19",
     "ninja",
+    "looseversion",
 ]
 build-backend = "setuptools.build_meta"
 
diff --git a/scripts/tsplot b/scripts/tsplot
index c64e7b18..10ddf3f4 100755
--- a/scripts/tsplot
+++ b/scripts/tsplot
@@ -11,7 +11,7 @@ import matplotlib as M
 import matplotlib.pyplot as P
 import numbers
 from functools import reduce
-from distutils.version import LooseVersion
+from looseversion import LooseVersion
 from itertools import chain, islice, cycle
 
 # Run-time check for matplot lib version for line style functionality.
```

I can provide a PR based on the release tag if you like.

penguinpee:
An alternative to introducing `looseversion` as anew dependency would be:

```patch
--- a/scripts/tsplot
+++ b/scripts/tsplot
@@ -11,7 +11,7 @@ import matplotlib as M
 import matplotlib.pyplot as P
 import numbers
 from functools import reduce
-from distutils.version import LooseVersion
+from setuptools._distutils.version import LooseVersion
 from itertools import chain, islice, cycle
 
 # Run-time check for matplot lib version for line style functionality.
```

I'm not sure which is better or more (backwards) compatible.

brenthuisman:
I have a much better idea: remove the check entirely. I don't think we need compatibility with matplotlib<1.5.0. Would that solve this issue?

brenthuisman:
We're about to release v0.9 by the way, so if you have more (we've not tested with Py3.12 yet), it would be good to get them in ASAP-ish :)

penguinpee:
> I have a much better idea: remove the check entirely. I don't think we need to check for ancient matplotlib anymore. Would that solve this issue?

I think it would.

> We're about to release v0.9 by the way, so if you have more (we've not tested with Py3.12 yet), it would be good to get them in ASAP-ish :)

Well, I just tried a one shot on building 0.8.1. That failed on something related to `ext-gtest`. I'm not too familiar with arbor. So, I didn't investigate further. It looked like a missing dependency. But it could easily be something related to our build environment.

Is there a certain tag / branch, I can check out to give 0.9 a test drive?

brenthuisman:
You can check out the master branch for the basis for our next release. There are a few PRs tagged with milestone v0.9, but nothing that touches dependencies afaik.

brenthuisman:
The code was deleted with #2158, so I'll close this PR. If there are any other modification you need for Py3.12, please let us know!

PR #2160, opened 2023-07-31
Break out pip and doc tests, exclude doc-change from eating up precious runner cycles

brenthuisman:
- Break out pip and doc tests from `test-everything.yaml`
- Only trigger docs build on docs changes, and not trigger code-CI on docs-only changes.
- Remove monthly submodule check. This will be handled differently, see #1731
- Enable Pypy wheel builds (they pass testing, see https://github.com/brenthuisman/arbor/actions/runs/5725359772)


thorstenhater (approved):
Lovely! Splitting the CI monolith is a good step!

Issue #952, opened 2020-02-04
Python: unify `__repr__` and `__str__`


akuesters:

Helveg:
Should they be?. `__repr__` serves to generate a string that can represent the object when passed to `eval()`, while `__str__` simply converts it to a string. This is best illustrated by this example:

```
>>> repr('r')
"'r'"
>>> str('r')
'r'
```

brenthuisman:
Are there any cases in which we want a `eval`able string representation? Off hand I can only think of `cable_cell` and components, `label_dict`. I'd be in favor of not implementing/removing `__repr__` for all types except those.

brenthuisman:
We should *not* unify `__repr__` and `__str__`, because they are meant for different purposes: the former must provide a full description of the object such that it can be recreated in full if passed to `eval()`. `__str__` is a way to show some info about the object instance. After our groupmeet on march 16, I'm updating this issue with the following:

* Python: Go over all `__str__` implementations and insure they print something informative about their class.
* C++: Something similar. Implement `ostream& operator<<(ostream& os, const Type& type)` for all Arbor classes and structs?
* Define in C++ land and simply wrap in Python.

Helveg:
Important note: Make it **clear** that it is still a print of an object, and make them unique even when their values used to inform you are the same! You can use Python's `id` function and follow `object`s `__str__` implementation to print strings like:

```
<arbor.recipe with 15 cells at 0x0045e44>
```

Otherwise it gets very confusing very quickly when debugging

brenthuisman:
Good point, I hadn't thought of that, even though I've run into this countless times ;) To avoid duplication, I'm probably going to pass `&obj` to `__str__`.

brenthuisman:
We have `arbor/util/strprintf.hpp` and `python/strprintf.hpp`. The latter seems an expansion of the former. I think it would be nice to remove this duplication.

Even nicer would be to use a library like https://github.com/fmtlib/fmt. Is there anything not covered in this lib? `strprintf.hpp` is a bit underdocumented and I'm struggling to find out what uses it has. `f"{}"` like functionality, but maybe more.

halfflat:
Historically, we haven't used `fmt` because it was an external dependency which we thought we didn't need, and besides, we'll essentially get it for free with C++20. That said, as far as external dependencies go, I understand that `fmt` is exceptionally well behaved.

[...]

PR #2168, opened 2023-08-03
Remove strprintf printing for {fmt}.

thorstenhater:
Use libfmt in more place.


arbor/CMakeLists.txt
```diff
@@ -135,6 +135,24 @@ add_library(arbor ${arbor_sources} ${arbor-builtin-mechanisms})
 target_link_libraries(arbor PRIVATE arbor-private-deps arbor-private-headers)
 target_link_libraries(arbor PUBLIC arbor-public-deps arbor-public-headers)
 
+if (ARB_USE_BUNDLED_FMT)
```
brenthuisman, line 138: Shouldn't this live in `ext/CMakeLists.txt`?

arbor/fvm_layout.cpp
```diff
@@ -29,10 +29,9 @@
 #include "util/unique.hpp"
 #include "util/strprintf.hpp"
```
brenthuisman, line 30: Here and elsewhere: remove the `util/strprintf.hpp` header?

arbor/util/strprintf.hpp
brenthuisman, line 1: Is the remainder still used anywhere? It is difficult to kill that usage too?

brenthuisman: Tests I see, for `to_string`. What's wrong with `std::to_string()`?

brenthuisman (changes requested):

arbor/fvm_layout.cpp
thorstenhater: No, we need it for the ADL-ized `to_string`.

arbor/util/strprintf.hpp
thorstenhater: Glad  you asked :D. `util::to_string` uses ADL (=argument dependent lookup) to select 
`to_string` if present in namespace global or `arb`. If not it'll try to use `<<` via `std::stringstream`.
`std::to_string` does not. This is why many classes in Arbor produce a `friend <<` but do not 
specialise `to_string`.

brenthuisman: Hmm, this is all a bit implicit. We should document that, for instance, all Arbor objects implement a public `operator<<` (as per https://github.com/arbor-sim/arbor/issues/952). Maybe `operator<<` isn't really what we want, I suppose often you want to have the string output, so we could say that all Arbor objects also implement `as_string()` and that it has the same output as `operator<<`. What do you think?

thorstenhater: I just learned that bit, so yes, a bit hidden. We can add something like this to our devdocs:
> Arbor classes _should_ implement `<<` for debugging and user interface purposes.

Should be a _should_ since it might never be needed and can be retrofitted easily.
I'd separate both concerns from this PR though, which is purely swapping out the substrate. 

Regarding stream ./. strings: 
- Strings do allocate memory (repeatedly!), thus recursively stringifiing classes leads to potentially bad blowup, while streaming data out just needs a subset of the memory. 
- The memory concern will be especially bad for arrays/containers.
- Streams have the issue of indirection via stringstream which `util::to_string` packages for you.

So, despite the slight convenience, I prefer the `<<` way.

thorstenhater: So, if the output should (MUST!) be the same, we end up with `as_string(t) = (stringstream{} << t).str()`
which is what `util::to_string` does anyhow. Also `util::to_string` _falls back_ to `std::to_string` where 
available (hence the ADL dance). See here

https://gcc.godbolt.org/z/r8KMEzYcj

- `Alice` provides `to_string` and `<<`
- `Bob` has just `<<`
- `double` has `std::to_string`.

Thus, everybody will just be repeating `util::to_string` anyhow.

brenthuisman: Sometimes inheritance from a base class is nice.

/ducks

brenthuisman (dismissed):
OK

arbor/util/strprintf.hpp
thorstenhater: Yeeees .... but `(std::)to_string` is standalone function and not a member or class.

brenthuisman:
Then mark it draft, I thought we had agreed to do that?

arbor/util/strprintf.hpp
brenthuisman: Yeeees ... but `.print()` could just be an inherited method containing the string construction from `<<` once and for all, without the separate-but-not-really function in `util`.

thorstenhater:
Closing as it diverged too much!

PR #2202, opened 2023-08-09
Bump to v0.9.0

brenthuisman:


pyproject.toml
```diff
@@ -42,7 +43,7 @@ wheel.packages = []
 
 [tool.black]
 line-length = 88
-target-version = ['py37', 'py38', 'py39', 'py310', 'py311']
+target-version = ['py37', 'py38', 'py39', 'py310', 'py311', 'py312']
```
thorstenhater, line 46: Do we want to ditch 3.7 here? Numpy stopped supporting it in Dec'21
https://endoflife.date/numpy

thorstenhater: See also in other python versions.

CHANGELOG.md
```diff
@@ -1,3 +1,111 @@
...
+* Arbor now supports checkpointing and resume from a previously stored checkpoint. Useful when a certain point in time needs to be explored in multiple directions, when you want to rewind to a previous state, etc. [See documentation for more.](https://docs.arbor-sim.org/en/latest/fileformat/serdes.html)
+* More painted parameters are now scalable through `iexpr`: temperature, capacitance, resistivity, membrane potential and the following ionic parameters: internal and external concentration, diffusivity, and reversal potential. [See documentation.](https://docs.arbor-sim.org/en/latest/concepts/decor.html#scaling-mechanism-and-membrane-parameters)
+* A set of ANN activation functions for NMODL have been added: `sigmoid(x)`, `relu(x)` and `tanh(x)`. Control volume `area` is exposed through NMODL.
+* A revamped backend for ARM CPU's that support Scalable Vector Extension (SVE). Arbor and modcc are now fully compatible, so users who have access to A64FX-based HPC can take full advantage of that hardware.
+
```
thorstenhater, line 14: I'd actually add `fixed-dt` here, maybe even as the top one.

CHANGELOG.md
```diff
@@ -1,3 +1,111 @@
...
+## Breaking changes since v0.8.1
+
+* Version bumps:
+  * CUDA 11 or higher is required for GPU builds.
+* Contexts are now constructed kwargs-only. So, `arbor.context(12, None)` is now `arb.context(threads=12, gpu=None)`. In the case no arguments are supplied, the context initialized to `default_allocation`, which means it will use the total number of threads and first GPU that are locally available. `arbor.simulation` constructed without an explicit `arbor.context` also initializes a `default_allocation`.
+* `set_binning_policy` has been removed due to the switch to a fixed dt.
```
thorstenhater, line 20: This needs more elaboration, possibly farther up.

thorstenhater (changes requested):

pyproject.toml
brenthuisman: I don't. We don't really depend on a particular version of numpy, so there's no real need to deprecate, as long as it is recentish. Plus, Google Colab is still stuck on 3.7, which is/was a goto for some to get started with Notebooks. As long as it works, I'd prefer to keep that option open. I'm guess next Arbor release things will be different anyhow.

CHANGELOG.md
thorstenhater: Then it should be a breaking change with more explanation ;)

pyproject.toml
thorstenhater: Colab -- while true -- is a good reason not to.

CHANGELOG.md
```diff
@@ -1,3 +1,111 @@
...
+* More painted parameters are now scalable through `iexpr`: temperature, capacitance, resistivity, membrane potential and the following ionic parameters: internal and external concentration, diffusivity, and reversal potential. [See documentation.](https://docs.arbor-sim.org/en/latest/concepts/decor.html#scaling-mechanism-and-membrane-parameters)
+* A set of ANN activation functions for NMODL have been added: `sigmoid(x)`, `relu(x)` and `tanh(x)`. Control volume `area` is exposed through NMODL.
+* A revamped backend for ARM CPU's that support Scalable Vector Extension (SVE). Arbor and modcc are now fully compatible, so users who have access to A64FX-based HPC can take full advantage of that hardware.
+
+## Breaking changes since v0.8.1
+
+* The ability to set binning and sampling policies have been removed, due to Arbor now having a fixed timestep. Removing exact delivery increases the speed of the simulation due to elimination of small steps, makes the numerics independent of presence of sampling, and also leads to a number of code simplifications.
```
thorstenhater, line 17:
```suggestion
* The ability to set binning and sampling policies has been removed, due to Arbor now having a fixed timestep. Removing exact delivery increases the speed of the simulation due to elimination of small steps, makes the numerics independent of presence of sampling, and also leads to a number of code simplifications.
```

thorstenhater (changes requested):
Grammar then good to go.

thorstenhater (dismissed):

thorstenhater (approved):

Issue #2145, opened 2023-06-26
Flaws in the diffusion process


jlubo:
There seem to be several flaws in the diffusion process (cf. the [requirements](https://github.com/arbor-sim/arbor/issues/2084#issuecomment-1498037774) here): 

1. the number of particles does not stay constant in the absence of particle injection and removal
2. the equilibrium concentration depends on the spatial discretization
3. negative concentrations can occur, which is not possible in physical reality

These points can be demonstrated with the attached [example](https://github.com/arbor-sim/arbor/files/11872731/diffusion_overview_2023-06-26.zip) (tested with master version d1139b7).

Point 1 is shown in the figures below by the dashed line termed "sum", which is the sum over the concentrations of all compartments weighted by their volume. The sum should exhibit a discontinuous increase upon each particle injection (or decrease upon each removal) and otherwise stay constant. Its final value should be 1500, because of twice an injection of 1000 particles and once a removal of 500 particles. The outcome does not match this assumption.

Point 2 can be seen from comparing the two figures below. For the first figure, the two dendrites each consisted of 3 compartments (CVs). For the second figure, no such fine discretization was used and each dendrite as well as the soma consisted of one compartment only. The final values of the concentrations do not match, which is opposed to the expectation.

Point 3 should become obvious from the top right plot in both figures.

![multi_compartment_dendrite_synapses_probed](https://github.com/arbor-sim/arbor/assets/33398515/4bfae3cd-3145-4dc7-8f50-094573f877e3)

![multi_compartment_dendrite_synapses_probed_no_fine_discr](https://github.com/arbor-sim/arbor/assets/33398515/99ee032b-8d24-438b-becc-d7c84d17c18b)

thorstenhater:
Hi @jlubo,

thanks for reporting. I sadly cannot run the example on current master for two reasons:
- `area` is being set from external, which is refused since `area` is built-in (since a few commits back)
- removing the settings result in further errors in the plotting code. 

Could try this with current master, potentially the area calculation is off. That might be also causing issue 2.

For the negative values, as far as I understand the issue, this can only happen if a mechanism directly accesses
`Xd` and doesn't check for underflow? In some interpretations of `Xd` this might be considered part of the design
(ie a fluctuation around the mean). Is that correct, or am I missing something here? What semantics would you have 
expected? 

As for the conservation issues: I cannot extract the injected values from the plots. In the second, the deviations seem
[...]

Issue #2084, opened 2023-01-26
Units and behavior of diffusive particles


jlubo:
In the [documentation](https://docs.arbor-sim.org/en/latest/fileformat/nmodl.html#units), `mmol/L` is mentioned as the unit for `ion X diffusive concentration`, but it is not stated what it refers to (e.g., to point mechanisms, as it is stated for other quantities on the same documentation page). More importantly, the unit of the amount of diffusive particles is neither stated in the documentation nor does it immediately become clear.

Here is an example to specify the issue. Consider a compartment with the surface area `A = 2⋅π⋅radius⋅length`. To get from a point measurement of the diffusive particle concentration `Xd` to the amount of particles on the whole surface of the compartment, we find that one needs to compute `Xd⋅A⋅10^9`. Then, with `Xd`  in units of `mmol/L`, `A` in units of `m^2`, and `10^9` in units of `µm`, the unit of the amount of diffusive particles would be `µmol`, which seems fine. Nevertheless, the exact definition of the amount of particles (and thus its biophysical meaning) remains unclear because the origin and interpretation of the huge factor `10^9 µm` is unclear. I guess, the problem may originate from the fact that the concentration is defined 'per volume' while all the dynamics currently considered by Arbor happen on a surface.

Furthermore, it is not clear from the documentation how the particle diffusion behaves in general (but some information has already been provided in [this PR](https://github.com/arbor-sim/arbor/pull/1729)).

To summarize, it would be nice to have a thorough description of the behavior of the particle diffusion in the documentation. Moreover, fixing the unit of the amount of particles (perhaps `µmol`) is likely necessary, and its definition should be mentioned in the documentation.

(edited)

thorstenhater:
So, on the behaviour the section you quoted says this:
> To enable diffusion of ion species along the morphology (axial diffusion) one sets the per-species diffusivity to a positive value. It can be changed per region and defaults to zero. This is strictly passive transport according to the diffusion equation X' = ß ∆X where X is the species’ diffusive concentration and ß the diffusivity constant.

and the NMODL docs teach how to manipulate the quantity X from a mechanism. The developer section talks about
the mathematical details, namely that X basically follows the cable equation.

It's a simple process so I couldn't think of anything else of interest.

jlubo:
Thanks @thorstenhater. Regarding the mathematical details, do you mean [this document](https://github.com/arbor-sim/arbor/tree/master/doc/math/model)?

[...]

PR #2209, opened 2023-08-15
Tests for particle diffusion

jlubo:
Python tests are added to ensure the working of the diffusion of arbitrary particles in setups with 1-3 segments (cf. issues https://github.com/arbor-sim/arbor/issues/2084 and https://github.com/arbor-sim/arbor/issues/2145). 

Besides showing that the diffusion seems to work if segments have the same radius, the tests that are added here also point to the fact that the diffusion across segments of different radius is still erroneous. Therefore, I left the tests for different radii commented until https://github.com/arbor-sim/arbor/issues/2145 has been resolved.


thorstenhater:
Hey @jlubo,

thanks for adding this. Could you run `black .` and push the results? That'll make the Linter
check pass.

python/test/unit/test_diffusion.py
```diff
@@ -0,0 +1,315 @@
...
+        morph = A.morphology(tree)
+
+        # ---------------------------------------------------------------------------------------
+        # decorate the morphology with mechanisms
+        dec = A.decor()
+        if num_segs < 3:
+            dec.discretization(
+                A.cv_policy(f"(fixed-per-branch {num_segs*num_cvs_per_seg} (branch 0))")
+            )
+        elif num_segs == 3:
```
thorstenhater, line 189: Does this result in different numbers of CVs? It seems like the proposed
morphology produces either one linear or  three branches in Y-shape. In both cases you're setting `num_cvs_per_seg` on all branches. Thus, this should be equivalent to just setting `fixed-per-branch`?!

thorstenhater (changes requested):
Hey, 
made some comments for improvements, rest looks great, thanks.

python/test/unit/test_diffusion.py
jlubo: Thanks for pointing that out. Indeed the branch specification is unnecessary here, I've removed it from the code. But the distinction between `num_segs < 3` and `num_segs == 3` still is necessary because for less than three segments, there is only one branch.

.gitignore
```diff
@@ -105,3 +105,6 @@ _skbuild
 
 # generated by test scripts
 results
+
+# environment setting script
+set_arbor_env*
```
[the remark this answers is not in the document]
jlubo: It's a file that I usually use to set the Arbor environment, I guess it can't be considered common :D

python/test/unit/test_diffusion.py
```diff
@@ -0,0 +1,315 @@
...
+"""
+Tests for the concentration and amount of diffusive particles across time and morphology.
+Three different morphological structures are considered: 1 segment ("soma only"), 2 segments
+("soma with dendrite"), and 3 segments ("soma with two dendrites").
+
+NOTE: Internally, Arbor only knows concentrations. Thus, particle amounts have to be computed
+      from concentrations by integrating over the volume of the morphology. The total amount
+      of particles should be conserved unless there is deliberate injection or removal of
+      particles.
+"""
```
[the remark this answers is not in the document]
jlubo: :)

jlubo:
> Hey, made some comments for improvements, rest looks great, thanks.

Thanks @thorstenhater! I've revised the code according to the comments.

python/test/unit/test_diffusion.py
thorstenhater: True. Question though: Do you want a fixed number/branch or really a fixed $\Delta x$?

ext/fmt
thorstenhater, line 1: I'd recommend `git fetch --all && git merge origin/master && git submodule sync` ... changing submodules in a feature PR is likely not what you want.

ext/pugixml
thorstenhater, line 1: Same as above.

thorstenhater (changes requested):

python/test/unit/test_diffusion.py
jlubo: Yeah, I wanted to control the number of CVs, however, I guess here it shouldn't matter which quantity is kept fixed.

ext/fmt
jlubo: Thanks. I must have missed that they were in.

python/test/unit/test_diffusion.py
thorstenhater: That seemed a bit unusual and so I am asking.

thorstenhater (approved):
Thank you very much!

thorstenhater:
cscs-ci run daint-gpu

thorstenhater:
Uh-oh
```
subprocess.CalledProcessError: Command 'make' returned non-zero exit status 2.
 Error:
/tmp/tmpqgxm6fo0/build/generated/diffusion/neuron_with_diffusion_gpu.cu(59): warning #177-D: function "arb::diffusion_catalogue::<unnamed>::multiply" was declared but never referenced
/tmp/tmpqgxm6fo0/build/generated/diffusion/synapse_with_diffusion_gpu.cu(74): error: identifier "n_" is undefined
/tmp/tmpqgxm6fo0/build/generated/diffusion/synapse_with_diffusion_gpu.cu(55): warning #177-D: function "arb::diffusion_catalogue::<unnamed>::multiply" was declared but never referenced
1 error detected in the compilation of "/tmp/tmpqgxm6fo0/build/generated/diffusion/synapse_with_diffusion_gpu.cu".
make[2]: *** [CMakeFiles/diffusion-catalogue.dir/build.make:183: CMakeFiles/diffusion-catalogue.dir/generated/diffusion/synapse_with_diffusion_gpu.cu.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/diffusion-catalogue.dir/all] Error 2
```

jlubo:
The different-radii and different-length tests are now uncommented. They should succeed if they are run with this new fix in the diffusion solver: https://github.com/thorstenhater/arbor/commit/7577092f89e9b54d8fde26ac6963a7c9965e43cc

thorstenhater:
Hey @jlubo,
 
could you leave them disabled, I'll add them to the to-be-made PR for the fixes.

jlubo:
Done :)

thorstenhater:
I'll fix the GPU problem in my follow-up #2226

python/test/unit/test_diffusion.py
```diff
@@ -351,31 +350,59 @@ def simulate_and_test_diffusion(
             delta=self.dev * sV_tot_max_expected,
         )  # max_{t}(s⋅V) [direct]
 
+        # additional tests for the case that there is only one segment of fixed radius
+        if num_segs == 1:
+            self.assertAlmostEqual(
```
thorstenhater, line 355: In the varying radius case: Can't we estimate the equilibrium concentration as:
$N = \sum\limits_{i\in C} V_i Xi(t=0)$ where $C$ is the set of CV indices
and 
$X_i(t\rightarrow\infty) = \frac{N}{V_i}$
with 
$X_i(t\rightarrow\infty) = X_j(t\rightarrow\infty)\quad\forall i,j\in C$?

thorstenhater (changes requested):
Nice! See also downstream #2226. Two questions:
1. It seems like lots of special casing for the different setups (num_sgs=1,2,3). Do you think it would be cleaner to split these into different tests altogether?
2. Question also inline: Can't we always compute the equilibrium concentration?

jlubo:
>     1. It seems like lots of special casing for the different setups (num_sgs=1,2,3). Do you think it would be cleaner to split these into different tests altogether?

I agree, I've now split the function `get_morph_and_decor()` into into individual functions for 1, 2, and 3 segments.

>     2. Question also inline: Can't we always compute the equilibrium concentration?

Yes we can! I've introduced a more general computation of CV volumes making this possible now.

python/test/unit/test_diffusion.py
jlubo: Yes! I've addressed this now, also see the comment below.

boeschf:
cscs-ci run daint-gpu

thorstenhater:
Hey @jlubo,

could I close this? It's subsumed in #2226.

jlubo:
Yes, sure. I'll check #2226 again in the next days.

PR #2213, opened 2023-08-24
Use Labels for Probes instead of Numerical Ids

thorstenhater:
# Overview

This rectifies a longstanding exception from Arbor's scheme of `(gid, "label")` to address 
point-like (or set-of-point-like) entities. Probes are up to now addressed as `(gid, lid)`, where
`lid` is computed from the position in the return value of `recipe::get_probes`. This is not
only a more uniform interface, but also a safer one (imagine adding an additional probe in the
middle).

That is, from now now on, each probes needs to add a stringy tag, eg
```cxx
  auto get_probes(cell_gid_type gid) const override { return {{cable_probe_voltage(...), "Um"}}; }
```
and we consume this as
```cxx
// ....
auto sim = simulation(rec);
sim.add_sampler({0, "Um"}, ...);
```

In addition the optional `tag` field was removed. It has been hardly ever been used (no examples
and no models in the wild I know of). It has also been hidden from Python. Thus, the two APIs are
now closer aligned.

# ⚠️ Breaking

- Each probes needs a string tag.
- Numerical ids are no longer allowed.
- (C++) `probe_info::tag` is gone

# TODO
- [x] Update docs.
- [x] Check python examples.
- [x] Check python tests.

# Future tasks
Here are some things I plan to do later on, but might overburden this PR. The idea is to streamline 
the interface presented to the user and easing their cognitive load.

- We should probe POINT processes by label, too.
- We should streamline the sampler interface, currently there are 4 cases
  1. POINT: single item metadata  of specialised struct
  2. POINT cell: vector of these structs.
  3. DENSITY/voltage/current/...: scalar float, `mlocation`
  4. DENSITY/voltage/current/... cell: vector of float, vector of `mcable` 
  
  Of these we could at least merge 1 & 2, potentially 3 & 4, and maybe 1-4. 
- Potentially adding pre-canned objects/hooks to dump data in certain formats (although I'd rather _not_ take additional dependencies, HDF5 or otherwise).


thorstenhater:
Yeah, that stuck out like a sore thumb.

arbor/fvm_lowered_cell.hpp
```diff
@@ -175,22 +179,40 @@ struct fvm_probe_data {
...
 
-    std::unordered_map<cell_member_type, probe_tag> tag;
-    std::unordered_multimap<cell_member_type, fvm_probe_data> data;
+    auto count(const cell_address_type& k) const { return data.count(k); }
 
-    std::size_t size() const {
-        arb_assert(tag.size()==data.size());
-        return data.size();
+    // Return range of fvm_probe_data values associated with probeset_id.
+    std::vector<const fvm_probe_data*> data_on(const cell_address_type& probeset_id) const {
```
AdhocMan, line 198: Maybe one can avoid creating and returning a vector here by just using a range constructed from the const_iterators form `equal_range`?

python/example/network_ring.py
```diff
@@ -6,6 +6,8 @@
 import seaborn  # You may have to pip install these
 from math import sqrt
 
+print(arbor.__path__)
```
AdhocMan, line 9:
```suggestion
```

arbor/include/arbor/simulation.hpp
```diff
@@ -62,15 +62,16 @@ class ARB_ARBOR_API simulation {
...
+                                           sampler_function f);
 
     void remove_sampler(sampler_association_handle);
 
     void remove_all_samplers();
 
     // Return probe metadata, one entry per probe associated with supplied probe id,
     // or an empty vector if no local match for probe id.
-    std::vector<probe_metadata> get_probe_metadata(cell_member_type probeset_id) const;
+    std::vector<probe_metadata> get_probe_metadata(const cell_address_type& probeset_id) const;
```
AdhocMan, line 74: This function seems to be missing in the docs. Maybe add it with the new signature?

arbor/include/arbor/common_types.hpp
```diff
@@ -97,6 +97,28 @@ struct cell_global_label_type {
     cell_global_label_type(cell_gid_type gid, cell_tag_type tag, lid_selection_policy policy): gid(gid), label(std::move(tag), policy) {}
 };
 
+// User facing handle for referring to probes and similar
+
+struct cell_address_type {
```
AdhocMan, line 102: This could also be added to the API docs.

AdhocMan (changes requested):
Nice work! I think it good to have a more concise way of addressing probes.
Just a few minor comments.

arbor/fvm_lowered_cell.hpp
thorstenhater: 1. I am a bit doubtful of the lifetime management; any modification to the underlying data structure might invalidate the range. Might just side effects from using too much Rust, though.
2. Can we even construct such a range? We get what is basically a range of `(k, v)` where `k` is constant...    We could use an iterator adapter for the occasion, but not a range, I think.

arbor/include/arbor/common_types.hpp
thorstenhater: 👍

doc/cpp/recipe.rst
```diff
@@ -135,17 +135,13 @@ See :ref:`cppinterconnectivity`.
...
 .. cpp:class:: probe_info
 
     Probes are specified in the recipe objects that are used to initialize a
     model; the specification of the item or value that is subjected to a
     probe will be specific to a particular cell type.
 
-    .. cpp:member:: probe_tag tag
+    .. cpp:member:: cell_tag_type tag
 
            Opaque key, returned in sample record.
```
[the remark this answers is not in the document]
thorstenhater: nope, it's 'your' label.

python/example/network_ring.py
thorstenhater: Done.

arbor/fvm_lowered_cell.hpp
AdhocMan: You are right, if modification after usage of `data_on` is expected, then this would be an issue. I'm fine with leaving it as is, since this would likely be just a small optimization.

PR #2216, opened 2023-09-05
Treat some corner cases in modcc

thorstenhater:
These tests and fixes cover a few cases related to kinetic blocks:

#  Non-reaction statement in between reaction statements
This leads to the 'normal' statements being hoisted to the top resulting 
in unexpected semantics. This is now an error.

#  Reaction statements under conditional
This has always been an error; add a test for it.

# Reaction/Linear statements in non-KINETIC/LINEAR context
This is now an error; did SEGV before.

# Derivative statements in KINETIC/LINEAR/PROCEDURE
These are now error, they were before, but resulted in a confusing ICE message.


AdhocMan (approved):
lgtm

PR #2218, opened 2023-09-11
Unify loading of Standard Morphology Files.

### LINKED ISSUE #1981: Clean-up Morphology Loaders
Currently we have three kinds of loader as categorised by their return value
- NML returns a really complex structure containing 
  - all labels and regions declared by the input
  - a list of morphologies declared in `<cell>` tags
  - a list of morphologies declared in `<morphologies>`
- ASC returns a structure containing the morphology and a fixed set of labels
- SWC returns a bare morphology and we have a list of predefined labels
- ASC and SWC can be loaded as segment_trees, but not NML

This is neither conforming to the least surprise principle nor ergonomic. Thus
- [x] SWC should return a structure similar to the current ASC loader (eg `labelled_morphology`, but maybe a pair is fine as well); subsuming the external tags
- [X] ASC should insert custom ~tags~ (these are currently _parsed_ but discarded)
  - ASC returns a list of markers and spines. 
- [x] NML should be a simpler structure to handle
  - [x] `<cell>` and `<morphology>` can be unified (but still tagged by the origin), 
  - [x] add `segment_tree` loading
  - [x] asking for a specific morphology in an `nml` object should give back a `labelled_morphology`
- [x] Finally: `cable_cell` should add a new constructor consuming a `labelled_morphology` and a `decor`
  - changed my mind here: it doesn't make the API any nicer and thus complicates matters.
- [x] ~Bonus: add convenience function `load_morphology` dispatching based on file suffix~
  - doesn't work since NML and asc+swc return fundamentally different objects

thorstenhater:
All morphology loaders now return a `loaded_morphology` object comprising
- `morphology`: the morphology corresponding to the data on disk
- `segment_tree`: the raw segment tree
- `label_dict`: all labels defined in the file
  -  `asc`/`swc`: `axon`, `soma`, `dend`, and `apic` based on tag 1, 2, 3, 4.
  - `nml`: the union of the label dicts contained in the metadata, prefixed by `grp:`, `nmd:`, and `seg:`.
- `metadata`: a set of items given by the fileformat
  - `swc`: nothing
  - `asc`: a list of spines and marker sets
  - `nml`: 
    - `named_segments` a label dict
    - `segment_groups` a label dict `name` -> segment ids
    - `group_segments` a map from segment id to group name(s)
    - `segments` a label dict, one entry per id
    - `cell_id` id of the cell defining this morphology, if any
    - `id` id of the morphology

Note that loading neuroml still goes via the `neuroml` object and the `cell_morphology`/`morphology`
accessors return an optional `loaded_morphology`.

## ⚠️ Breaking ⚠️ 

- Removed the `load_*_raw` functions, as `loaded_morphology` bundles the segment tree.
- The interface of `load_*` now has an extra indirection.
- The interface of `neuroml` changed.
- `label_dict::import` is now called `extend` for future safety (=C++ modules) and Python; calling things `import` (=a keyword) is not well received by `black` et al.

## Issues

Closes #1981


arborio/include/arborio/loaded_morphology.hpp
```diff
@@ -0,0 +1,75 @@
...
+    std::vector<arb::mpoint> locations;
+};
+
+struct ARB_SYMBOL_VISIBLE asc_metadata {
+    std::vector<asc_marker_set> markers;
+    std::vector<asc_spine> spines;
+};
+
+// Bundle some detailed metadata for neuroml ingestion.
+struct ARB_SYMBOL_VISIBLE nml_metadata {
```
boeschf, line 40: consider including `<arborio/export.hpp>` and use `ARB_ARBORIO_API` for "normal" classes which don't need rtti, see [docs](https://docs.arbor-sim.org/en/latest/dev/export.html).

python/label_dict.hpp
```diff
@@ -48,11 +48,11 @@ struct label_dict_proxy {
         return locsets.size() + regions.size() + iexpressions.size();
     }
 
-    void import(const label_dict_proxy& other, std::string prefix = "") {
-        dict.import(other.dict, prefix);
-
+    auto extend(const label_dict_proxy& other, std::string prefix = "") {
```
boeschf, line 51: copy of `*this` intended when returning?

python/identifiers.cpp
```diff
@@ -50,9 +50,14 @@ void register_identifiers(py::module& m) {
...
-               if (py::len(t)!=2) throw std::runtime_error("tuple length != 2");
-               return arb::cell_local_label_type{t[0].cast<arb::cell_tag_type>(), t[1].cast<arb::lid_selection_policy>()};
+        .def(py::init([](const std::tuple<arb::cell_tag_type, arb::lid_selection_policy>& t) {
+               return arb::cell_local_label_type{std::get<arb::cell_tag_type>(t), std::get<arb::lid_selection_policy>(t)};
+             }),
+             "Construct a cell_local_label identifier with tuple argument (label, policy):\n"
+             "  label:  The identifier of a group of one or more items on a cell.\n"
+             "  policy: The policy for selecting one of possibly multiple items associated with the label.\n")
+        .def(py::init([](const std::pair<arb::cell_tag_type, arb::lid_selection_policy>& t) {
+               return arb::cell_local_label_type{std::get<arb::cell_tag_type>(t), std::get<arb::lid_selection_policy>(t)};
```
boeschf, line 60: do we really need both `tuple` and `pair`?

python/label_dict.hpp
thorstenhater: no, good catch.

python/identifiers.cpp
thorstenhater: Probably no ... but what's the harm?

boeschf: nothing really, was just wondering

boeschf (approved):
looks good!

Issue #2129, opened 2023-06-02
Inject/decay of diffusing particles does not work with custom mechanism


jlubo:
Injecting diffusing particles with the built-in mechanism `inject` and letting their concentration decay with the built-in mechanism `decay` works fine. But if, instead of the built-in mechanism `inject`, an identical mechanism `custom_inject` from a custom catalogue is used, diffusion does not work anymore.

Please see the attached plots and code: [diffusion_with_inject_and_decay_2023-06-02.zip](https://github.com/arbor-sim/arbor/files/11636554/diffusion_with_inject_and_decay_2023-06-02.zip). The example is somewhat similar to the existing [diffusion example](https://github.com/arbor-sim/arbor/blob/9627b8bfd94a30e29c16714e72ced912b9e7da25/python/example/diffusion.py). I derived it from Shirin's code for calcium diffusion. I've tested this with Arbor v0.8.1.

thorstenhater:
That would be extremely strange, but I'll take a look. Did you try adding custom_inject to the vanilla example?

jlubo:
For the [vanilla example](https://github.com/arbor-sim/arbor/blob/9627b8bfd94a30e29c16714e72ced912b9e7da25/python/example/diffusion.py) the results differ as well:

[results_builtin.pdf](https://github.com/arbor-sim/arbor/files/11642894/results_builtin.pdf)
[results_custom.pdf](https://github.com/arbor-sim/arbor/files/11642895/results_custom.pdf)

thorstenhater:
Ok, that is curious. Just to double check: Are both build using the same compiler flags?
That is in particular: ARB_VECTORIZE?

jlubo:
Yes, the only thing that is changed is the mechanism, i.e.
```
mech_inject = arbor.mechanism('inject/x=test_ion', {'alpha' : 100})
```
is exchanged with
```
mech_inject = arbor.mechanism('custom_inject/x=test_ion', {'alpha' : 100})
```

Shirin1993:
Here is another version of the problem :
 I tried a _custom_ `injection` in combination with a _custom_ `decay` and also a _built-in_ `decay`. 
It seems that :

- the `built-in decay` doesnt work, since it results in zero concentration.

- the `custom decay` also doesnt work since changing the parameter `tau` has no effect on diffusion. 

I have attached the related codes and plots

[...]

PR #2221, opened 2023-09-13
Allow taking derivatives of `Xd`

thorstenhater:
# Introduction 

The core problem is that `Xd` is not a `STATE` and only those can appear in derivatives/ODEs.

This requires rewriting the idiomatic
```
 NEURON {
    SUFFIX decay
    USEION x WRITE xd
    RANGE tau
}

PARAMETER { tau = 5 }

BREAKPOINT {
   SOLVE dX METHOD cnexp
}

DERIVATIVE dX {
   xd' = -tau*xd
}
```
as this
```
NEURON {
    SUFFIX decay
    USEION x WRITE xd
    RANGE tau
}

STATE { F }

INITIAL { F = xd }

PARAMETER { tau = 5 }

BREAKPOINT {
   SOLVE dX METHOD cnexp
   xd = F
}

DERIVATIVE dX {
   F = xd
   F' = -tau*F
}
```

_However_, `BREAKPOINT` is translated into _two_ C++ calls, not one:
1. compute $F(t)$ from $F'(t)$
2. update `xd` based on `F`

and other things -- like the **solver** -- happen in between. This has proven
not only ugly, but also unsound, see #2129 and possibly more.

# Fixing the Problem

Simply allow `Xd` to be differentiated. As this is not an ionic _current_ nothing 
bad will happen., right? Right? This could still be awkward when making mechanisms like
this
```
BREAKPOINT {
   SOLVE dX METHOD cnexp
   xd = xd + 42
}

DERIVATIVE dX {
   xd' = -tau*xd
}
```

### DIFF
--- a/doc/fileformat/nmodl.rst
+++ b/doc/fileformat/nmodl.rst
@@ -84,9 +84,7 @@ Ions
   not made ``STATE`` they may be written to, but their values will be reset to
   their initial values every time step.
 * The diffusive concentration ``Xd`` does not share this semantics. It will not
-  be reset, even if not in ``STATE``, and may freely be written. This comes at the
-  cost of awkward treatment of ODEs for ``Xd``, see the included ``decay.mod`` for
-  an example.
+  be reset, even if not in ``STATE``, and may freely be written.
 * ``Xd`` is present on all cables iff its associated diffusivity is set to a
   non-zero value.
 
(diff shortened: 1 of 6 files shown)


AdhocMan (approved):
The code changes seems fine to me.  I can't fully judge the impact of the feature itself. Up to you if you'd like someone else to take a look at it as well.

thorstenhater:
Our primary users seem too happy not to merge it ;)

jlubo:
I can confirm that this works for me, solving the issue #2129 as well as the negative concentration issue mentioned in #2145 :)

PR #2225, opened 2023-10-06
CI: Fix spack url parsing

AdhocMan:
Github seems to have changed their json tags, which identify the url to download the latest release of spack.


thorstenhater (approved):
👍 Thanks

Issue #1832, opened 2022-02-07
Have more cell types builtin (e.g. Izhikevich) to improve NeuroML compat


brenthuisman:
NeuroML relies on the simulator to have implementations for cell kinds, one of which is [izhikevichCell](https://docs.neuroml.org/Userdocs/Schemas/Cells.html#izhikevichcell). The rather short list of point neuron models in Arbor has been mentioned before, so that makes two reasons to consider expanding the list. For reference: [https://docs.neuroml.org/Userdocs/Schemas/Cells.html](here) (scroll down on the RHS) the list that NeuroML expects.

* [ ] Get feedback on a list of most wanted cell kinds.
  * [ ] Bring up in the upcoming Developer Meeting.
* [ ] Are there any cell kinds needed to show off our NML support? Anything that would to well in a publication?

Whoever reads this, please add your cell kinds in a comment.

thorstenhater:
This is only a small step on the road to a larger project:
- allow users to bring their own Cell(Group)s #1605 
  - this is either an extension to our current ABI or a new ABI aimed at cell groups
- unify probing across cell groups
  - started here for LIF #1263 
- (optional and hard) extract cells from NML2 description and compile them to whatever #1605 comes up with.

brenthuisman:
#1605 is ambitious and long term. Earlier we advised people to write a mechanism to implement their own single compartment neuron, e.g. https://github.com/tetzlab/FIPPA

PR #2230, opened 2023-11-07
Add AdEx cells.

thorstenhater:
Add Adaptive Exponential Cells to Arbor. Motivation here is to add another useful cell kind before considering two major features:
1. Branching out Cell Group support by offering a plugin interface like the mechanism ABI. Finding a proper abstraction should be easier by having 3+2 examples instead of 2+2.
2. Adding support for custom synapse dynamics on point models.

Logically though, 2. should go before 1. since a plugin interface should declare how and if custom synapses are handled...

Closes https://github.com/arbor-sim/arbor/issues/1832

More on request.


thorstenhater:
Closes #1832 

More on request.

ErbB4 (approved):
looks good and I will come back to comment again after I use it.

PR #2231, opened 2023-11-14
better spack workflow

boeschf:
Previous spack workflow doesn't run through anymore. A possible reason might be the spack cache which is cached with github's own cache feature. This PR's intent:
- simplify spack workflow
- use external build-cache (https://github.com/spack/github-actions-buildcache)


thorstenhater (approved):
lgtm

PR #2232, opened 2023-11-24
rename py_recipe to recipe.

### LINKED ISSUE #1634: Python Recipe (abstract) base class.
It would be nice to have `arbor.recipe` do a few more things for the user:

* Throw error when required methods are not overridden.
  * `PYBIND11_OVERRIDE_PURE` is not throwing in the current `arbor.recipe` impl. Only upon use of the instantiated class derived from `arbor.recipe`. Not entirely sure why, but earlier surely would be nice.
* Do away with manually having to call `__init__` on the `arbor.recipe` base class.
* Maybe: have `print()` do something useful.
* Question: does `arbor.recipe` need to be a abstract base class [2]?

See [1] for the trampoline pattern.

[1] https://pybind11.readthedocs.io/en/stable/advanced/classes.html
[2] https://docs.python.org/3/library/abc.html

thorstenhater:
Solve name mismatch Python ./. C++.

Closes #1634

### DIFF
--- a/python/domain_decomposition.cpp
+++ b/python/domain_decomposition.cpp
@@ -1,6 +1,5 @@
 #include <limits>
 #include <string>
-#include <sstream>
 
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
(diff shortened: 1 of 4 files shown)


boeschf (approved):
Looks good to me. Just for clarification: we rename it to have clearer error messages?

thorstenhater:
Yeah, essentially. This eliminates the mismatch we observe here:
```
arbor on  py/rename-trampolines:master [$?] is 📦 v0.9.1-dev via △ v3.27.8 via 🐍 v3.11.6 (python-3.11)
❯ /opt/homebrew/bin/python3.11 -c 'import arbor as A; A.recipe().num_cells()'
RuntimeError: Tried to call pure virtual function "py_recipe::num_cells"
```
to the slightly more palatable
```
arbor on  py/rename-trampolines:master [$?] is 📦 v0.9.1-dev via △ v3.27.8 via 🐍 v3.11.6 (python-3.11)
❯ /opt/homebrew/bin/python3.11 -c 'import arbor as A; A.recipe().num_cells()'
RuntimeError: Tried to call pure virtual function "recipe::num_cells"
```

thorstenhater:
and that was all left to be done in #1634; mainly due to the workings of pybind11 we cannot fix the rest better
than a RuntimeError.

boeschf:
thanks for the explanation

thorstenhater:
Sure! I am on a clean-up spree through the issue tracker; closing redundant/unfixable stuff and directly fixing those that are sufficiently low effort.

PR #2237, opened 2023-12-05
Update `segment` to `msegment` as name present in arbor module

Helveg:
It seems the `segment` class has been renamed to `msegment`, I updated the morphology docs to reflect this (probably breaking a couple labels elsewhere?)

### DIFF
--- a/doc/python/morphology.rst
+++ b/doc/python/morphology.rst
@@ -118,7 +118,7 @@ Cable cell morphology
 
         Radius of the cable (μm)
 
-.. class:: segment
+.. class:: msegment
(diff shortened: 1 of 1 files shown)


thorstenhater:
Interesting question, I stumbled on this during #2233, we might want to rename:
- i_clamp / iclamp
- msegment / segment
- mpoint / point
- mcable / cable

which are either currently divergent _or_ inconsistent. I personally dislike the namespacing on a
budget introduce by the `m` prefix.

thorstenhater (approved):

PR #2238, opened 2023-12-05
Typestub

Helveg:
I added a `.pyi` file (first 3 hours by hand, then stumbled upon `pybind11-stubgen`, and another hour of linting). But now I think we have a type stub?

I think using this stub and [`sphinx.ext.autodoc`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html) we could replace large chunks of the manually generated documentation. I also added a tiny check to confirm that everything that's in the `arbor` module has a counterpart in the `.pyi` to CI.


Helveg:
@thorstenhater it seems my 2 new files in the `/python` folder aren't included in the installed package. Do you know how to get that picked up?

thorstenhater:
See the infamous #2233, added via the same generator and fixing all the 'mishaps'

Helveg:
I already had the feeling this was going to be a dupe, but I couldn't stop myself ;) For automation, it seems like it mostly put `...` where it didn't know what to do; what's your experience?

If so, what I could do for now, in terms of automation, is to generate a new stub, and compare them via an ast visitor, that only overwrites non-ellipsis diffs?

PR #2241, opened 2023-12-08
Fix a few integer type mismatches in for loops.

### LINKED ISSUE #2240: int size, signedness mismatches
In `arbor/include/arbor/util/hash_def.h`, the `internal_hash` function has two loops with a `size_t` bound but an `int` iterator:
```
for (int ix = 0; ix < sizeof(data); ++ix)
```
While it is hard to imagine triggering the limited range bug, it also gives us a signedness comparison warning.

In `test/unit/test_probe.cpp` there are two loops with an `unsigned` bound but an `int` iterator:
```
for (int ix = 0; ix < n_cv; ++ix) i_memb[ix] = *s_beg++;
```
Making `ix` an `unsigned` is sufficient here.

halfflat:
* Use size_t to iterate size_t-bounded loops in `internal_hash()`.
* Use unsigned to iterate unsigned-bounded loops in `test_probe.cpp`.

Fixes #2240

### DIFF
--- a/arbor/include/arbor/util/hash_def.hpp
+++ b/arbor/include/arbor/util/hash_def.hpp
@@ -55,7 +55,7 @@ inline constexpr std::size_t internal_hash(T&& data) {
     if constexpr (std::is_integral_v<D>) {
         unsigned long long bytes = data;
         std::size_t hash = offset_basis;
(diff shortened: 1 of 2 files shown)


thorstenhater (approved):
Yes, thank you.

PR #2246, opened 2024-01-10
disable spack latest_release

boeschf:
The spack CI test fails when a point release (`latest_release`) for `spack` is selected, but works for the `develop` branch. This is most likely connected to the external `spack` buildcache, maintained by spack developers (https://github.com/spack/github-actions-buildcache).
Therefore, we disable the `latest_release` case until a workaround is implemented or a compatible `spack` buildcache is available.


thorstenhater (approved):
👍

PR #2247, opened 2024-01-11
fix gpu warp intrinsics

boeschf:
`reduce_by_key` depends on warp level intrinsics to transfer values between different threads (lanes) participating in the reduction. The pertinent intrinsic is `__shfl_down_sync` which is accessed through Arbor's wrapper function `shfl_down`. However, the contribution from each thread to the reduction was erroneously truncated to an integer value. This PR fixes the signature of the respective wrapper functions and modifies the unit test in order to check that floating point reductions are not truncated. 
While cleaning up the cuda code path, the workaround using two 32-bit shuffle instructions for 64 bit data types (doubles) was removed - this was probably a leftover from cuda versions prior to 9.0.

### DIFF
--- a/arbor/include/arbor/gpu/cuda_api.hpp
+++ b/arbor/include/arbor/gpu/cuda_api.hpp
@@ -139,17 +139,6 @@ inline float gpu_atomic_sub(float* address, float val) {
 
 /// Warp-Level Primitives
 
-__device__ __inline__ double shfl(unsigned mask, double x, int lane)
-{
-    auto tmp = static_cast<uint64_t>(x);
-    auto lo = static_cast<unsigned>(tmp);
-    auto hi = static_cast<unsigned>(tmp >> 32);
-    hi = __shfl_sync(mask, static_cast<int>(hi), lane, warpSize);
-    lo = __shfl_sync(mask, static_cast<int>(lo), lane, warpSize);
-    return static_cast<double>(static_cast<uint64_t>(hi) << 32 |
-                               static_cast<uint64_t>(lo));
-}
-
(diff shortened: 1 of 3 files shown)


arbor/include/arbor/gpu/cuda_api.hpp
```diff
@@ -139,17 +139,6 @@ inline float gpu_atomic_sub(float* address, float val) {
 
 /// Warp-Level Primitives
 
-__device__ __inline__ double shfl(unsigned mask, double x, int lane)
```
bcumming, line 142: Great - these are indeed a hang over from K20X and ancient versions of CUDA.

arbor/include/arbor/gpu/cuda_api.hpp
```diff
@@ -158,24 +147,15 @@ __device__ __inline__ unsigned any(unsigned mask, unsigned width) {
...
 
-#ifdef __NVCC__
-__device__ __inline__ double shfl_up(unsigned mask, int idx, unsigned lane_id, unsigned shift) {
-    return __shfl_up_sync(mask, idx, shift);
-}
-
-__device__ __inline__ double shfl_down(unsigned mask, int idx, unsigned lane_id, unsigned shift) {
-    return __shfl_down_sync(mask, idx, shift);
+template<typename T>
+__device__ __inline__ T shfl_up(unsigned mask, T var, unsigned lane_id, unsigned shift) {
```
bcumming, line 151: Good catch.

bcumming (approved):
Thankyou @boeschf for catching this.

I am deeply embarrassed by both my bug, and low-quality unit test that didn't catch the bug.

PR #2249, opened 2024-01-17
Create less intermediate data from events

thorstenhater:
This PR optimizes event handling by removing the intermediate structure created by this chain
of transformations:
1. `event_lanes := vector<vector<event>>` (one lane per local cell, each lane sorted by time)
2. `staged_events_per_mech_id := vector<vector<vector<deliverable_event>>>` (one vector per mech id and time step, sorted by time)
3. `vector<event_stream>` one stream per mech id

The following optimisations where performed:
- cut out the middle step (2) completely as it is wholly unneeded and sort directly into event streams
- remove a spurious index structure from `cable_cell_group`
- slim down `deliverable_event` and `deliverable_event_data`
- `event_stream` now uses a partition instead of a vector of ranges for splitting its data into `dt` buckets. (Save 8B per `dt` ;))

The result is that the quite pathological example `calcium_stdp.py` (as given in which generates immense amounts
of spikes using a single cell group and a single epoch drops from 3.8GB heap to 1.9GB heap usage at peak
using the same runtime or slightly less (<5% difference).

## TODO
- [x] Port to GPU.
 - [X] Tests pass
 - [x] Examples run through
- [x] Fix tests... Interestingly _locally_ all tests pass on my dev machine regardless of optimisation, vectorisation, and assertion settings.


boeschf:
cscs-ci run daint-mc

boeschf (commented):
looks good, some code is duplicated between the backends, see [this PR](https://github.com/thorstenhater/arbor/pull/3).

boeschf (approved):

PR #2255, opened 2024-02-07
Tentatively fix one instance.

thorstenhater:

### DIFF
--- a/test/unit/test_sde.cpp
+++ b/test/unit/test_sde.cpp
@@ -378,7 +378,7 @@ public:
 // first argument indicates the number of random variables
 void advance_common(unsigned int n_rv, arb_mechanism_ppack* pp) {
     const auto width = pp->width;
-    arb_value_type* ptr = archive_ptr->claim(width * n_rv);
+    arb_value_type* ptr = archive_ptr->claim(std::size_t{width} * n_rv);
     for (arb_size_type j=0; j<n_rv; ++j) {
         for (arb_size_type i=0; i<width; ++i) {
             ptr[j*width+i] = pp->random_numbers[j][i];


halfflat:
I was wondering about some of these `size_t` casts in calls to `std::vector::reserve()`; they create an inconsistency with prior checks in `arb_assert` and the explicit casting here makes it look like we're working around some sort of range issue.

Are these there to avoid a potential `int32_t` overflow? If so, we should consider using different types (viz. `std::size_t` or `std::ptrdiff_t`) to represent quantities such as `n_sample` etc.

For the changes in `modcc/io/prefixbuf.cpp`, storing the `unsigned` values from the indent stack and iword-stored tabwidth as `size_t` is a bit misleading: they really are `unsigned` values. And while there is a potential unsigned wraparound from their multiplication if the indent code is somehow being abused, I'd assert that it's safer to let it wrap in an `unsigned` then to start allocating strings of length greater than 4GB.

halfflat:
I don't recall why it was signed; I'd have to go look at the git blame :) It might have been because we use them to store differences. It's possible but less likely they get SIMDed somewhere and we don't yet have unsigned SIMD in the library.

halfflat:
But on the CodeQL front, what is its complaint? Potential signed overflow, or is it complaining about implicit conversion of signed to size_t in the `reserve()` calls? If the latter, I don't think that should be something that's flagged: it's a reasonable thing to allow if we are confident that the non-negative precondition is met, and casting to an unsigned type first doesn't fix the problem.

Is there some way of decorating `arb_assert` so as to make that assertion visible to CodeQL? E.g. if we had
```
    int n = something.last-something.first;
    arb_assert(n>=0); // macro includes some magic for CodeQL
    vec.reserve(n); // this is fine because we made an assertion that n>=0
```
it would all be fine?

modcc/io/prefixbuf.cpp
```diff
@@ -75,10 +75,14 @@ int indent_manip::xindex() {
 
 static void apply_indent_prefix(std::ios& s, int index) {
     if (auto pbuf = dynamic_cast<prefixbuf*>(s.rdbuf())) {
-        indent_stack* stack_ptr = static_cast<indent_stack*>(s.pword(index));
-        size_t tabs = (!stack_ptr || stack_ptr->empty())? 0 : stack_ptr->top();
-        size_t tabwidth = s.iword(index);
-        pbuf->prefix = std::string(tabs*tabwidth, ' ');
+        if (auto ptr = static_cast<indent_stack*>(s.pword(index));
+            ptr && !ptr->empty()) {
+            unsigned n_tab = ptr->top();
```
halfflat, line 80: Argh! This is terribly misleading formatting

thorstenhater:
Possibly, but I do not think it's worth it even. CodeQL is _technically_ correct in pointing out that possibility, but
it doesn't flag the project as endangered. It will complain, though, if a PR introduces _new_ problems.

The urge to 'fix' this was merely my completionist attitude.

modcc/io/prefixbuf.cpp
thorstenhater: I knew you were going to love it. Isn't it beautiful how the second line of the condition and the first line of the condition line up?

test/unit/test_sde.cpp
```diff
@@ -377,8 +377,8 @@ class sde_recipe: public simple_recipe_base {
 // generic advance method used for all pertinent mechanisms
 // first argument indicates the number of random variables
 void advance_common(unsigned int n_rv, arb_mechanism_ppack* pp) {
-    const auto width = pp->width;
-    arb_value_type* ptr = archive_ptr->claim(std::size_t{width} * n_rv);
+    auto width = static_cast<size_t>(pp->width);
```
halfflat, line 380: Why not just `size_t width = pp->width`?

halfflat:
Regarding the CodeQL flagging the size_t conversion, I'd say we should either ignore it or address it in a robust way (with actual range checks). My inclination is to ignore it.

modcc/io/prefixbuf.cpp
halfflat: It has a certain horrible charm to it.

arbor/communication/dry_run_context.cpp
```diff
@@ -49,25 +47,24 @@ struct dry_run_context_impl {
     void remote_ctrl_send_done() const {}
     gathered_vector<cell_gid_type>
     gather_gids(const std::vector<cell_gid_type>& local_gids) const {
-        using count_type = typename gathered_vector<cell_gid_type>::count_type;
-
-        count_type local_size = local_gids.size();
+        size_t local_size = local_gids.size();
```
boeschf, line 50: `std::size_t`?

test/unit/test_sde.cpp
```diff
@@ -377,8 +377,8 @@ class sde_recipe: public simple_recipe_base {
 // generic advance method used for all pertinent mechanisms
 // first argument indicates the number of random variables
 void advance_common(unsigned int n_rv, arb_mechanism_ppack* pp) {
-    const auto width = pp->width;
-    arb_value_type* ptr = archive_ptr->claim(width * n_rv);
+    std::size_t width = static_cast<size_t>(pp->width);
```
boeschf, line 380:
```suggestion
    auto width = static_cast<std::size_t>(pp->width);
```

arbor/include/arbor/common_types.hpp
```diff
@@ -138,8 +138,8 @@ using time_type = double;
 constexpr time_type terminal_time = std::numeric_limits<time_type>::max();
 
 // For holding counts and indexes into generated sample data.
-
-using sample_size_type = std::int32_t;
+using sample_index_type = std::int32_t;
```
boeschf, line 141: where is this type actually used?

thorstenhater: It -- currently -- isn't, but it's useful when we are considering things like:
```cxx
sample_size_type a = ..., b = ..., c = a - b;
```
due to the potential overflow.

arbor/communication/dry_run_context.cpp
thorstenhater: Yes and all other instance in that file.

test/unit/test_sde.cpp
thorstenhater: Yes, that's fine, too.

arbor/cable_cell_group.cpp
```diff
@@ -215,9 +210,9 @@ void run_samples(
     fvm_probe_scratch& scratch)
 {
     const sample_size_type n_raw_per_sample = p.raw_handles.size();
-    sample_size_type n_sample = (sc.end_offset-sc.begin_offset)/n_raw_per_sample;
-    arb_assert((sc.end_offset-sc.begin_offset)==n_sample*n_raw_per_sample);
-    arb_assert((unsigned)n_raw_per_sample==p.weight.size());
+    sample_size_type n_sample = (sc.end_offset - sc.begin_offset)/n_raw_per_sample;
+    arb_assert((sc.end_offset - sc.begin_offset)==n_sample*n_raw_per_sample);
+    arb_assert((unsigned)n_raw_per_sample == p.weight.size());
```
halfflat, line 215: Are these changes to make CodeQL happy, or just for readability?

arbor/communication/dry_run_context.cpp
```diff
@@ -26,7 +24,7 @@ struct dry_run_context_impl {
         count_type local_size = local_spikes.size();
 
         std::vector<spike> gathered_spikes;
-        gathered_spikes.reserve(local_size*num_ranks_);
+        gathered_spikes.reserve(std::size_t{local_size}*num_ranks_);
```
halfflat, line 27: We shouldn't change the (effective) maximum size of the gathered_spikes vector without also changing the range of the indices in the loop 35-37.

We ultimately make a partition of the indices into the variable `partition` which has `count_type` as its value type, and this vector is used to construct the `gathered_vector` return type, which also expects things in terms of `count_type`. I believe the correct fix is not to cast `local_size` or use `std::size_t` here but instead
1. Add an assertion which confirms that the `local_size*num_ranks_` doesn't wrap in `count_type`, and
2. Uses `std::size_t{local_size*num_ranks_}` to keep CodeQL happy or else we tell CodeQL that this is not a bug (my preference).

arbor/communication/dry_run_context.cpp
```diff
@@ -49,25 +47,24 @@ struct dry_run_context_impl {
...
+            for (std::size_t j = i*local_size; j < (i+1)*local_size; j++){
                 gathered_gids[j] += num_cells_per_tile_*i;
             }
         }
 
+        using count_type = typename gathered_vector<cell_gid_type>::count_type;
         std::vector<count_type> partition;
-        for (count_type i = 0; i <= num_ranks_; i++) {
+        for (std::size_t i = 0; i <= num_ranks_; i++) {
             partition.push_back(static_cast<count_type>(i*local_size));
```
halfflat, line 68: There's no point doing everything in `size_t` only to cast back down to `count_type` here; the alterations just make the potential bug more obscure. IMO the only correct resolution is to have an assert that checks that the product doesn't wrap and then to keep using `count_type`.

halfflat (changes requested):
One query and some requests for changing how we deal with the wrapping product warning. Generally, I suspect the cost of a range-checking assert is cheap and provides the guarantees we need while casting up and down the unsigned integer sizes keeps the potential bug while changing its clothes.

Currently `arb_assert` is a build-time option, and I'm wondering if we could perhaps split its functionality into asserts which should always be checked and those which are really there to be enabled when we're tracking down bugs in our code logic.

arbor/cable_cell_group.cpp
thorstenhater: Readability. I can get behind `a*b + c` for similarities with mathematical notation, but `=`, `==`, and especially `-` tend to merge visually into one token. Or: Too similar to `snake_case` for my liking.
As much of the code, especially newer parts is in this style, I tend to adjust these expression in passing when editing files. Same for removing redundant headers, which my IDE picks up on these days.

modcc/io/prefixbuf.cpp
thorstenhater: Fixed it.

modcc/io/prefixbuf.cpp
```diff
@@ -75,11 +75,14 @@
...
-        indent_stack* stack_ptr = static_cast<indent_stack*>(s.pword(index));
-        unsigned tabwidth = s.iword(index);
-
-        unsigned tabs = (!stack_ptr || stack_ptr->empty())? 0: stack_ptr->top();
-        pbuf->prefix = std::string(tabs*tabwidth, ' ');
+        if (auto ptr = static_cast<indent_stack*>(s.pword(index));
+            ptr && !ptr->empty()) {
+            unsigned n_tab = ptr->top();
+            unsigned width = s.iword(index);
+            pbuf->prefix = std::string(n_tab*width, ' ');
```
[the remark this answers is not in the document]
thorstenhater: Dismissed.

boeschf (approved):

PR #2257, opened 2024-02-22
fix auto detect of sve length

boeschf:
- apply pertinent compiler flags when trying to compile the detection code
- improve compiler error message when sve length is not set


bcumming (approved):
Looks good to me.

PR #2264, opened 2024-04-02
Tiny fix for |epoch| < dt.

### LINKED ISSUE #2263: Connection delay of `0` causes unexpected behavior
Setting the delay time of a connection to`0` causes unexpected behavior. Specifically, the probing of variables from mechanisms fails by yielding an empty array (see the code [here](https://github.com/jlubo/arbor/blob/70840741e2f19a2173b963480b3dec71245d6f87/python/test/delay_zero_bug.py)).

I've encountered the issue on Arbor commit version 2ec4db34bdc67ddb9d2b16b1946ee40d974eb413. According to @thorstenhater, it has not yet been fixed by the currently most recent version (08a157fb52455c3a45b7ffae1bf6aabd33a43023).

The issue can be circumvented by setting the delay to the duration of one timestep. Note that it is related to what is mentioned [here](https://github.com/arbor-sim/arbor/issues/1997).

thorstenhater:
Explicitly use epochs of length of at least $\Delta t$. 

Fixes #2263

### DIFF
--- a/arbor/simulation.cpp
+++ b/arbor/simulation.cpp
@@ -355,7 +355,7 @@ void simulation_state::reset() {
 time_type simulation_state::run(time_type tfinal, time_type dt) {
     // Progress simulation to time tfinal, through a series of integration epochs
     // of length at most t_interval_. t_interval_ is chosen to be no more than
-    // than half the network minimum delay.
+    // than half the network minimum delay and minimally the timestep `dt`.
     //
     // There are three simulation tasks that can be run partially in parallel:
     //
@@ -398,6 +398,8 @@ time_type simulation_state::run(time_type tfinal, time_type dt) {
     if (!std::isfinite(dt) || tfinal < 0) throw std::domain_error("simulation: dt must be finite, positive, and in [ms]");
 
     if (tfinal<=epoch_.t1) return epoch_.t1;
+
+    auto epoch_length = std::max(t_interval_, dt);
 
     // Compute following epoch, with max time tfinal.
     auto next_epoch = [tfinal](epoch e, time_type interval) -> epoch {
@@ -485,8 +487,8 @@ time_type simulation_state::run(time_type tfinal, time_type dt) {
     };
 
     epoch prev = epoch_;
-    epoch current = next_epoch(prev, t_interval_);
-    epoch next = next_epoch(current, t_interval_);
+    epoch current = next_epoch(prev, epoch_length);
+    epoch next = next_epoch(current, epoch_length);
 
     if (epoch_callback_) epoch_callback_(current.t0, tfinal);
 
@@ -507,7 +509,7 @@ time_type simulation_state::run(time_type tfinal, time_type dt) {
         for (;;) {
             prev = current;
             current = next;
-            next = next_epoch(next, t_interval_);
+            next = next_epoch(next, epoch_length);
             if (next.empty()) break;
 
             g.run([&]() { exchange(prev); enqueue(next); });


bcumming:
In this case, `dt > min_delay/2` will lead to invalid output because spikes won't be delivered on time. I see that the desire is to "simulate" a zero-delay connection by picking a very small connection delay, so why not just require that they user selects `delay = dt/2`?

According to Arbor's model, shouldn't `dt > min_delay/2` be a hard error?
#2263 mentions that a connection delay of 0 creates invalid output. According to Arbor's model, shouldn't the following always be true:

* `dt <= min_delay/2`
* `dt > 0`
* `min_delay > 0`

thorstenhater:
Hi @bcumming, currently I am looking at alternatives, but so far I haven't anything I can get behind:
1. Implicitly changing all delays to at least `dt`. Altering user input without consent
2. Allocating a minimum delay like EPSILON. Makes for _really_ slow simulations.
3. The variant above. Incorrect, as you noted.
4. Your suggestion of throwing on zero delay. 

Out of those, 4. is probably the least worst option.

halfflat:
My feelings about this:
* The dt provided to `simulation::run` should be regarded as firstly a maximum dt and secondly as being up to cell group implementations to interpret in their context of finding a solution (including being ignored).
* Maximum dt should be greater than zero (we can throw if not a positive number).
* Because of the above, a cell group should be free to internally set its dt in integration to less than the epoch duration.

arbor/include/arbor/recipe.hpp
```diff
@@ -54,8 +54,8 @@ struct cell_connection_base {
 
     cell_connection_base(L src, cell_local_label_type dst, float w, const U::quantity& d):
         source(std::move(src)), target(std::move(dst)), weight(w), delay(d.value_as(U::ms)) {
-        if (std::isnan(weight)) throw std::out_of_range("Connection weight must be finite.");
-        if (std::isnan(delay) || delay < 0)  throw std::out_of_range("Connection delay must be non-negative and infinite in units of [ms].");
+        if (std::isnan(weight)) throw std::domain_error("Connection weight must be finite.");
+        if (std::isnan(delay) || delay <= 0)  throw std::domain_error("Connection delay must be positive, finite, and given in units of [ms].");
     }
```
halfflat, line 59: We have quite specific exceptions such as `arb::bad_connection_source_gid` defined in `<arbexcept.hpp>`; it would be consistent to define some exceptions to throw here that derive from `arb::arbor_exception`.

thorstenhater: I am a bit torn here, `std::domain_error` fits the basic checks for `isnan` etc well and has the advantage that it plays nice with pybind11 (it'll automatically convert to `ValueError`).

arbor/include/arbor/recipe.hpp
```diff
@@ -54,8 +54,8 @@ struct cell_connection_base {
 
     cell_connection_base(L src, cell_local_label_type dst, float w, const U::quantity& d):
         source(std::move(src)), target(std::move(dst)), weight(w), delay(d.value_as(U::ms)) {
-        if (std::isnan(weight)) throw std::out_of_range("Connection weight must be finite.");
-        if (std::isnan(delay) || delay < 0)  throw std::out_of_range("Connection delay must be non-negative and infinite in units of [ms].");
+        if (std::isnan(weight)) throw std::domain_error("Connection weight must be finite.");
+        if (std::isnan(delay) || delay <= 0)  throw std::domain_error("Connection delay must be positive, finite, and given in units of [ms].");
```
halfflat, line 58: Inasmuch as we're presuming NaNs work as expected, we could just have the test `!(delay > 0)` instead of `std::isnan(delay) || delay <= 0`.

Do we really need to enforce that delay is finite? If so, then the test should include that.

Also, not being familiar (yet) with how LLNL units works, why do we need to specify that the quantity is in milliseconds? Can't we just convert as required or else assert in the type?

arbor/include/arbor/recipe.hpp
```diff
@@ -69,7 +69,7 @@ struct gap_junction_connection {
 
     gap_junction_connection(cell_global_label_type peer, cell_local_label_type local, double g):
         peer(std::move(peer)), local(std::move(local)), weight(g) {
-        if (std::isnan(weight)) throw std::out_of_range("Gap junction weight must be finite.");
+        if (std::isnan(weight)) throw std::domain_error("Gap junction weight must be finite.");
```
halfflat, line 72: Similar comments as above: testing finite or testing Nan? We should use an arbor exception.

arbor/include/arbor/recipe.hpp
thorstenhater: Well, the two issues go hand in hand: `(42 * U.ms).value_as(U.mV) == NaN`. So, receiving a nan can mean either we got `nan * U.ms` or an erroneous unit. This is why the message allows for both.

arbor/simulation.cpp
```diff
@@ -394,10 +395,12 @@ time_type simulation_state::run(time_type tfinal, time_type dt) {
...
-    if (!std::isfinite(tfinal) || tfinal < 0) throw std::domain_error("simulation: tfinal must be finite, positive, and in [ms]");
-    if (!std::isfinite(dt) || tfinal < 0) throw std::domain_error("simulation: dt must be finite, positive, and in [ms]");
-
-    if (tfinal<=epoch_.t1) return epoch_.t1;
+    // Compare up to picoseconds
+    time_type eps = 1e-9;
+    if (!std::isfinite(dt) || dt < eps) throw std::domain_error("simulation: dt must be finite, positive, and in [ms]");
+    if (dt - t_interval_ > eps) throw std::domain_error(util::pprintf("simulation: dt={}ms is larger than epoch length by {}, chose at most half the minimal connection delay {}ms.", dt, dt - t_interval_, t_interval_));
+    if (!std::isfinite(tfinal) || tfinal < eps) throw std::domain_error("simulation: tfinal must be finite, positive, and in [ms]");
+    if (tfinal - epoch_.t1 < dt) throw std::domain_error(util::pprintf("simulation: tfinal={}ms doesn't make progress of least one dt; current time of simulation is {}ms and dt {}ms", tfinal, epoch_.t1, dt));
```
halfflat, line 403: This seems all a bit fiddly; as mentioned in the general comments, I think we should just leave dt interpretation up to the integrators and they can make a sensible choice, e.g. clip it by epoch duration or interpret very small dt values as being larger, based on whatever they have to do. Having eps here is untidy because it splits the responsibility for sane dt interpretation between the main loop and the integrators.

halfflat: Also, if someone wants to run arbor for ever, do we really need to stop them? We already have a check for tfinal being less than the end time of the preceding epoch, where we return the correct 'simulated up to' time, so we don't really need tests for it being zero, or negative; in fact a tfinal of zero should be a valid no-op in my opinion.

In short, a isnan test for tfinal remains sufficient, I believe.

halfflat: No we don't have to pass it down the stack: we know epochs are at most min_delay/2 long. The epoch already has all the info they need, and if there is more info it will be in the cell kind global data, which they also have.

halfflat: #2053 is just cable cells, and I still want to partially revert it so that we can use flexible time steps with a different cable cell integrator, even if it's just to make them line up correctly with epoch intervals.

halfflat: As an example, if we have the last epoch (ending on tfinal) having duration 3.1 dt, wouldn't it be best to set the fixed dt for that epoch to be e.g. the duration/4 and then we would know that all cell states across cell groups were actually at the same integration time?

thorstenhater: > Also, if someone wants to run arbor for ever, do we really need to stop them? We already have a check for tfinal being less than the end time of the preceding epoch, where we return the correct 'simulated up to' time, so we don't really need tests for it being zero, or negative; in fact a tfinal of zero should be a valid no-op in my opinion.

Technically I agree. Practically I know that I'd spent way too much time looking for the reason why my simulation did nothing in that situation. Especially since our `time_type` doesn't distinguish time points from durations, so `run(5 *ms)` really could mean _run until t=5ms_ or _run for 5ms_. Thus I think it's friendlier towards the user to tell them 'this is a no-op and you likely didn't mean that'. Especially, as I cannot currently imagine situations where this is a semantically meaningful request.

halfflat: As a general principle, I think we should keep things general unless there is a good reason not to. I mean, the variable is literally called tfinal, and we return the simulation state time - it's a simple interface with simple semantics (from the outside) and there's nothing stopping us adding checks on the outside of that if we want to provide more hand holds in the Python interface (though I don't think we should there either).

thorstenhater: Damn you and your arguments :D

arbor/include/arbor/recipe.hpp
jlubo: Here, one might provide additional information for users who want to have a minimum delay. For example:
```
if (!std::isfinite(delay) || delay <= 0)  throw std::domain_error("Connection delay must be positive, finite, and given in units of [ms]. The minimum possible value is one timestep.");
```

jlubo: And another thing: shouldn't there be a machine epsilon for comparisons with float values like `delay`? To avoid that some value very close to `0`, which still meets `>0`, is used?

jlubo (commented):
@thorstenhater please see my comments above. Apart from that, looks good to me.

arbor/include/arbor/recipe.hpp
thorstenhater: > `if (!std::isfinite(delay) || delay <= 0)  throw std::domain_error("Connection delay must be positive, finite, and given in units of [ms]. The minimum possible value is one timestep.");`

I decided to follow Sam's argument and practicality here. Knowing `dt` isn't feasible at connection construction, since _that_ is only declared with `sim.run(dt=...)`. I see the point though and compromise a compromise: 
Let `sim.run` fail if `dt > min_delay`. For now. If we want to get cutesy later, we can revert that error state and add whatever smarts to the cell groups we desired. @halfflat @jlubo ? (Note that this is the last PR needed for 0.11)

thorstenhater: > And another thing: shouldn't there be a machine epsilon for comparisons with float values like delay? To avoid that some value very close to 0, which still meets >0, is used?

Comparing against 0 is fine, for this at least, since users likely type `0` by accident and not `1e-306`. _Also_ this would be absorbed by the check during `run` above.

jlubo: I see, that makes sense, thanks for the clarification.

jlubo (approved):

PR #2265, opened 2024-04-03
Remove `simulation::inject_events`

thorstenhater:
This removes `simulation::inject_events` which is made obsolete by `recipe::event_generators`. 
The method exposed internals by giving access to the `lid` of synapses (over labels) and broke 
abstractions (`gid` must be present on this MPI rank and all events must be deliverable during
the current time (as per documentation)). It was used only in a few tests and examples and these
uses were replaced by `event_generator`s.


bcumming (approved):
lgtm

PR #2267, opened 2024-04-12
Use CPM and simplify CMake.

thorstenhater:
There. I finally cracked. While trying to get CMake to play ball with Tracy I had to devise the next
bespoke solution for adding a dependency. Now, we are just using CPM https://github.com/cpm-cmake/CPM.cmake
which makes things significantly less painful, if not painless, it's still CMake after all.

This means:
- No more submodules. Everything is pulled in via CPM.
- No more toggling between system packages and submodules. System is preferred, if compatible.
- No more ten ways of adding a dependency. It's two now. (I'd prefer one, but what can we do?)
- Significantly less code in our build system.


ErbB4 (dismissed):

CMakeLists.txt
```diff
@@ -363,8 +363,10 @@ CPMAddPackage(NAME units
                       "UNITS_BUILD_WEBSERVER OFF")
 add_library(ext-units INTERFACE)
 if(units_ADDED)
+    message(WARNING "UNITS Vendored")
```
ErbB4: the warning message is not easy to follow...

CMakeLists.txt
```diff
@@ -363,8 +363,10 @@ CPMAddPackage(NAME units
...
     target_link_libraries(ext-units INTERFACE units::units)
 else()
+    message(WARNING "UNITS Package")
```
ErbB4: package?

boeschf (approved):

PR #2271, opened 2024-04-23
build and install with miniconda

ErbB4:
from source

### DIFF
--- a/doc/install/build_install.rst
+++ b/doc/install/build_install.rst
@@ -1019,7 +1019,14 @@ Replace <site-packages> with the path you get in the previous operation before â
 ln -sf /lib/x86_64-linux-gnu/libstdc++.so.6 ~/miniconda3/envs/arbor_test/bin/../lib/libstdc++.so.6
 
 
-#go to any working directory to try if you successfully installed arbor, by starting python and importing arbor
+#go to any working directory to try if you successfully installed arbor, by starting python and importing arbor.
+
+One thing to add here could be testing for the version, i.e.
+
+python -c 'import arbor; print(arbor.__version__)'
+
+should work without errors and print something like 0.91-dev.
+
 
 python
 import arbor
--- a/doc/install/build_install.rst
+++ b/doc/install/build_install.rst
@@ -965,7 +965,7 @@ need to be `updated <install-downloading_>`_.
 Build in conda virtual environment using miniconda
 ==================================================
 
-If you hope to install Arbor from source in a virtual environment in order not to interfere with other Arbor versions or software you have installed, or if you want to use features that are not included in the official release, you can use miniconda for this purpose. Otherwise, type pip install arbor in your terminal to install arbor. But you will have to fix two issues, one is feeding the correct Python path, and the other is 'GLIBCXX 3.4.30 not found in conda environment'.
+If you hope to install Arbor from source in a virtual environment in order not to interfere with other Arbor versions or software you have installed, or if you want to use features that are not included in the official release, you can use miniconda for this purpose. Otherwise, type pip install arbor in your terminal to install arbor. But you will have to fix two issues, one is feeding the correct Python path, and the other is 'GLIBCXX 3.4.30 not found in conda environment'. (Please note: This is assuming Linux and MacOS is not yet covered.)
 
 .. code-block:: bash
 
--- a/doc/install/build_install.rst
+++ b/doc/install/build_install.rst
@@ -960,3 +960,7 @@ need to be `updated <install-downloading_>`_.
         git submodule update
     Or download submodules recursively when checking out:
         git clone --recurse-submodules https://github.com/arbor-sim/arbor.git
+
+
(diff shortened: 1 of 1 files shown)


doc/install/build_install.rst
```diff
@@ -960,3 +960,73 @@ need to be `updated <install-downloading_>`_.
...
+#redirect the libc files such that the miniconda environment can access it
+
+ln -sf /lib/x86_64-linux-gnu/libstdc++.so.6 /home/han/miniconda3/envs/arbor_test/bin/../lib/libstdc++.so.6
+
+
+#go to any working directory to try if you successfully installed arbor, by starting python and importing arbor
+
+python
+import arbor
+
```
thorstenhater, line 1026: One thing to add here could be testing for the version, i.e.
```
python -c 'import arbor; print(arbor.__version__)'
```
should work without errors and print something like `0.91-dev`.

doc/install/build_install.rst
```diff
@@ -960,3 +960,73 @@ need to be `updated <install-downloading_>`_.
...
+python -c 'import numpy; print(numpy.__path__)'
+
+#load the right path to the one used for installing
+
+cp -r ~/miniconda3/envs/arbor_test/src/arbor/build/python/arbor <site-packages>
+Replace <site-packages> with the path you get in the previous operation before ‘/numpy’
+
+#redirect the libc files such that the miniconda environment can access it
+
+ln -sf /lib/x86_64-linux-gnu/libstdc++.so.6 /home/han/miniconda3/envs/arbor_test/bin/../lib/libstdc++.so.6
```
thorstenhater, line 1019:
```suggestion
ln -sf /lib/x86_64-linux-gnu/libstdc++.so.6 ~/miniconda3/envs/arbor_test/bin/../lib/libstdc++.so.6
```

doc/install/build_install.rst
```diff
@@ -960,3 +960,73 @@ need to be `updated <install-downloading_>`_.
         git submodule update
     Or download submodules recursively when checking out:
         git clone --recurse-submodules https://github.com/arbor-sim/arbor.git
+
+
+Build in conda virtual environment using miniconda
+==================================================
+
+If you hope to install Arbor from source in a virtual environment in order not to interfere with other Arbor versions or software you have installed, or if you want to use features that are not included in the official release, you can use miniconda for this purpose. Otherwise, type pip install arbor in your terminal to install arbor. But you will have to fix two issues, one is feeding the correct Python path, and the other is 'GLIBCXX 3.4.30 not found in conda environment'.
```
thorstenhater, line 968: Maybe note that this is assuming Linux and MacOS is not yet covered.

doc/install/build_install.rst
```diff
@@ -960,3 +960,73 @@ need to be `updated <install-downloading_>`_.
...
+mkdir build
+cd build
+
+cmake .. -GNinja -DCMAKE_CXX_COMPILER=$(which g++) -DCMAKE_C_COMPILER=$(which gcc) -DARB_WITH_PYTHON=ON -DARB_VECTORIZE=ON -DPython3_EXECUTABLE=$(which python3) -DARB_USE_BUNDLED_LIBS=ON
+
+
+#activate ninja to install
+
+ninja
+sudo ninja install
```
thorstenhater, line 1004: You don't need this and it'll install to the full system, i.e. outside any conda env, so maybe not recommend it ;)

doc/install/build_install.rst
```diff
@@ -960,3 +960,73 @@ need to be `updated <install-downloading_>`_.
...
+
+#create a virtual environment
+
+conda create --name arbor_test
+conda activate arbor_test
+
+
+#go to the folder and clone the Arbor source package from GitHub
+
+cd ~/miniconda3/envs/arbor_test/
```
thorstenhater, line 980: This particular path is not required, a note towards might be helpful for a beginner. Something
like 'you can put this anywhere you like, but we chose this path.'

doc/install/build_install.rst
```diff
@@ -960,3 +960,73 @@ need to be `updated <install-downloading_>`_.
...
+#go to any working directory to try if you successfully installed arbor, by starting python and importing arbor
+
+python
+import arbor
+
+#then deactivate the environment if no more actions are planned. In the future, always first activate the virtual environment with and then use arbor in this environment with:
+
+conda activate arbor_test
+python
+import arbor
```
thorstenhater, line 1031: you can add `>>>` here, which is the usual prompt from python to show that this command goes to the Python interpreter.

thorstenhater (changes requested):

doc/install/build_install.rst
ErbB4: added to the main text

doc/install/build_install.rst
ErbB4: added. thanks for the advice.

thorstenhater (approved):
Thanks, look great

PR #2274, opened 2024-05-21
🧹  Major housekeeping

### LINKED ISSUE #2273: Weekly pip builds failing (for some time now)
- Python version bump causes installation of packages to fail outside of venvs.
- Artifact actions need an update.
- The notification goes to who last touched the workflow file, so that may be why you've been missing it. Any change pushed by an Arbor team member will cause them to go to you, which is better than going to me :)

**Describe the bug**

See: https://github.com/arbor-sim/arbor/actions/runs/8955059490
Looks like a bump in Python versions, which changed behavior to not allow installation outside venvs. That's probably the only change you need to make, setup a venv first.

thorstenhater:
Housekeeping before v0.10.0

## bump and deprecate dependencies

Go through all submodules, build docs, and other dependencies to adjust version numbers.
Submodules go to Spack's latest

## `pip` now protects managed environments
MacOS `brew` has becomes finicky lately when it comes to installing globally via `pip`,
as new versions of `pip` are no longer allowing the user to break global environments.
Fix by forcing `pip` to disregard that rule, as we set up and tear down VMs for precisely
this purpose.

Fixes #2273 

## spack develop started to misbehave

Massage code in package.py around handling optimization flags. Output is a `version` object,
input expects a string. Adjust by converting to string

## fix dubious type punning

`network.cpp` uses `reinterpret_cast` to type pun. Use `memcpy` instead as recommended.
Also got flagged by compiler warning with newer GCC

## `result_of` -> `invoke_result`

Flagged as deprecated.

## fix includes in `test_*`

Newer versions of GCC require including `algorithm` directly.


doc/dependencies.csv
```diff
@@ -2,29 +2,22 @@ Build option/target,Tool name,Minimum version,Notes,Update criteria
 --,CMake,3.19,,"* Is available as a module on HPC systems
 * When arbitrary changes in more recently released versions make supported"
 --,C++ compiler,,C++17 support. See below,
---,GCC,9.0.0,,"* it is supported by the minimum version of CUDA.
+--,GCC,12.0.0,,"* it is supported by the minimum version of CUDA.
 * it is available either by default or can be installed using standard package manager on the supported Linux versions."
---,Clang,10.0,Needs GCC 9 or later for standard library.,
---,Apple Clang,11,Apple LLVM version 11.0.0 (clang-1100.0.33.8),
+--,Clang,12.0,,
+--,Apple Clang,13,Apple LLVM version 13.0.0 (clang-1300.x.y.z),
```
ErbB4, line 8: how do I find the more recent Clang version supported by Apple? I read something here https://forums.developer.apple.com/forums/thread/746509

But the numbers do not match exactly

doc/dependencies.csv
```diff
@@ -2,29 +2,22 @@ Build option/target,Tool name,Minimum version,Notes,Update criteria
...
+ARB_GPU,CUDA,12.0,,"* It is available on all of the target HPC systems (Piz Daint, Juwels Booster)"
 ARB_WITH_MPI,MPI,,Many MPI implementations are supported.,
 "ARB_WITH_MPI, ARB_WITH_PYTHON ",mpi4py,3.1.0,Minimum version depends on compatiblity with your Python and MPI versions. 3.1.0 is the first release to support Python > 3.8.,
 html,     .. literalinclude:: ../requirements.txt,,See ``doc/requirements.txt``,
 unit,googletest,1.12.1,submodule ``ext/gtest.*``,
 bench,Google-benchmark,,submodule ``ext/google-benchmark``,
---,json,3.11.2,submodule ``ext/json``,
---,pugi,1.13,submodule ``ext/pugixml``,
+--,json,3.11.3,submodule ``ext/json``,
+--,pugi,1.14,submodule ``ext/pugixml``,
```
ErbB4, line 17: what are the cretiria for these dependencies? Where do I find the most recent versions?

.github/workflows/build-pip-wheels.yml
```diff
@@ -6,7 +6,6 @@ on:
       - ciwheel
     tags:
       - v*
-
   schedule:
     - cron: '0 2 * * 0' # run at 2 AM every sunday
```
ErbB4, line 10: why regularly and so specific?

.github/workflows/test-matrix.yml
```diff
@@ -17,70 +17,70 @@ jobs:
...
         - {
             name:  "Linux Min GCC",
-            os:    "ubuntu-20.04",
-            cc:    "gcc-9",
-            cxx:   "g++-9",
-            py:    "3.8",
+            os:    "ubuntu-22.04",
+            cc:    "gcc-12",
+            cxx:   "g++-12",
+            py:    "3.9",
```
ErbB4, line 23: not 3.10?

.github/workflows/test-matrix.yml
```diff
@@ -17,70 +17,70 @@ jobs:
...
             name:  "MacOS Max",
-            os:    "macos-13",
+            os:    "macos-14",
             cc:    "clang",
             cxx:   "clang++",
-            py:    "3.11",
+            py:    "3.12",
             cmake: "3.22.x",
             mpi:   "ON",
             simd:  "OFF"
```
ErbB4, line 86: how do you remember all the versions for different systems? awesome...

.github/workflows/test-pip.yml
```diff
@@ -14,8 +14,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-20.04]
-        python-version: [3.8]
+        os: [ubuntu-24.04]
+        python-version: [3.12]
```
ErbB4, line 18: now 3.12?

doc/dependencies.csv
thorstenhater: These bundled libraries should be installable via spack, thus we must not be more recent then
the latest version there. If this is not the case, the CI workflow for spack will fail, so we got an
automated check there. 

Other than that, we try to track the latest stable releases.

.github/workflows/test-matrix.yml
thorstenhater: As 22.04 has become the 'older' version to support, we test the lowest version of Python we support here.

doc/dependencies.csv
thorstenhater: That's a lot of manual I am afraid. I source the versions here
https://en.wikipedia.org/wiki/Xcode
https://en.wikipedia.org/wiki/Clang
https://gcc.gnu.org/releases.html

and for CUDA I use the NVIDIA blog (I found no handy table, sadly)

ErbB4 (approved):

ErbB4:
well cleaned ;-)

PR #2277, opened 2024-06-07
added a comment

ErbB4:
bored and starting over my learning process without leaving any single detail this time.

### DIFF
--- a/doc/python/morphology.rst
+++ b/doc/python/morphology.rst
@@ -119,6 +119,7 @@ Cable cell morphology
         Radius of the cable (μm)
 
 .. class:: msegment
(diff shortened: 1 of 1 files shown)


doc/python/morphology.rst
```diff
@@ -119,6 +119,7 @@ Cable cell morphology
         Radius of the cable (μm)
 
 .. class:: msegment
+Consider adding a sentence here explaining what this class does? (I don't know how to put comments in .rst file)
```
thorstenhater, line 122: This is just text and `.. class::` markup is extra. You should be able to just write any text here you like.

ErbB4 (commented):
resolved

PR #2289, opened 2024-07-24
fixed the writing

ErbB4:


thorstenhater (changes requested):
Thanks, but if I may, I'd request one additional minor change.

thorstenhater (approved):

PR #2302, opened 2024-07-26
Use `-march` correctly

thorstenhater:
Selecting tuning flags depends on compilers and architectures. 
Account for this complication which became more apparent w/
proliferation of Arm64 based chips.

Following from: https://maskray.me/blog/2022-08-28-march-mcpu-mtune

Also fix a new --- sensible --- warning.


cmake/CompilerOptions.cmake
```diff
@@ -110,18 +110,26 @@ function(set_arch_target optvar optvar_cuda_guarded arch)
         endforeach(line)
         string(REGEX REPLACE "-.*" "" target_model "${target}")
 
-        # Use -mcpu for all supported targets _except_ for x86 and Apple arm64, where it should be -march.
-
+        # Figure out which flags to pass to compiler to tune for concrete
+        # architecture.
+        # See clang / gcc manuals and:
+        # https://maskray.me/blog/2022-08-28-march-mcpu-mtune
         if (CMAKE_CXX_COMPILER_ID MATCHES "AppleClang" AND CMAKE_CXX_COMPILER_VERSION LESS 15)
```
ErbB4, line 117: doesn't look like the 'if' syntax I have learned in C++. sure correct?

thorstenhater: Yes, this is a CMake script, not C++.

cmake/CompilerOptions.cmake
```diff
@@ -110,18 +110,26 @@ function(set_arch_target optvar optvar_cuda_guarded arch)
...
+                # on AArch64, this is correct, ...
+                set(arch_opt "-mcpu=${arch} -mtune=${arch}")
             else()
-                set(arch_opt "-mcpu=${arch}")
-            endif()
+                # ... however on x86 mcpu _is_ mtune _and_ deprecated (since 2003!), but ...
+                set(arch_opt "-march=${arch}")
+            endif ()
+        elseif (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+            # ... clang likes mcpu!
```
ErbB4, line 128: the comment says sth different from the code...

ErbB4 (approved):

halfflat:
Just a query: what was the warning that d571598 fixed? A quick test on Godbolt gave me nothing, and I'm curious.

thorstenhater:
I can reproduce it, but if it is sufficient, as far as I remember GCC 14 complained about `memcpy` used on a dubious pointer in STL. Seems likely to be linked to working on MacOS. I wouldn't even have fixed it, if I didn't find the new version clearer.

PR #2303, opened 2024-07-26
simply typeseting

ErbB4:

### DIFF
--- a/doc/tutorial/network_ring_mpi.rst
+++ b/doc/tutorial/network_ring_mpi.rst
@@ -4,7 +4,7 @@ Distributed ring network (MPI)
 ==============================
 
 In this example, the ring network created in an :ref:`earlier tutorial <tutorialnetworkring>` will be used to run the model in
-a distributed context using MPI. Only the differences with that tutorial will be described.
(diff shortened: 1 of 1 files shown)


doc/tutorial/network_ring_mpi.rst
```diff
@@ -4,7 +4,7 @@ Distributed ring network (MPI)
 ==============================
 
 In this example, the ring network created in an :ref:`earlier tutorial <tutorialnetworkring>` will be used to run the model in
-a distributed context using MPI. Only the differences with that tutorial will be described.
+a distributed context using MPI. Only the differences between that tutorial and the other will be described.
```
thorstenhater, line 7: I'd suggest re-writing the whole paragraph, for example:
```
In this example, we will build on an :ref:`earlier tutorial <tutorialnetworkring>`,
to demonstrate usage of distributed contexts. The network will be unaltered, 
but its execution is parallelized using MPI. As these concerns are largely orthogonal
we will discuss only the differences between the two simulations.

thorstenhater (changes requested):

thorstenhater (approved):

PR #2329, opened 2024-07-30
grammar fix

ErbB4:

### DIFF
--- a/doc/fileformat/swc.rst
+++ b/doc/fileformat/swc.rst
@@ -119,7 +119,7 @@ joined at the soma center. It differs in two ways:
 
 **The axon, dendrite, and apical sub-trees follow special rules for attachment to
 the soma**: By default, the sub-tree starts at the first sample with the
-dendrite, axon, or apical tag [what is apical tag?], and not at the parent location on the soma, and
+a tag indicating a non-soma segment, and not at the parent location on the soma, and
 the sub-tree is connected to its parent with a "zero resistance wire".
 **Except** when the sub-tree is defined by a single child sample. In which case
 the sub-tree is composed of a single segment from the parent location on the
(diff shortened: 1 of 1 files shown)


doc/fileformat/swc.rst
```diff
@@ -96,7 +96,7 @@ axons and apical dendrites are attached to it that are not described explicitly
 
 **Every sample must have the same SWC identifier (tag) as its parent, except for
 samples whose parent is tagged as soma**:
-This enforces that axons, dendrites and apical dendrites can only attach to the soma.
+This enforces that axons, [basel? is missing] dendrites, and apical dendrites can only attach to the soma.
```
thorstenhater, line 99:
```suggestion
This enforces that axonal and dendritic segments can only attach to the soma.
```
Yes, both apical and basal, so let's compress this to encompass both.

doc/fileformat/swc.rst
```diff
@@ -117,24 +117,24 @@ joined at the soma center. It differs in two ways:
   ``x=x0`` and the third at ``x=x0+r``, to form a single section.
 * All dendrites, axons, and apical dendrites are attached to the center of the soma with "zero resistance wires".
 
-**The axon, dendrite and apical sub-trees follow special rules for attachment to
+**The axon, dendrite, and apical sub-trees follow special rules for attachment to
 the soma**: By default, the sub-tree starts at the first sample with the
-dendrite, axon or apical tag, and not at the parent location on the soma, and
+dendrite, axon, or apical tag [what is apical tag?], and not at the parent location on the soma, and
```
thorstenhater, line 122:
```suggestion
a tag indicating a non-soma segment, and not at the parent location on the soma, and
```
It's a slightly confusing way of stating what a segment layout like this
```
(soma) - (soma) - (axon) - (axon)
```
should be, where `()` indicates a sample of `(x, y, z, r, tag)` where I have just shown
the tags. We need to convert this from a list of points to a list of segments with attached 
regions (soma, axon, ...). In particular we get this
```
(soma) [soma] (soma) [soma] (axon) [axon] (axon)
```
and not
```
(soma) [soma] (soma) [axon] (axon) [axon] (axon)
```
Both would have been possible solutions.

thorstenhater (changes requested):

doc/fileformat/swc.rst
thorstenhater: We are really getting into the intricacies of both Neuron and Arbor here...

thorstenhater (approved):

PR #2332, opened 2024-07-30
fixed sth but not sure if correct

ErbB4:

### DIFF
--- a/doc/fileformat/cable_cell.rst
+++ b/doc/fileformat/cable_cell.rst
@@ -66,7 +66,7 @@ The components of the label dictionary are the following:
    This expression identifies the radius iexpr with a scaling factor of 0.5.
 
 
-For any number of locset or region, an iexpr definitions can be grouped in a label dictionary as follows:
+Any  definition of a locset, region, or iexpr can be stored in a label dictionary as follows.
 
 .. label:: (label-dict [...def:region-def/locset-def/iexpr-def])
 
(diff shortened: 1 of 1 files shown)


doc/fileformat/cable_cell.rst
```diff
@@ -63,14 +63,14 @@ The components of the label dictionary are the following:
 
       (iexpr-def "my_iexpr" (radius 0.5))
 
-   This expression identifies the radius iexpr with a scaling factor 0.5.
+   This expression identifies the radius iexpr with a scaling factor of 0.5.
 
 
-Any number of locset, region an iexpr definitions can be grouped in a label dictionary as follows:
+For any number of locset or region, an iexpr definitions can be grouped in a label dictionary as follows:
```
thorstenhater, line 69:
```suggestion
Any  definition of a locset, region, or iexpr can be stored in a label dictionary as follows.
```
Label dictionaries give names to expressions that might be sensible to reuse. E.g.
```
ld = A.label_dict()
ld['term-dend'] = '(restrict (tag 3) (terminal))'
```
and we can then use these in placing things
```
decor.place('"term-dend"', A.synapse(...))
```
or in other expressions
```
ld['int'] = '(proximal-interval (locset "term-dend") 10)'
A.paint('"int", ...)
```
If any of that is not sufficiently explained here or in the labels discussion, we/I can update those places.

thorstenhater (changes requested):

thorstenhater (approved):

PR #2342, opened 2024-07-31
fix grammar, type, code styling

ErbB4:

### DIFF
--- a/doc/python/labels.rst
+++ b/doc/python/labels.rst
@@ -26,36 +26,36 @@ Cable cell labels
 
    .. attribute:: regions
 
-      The region definitons in the dictionary.
+      The region definitions in the dictionary.
 
    .. attribute:: locsets
 
-      The locset definitons in the dictionary.
+      The locset definitions in the dictionary.
 
 The ``arbor.label_dict`` type is used for creating and manipulating label dictionaries,
 which can be initialised with a dictionary that defines (label, :ref:`expression <labels-expressions>`)
 pairs. For example, a dictionary that uses tags that correspond to SWC
 `structure identifiers <http://www.neuronland.org/NLMorphologyConverter/MorphologyFormats/SWC/Spec.html>`_
-to label soma, axon, dendrite and apical dendrites is:
+to label soma, axon, basal dendrites, and apical dendrites is:
 
 .. code-block:: python
 
-    import arbor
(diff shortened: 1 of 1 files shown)


doc/python/labels.rst
```diff
@@ -103,26 +103,26 @@ of both the *'dend'* and *'apic'* regions.
             # equivalent to (join (tag 3) (tag 4))
             'tree': '(join (region "dend") (region "apic"))'})
 
-The order that labels are defined does not matter, so an :ref:`expression <labels-expressions>` can refer to a
+The order which labels are defined does not matter, so an :ref:`expression <labels-expressions>` can refer to a
```
thorstenhater, line 106:
```suggestion
The order which labels are defined in does not matter, so an :ref:`expression <labels-expressions>` can refer to a
```
Not sure, but as is seems incomplete.

doc/python/labels.rst
```diff
@@ -142,9 +142,9 @@ two labels refer to one another:
     # definition of 'reg'.
 
 .. note::
-    In the example above there will be no error when the label dictionary is defined.
+    In the example above, there will be no error when the label dictionary is defined.
     Instead, there will be an error later when the label dictionary is applied to
-    a morphology, and the cyclic dependency is detected when thingifying the locations
+    a morphology, and the cyclic dependency is detected when thingifying (?) the locations
```
thorstenhater, line 147:
```suggestion
    a morphology, and the cyclic dependency is detected when thingifying the locations
```
*'Thingification'* is the term the elders agreed upon after a long and arduous fight. 
Correct would be *'reification'* (rei = greek for thing); the process of turning abstract
into --- well --- a thing. It's used in CS when you need to manipulate a concept that is normally
not an object in your language, e.g. you cannot pass a type to a (non-template) function
in C++, but you _can_ pass a numerical tag naming that type.

So, thingify is now a term in our API and thus used in the documentation. Concretely, it's used
to describe the process of mapping this `'(proximal (tag 3))'` **and** a morphology to a set of
points.

thorstenhater (changes requested):

doc/python/labels.rst
thorstenhater: As to why it wasn't 'reification', that was deemed to arcane for common use.

PR #2353, opened 2024-07-31
Update interconnectivity.rst

ErbB4:

### DIFF
--- a/doc/cpp/interconnectivity.rst
+++ b/doc/cpp/interconnectivity.rst
@@ -50,8 +50,14 @@ Interconnectivity
 
     .. cpp:member:: cell_remote_label_type source
 
-        Source end point, represented by a :cpp:type:`cell_remote_label_type` which packages
-        a cell gid, integral tag of a group of sources on the cell, and source selection policy. [hard to read this line]
+        Source end point, comprising global id, uniquely identifying the source object inside the remote simulation, local id, uniquely identifying the location of the source on the source object
+    selection policy, how to map sources (gid, lid) to the targets in Arbor.
+
+    .. Note::
+    The tuple (gid, lid) must be unique for each source and follow Arbor's model of multiple
+    sources (threshold detectors) per cell. If the remote simulation does not provide multiple items per
+    source object, lid can be ignored and set to zero.
+
 
     .. cpp:member:: cell_local_label_type target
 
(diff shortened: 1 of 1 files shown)


doc/cpp/interconnectivity.rst
```diff
@@ -45,13 +45,13 @@ Interconnectivity
     a cell or a spike source. The target is a synapse on the post-synaptic cell.
 
     The :cpp:member:`target` does not include the gid of a cell, this is because a
-    :cpp:class:`ext_cell_connection` is bound to the target cell which means that the gid
+    :cpp:class:`ext_cell_connection` is bound to the target cell, which means that the gid
     is implicitly known.
 
     .. cpp:member:: cell_remote_label_type source
 
         Source end point, represented by a :cpp:type:`cell_remote_label_type` which packages
```
thorstenhater, line 53: Agreed, please advise to make it better. How about:

Source endpoint, comprising
- global id; uniquely identifying the source object inside the remote simulation
- local id; uniquely identifying the location of the source on the source object
- selection policy; how to map sources `(gid, lid)` to the targets in Arbor.

Note: The tuple `(gid, lid)` must be unique for each source and follows Arbor's model of multiple
sources (threshold detectors) per cell. If the remote simulation does not provide multiple items per
source object, `lid` can be ignored and set to zero.

doc/cpp/interconnectivity.rst
```diff
@@ -359,13 +359,13 @@ Interconnectivity
...
-        Label dictionary for named selecations and values.
+        Label dictionary for named selections and values.
 
 
 .. function:: generate_network_connections(recipe, context, decomp)
 
         Generate network connections from the network description in the recipe. Only generates connections 
-        with local gids in the domain composition as target.  Does not include connections from
+        with local gids in the domain composition as the target.  Does not include connections from
         the "connections_on" recipe function.
```
thorstenhater, line 369:
```suggestion
        the ``connections_on`` recipe function.
```

doc/cpp/interconnectivity.rst
```diff
@@ -45,13 +45,13 @@ Interconnectivity
     a cell or a spike source. The target is a synapse on the post-synaptic cell.
 
     The :cpp:member:`target` does not include the gid of a cell, this is because a
-    :cpp:class:`ext_cell_connection` is bound to the target cell which means that the gid
+    :cpp:class:`ext_cell_connection` is bound to the target cell, which means that the gid
```
thorstenhater, line 48:
```suggestion
    :cpp:class:`ext_cell_connection` is bound to the target cell, which means that the ``gid``
```

thorstenhater (changes requested):

thorstenhater (changes requested):

PR #2383, opened 2024-08-14
added E_R to lif cell model

ErbB4:
added a missing construction argument E_R to LIF cell in python


thorstenhater:
One thing before we review / merge this: #2384 will automate type stubs. As written this PR (#2383) has outdated stubs in `python/stubs`, but we shouldn't rely on contributors to manually update them. 
Thus suggestion: review / merge #2384 first, then this.

thorstenhater (approved):

PR #2385, opened 2024-08-14
Allow unwrapping of wrappers.

### LINKED ISSUE #2254: Improve arithmetic interface of quanity stubs
There's some stubs like `arbor.membrane_capacitance` that aren't very Python friendly, to retrieve the value it seems I have to parse their string representations:

```
>>> x = arbor.membrane_capacitance(1 * units.F / units.m2)
>>> x + 1
TypeError: unsupported operand type(s) for +: 'arbor._arbor.membrane_capacitance' and 'int'
>>> x * 2
TypeError: unsupported operand type(s) for *: 'arbor._arbor.membrane_capacitance' and 'int'
>>> int(x)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'arbor._arbor.membrane_capacitance'
>>> float(x)
TypeError: float() argument must be a string or a real number, not 'arbor._arbor.membrane_capacitance'
>>> dir(x)
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__']
>>> x < 1
TypeError: '<' not supported between instances of 'arbor._arbor.membrane_capacitance' and 'int'
>>> x >= 1
TypeError: '>=' not supported between instances of 'arbor._arbor.membrane_capacitance' and 'int'
>>> float(x)
TypeError: float() argument must be a string or a real number, not 'arbor._arbor.membrane_capacitance'
>>> float(str(x))
ValueError: could not convert string to float: 'Cm=1'
>>> float(str(x)[3:])
1.0
```

There's also `arbor.axial_resistivity` and probably others. These are returned from `decor.paintings()` and make it pretty opaque to inspection.

thorstenhater:
Fixes #2254

### DIFF
--- a/python/cells.cpp
+++ b/python/cells.cpp
@@ -410,46 +410,62 @@ void register_cells(py::module& m) {
     );
 
     membrane_potential
-        .def(py::init([](const U::quantity& v) -> arb::init_membrane_potential { return {v }; }))
-        .def("__repr__",
-             [](const arb::init_membrane_potential& d){
-                 return "Vm=" + to_string(d.value) + " scale=" + to_string(d.scale);});
+        .def(py::init([](const U::quantity& v) -> arb::init_membrane_potential { return {v}; }))
+        .def_property_readonly("value", [](const arb::init_membrane_potential& d) -> U::quantity { return d.value * U::mV; })
+        .def_property_readonly("scale", [](const arb::init_membrane_potential& d) -> arb::iexpr { return d.scale; })
+        .def("__repr__", [](const arb::init_membrane_potential& d) { return "Vm=" + to_string(d.value) + "mV scale=" + to_string(d.scale);});
 
     revpot_method
-        .def(py::init([](const std::string& ion,
-                         const arb::mechanism_desc& d) -> arb::ion_reversal_potential_method {
-            return {ion, d};
-        }))
-        .def("__repr__", [](const arb::ion_reversal_potential_method& d) {
-            return "ion" + d.ion + " method=" + d.method.name();});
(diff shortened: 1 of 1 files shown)


Helveg:
Ok so if I scanned the changes correctly, one could now `arbor.membrane_capacitance(...).value` and that would be a `quantity` object, so `x.value.value` would be the magnitude?

thorstenhater:
Exactly. It's a bit awkward, but there's also the `scale`, which is a scalar (not a quantity) potentially varying along the dendrite. Thus, we have the split into `(value=(magnitude, unit), scale=iexpr)`.

Helveg:
I don't have arbor in my software environment anymore, and no bandwidth really :(

boeschf:
`x.value.value` looks indeed a bit awkward. Wouldn't it be more consistent with the canonical use of units if the `magnitude`/`value` can be queried by `x.value` and the `quantity` by `x.quantity`  (i.e. `x.value == x.quantity.value`)? Not sure if `quantity` can be used as a property here, though.

thorstenhater:
Yes, that seem like a good solution. As we use a base value and a potentially varying scale, how about
```
base :: double
base_as :: unit -> double
scale :: iexpr
unit :: unit
```
that should cover all the bases (!!).

boeschf:
yes, better still. should it be base or basis?

thorstenhater:
I wonder about that, too, but I tend towards base (as in 'base ten').

Issue #2382, opened 2024-08-10
0.10.1: The library lib/libunits.a shouldn't be installed


yurivict:
This is a bundles library that should only be used internally.

thorstenhater:
Hi @yurivict,

thanks for catching.

thorstenhater:
Hi @yurivict,

is this still an issue (in 0.11)?

PR #2390, opened 2024-08-21
Don't install unneeded stuff for units.

thorstenhater:
Closes: #2382


thorstenhater:
Subsumed by #2267 and incompatible if not.

PR #2393, opened 2024-08-26
Reduce redundant GPU allocations

thorstenhater:
# Introduction

- Do not allocate and fill storage for resetting ion concentrations if these are never changed.
  Reasoning: If concentrations are never changed, we do not reset them and thus do not need
  to store the values. 
- Remove a duplicate of the CV area in solver and shared state and pass a reference during solving
- Remove redundant `solution` from GPU solver
- Do not store / allocate diffusion values if not needed.

This saves **per CV**
- `1 x 8B` for `cv_area` unconditionally
- `1 x 8B` for `Xd` for each ion with no diffusion is in use (majority of cases)
- `2 x 8B` for `Xi` for each ion (`reset` and `init`) if not written (reasonably often)
- `2 x 8B` for `Xo` for each ion (`reset` and `init`) if not written (majority of cases)
- `1 x 8B` for `eX` reset for each ion if not read (majority)
- `1 x 8B` for `eX` for each ion if not read (rarely)

In my standard benchmark, `busyring` with complex cells, this saves about 18% of the total GPU
allocation for the cell data (`shared_state`).

This has become a mixed bag, fixing a few additional things that came up during testing this:
- a bug in event handling on GPU
- pybind11 stub generation


jlubo:
This is nice!

The previously high memory consumption prevented the execution of my large network application with multi-compartment neurons (see [here](https://github.com/jlubo/arbor_network_consolidation) for the code of the single-compartment case, the code for the multi-compartment case is not public yet) on common GPUs. The present PR reduces the memory consumption for that application from >> 8 GB to 112 MB. 

Also, the whole variety of tests for my network application do pass (for both the single-compartment and the multi-compartment case).

thorstenhater:
You can also try #2394, which takes the concept one step further.

arbor/backends/multicore/shared_state.cpp
```diff
@@ -91,10 +103,13 @@ void ion_state::zero_current() {
 
 void ion_state::reset() {
     zero_current();
-    std::copy(reset_Xi_.begin(), reset_Xi_.end(), Xd_.begin());
     if (write_Xi_) std::copy(reset_Xi_.begin(), reset_Xi_.end(), Xi_.begin());
     if (write_Xo_) std::copy(reset_Xo_.begin(), reset_Xo_.end(), Xo_.begin());
     if (write_eX_) std::copy(init_eX_.begin(), init_eX_.end(), eX_.begin());
+    // This goes _last_ or at least after Xi since we might have removed reset_Xi
+    // when Xi is constant. Thus conditionally resetting Xi first and then copying
+    // Xi -> Xd is save in all cases.
```
jlubo, line 111:
```suggestion
    // Xi -> Xd is safe in all cases.
```

arbor/backends/multicore/shared_state.cpp
```diff
@@ -91,10 +103,13 @@ void ion_state::zero_current() {
 
 void ion_state::reset() {
     zero_current();
-    std::copy(reset_Xi_.begin(), reset_Xi_.end(), Xd_.begin());
```
jlubo, line 94: Is reading out the ion concentration (in the case of `USEION X READ Xd`) unaffected by this?

thorstenhater: Yes, this is just for resetting it to the starting value.

jlubo (approved):

jlubo (dismissed):

jlubo:
The failing pipeline might have to be fixed or removed.

thorstenhater:
Spack is failing due to pb11-stubgen not being updated. I made the corresponding PR here
https://github.com/spack/spack/pull/47162

boeschf (commented):
the conditional initialization and allocation of the different ion state arrays is not very easy to follow. There must be a better way to do this, but that's maybe for another PR.

spack/package.py
```diff
@@ -122,6 +122,7 @@ class Arbor(CMakePackage, CudaPackage):
         depends_on("py-pybind11@2.10.1:", when="@0.7.1:", type="build")
         depends_on("py-pybind11@2.10.1:", when="@0.7.1:", type="build")
         depends_on("py-pybind11@2.10.1:", when="@2.11.1:", type="build")
+        depends_on("py-pybind11-stubgen@2.5.1", type="build")
```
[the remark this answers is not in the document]
thorstenhater: It's the latest version, but we can relax this, yes.

arbor/backends/gpu/shared_state.cpp
```diff
@@ -50,26 +50,20 @@ ion_state::ion_state(const fvm_ion_config& ion_data,
...
-        Xi_ = make_const_view(ion_data.init_iconc);
-    }
-    if (read_Xo_) {
-        Xo_ = make_const_view(ion_data.init_econc);
-    }
+    // Allocate reset data only if anybody overwrites the resettable ararys
+    // This is used by internal and diffusive concentrations
     if (write_Xi_ || write_Xd_) {
-        // ... but this is used by Xd and Xi!
         reset_Xi_ = make_const_view(ion_data.reset_iconc);
```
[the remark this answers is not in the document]
thorstenhater: all these should be clearer (and correcter) now

boeschf:
can we add a spack variant that would disable the pybind11 stubgen? This would be consistent with the cmake and would allow the spack CI to still pass.

boeschf (approved):
thanks, looks good to me

PR #2394, opened 2024-08-26
Reduce even more GPU allocations

thorstenhater:
Bubbling up ion access information from modcc allows us to skip allocation of `Xi` and `Xo` iff
no mechanism reads those values. 

A minor problem: Sampling may try to touch data that doesn't exist; however who'd ask for `Xi`
if they never use it? 

Stacking this PR on top of #2393 the memory consumption drops by a further 5% down to 77% 
if the original value.

### Todo
- [x] Guard against sampling  non-existing `Xi`/`Xo`


jlubo:
I tested this analogously to #2393 and also here, all my tests pass. For the specific case, however, the further reduction in memory consumption is quite small (from 111.5 MB to 109.8 MB).

thorstenhater:
Well, that's expected as this reduction is conditional on certain commonly used things not being used.

jlubo (approved):
With many neuronal compartments, I just encountered very high *main* memory usage, but I guess that is a separate issue.

Regarding this PR, I don't have any further complaints.

jlubo (approved):

jlubo (approved):
Also here, the failing pipeline might have to be fixed or removed.

PR #2396, opened 2024-08-27
Allow for units to be sourced from outside.

### LINKED ISSUE #2392: Build fails with -DARB_USE_BUNDLED_UNITS=OFF because the install(units) statement is unconditional
Please allow arbor to cleanly build with an externally installed units library.

thorstenhater:
Closes #2392


yurivict:
There is the failure with this patch:
```
CMake Error at CMakeLists.txt:313 (install):
  install TARGETS given target "units" which does not exist.
```

units shouldn't be installed by arbor's cmake scripts.

thorstenhater:
This should be subsumed by #2267 and would need a rewrite if not.

PR #2404, opened 2024-09-11
Add debug print to flaky JSON test

thorstenhater:
Try to catch a flaky JSON test in CI.


boeschf (approved):
Thanks for finding this! Shouldn't CodeQL warn about stuff like this?

thorstenhater:
Clang-tidy and friends do warn about this, but no one had the stomach to sit down, work through all the warnings
deciding between fixing and ignoring them and make the setup part of Arbor. So far...
CodeQL doesn't seem to mind that particular issue.

PR #2406, opened 2024-09-12
Use clang-tidy. 

thorstenhater:
Allow (opt-in) to run `clang-tidy` on Arbor.

Limitations currently are:
- excluded `lmorpho`
- excluded `modcc`
- only the default set of check is in use `clang-analyzer-*`

Sensible suggestions were implemented and as only two
weird ones remained, I massaged the code to appease the
Linter despite being quite sure of a false positive.


boeschf:
cscs-ci run default

PR #2408, opened 2024-09-18
Remove clock for std::chrono

thorstenhater:
`clock.hpp` has become redundant over `std::chrono` and was replaced with `monotonic_clock`.


boeschf (approved):
lgtm

PR #2409, opened 2024-09-18
Clean-up catalogue extension.

thorstenhater:
1. rename C++ `catalogue::import` to `catalogue::extend` as `import` will likely be a reserved word due to modules
2. Use an empty default for the prefix. 99% real world examples use an empty prefix.
3. Adjust tests and examples accordingly.

### DIFF
--- a/arbor/include/arbor/mechcat.hpp
+++ b/arbor/include/arbor/mechcat.hpp
@@ -96,7 +96,7 @@ public:
     }
 
     // Copy over another catalogue's mechanism and attach a -- possibly empty -- prefix
-    void import(const mechanism_catalogue& other, const std::string& prefix);
+    void extend(const mechanism_catalogue& other, const std::string& prefix = "");
 
     ~mechanism_catalogue();
 
--- a/arbor/mechcat.cpp
+++ b/arbor/mechcat.cpp
@@ -1,5 +1,4 @@
 #include <cstdlib>
-#include <map>
 #include <memory>
 #include <string>
 #include <vector>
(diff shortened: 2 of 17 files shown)


arbor/include/arbor/mechcat.hpp
```diff
@@ -96,7 +96,7 @@ class ARB_ARBOR_API mechanism_catalogue {
     }
 
     // Copy over another catalogue's mechanism and attach a -- possibly empty -- prefix
-    void import(const mechanism_catalogue& other, const std::string& prefix);
+    void extend(const mechanism_catalogue& other, const std::string& prefix = "");
```
ErbB4, line 99: nice. I think I understood this perfectly hehe

arbor/mechcat.cpp
```diff
@@ -114,19 +113,19 @@ struct catalogue_state {
     catalogue_state() = default;
 
     catalogue_state(const catalogue_state& other) {
-        import(other, "");
+        extend(other, "");
```
ErbB4, line 116: What is the difference between import and extend, since it is used everywhere

thorstenhater: I renamed the method to avoid clashing with an upcoming in the C++ language (`modules` will bring `import`).
So, now all uses of `catalogue::import` (and `catalogue_state::import`) must be re-written as `extend` instead.

ErbB4: got it!

ErbB4 (approved):

PR #2410, opened 2024-09-18
Clean-up redundant items: files and functions.

thorstenhater:
Some justifications:
- power & Co: only useful on Cray, likely outdated. Use external power measurements if this is relevant.
- trace.hpp was never used and was locked_ostream's only user.


boeschf (approved):
thanks - 1243 lines gone!

thorstenhater:
and good riddance.

PR #2411, opened 2024-09-19
Check for supported compilers.

thorstenhater:
Utter a dire warning for those who try to build with outdated and/or weird compilers.
Also: I guess XLC is no longer a thing.


boeschf (approved):
lgtm

PR #2417, opened 2024-09-30
Allow probing of point state by tag.

thorstenhater:
This concludes the introduction of 'labels' for synapse identification / placements started back in 
PR #1504 by @noraabiakar. Labels are now used when probing a synapse's state.

### DIFF
--- a/arbor/fvm_lowered_cell_impl.hpp
+++ b/arbor/fvm_lowered_cell_impl.hpp
@@ -14,6 +14,7 @@
 
 #include <arbor/assert.hpp>
 #include <arbor/common_types.hpp>
+#include <arbor/cable_cell.hpp>
 #include <arbor/cable_cell_param.hpp>
 #include <arbor/recipe.hpp>
 #include <arbor/util/any_visitor.hpp>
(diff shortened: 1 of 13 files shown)


arbor/fvm_lowered_cell_impl.hpp
```diff
@@ -913,17 +915,27 @@ void resolve_probe(const cable_probe_point_state& p, probe_resolution_data<B>& R
...
+            if (handle.mech_id != mech_id) return;
+            auto mech_index = handle.mech_index;
+            R.result.push_back(fvm_probe_scalar{{data + mech_index},
+                                                 point_info_of(target,
+                                                               lid,
+                                                               mech_index,
+                                                               synapses.at(mech),
+                                                               R.M.mechanisms.at(mech).multiplicity)});
+        }
+    }
```
boeschf, line 938: just for my understanding - can there ever be more than one probe resolved and pushed to result?

boeschf (commented):
lgtm, thanks.

arbor/fvm_lowered_cell_impl.hpp
thorstenhater: Formally, yes, as `synapse_ranges` is a multimap of ranges. This _could_ happen 
if two synapses are given the same label. We could forbid that, but currently don't.

boeschf: sure, sounds good

boeschf (approved):

arbor/fvm_lowered_cell_impl.hpp
thorstenhater: Ah, that's impossible, as the probe contains the mechanism type (the id in the resolution step):
```
if (handle.mech_id != mech_id) return; 
```
So, multiple synapses of the **same** type can appear in the same probe, here.

PR #2401, opened 2024-09-10
C++20 Spaceship operator

thorstenhater:
First step along the road to C++20: Use the `<=>` operator instead of homebaked macros.
Also says goodbye to MacOS 12 and associated apple-clang, which is completely out of touch.

PR #2421, opened 2024-10-16
event handling: fix a bug and clean house

boeschf:
- fix event stream when more than one cell in a cell group have same synapse
  - events would previously no longer necessarily be sorted by time
  - in order to simplify: also sort with respect to mechanism index (as was previously only required for the gpu backend)
  - add pertinent test
- while cleaning up: overhauled the event related files and data structures
  - removed dead code
  - made event handling less generic (this feature was not used anywhere)


arbor/backends/event_stream_base.hpp
```diff
@@ -94,16 +98,75 @@ struct event_stream_base {
...
                 const auto& handle = handles[div + target];
-                streams[handle.mech_id].ev_data_.push_back({handle.mech_index, weight});
-                dt_sizes[handle.mech_id][step]++;
+                auto& stream = streams[handle.mech_id];
+                stream.spikes_.push_back(spike_data{step, time, handle.mech_index, weight});
+                // insertion sort with last element as pivot
+                auto first = stream.spikes_.begin();
+                auto last = stream.spikes_.end();
+                auto pivot = std::prev(last, 1);
+                std::rotate(std::upper_bound(first, pivot, *pivot, [](auto const& l, auto const& r) noexcept {
```
thorstenhater, line 107: Since #2401 this should be simply `<` with all the same semantics and performance, no?

arbor/backends/event_stream_base.hpp
```diff
@@ -94,16 +98,75 @@ struct event_stream_base {
...
+            static_cast<spike_event_stream_base&>(stream).init();
         }
     }
+
+  protected: // members
+    struct spike_data {
+        arb_size_type step = 0;
+        time_type time = 0;
+        cell_local_size_type mech_index = 0;
+        float weight = 0;
```
thorstenhater, line 137:
```suggestion
        float weight = 0;
        auto operator<=>(const spike_data&) const = default;
```

arbor/backends/event_stream_base.hpp
```diff
@@ -20,8 +18,7 @@ template <typename Event>
 struct event_stream_base {
     using size_type = std::size_t;
```
thorstenhater, line 19: This seem redundant or at least underutilized below.

thorstenhater (changes requested):
A few minor requests, but overall, this is great simplification!
A question that didn't fit into the PR itself: Threaded spike sorting
as on GPU is gone now? Why?

arbor/backends/event_stream_base.hpp
```diff
@@ -62,24 +59,31 @@ struct event_stream_base {
...
+        // reset streams and allocate sufficient space for temporaries
         auto n_steps = steps.size();
-        std::unordered_map<unsigned, std::vector<std::size_t>> dt_sizes;
         for (auto& [k, v]: streams) {
             v.clear();
-            dt_sizes[k].resize(n_steps, 0);
+            v.spike_counter_.clear();
+            v.spike_counter_.resize(steps.size(), 0);
+            v.spikes_.clear();
+            v.spikes_.reserve(v.ev_data_.capacity());
```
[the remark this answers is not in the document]
boeschf: this is intentional, as `ev_data_` has been cleared before. I should probably make a comment. Otherwise, I can reorder the operations, so that `v.clear()` is called afterwards.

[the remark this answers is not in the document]
thorstenhater: I think that would be confusing, just make a note.

boeschf:
> A few minor requests, but overall, this is great simplification! A question that didn't fit into the PR itself: Threaded spike sorting as on GPU is gone now? Why?

Since we need sorting in any case due to the events not being in order w.r.t. time under certain conditions (when more than one cell in a cell group have same synapse) I just rolled the sorting w.r.t. `mech_index` into the same insertion sort. This is not strictly necessary on CPU, but it doesn't change the semantics. The GPU implementation previously did this additional sorting with the mentioned thread-parallel implementation.

arbor/backends/event_stream_base.hpp
boeschf: good point

arbor/backends/event_stream_base.hpp
boeschf: yes

arbor/backends/event_stream_base.hpp
boeschf: absolutely, thanks for pointing this out

thorstenhater:
Understoo!

thorstenhater (approved):
When you're satisfied with the performance impact, we can merge.
🎉

thorstenhater:
I'll merge this now, correctness is more important, we can address potential performance later.

PR #2422, opened 2024-10-22
use supported compiler for codeql

boeschf:
We require at least gcc12


thorstenhater (approved):
Yes, that's also in #2393, we can frontport it ;)

PR #2428, opened 2024-12-10
Clean up K_T and K_P.

thorstenhater:
- Simplify logic and maths
- Remove double application of `qt`
- Align style and usage of both files.

### DIFF
--- a/mechanisms/allen/K_P.mod
+++ b/mechanisms/allen/K_P.mod
@@ -1,6 +1,7 @@
-: Comment: The persistent component of the K current
-: Reference:        Voltage-gated K+ channels in layer 5 neocortical pyramidal neurones from young rats:subtypes and gradients,Korngreen and Sakmann, J. Physiology, 2000
-
+: Comment:   The persistent component of the K current
+: Reference: Voltage-gated K+ channels in layer 5 neocortical pyramidal neurones from young rats:subtypes and gradients
+:            Korngreen and Sakmann
+:            J. Physiology, 2000
 
 NEURON {
     SUFFIX K_P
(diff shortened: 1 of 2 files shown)


mechanisms/allen/K_P.mod
```diff
@@ -30,29 +31,32 @@ BREAKPOINT {
...
-        m_rho = qt/(tauF * (1.25 +  13.00*exp(-(v - vshift) * 0.026)))
+    if (vs < -50) {
+        m_z = 175.03*exp( 0.026*vs)
+    } else {
+        m_z =  13.00*exp(-0.026*vs)
     }
 
-    h_rho =  qt/(360 + (1010 + 24*(v + 55 - vshift))*exp(-((v + 75 - vshift)/48)^2))
+    m_rho = qt/(tauF*(1.25 + m_z))
+    h_rho = qt/((24*vs + 2690)*exp(-((vs + 75)/48)^2))
```
beaherrera, line 49: It should be h_rho = qt/(360 + (24*vs + 2330)*exp(-(((vs + 75)/48)^2))) since '360' doesn't multiply the exponential term in the original definition.

beaherrera (commented):
Hi @thorstenhater,
There is a mistake in the K_P.mod definition (see comment on L49). The rest of the mechanisms look good.
Best,
Beatriz

mechanisms/allen/K_P.mod
thorstenhater: Yes, good catch, thank you.

thorstenhater: Fixed

jlubo (approved):

PR #2430, opened 2024-12-18
Add cable cell option to Brunel example

thorstenhater:
1. Add the option to use a simple HH cell to the Brunel example. 
2. Add correct group size controls to Brunel
3. Add more integration tests
4. Make `lif_cell` an aggregate, which enables C++20 designated initializers i.e. `lif_cell {.Vm=42_mV}`.

This enables us to
- test/benchmark the event delivery in the cable cell path (Brunel is quite event heavy)
- add a few integration tests checking the invariance across CPU/GPU and group size with both paths
- add optimizations of said code paths later.

### DIFF
--- a/arbor/include/arbor/lif_cell.hpp
+++ b/arbor/include/arbor/lif_cell.hpp
@@ -12,8 +12,8 @@ using namespace U::literals;
 
 // Model parameters of leaky integrate and fire neuron model.
 struct ARB_SYMBOL_VISIBLE lif_cell {
-    cell_tag_type source; // Label of source.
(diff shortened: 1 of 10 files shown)


ErbB4 (approved):

thorstenhater:
Note that there are some tests failing. These occur consistently when building with the Apple Clang Compiler Suite.
Potentially this is an instance of:
https://stackoverflow.com/questions/76407947/trivial-c-program-yields-different-result-in-clang-macos-arm64-and-clang-macos-x
I'll still merge, since nothing functional changed here, but spin off an issue at high priority to investigate.

PR #2439, opened 2025-01-27
Adjust Diffusion Scaling

thorstenhater:
Add missing scaling to diffusivity parameter.

### DIFF
--- a/.github/workflows/test-pip.yml
+++ b/.github/workflows/test-pip.yml
@@ -29,7 +29,7 @@ jobs:
           python -m pip install --upgrade pip
           python -m venv ~/env
           source ~/env/bin/activate
-          pip install numpy pybind11-stubgen
+          pip install numpy scipy pybind11-stubgen
       - name: Debug info Python
         run: |
           source ~/env/bin/activate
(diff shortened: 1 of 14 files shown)


jlubo:
Looks good to me. This also receives support from according tests that I wrote based on the last comment in issue https://github.com/arbor-sim/arbor/issues/2084. I've added the tests to the branch here: https://github.com/jlubo/arbor/commits/bug/diffusion-scaling/ @thorstenhater

jlubo (dismissed):

jlubo (commented):
The tests should be included, otherwise, can be merged.

thorstenhater:
Hi @jlubo,

thanks for the extra tests, I am a bit worried about adding SciPy to the dependencies, though.
Could we write the same in pure NumPy (which is already included)?

thorstenhater:
Thanks, but I think checking for scipy and executing the tests conditionally is slightly better. The CI suite now has SciPy installed, so that it will run the relevant tests always before merging and users are free to install SciPy or not.

jlubo (approved):

PR #2441, opened 2025-02-10
No more locks in sampler add/del

thorstenhater:
Recognize that sampler handling is a simulation-level method, which is not callable by multiple threads.

### DIFF
--- a/arbor/cable_cell_group.cpp
+++ b/arbor/cable_cell_group.cpp
@@ -389,9 +389,7 @@ void cable_cell_group::advance(epoch ep, time_type dt, const event_lane_subrange
     sample_size_type n_samples = 0;
     sample_size_type max_samples_per_call = 0;
 
-    if (!sampler_map_.empty()) { // NOTE: We avoid the lock here as often as possible
-        // SAFETY: We need the lock here, as _schedule_ is not reentrant.
-        std::lock_guard<std::mutex> guard(sampler_mex_);
+    if (!sampler_map_.empty()) { // NOTE: We avoid work here as often as possible
         for (auto& [sk, sa]: sampler_map_) {
             if (sa.probeset_ids.empty()) continue; // No need to make any schedule
             auto sample_times = util::make_range(sa.sched.events(tstart, ep.t1));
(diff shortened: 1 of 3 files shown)


halfflat:
Just for context, the locking was put in place so that samplers could be changed dynamically in response to observed samples, and so in fact could be called by multiple threads.

Not sure how best to enforce that sampler callbacks (or some other thread client side that is acting independently) can't change sampler–probe binding. How much do the locks themselves (as opposed to the other optimizations) slow things down? Keeping them is the easy option for safety and also for this functionality. I guess an alternative would be to have a check in the simulation method to e.g. throw if `run()` is underway in another thread. 

An alternative to keep the functionality would be to use that steering interface we mooted way back and put these requests on a queue which is consumed by the simulation object between epochs, but that's of course a big deal.

For applying many sampler/probe bindings at once but keeping locks, how about a bulk interface?

thorstenhater:
> Just for context, the locking was put in place so that samplers could be changed dynamically in response to observed samples, and so in fact could be called by multiple threads.

Interesting, I didn't have that background. Given the actual problem reported by @jlubo and the payoff I found, I'd be willing to accept the degradation in functionality, i.e. loss of dynamic sampling. To my knowledge there is neither an active use case nor is it tested or documented. For the actual simulation, it adds 4,000,000 samplers (!) in total, so any speed-up is critical.

> How much do the locks themselves (as opposed to the other optimizations) slow things down?

Removing locks is a small factor in speed-up for adding sampler, but will also speed-up the simulation time as locking is needed when pulling samplers from the map, too.

As for the for the suggestions, I like them, but given time / personnel constraints, let's reserve those for an actual use case. Although, I suspect a concurrent hash map (or using a straight-up vector) might also an option.

thorstenhater:
As for the dominant fraction in profiling: I am often seeing -- in benchmarks, thus biased towards shorter times -- 1:1 splits initialisation ./. simulation. Label resolution and connection sorting being major cost centres. For the first one: #2447. For the latter I am brewing something.

The use case that inspired this particular overhaul: @jlubo dumps all the state at all the places at the simulation's end. As that includes all synapses, $\mathcal{O}(n^2)$ samplers with a trivial schedule are set.

PR #2443, opened 2025-02-19
Remove obsolete show() from morphology and segment_tree classes

jeffreytepper:
When developing a cell in version 0.10.0, I found that the show() method does not exist for both the segement_tree and morphology classes. I removed them from the documentation.

### DIFF
--- a/doc/python/morphology.rst
+++ b/doc/python/morphology.rst
@@ -299,12 +299,6 @@ Cable cell morphology
 
         A list of the segments.
 
-    .. method:: show
-
-        Return a string containing an ASCII rendering of the tree.
-
-        :return: string
-
 .. py:class:: morphology
 
     A *morphology* describes the geometry of a cell as unbranched cables
@@ -357,12 +351,6 @@ Cable cell morphology
 
             :param int i: branch index
             :rtype: list[msegment]
-
-    .. method:: show
-
-        Return a string containing an ASCII rendering of the morphology.
-
-        :return: string
 
 .. py:class:: place_pwlin
 


thorstenhater:
Hi @jeffreytepper,
thanks for taking the time for making a PR, but these methods are defined on the Python classes:
https://github.com/arbor-sim/arbor/blob/79983144e6f20b268950cb6b3b5e8c1496ef43a0/python/morphology.cpp#L296
https://github.com/arbor-sim/arbor/blob/79983144e6f20b268950cb6b3b5e8c1496ef43a0/python/morphology.cpp#L328
and on my local installation they work just fine:
```
Python 3.13.2 (main, Feb  4 2025, 14:51:09) [Clang 16.0.0 (clang-1600.0.26.6)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import arbor as A
>>> t = A.segment_tree()
>>> t.show()
''
>>>
```
So, if this is not working for you, maybe something else is amiss?

jeffreytepper:
Hello Thorsten,

I am working with version 0.10.0 that I installed using pip on MacOS with Apple Silicon. 

> Python 3.12.4 | packaged by Anaconda, Inc. | (main, Jun 18 2024, 10:07:17) [Clang 14.0.6 ] on darwin
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import arbor as A
> >>> t = A.segment_tree()
> >>> t.show()
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> AttributeError: 'arbor._arbor.segment_tree' object has no attribute ‘show'

Just to further verify:

> >>> 'show' in dir(A.segment_tree)
> False

Whereas it works for other methods in the docs

> >>> 'size' in dir(A.segment_tree)
> True

In fact, here is the list of all attributes under segment_tree on my system:

> >>> [attr for attr in dir(A.segment_tree) if not attr.startswith('__')]
> ['append', 'apply_isometry', 'empty', 'equivalent', 'is_fork', 'is_root', 'is_terminal', 'join_at', 'parents', 'reserve', 'segments', 'size', 'split_at', 'tag_roots’]

I see another attribute ‘reserve’ that is not listed on the docs, perhaps this may be a clue?

The attributes for morphology:

> >>> [attr for attr in dir(A.morphology) if not attr.startswith('__')]
> ['branch_children', 'branch_parent', 'branch_segments', 'empty', 'num_branches', 'to_segment_tree']

Thank you for looking this over. By the way, I am a grad student that has been working with Jannik is his lab, it is a pleasure to meet you. 

Thank you for looking this over,

Max

thorstenhater:
Nice to meet you, too. This seems to be an issue with the docs, then. Are you sure you are looking at the same versions?
That is: docs of v0.10 ./. arbor v0.10. I am asking since the ASCII art feature was merged _after_ v0.10

thorstenhater:
@jeffreytepper any updates here? I'd like to close this, if possible.

PR #2447, opened 2025-03-14
Re-factor label-resolution.

thorstenhater:
Using our standard Brunel example with cable cells instead of LIF cells we notice that 
simulation setup take roughly 10x longer than the actual execution. By taking a hard look
at the label resolver functionality, we can reduce the setup time by more than half and reduce
the total runtime by half (was 15s)

Result:
```
bin/brunel --n-excitatory 10000 --n-inhibitory 2500 --tfinal 10 --in-degree-prop 0.1 --dt 0.01 --lambda 1.5 -g 0.8 --delay 1.0 --weight 0.00010 -G 10 --use-cable-cells
==========================================
  Brunel model miniapp
  - distributed : 1 (mpi)
  - threads     : 8
  - gpus        : no
==========================================


There were 38956 spikes

meter                         time(s)
setup                           0.000
model-init                      5.366
model-simulate                  1.549
meter-total                     6.915
```

Changes:
- Notice that neither `assert-univalent` nor `round-robin-halt` actually have a state and remove those.
- State is now stored in a one-layer hash map with key `(gid, tag)`.
- Remove `variant` and associated polymorphism, replace by `switch`.
- Remove the partition from `range_set` and use linear search.
- Make resolver able to reset, avoiding re-allocation


thorstenhater:
Merged via #2449

PR #2448, opened 2025-03-17
CSCS CI 

thorstenhater:
💀


ibanezbm (approved):
Approved. GitLab CI is no longer needed after migrating to GitHub.

thorstenhater:
Technically this was intended to run Arbor's CI on an HPC system incl GPU, so it had some value.
However, never worked correctly and never was finished.

PR #2459, opened 2025-05-08
Add an escape hatch to label resolution

thorstenhater:
Allow users to explicitly opt out of label resolution when needed. This is available
in two steps: using some manually resolved connections and disabling resolution
completely, turning string labels into hard errors. The latter will also skip the MPI
step. Manual connections will not be checked for existence of the object.

This is an optimisation for the rare cases at extreme scale where initialisation time
is important. The performance impact is quite dramatic:

```json
{
    "name": "run_n=128_d=10-complex=true",
    "num-cells": 8192,
    "num-tiles": 4096,
    "synapses": 1000,
    "min-delay": 5,
    "duration": 250,
    "ring-size": 4,
    "event-weight": 0.2,
    "record": false,
    "spikes": false,
    "dt": 0.1,
    "depth": 2,
    "complex": false,
    "resolve-sources": false,
    "cpu-group-size": 1024,
    "branch-probs": [
        1,
        0.1
    ],
    "compartments": [
        2,
        1
    ],
    "lengths": [
        2,
        1
    ]
}

```

Before (or `"resolve-sources": true`)
```
gpu:      no
threads:  8
mpi:      no
ranks:    4096

running simulation

100% |--------------------------------------------------|           250ms

394264576 spikes generated at rate of 6.34092e-07 ms between spikes

meter                         time(s)
model-init                     55.449
model-run                      15.519
meter-total                    70.968
```

After  (or `"resolve-sources": false`)
```
gpu:      no
threads:  8
mpi:      no
ranks:    4096

running simulation

100% |--------------------------------------------------|           250ms

394264576 spikes generated at rate of 6.34092e-07 ms between spikes

meter                         time(s)
model-init                      3.717
model-run                      15.392
meter-total                    19.109
```


thorstenhater:
(docs currently broken due to faulty snowball stemmer release 3.0.0)

arbor/communication/communicator.cpp
```diff
@@ -170,6 +172,30 @@ void communicator::update_connections(const recipe& rec,
     }
     PL();
 
+    PE(init:communicator:update:connections:raw);
```
ibanezbm, line 175: To avoid duplicating code here, one idea could be to use a lambda function, as shown below. That said, if you feel it makes the code harder to read (which is likely), we can just leave it as is no problem:

``` cxx
PE(init:communicator:update:connections:local);

std::size_t n_con = 0;
std::vector<std::vector<connection>> connections_by_src_domain(num_domains_);
target_resolver.clear();

bool resolution_enabled = rec.resolve_sources();

auto process_connections = [&](const auto& get_connections, const auto& get_src_lid) {
    for (const auto tgt_gid: gids) {
        auto iod = dom_dec.index_on_domain(tgt_gid);
        source_resolver.clear();

        for (const auto& conn: get_connections(tgt_gid)) {
            auto src_gid = conn.source.gid;

            if (src_gid >= num_total_cells_) {
                throw arb::bad_connection_source_gid(tgt_gid, src_gid, num_total_cells_);
            }

            auto src_dom = dom_dec.gid_domain(src_gid);
            auto src_lid = get_src_lid(conn.source);
            auto tgt_lid = target_resolver.resolve(tgt_gid, conn.target);

            // Compatible with older compilers (no designated initializers)
            connections_by_src_domain[src_dom].emplace_back(connection{
                {src_gid, src_lid}, tgt_lid, conn.weight, conn.delay, iod
            });

            ++n_con;
        }
    }
};

if (!resolution_enabled) {
    throw resolution_disabled{gids.front()};
}

// Process resolved connections
process_connections(
    [&](auto gid) -> const auto& { return rec.connections_on(gid); },
    [&](const auto& src) { return source_resolver.resolve(src); }
);

PL();

PE(init:communicator:update:connections:raw);

// Process raw connections
process_connections(
    [&](auto gid) -> const auto& { return rec.raw_connections_on(gid); },
    [&](const auto& src) { return src.index; }
);

PL();
```

python/recipe.hpp
```diff
@@ -25,40 +25,23 @@ namespace pyarb {
 // of util::unique_any used by the C++ recipe interface.
 // The recipe_shim unwraps the python objects, and forwards them
 // to the C++ back end.
-
-class recipe {
-public:
-    recipe() = default;
+struct recipe {    recipe() = default;
```
ibanezbm, line 28: Line break

ibanezbm (approved):
Approve!!

PR #2460, opened 2025-05-08
Replace MPI_Allgatherv with MPI_Alltoallv for communication pattern

ibanezbm:
Change the internal communication model of the simulator to improve scalability in scenarios involving hundreds of ranks.


arbor/communication/communicator.cpp
```diff
@@ -147,7 +149,12 @@ void communicator::update_connections(const recipe& rec,
     std::size_t n_con = 0;
     std::vector<std::vector<connection>> connections_by_src_domain(num_domains_);
     target_resolver.clear();
+    auto my_rank = ctx_->distributed->id();
+    std::unordered_map<cell_gid_type, std::unordered_set<cell_size_type>> outgoing_remote_targets;
```
thorstenhater, line 153: Two notes here:
1. For better or worse 'remote' has been used for co-simulation partners elsewhere, ie a separate program we exchange spikes with. To avoid confusion with those items I'd suggest a different name here.
2. Using a `set` is likely not the most frugal choice in terms of iteration speed and memory consumption, a vector should fit the usage patterns a bit better.

arbor/communication/communicator.cpp
```diff
@@ -218,28 +241,45 @@ time_type communicator::min_delay() {
...
+    for (const auto& subvec : local_spikes) {
+        flat_spikes.insert(flat_spikes.end(), subvec.begin(), subvec.end());
     }
-    auto remote_spikes = ctx_->distributed->remote_gather_spikes(local_spikes);
-    PL();
 
+    std::sort(flat_spikes.begin(), flat_spikes.end()); 
+    flat_spikes.erase(std::unique(flat_spikes.begin(), flat_spikes.end()), flat_spikes.end());
+    
+    auto remote_spikes = ctx_->distributed->remote_gather_spikes(flat_spikes);
```
thorstenhater, line 281: This seems like one unfortunate consequence of the nested vector datastructure, especially as the overhead is payed even if no cosimulation partner exists.

arbor/include/arbor/simulation.hpp
```diff
@@ -79,7 +81,7 @@ class ARB_ARBOR_API simulation {
 
     // Register a callback that will perform a export of the rank local
     // spike vector.
-    void set_local_spike_callback(spike_export_function = spike_export_function{});
+    void set_local_spike_callback(rank_spike_export_function = rank_spike_export_function{});
```
thorstenhater, line 84: This is a user supplied callback and thus is breaking API, which is currently not a showstopper since Arbor is <1.0 as such.
However, since the user will now be presented with duplicates of spikes (one for each rank we will transmit it to), we essentially expose internal details to them. 
That's problematic on a few levels (happy to discuss in detail, if you want)

thorstenhater (changes requested):
Hi @ibanezbm,

thank you for this effort, it'll be central to scaling up to the large machines coming down the pipeline. I have left some general, high-level comments about design and datastructures. The current implementation -- storing spikes in nested vectors -- is a bit problematic, as it will expose the A2A algorithm to users of the high-level interface. In particular, Python and C++ users will 'see' the bucket structure through duplicate spikes. I'd suggest keeping the spikes in a flat vector as before and construct the bucketed vector just before exchange. If the performance is worse, don't worry for now and focus on correctness and interface first. If it scales better, we will find a way to make it worthwhile.

PR #2469, opened 2025-07-01
use better builtins for double max/min 

thorstenhater:
`std::max/min` don't generate specialized PTX for double/float, so use `std::fmax/fmin/fmaxf/fminf` instead.

Following https://karimsayedre.github.io/RTIOW.html 
Used https://godbolt.org/z/bj9YWEGqG


ibanezbm (approved):
Approved

PR #2513, opened 2026-06-09
Fixes broken build on aarch64+SVE when ARB_VECTORIZE=off

bcumming:
The __ARM_FEATURE_SVE preprocessor variable is always set when building on Grace-Hopper - so the vls_sve_bits.h file will always be configured and included in simd/native.hpp

The value of ARB_SVE_BITS must be determined by CMake when vls_sve_bits.h is included, but it was only set in CMake when ARB_VECTORIZE=on.


thorstenhater (approved):
Hi @bcumming,

thanks for catching that!

PR #2518, opened 2026-06-24
Replace outdated method for exporting CUDA symbols.

thorstenhater:
Adjust CMake to still export CUDA symbols despite removal of outdated method.


thorstenhater:
Hey @bcumming,

this fixes the other issue you encountered on GH200 while preserving the
export of CUDA-RT symbols.

bcumming (approved):
Looks good!

PR #2519, opened 2026-07-14
Method to retrieve Arbor configuration in C++ / cleanup of busyring example 

jlubo:
This PR introduces a method to get a string representation of the Arbor configuration in C++ (analogously to the `config()` method that has already been available for the Python frontend).

While the C++ function was already introduced by PR #2489 within the busyring example, this PR makes it globally available as a member of the `arb` class.

The PR further introduces small fixes for the busyring example (in the most recent version, the synapse number was off by 1) and for the profiler (missing header file).

### DIFF
--- a/arbor/version.cpp
+++ b/arbor/version.cpp
@@ -18,7 +18,7 @@ ARB_ARBOR_API const char* full_build_id = ARB_FULL_BUILD_ID;
 
 using config_map = std::map<std::string, std::variant<std::string, bool, int>>;
 
-// returns Arbor config map
+// returns Arbor configuration map
 ARB_ARBOR_API config_map get_arbor_config() {
     config_map config;
 
@@ -67,7 +67,7 @@ ARB_ARBOR_API config_map get_arbor_config() {
     return config;
 }
 
-// pretty-print function for the Arbor config map
+// return pretty-printed string representation of the Arbor configuration map
 ARB_ARBOR_API std::string get_arbor_config_str() {
     config_map config = get_arbor_config();
     std::string config_str = "";
@@ -78,10 +78,7 @@ ARB_ARBOR_API std::string get_arbor_config_str() {
         }
 
         std::string operator()(bool value) const {
-            if (value)
-                return "true";
-            else
-                return "false";
+            return value ? "true" : "false";
         }
 
         std::string operator()(int value) const {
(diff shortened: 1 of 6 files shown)


thorstenhater:
Hi @jlubo,

thank you, having the configuration as a table of values at runtime is really helpful. However, I am not sure though whether locking this to a string is necessarily the most useful way of implementing the functionality. My feeling is 
that providing a `map<string, value>` where `value = variant<string, int>` or possibly just `value = string` and having a pretty printer on top of that might be more natural. What do you think?

jlubo:
Thanks for the suggestion @thorstenhater I've now implemented it in such a way.

Just, the order of the listed config parameters is now different than before (now alphabetically ordered). But I guess that shouldn't be an issue.

arbor/version.cpp
```diff
@@ -13,4 +15,88 @@ ARB_ARBOR_API const char* version_dev = ARB_VERSION_DEV;
...
+ARB_ARBOR_API std::string get_arbor_config_str() {
+    config_map config = get_arbor_config();
+    std::string config_str = "";
+
+    struct value_str {
+        std::string operator()(const std::string& value) const {
+            return "'" + value + "'";
+        }
+
+        std::string operator()(bool value) const {
```
thorstenhater, line 80: Really a nitpick, but if you touch this code (see below), please use `{}` or `value ? "true" : "false"`.
Our formatting rule is to omit the curlies only if the statement goes on a single line 🤓

arbor/version.cpp
```diff
@@ -13,4 +15,88 @@ ARB_ARBOR_API const char* version_dev = ARB_VERSION_DEV;
...
+            else
+                return "false";
+        }
+
+        std::string operator()(int value) const {
+            return std::to_string(value);
+        }
+    };
+
+    for (auto it = config.begin(); it != config.end(); ++it) {
```
thorstenhater, line 89: Question here is whether you might want a format closer to JSON?
Meaning
- wrap in `{...}`,
- add `"..."` around string values instead of `'...'`,
- and add `"..."` around keys.
That would allow machine reading of this string.

example/busyring/ring.cpp
```diff
@@ -473,7 +426,7 @@ arb::cable_cell complex_cell(arb::cell_gid_type gid, const cell_parameters& para
     auto dend = tagged(3);
     auto apic = tagged(4);
     auto cntr = location(0, 0.5);
-    auto syns = arb::ls::uniform(rall, 0, params.synapses-1, gid);
+    auto syns = arb::ls::uniform(rall, 0, params.synapses-2, gid);
```
thorstenhater, line 429: Why -2? We place one synapse for the ring connectivity, thus I'd expect -1 here?

thorstenhater (changes requested):

example/busyring/ring.cpp
jlubo: This is now just how it was before PR #2489, where it was falsely changed (the single ring-internal synapse should not be included here, as it is added separately).

arbor/version.cpp
jlubo: Adjusted that, thanks for the reminder :)

arbor/version.cpp
jlubo: I've added an additional function that returns a JSON representation of the configuration.

thorstenhater (approved):

PR #2522, opened 2026-08-26
Fix profiler calls field

marvinKaster:
The number of calls to a profile region had a bug because we didn't check whether it had been initialized. This PR fixes this and also does some refactoring.


thorstenhater (commented):
needs an `#include <format>` ;)

marvinKaster:
I also added a fix for the wall timers. Sometimes the time of children timer was higher than their parents

thorstenhater (approved):
Thanks for fixing this. One thing we will have to look at rather soon is the `timer_stack` construction; a lot of `vector<string>` are made, copied, and destroyed. While it increases accuracy, it also shows in the overhead.

